PR #30, opened 2018-11-05
Init weights and bias of gcn using glorot and zeros

riccardobucco:
As you can see in the official repo weights are initialized using glorot and bias using zeros
Check [layers.py](https://github.com/tkipf/gcn/blob/master/gcn/layers.py) (commit n. 9b8bd4b) at lines 155 and 158.

I simply added two new init functions (in the nn/init.py file) and updated the reset_parameters method of the GCNConv according to Thomas Kipf's code


rusty1s:
Thank you. Looks amazing. Will merge ASAP.

PR #58, opened 2018-12-18
Added None case to repeat

Bawaw:
If this is not added the line 8 crashes: 
```
if (len(src) > length): 
TypeError: object of type 'NoneType' has no len()
```

When invoking something like: 
```
voxel_grid(data.pos, data.batch, size = 10)
```


rusty1s:
Thanks :)

PR #62, opened 2018-12-20
Fix typo in MetaLayer and add test for it

AntoinePrv:
In `MetaLayer`, `global_x` was returned but not defined when `global_model` was None.
Added simple test for the fix.


AntoinePrv:
I don't know if you think adding a test dependency for python 2.7 is worth it...

rusty1s:
Thank you so much for spotting this bug. I will merge this pull request and slightly modify it :)

PR #63, opened 2018-12-26
Discrete and continuous node attributes in TUDataset

bknyaz:
TUDataset sometimes have both discrete and continuous (float-valued) node attributes and currently they are just concatenated. Some algorithms like Weisfeiler-Lehman (WL) Graph Kernels cannot make use of continuous node attributes. In some cases, we also may want to test algorithms with or without these node attributes to compare fairly to WL. I think it is important to address this to avoid future confusion as many people might use this toolbox to benchmark their algorithms.

1. I suggest to add flag to TUDataset to let the user decide whether to use them or not. 
2. I also would store their dimensionality since it is lost during concatenation (see https://github.com/rusty1s/pytorch_geometric/blob/eb0d7bad2f220d8ba277a17adba3f4aee6f7f72c/torch_geometric/read/tu.py#L30)

### DIFF
--- a/test/datasets/test_tu_dataset.py
+++ b/test/datasets/test_tu_dataset.py
@@ -17,7 +17,7 @@ def test_enzymes():
     dataset = TUDataset(root, 'ENZYMES')
 
     assert len(dataset) == 600
-    assert dataset.num_features == 21
+    assert dataset.num_features == 3
     assert dataset.num_classes == 6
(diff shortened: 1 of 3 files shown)


rusty1s:
Thank you! This is something I already had in mind, especially because in most papers it is not clear whether continuous attributes were used or not. I will check the code and merge when I get back home.

PR #67, opened 2019-01-08
Add k - N zero rows if N < k in global_sort_pool

todoaskit:
- Ref: Zhang et al. page 4


rusty1s:
Looks really good. Thank you very much.

PR #90, opened 2019-02-08
Add geoemtric shapes

Yannick-S:
Hi,
as promised, I created a geometric dataset. The set is not 100%, but I guess good for a start.

I had to change the read_off function because blender output off files with not just triangles, but also rectangle faces. 

About the geomtry data class, I am not sure if it is necessary to have the
@property
    def raw_file_names(self):
        return ['cone', 'cube']

as part of the code. I am not exactly sure what the code does, but I think it can safely be removed.

And the dataset is right now saved in a public repository on my github.

### DIFF
--- a/torch_geometric/datasets/geometry.py
+++ b/torch_geometric/datasets/geometry.py
@@ -1,0 +1,69 @@
+import os
+import os.path as osp
+import glob
+from os.path import basename
+import torch
+from torch_geometric.data import InMemoryDataset, download_url, extract_zip
+from torch_geometric.read import read_off
+
+class GeometricShapes(InMemoryDataset):
+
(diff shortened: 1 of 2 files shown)


Yannick-S:
oh and also: If you tell me how you did the documentation for the other datasets, I can do it for this one too.

rusty1s:
Thank you! I think documentation is necessary and it would be great if you could add it. The documentation is auto-generated from doc strings ([example](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/datasets/faust.py). You can test and visualize it by running `make html` in the `docs` dir.

rusty1s:
I made a few changes to your code, e.g., I couldn't download the dataset given the URL you defined.
The data looks cool so far. Thank you very much!

PR #105, opened 2019-02-25
Fix inconsistent super init call within in the class definition of GC…

lvermue:
…NConv

	modified:   docs/source/notes/create_gnn.rst


rusty1s:
Cool, thanks!

PR #130, opened 2019-03-14
Fix PointCNN benchmark; sklearn is replaced by scikit-learn.

aywi:
Fix PointCNN benchmark (dim (int) is required); [sklearn](https://pypi.org/project/sklearn/) is replaced by [scikit-learn](https://pypi.org/project/scikit-learn/).


rusty1s:
Thank you!

PR #151, opened 2019-03-25
Added more autoencoder models. Need help debugging them and making them work with my + other's datasets.

jlevy44:
Thanks! I'll be here to help develop. Seems like some of your autoencoder test scripts need to be fixed as well.


jlevy44:
Let's see if we can finish this PR soon :)

rusty1s:
Thanks! Looks good! I will fix sone things and merge ASAP, e.g., you need to call BCEWithLogitsLoss when omitting the sigmoid.

rusty1s:
Updated models in master. I will also add an example.

rusty1s:
I think your implementation of adversarial autoencoders were wrong. You need two optimizers for adversarial optimization: One for the encoder, and one for the discriminator.

jlevy44:
Ah thanks. I was going to mention that they had bugs and were not fully tested.

jlevy44:
You’re implementation looks great though. Can’t wait to use

PR #165, opened 2019-03-28
Adding Line Graph Transformation

pgtgrly:
Added Functionality for Line Graph Transformations for Directed and undirected graphs.

New data.x is set to old data.edge_attr and the data.edge_attr is set to None

### DIFF
--- a/test/transforms/test_line_graph.py
+++ b/test/transforms/test_line_graph.py
@@ -1,0 +1,36 @@
+import torch
+from torch_geometric.transforms import LineGraph
+from torch_geometric.data import Data
+
+
+def test_line_graph():
+    assert LineGraph().__repr__() == 'LineGraph'
+
+
+def test_line_graph_directed():
+    edge_index = torch.tensor([[0, 1, 2, 2, 3],
+                               [1, 2, 0, 3, 0]], dtype=torch.long)
+
+    expected_row = [0, 1, 1, 2, 3, 4]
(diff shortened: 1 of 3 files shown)


pgtgrly:
In case of Directed graph I have made an assumption (in interest of speed) that the user must define the edge index as:
[[0,1,1,2,2,0,]
 [1,0,2,1,0,2]]
instead of:
[[0,1,2,1,2,0]
 [1,2,0,2,1,2]]

rusty1s:
Thank you! This looks awesome. I will check and merge ASAP.

rusty1s:
Hi, I further improved on your version and made it a bit more "vectorized". Especially your "assumption" does not need to be true anymore. Running times are now VERY fast.

pgtgrly:
This is awesome! I will go through the code and understand the vectorized implementation as well as the bug.
I would appreciate it if you can share any resources that I can use to to learn how to write more vectorized code.
Thanks!

rusty1s:
I'd say it's all about experience. Generally you should force yourself to avoid Python-for loops as good as you can. IMO, it helps to debug/print out `numpy` code to understand what exactly is going on. That's how I learned it. Once you feel confident, you will start to exploit those methods for specific use-cases. Nonetheless, you're code was great and it helped me a lot. Thank you for your contribution!

pgtgrly:
Thank you for the advice!

PR #172, opened 2019-03-31
Add a dockerfile

shengwenLeong:
This dockerfile could be used to create a container and also allow users to add more software by themselves.


rusty1s:
Thank you!

PR #174, opened 2019-04-01
Added knn interpolation

kuynzereb:
Hi, I implemented feature propagation level from [PointNet++: Deep Hierarchical Feature Learning on Point Sets in a Metric](https://arxiv.org/abs/1706.02413), which can be used for segmentation tasks. It seems that now there is no correct place for such a function, so I created unpool module.


rusty1s:
Awesome contribution with good code quality. I like it. Thanks :)

PR #178, opened 2019-04-01
PointConv fix

kuynzereb:
It seems that confusion with indices are happened.


rusty1s:
Oh s**t. Thank you :)

Issue #182, opened 2019-04-02
PointNet++ implementation


kuynzereb:
## ❓ Questions & Help

Hi, I have some questions about implementation of PointNet++ in examples. First one: now there are the following lines:
```
idx = fps(pos, batch, ratio=0.5)  # 512 points
edge_index = radius(
    pos[idx], pos, 0.1, batch[idx], batch, max_num_neighbors=64)
```

If I understand correctly, it means that we find 64 sampled points for each point from initial point cloud. But in original pointnet++ we should find 64 points from initial point cloud for each sampled point (which is not the same thing). So it seems for me that more correct way is:
```
idx = fps(pos, batch, ratio=0.5)  # 512 points
edge_index = radius(
    pos, pos[idx], 0.1, batch, batch[idx], max_num_neighbors=64)
```
The second question. It seems for me that there is a problem with indices.
```
idx = fps(pos, batch, ratio=0.25)  # 128 points
edge_index = radius(
    pos[idx], pos, 0.2, batch[idx], batch, max_num_neighbors=64)
N, M = pos.size(0), idx.size(0)
```
Let's assume that `idx` has size 1, so there is only 1 sampled point. Then `pos[idx]` also has size 1, which means, that `edge_index[1]` is `[0, ..., 0]`. Now, look at the next line
```
x = F.relu(self.local_sa2(x, pos, edge_index, size=(N, M)))
```
Inside PointConv propagate we now have `x_i = [x[0], ..., x[0]]`, whereas it should be `[x[idx[0]], ..., x[idx[0]]]`. If my understanding is correct, then there is a problem with local and global indices of sampled points.

kuynzereb:
Maybe should be made something like
```
edge_index = torch.stack(edge_index[0], idx[edge_index[1]])
```

rusty1s:
Hi, Thank you very much for being so watchful.

To your first concern: The `radius` op should be permutation-invariant due to being based on distances. However, I see that this is not really convincing, as it is described the other way around in the documentation.

To your second concern: It seems that you are right and `PointConv` is totally bugged on bipartite graphs (it still works though :D). I like your suggestion. Will fix it ASAP.

kuynzereb:
Now when you said it I understand that radius really should be permutation-invariant with current implementation. Thank you!

It seems for me that it is a general problem of this idea of bipartite graph. I mean, if we want two sets of points with independant indices, then we should have two feature matrices and two position matrices. 

[...]

PR #183, opened 2019-04-02
New MessagePassing for bipartite graphs

kuynzereb:
I propose the following modification of MessagePassing class, which should fix the problem mentioned in #182


rusty1s:
Awesome! I made some minor changes and added tests. Thank you!

PR #204, opened 2019-04-12
Possible subscript typo for GATConv formula

liviu-:
I think there may be a small typo in the [`GATConv`](https://rusty1s.github.io/pytorch_geometric/build/html/_modules/torch_geometric/nn/conv/gat_conv.html#GATConv) formula in the first term where I believe the **x** vector should refer to the features of the current node _i_, and _j_ is not yet defined.


rusty1s:
Yes, thank you :)

PR #310, opened 2019-05-18
add supported type

sw-gong:
Situation:
The meshes I'm working on hold an attribute with the size [N, ?, L], represents N points, ? neighbors, L lengths. The below code would fix the 'Unsupported attribute type' error.

### DIFF
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -1,5 +1,6 @@
 import re
 import warnings
+import numpy as np
 
 import torch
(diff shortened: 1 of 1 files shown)


sw-gong:
There are still other conflicts.

rusty1s:
Hi, couldn‘t you just convert your tensor to PyTorch?

sw-gong:
I can't as I said. The second dimension is not equal of each point.

rusty1s:
Sure, that is not possible. You should encode those as edge features.

PR #315, opened 2019-05-18
Add missing variable in Introduction

ekagra-ranjan:
Given the outputs, the variable `key` must have been used while printing


rusty1s:
Thanks :)

PR #316, opened 2019-05-18
Fix typo in Attention equation in DNA

ekagra-ranjan:
Replace Q with Theta in equation of Attention in DNA. Improved comments on args.


rusty1s:
Thanks :)

PR #324, opened 2019-05-20
Size of NoneType does not exist

jzhou316:
better check "if tensor is not None" at the very beginning


rusty1s:
Perfect, thanks!

PR #330, opened 2019-05-22
make `correct` float to avoid truncating

ekagra-ranjan:
Without float conversion the `acc` variable always becomes `0`.


rusty1s:
Well, this is only needed for Python 2.7.

PR #339, opened 2019-05-24
Removed checks before applying dropout.

fadel:
These arguments are already checked before applying dropout inside [ATen](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/Dropout.cpp#L44). Do we need those checks here?


rusty1s:
No, thank you :)

PR #352, opened 2019-05-30
Use globally defined `device` instead of passing as arg

ekagra-ranjan:
The PR reduces the number of args in functions by using globally defined `device` variable.


rusty1s:
Thank you.

Issue #329, opened 2019-05-22
RuntimeError: cuda runtime error (6) 


FrancescoSaverioZuppichini:
## 🐛 Bug
Hi,

it's me again with a new bug :) So I am trying to train DGCNN on ModelNet10

```

#imports

pre_transform = T.NormalizeScale()

train_ds = ModelNet(root='./',
             train=True,
             pre_transform=pre_transform)

test_ds = ModelNet(root='./',
             train=True,
             pre_transform=pre_transform)


device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

train_dl = DataLoader(train_ds, batch_size=16)

test_dl = DataLoader(test_ds, batch_size=16)

class DynamicEdgeConv(gnn.EdgeConv):
    def __init__(self, k=6, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.k = k

    def forward(self, pos, batch):
        edge_index = knn_graph(pos, self.k, batch, loop=False)
        return super().forward(pos, edge_index)


model = ... # here I use DynamicEdgeConv

model = DGCNNClassification(3,10).to(device)

optimizer = Adam(model.parameters(), 0.001)
criterion = nn.CrossEntropyLoss()

for data in train_dl:
    start = time.time()
    optimizer.zero_grad()
    data = data.to(device)
    print(data)
    out = model(data.pos, data.batch) #HERE BOOM
    loss = criterion(out, data.y)
    loss.backward()
    print('[INFO] loss={} elapsed={:.2f}'.format(loss.item(), time.time() - start))
    optimizer.step()

```
THCudaCheck FAIL file=/opt/conda/conda-bld/pytorch_1556653114079/work/aten/src/THC/THCReduceAll.cuh line=327 error=6 : the launch timed out and was terminated
Traceback (most recent call last):
  File "DGCNN.py", line 144, in <module>
    out = model(data.pos, data.batch)
  File "/home/fzuppic/anaconda3/envs/gdl/lib/python3.7/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "DGCNN.py", line 115, in forward
    out = conv(out, batch)
  File "/home/fzuppic/anaconda3/envs/gdl/lib/python3.7/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "DGCNN.py", line 68, in forward
[...]

PR #357, opened 2019-05-31
Upload dgcnn for segementation

GrumpyZhou:
Hi @rusty1s ,

This is the DGCNN segmentation script, it also performs good in my environment. I somehow can not import mean_iou method and i leave it as it is, since I think you are going to change it anyway :) 

But it now only works per category. So later i will upload a shapenet dataset version that process all categories together.

### DIFF
--- a/examples/dgcnn_segmentation.py
+++ b/examples/dgcnn_segmentation.py
@@ -1,0 +1,193 @@
+import os.path as osp
+
+import torch
+import torch.nn as nn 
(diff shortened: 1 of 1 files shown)


jeannotes:
any results? @GrumpyZhou

GrumpyZhou:
@jeannotes  What do you mean by any results? Regarding your question, I just answered it in #329.

PR #359, opened 2019-06-01
Shapenet version that loads all categories together

GrumpyZhou:
Hi @rusty1s ,

This dataset relabels the label of each category in preprocess step, so that one can use it to train on all categories together and from my experiments it also works with the segmentation script that i uploaded in my last pull request.

### DIFF
--- a/torch_geometric/datasets/shapenet_all.py
+++ b/torch_geometric/datasets/shapenet_all.py
@@ -1,0 +1,162 @@
+import os
+import os.path as osp
+import glob
+
+import torch
+from torch_geometric.data import (Data, InMemoryDataset, download_url, extract_zip)
+from torch_geometric.read import read_txt_array
+from torch_geometric.data.makedirs import makedirs
+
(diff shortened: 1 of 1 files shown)


rusty1s:
Hi and thank you. Just wondering if this is common practice to train this dataset all at once?

GrumpyZhou:
Hi, 

Yes i think it is the common practice. You can look at [pointnet++ original implementation](https://github.com/charlesq34/pointnet2/blob/master/part_seg/part_dataset_all_normal.py).
Simply by looking the code, first there's no where you can define the category for training, secondly, looking at line 74-75 it is obvious they need to remap the labels. Finally, it also doesn't make much sense to evaluate a model on a single category... like we never train classifier for only 1 class.

rusty1s:
I unified the `ShapeNet` and `ShapeNetAll` datasets. Users can now pass in a list of categories (or `None` if they want to load all categories together).

GrumpyZhou:
Cool, thanks!

PR #361, opened 2019-06-01
Add JK to GCN in kernel benchmarking

ekagra-ranjan:
This PR adds to the kernel benchmark the following:
1) Jumping Knowledge variant to GCN
2)  Generalises JK on SAGE by importing JumpingKnowledge to support pooling operations other that concat
3) Proposes naming of model with JK to be named as `modelWithJK` and models without JK to be named as `model` to avoid any confusion.


rusty1s:
Really cool. I like that.

ekagra-ranjan:
Thanks @rusty1s! I think I messed up with the naming convention although. Just now I realised that all the models were having JK style implementation. In that case if we were to follow the proposed naming convention then I guess I have to similarly modify all the nets to have both version of JK and without JK. Should I send another PR making the necessary changes or do want me to revert the naming convention of GCN and GraphSAGE?

rusty1s:
I think we just have to rename GIN to GINWithJK. I can do that.

ekagra-ranjan:
Isn't Graclus, topK and diffPool also having JK style of concatenating the node representations from each layer?

rusty1s:
I guess, but those are hierarchical models with a multi-scale style concatenation style. I think we can leave them as they are.

rusty1s:
Done :)

ekagra-ranjan:
Regarding topK and diffPool, I couldn't find any mention of the concatenation of node to enable multi-scale classification. Will it be okay to allow the benchmark to represent models which are better (due to multi-scale) than the model proposed by the authors in their papers?

rusty1s:
Actually, the use of JK concatenation is mentioned in the TopK paper. For DiffPool, you can find the use of concatenation in the official implementation.

in addition, for TopK pooling, authors actually perform both global mean and max pooling to obtain graph-level representations, but due to uniform network architectures, we avoided this in our benchmark scenario.

ekagra-ranjan:
Yes, you are right. I should have looked at this [paper](https://arxiv.org/pdf/1811.01287.pdf) instead of this [one](https://openreview.net/pdf?id=HJePRoAct7). The diffpool implementation does have the `concat` arg although they never mentioned in their paper. Thank for pointing it out!

PR #368, opened 2019-06-02
Remove the need to calculate x_min_value in topK

ekagra-ranjan:
The role of `x_min_value` was to prevent nodes not present in a graph from being selected in topK. This was done by setting the values in the positions outside the nodes in a graph as `x_min_value`-1. 

The PR applies `tanh` to score before passing to `topk` which bounds the scores between -1 and 1. Now by replacing `x_min_value` with -2 we achieve the same intended result.


rusty1s:
LGTM. Thank you.

PR #377, opened 2019-06-05
Make GAT example code consistent with paper's implementation

soumyasanyal:
In GAT, the last layer uses `concat=False`. Also, specifically for Pubmed dataset, the number of heads in out layer is set to 8. So, it'll be better to have this comment for easier reproducibility of Pubmed results.


rusty1s:
LGTM. Thank you.

PR #385, opened 2019-06-07
Fix error in Diffpool

ekagra-ranjan:
The PR fixes the ValueError due to missing variable during unpacking of `dense_diff_pool`


rusty1s:
Thanks :)

PR #390, opened 2019-06-08
Add `ratio` and `JK` as args to kernels

ekagra-ranjan:
This PR introduces `ratio` and `JK ` as args to various models in kernels.

### DIFF
--- a/benchmark/kernel/diff_pool.py
+++ b/benchmark/kernel/diff_pool.py
@@ -7,7 +7,7 @@ from torch_geometric.nn import DenseSAGEConv, dense_diff_pool, JumpingKnowledge
 
 
 class Block(torch.nn.Module):
-    def __init__(self, in_channels, hidden_channels, out_channels):
+    def __init__(self, in_channels, hidden_channels, out_channels, mode='cat'):
         super(Block, self).__init__()
 
         self.conv1 = DenseSAGEConv(in_channels, hidden_channels)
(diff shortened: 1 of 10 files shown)


rusty1s:
Thank you. I guess the following Linear layer after the JK layer needs adjustment to, whether one uses the concat method or not.

ekagra-ranjan:
I am on it!

ekagra-ranjan:
@rusty1s  Done! Please have a look at it.

rusty1s:
Hi and thank you, I'm actually not that big of a fan of allowing different JK modes for hierarchical models because we are already on the graph-level. For modes `max` and `lstm`, their big strengths are their node-adaptive aggregation, something which is already lost after global pooling. We should just use `cat` and shouldn't allow any modifications here. Other than that, it looks good to me :)

ekagra-ranjan:
Cool! When you asked me to make changes on linear layers after JK I though that `max` and `lstm` would be meaningful. But I guess you are right about their strengths in node-adaptive aggregations which might not make sense after global pooling. I will make the changes you suggested.

ekagra-ranjan:
Can we leave the `max` and `lstm` in the `Block` class in Diffpool or should I remove that too?

rusty1s:
You can leave it there.

ekagra-ranjan:
Done.

rusty1s:
Thank you :)

PR #399, opened 2019-06-11
update num_node_features

AnirudhDagar:
This fixes the mismatch between documentation and dataset.py.
Update dataset.py to match with the [tutorial](https://github.com/AnirudhDagar/pytorch_geometric/blame/4c05d9b86135139da24e9942ddb448b23c817f4a/docs/source/notes/introduction.rst#L139) on datasets where dataset.num_node_features replaced dataset.num_features.


rusty1s:
Thanks :)

PR #401, opened 2019-06-11
fix docs

AnirudhDagar:
This PR fixes some outdated documentation in `introduction.rst`.
1. Updated and fixed outputs to the code.
2. Passing `use_node_attr=True` as an arg in `TUDataset` when using `Enzymes` dataset to use the 21 features.


rusty1s:
Awesome :)

PR #403, opened 2019-06-12
Reduce compute of diff_pool

ekagra-ranjan:
Prevents `dense_diff_pool` to be computed for last block iteration.


rusty1s:
Thanks :)

PR #410, opened 2019-06-14
Remove size from reset_parameters

ekagra-ranjan:
This PR removes the arg `size` from `reset_parameters()` from models which are not doing any arithmetic operation on size while using them in `reset_parameters()`.


rusty1s:
Sounds reasonable. Thank you.

PR #423, opened 2019-06-18
add FeaStConv

sw-gong:
Please check if there is any issue. https://arxiv.org/pdf/1706.05206.pdf

I used translation invariant assignments which reported to have better performance instead of the formula (2).

### DIFF
--- a/test/nn/conv/test_feast_conv.py
+++ b/test/nn/conv/test_feast_conv.py
@@ -1,0 +1,13 @@
+import torch
+from torch_geometric.nn import FeaStConv
+
+
+def test_feast_conv():
+    in_channels, out_channels = (16, 32)
+    edge_index = torch.tensor([[0, 0, 0, 1, 2, 3], [1, 2, 3, 0, 0, 0]])
+    num_nodes = edge_index.max().item() + 1
+    x = torch.randn((num_nodes, in_channels))
+
+    conv = FeaStConv(in_channels, out_channels, heads=2)
+    assert conv.__repr__() == 'FeaStConv(16, 32, heads=2)'
+    assert conv(x, edge_index).size() == (num_nodes, out_channels)
--- a/torch_geometric/nn/conv/__init__.py
+++ b/torch_geometric/nn/conv/__init__.py
@@ -21,6 +21,7 @@ from .edge_conv import EdgeConv, DynamicEdgeConv
 from .x_conv import XConv
 from .ppf_conv import PPFConv
 from .hypergraph_conv import HypergraphConv
+from .feast_conv import FeaStConv
 
 __all__ = [
     'MessagePassing',
@@ -48,4 +49,5 @@ __all__ = [
     'XConv',
     'PPFConv',
     'HypergraphConv',
+    'FeaStConv'
 ]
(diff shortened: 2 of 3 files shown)


sw-gong:
I don't understand what these two codecov fails mean. Can you please explain about this? Thank you.

rusty1s:
Test Coverage decreased.

rusty1s:
If you do not have time, I can write a test for it.

sw-gong:
It finally passed. I guess I will know how to give a good PR next time. :-) Btw, please check the implementation with the paper. Thanks.

rusty1s:
For linting, I do autoformatting with`yapf`, and run `python setup.py test` before comitting. We should probably create a `CONTRIBUTE.md` for all other contributors.

rusty1s:
Will check the code by tomorrow. Thanks again for your awesome contributions :)

sw-gong:
> For linting, I do autoformatting with`yapf`, and run `python setup.py test` before comitting. We should probably create a `CONTRIBUTE.md` for all other contributors.

Thank you. I'll have a look it.

rusty1s:
Hi, just some minor questions regarding this line: `q = torch.mm((x_i - x_j), self.u) + self.c`.

The paper states that it computes `q` by two parameters ` u` and `v` instead of just one. In addition, we do not really need `matmul` here, a basic `(x_j * self.u).sum(dim=1)` should be way faster (as far as I know).

sw-gong:
> Hi, just some minor questions regarding this line: `q = torch.mm((x_i - x_j), self.u) + self.c`.
> 
> The paper states that it computes `q` by two parameters ` u` and `v` instead of just one. In addition, we do not really need `matmul` here, a basic `(x_j * self.u).sum(dim=1)` should be way faster (as far as I know).

Hi, as I said at the beginning I used  translation invariant assignments which reported to have better performance instead of the formula (2) in the paper. As for using dot product replacing torch.matmul, I agree with you. 

image
image

rusty1s:
Ah, I'm sorry, you are right. I should read more closely :P

sw-gong:
No worries. :) 

Below is the repository posted by the paper author. I'm not familiar with tf but maybe you can have a look at it if you want. I ran experiments on FAUST with the same architecture shown in the paper (don't know hyperparameter setup), the accuracy of which is around 8-10% lower than reported 92.2%. 
https://github.com/nitika-verma/FeaStNet

dhorka:
I have one doubt related to this implementation. I was checking the paper I do not see where they use attention heads and how many they are using.

rusty1s:
Well, this denotes the number of weight matrices the convolution will use. A more appropriate name would be indeed `kernel_size`. I can change the naming. WDYT @sw-gong?

dhorka:
Then the output of the layer it will be weight_matrices*out_channels? I'm a bit confusing about that.

rusty1s:
No, the output will be `[num_nodes, out_channels]`. We sum up the results of all weight matrices, see [here](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.FeaStConv).

dhorka:
Then maybe it is not appropriate kernel_size, right? Normally kernel_size is used to define the filter of a CNN, in this case we have multiple filters. Maybe the nomenclature of the paper can be preserved and use M.

rusty1s:
If I am not totally mistaken, `kernel_size` is the appropriate name. In a traditional CNN, you also have multiple filters, with each kernel point having `[in_channels, out_channels]` weights. This is also discussed in the paper:
Screenshot 2019-07-22 at 09 44 43

PR #425, opened 2019-06-18
Update isolated.py

yutaizhou:
Some connected vertex may be contained in col only and not row in line 19, so they may get lost in the count during the unique() operation. The change ensures all nodes are accounted for.


rusty1s:
Yes, you are right. Thank you :)

PR #426, opened 2019-06-18
Update data.py

yutaizhou:
Description does not reflect intention of method name.


yutaizhou:
Please do update the doc as well!

rusty1s:
Sure!

PR #432, opened 2019-06-19
Minor optimization in TopK Pool

ekagra-ranjan:
Replaces `len(num_nodes)` with already computed `batch_size`.


rusty1s:
Thank you.

PR #450, opened 2019-06-25
Closed previous PR. Didn't import torch. Now it is fixed.

yutaizhou:
Original implementation assumes the edge index belongs to a graph that is not an induced subgraph. New version returns the number of unique node entries in edge_index, rather than the max node label + 1.

### DIFF
--- a/torch_geometric/utils/num_nodes.py
+++ b/torch_geometric/utils/num_nodes.py
@@ -1,2 +1,4 @@
+import torch
+
 def maybe_num_nodes(index, num_nodes=None):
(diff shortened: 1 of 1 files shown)


rusty1s:
I cannot accept this PR. The number of nodes still need to include isolated nodes.

yutaizhou:
Gotcha. After closer inspect, I realize that I did not understand the intended use. I must have mistaken the function as something a user can call, rather than for internal use only. Sorry!

rusty1s:
No problem, thank you for all your contributions :)

PR #472, opened 2019-07-03
Add graphConv to SAGPool

ekagra-ranjan:
There was a typo with the GCN option in the SAGPool. Instead of using GCN it was using GraphConv (as the default conv in benchmark currently is GraphConv). This PR fixes the above and also adds GraphConv as a separate option and makes it default to maintain consistency with the kernel benchmarking. 

I wasn't able to find a short form for GraphConv as  GCN is already taken :) so I have used gConv as the abbreviation. If we have to use 3 letter then GRC could be a possiblity. Please change the abbreviation if you have a better naming for GraphConv and do let me know too!


rusty1s:
Thank you :)

PR #475, opened 2019-07-03
Add Readout Pooling

ekagra-ranjan:
This PR adds the readout pooling from [Towards Sparse Hierarchical Graph Classifiers](https://arxiv.org/abs/1811.01287) to PyG. 

The following methods have been added:
1) `global_readout()` has been added to `glob/glob.py`.
2) `readout_x` and `readout` has been added to a new file `pool/readout.py`
3) tests and docs have also been added in the respective files.

### DIFF
--- a/test/nn/glob/test_glob.py
+++ b/test/nn/glob/test_glob.py
@@ -27,5 +27,3 @@ def test_global_pool():
     assert out.size() == (2, 8)
     assert out[0].tolist() == torch.cat((x[:4].mean(dim=0), x[:4].max(dim=0)[0]), dim=-1).tolist()
     assert out[1].tolist() == torch.cat((x[4:].mean(dim=0), x[4:].max(dim=0)[0]), dim=-1).tolist()
-
-test_global_pool()
--- a/test/nn/glob/test_glob.py
+++ b/test/nn/glob/test_glob.py
@@ -1,6 +1,6 @@
 import torch
 from torch_geometric.nn import (global_add_pool, global_mean_pool,
-                                global_max_pool)
+                                global_max_pool, global_readout)
 
 
 def test_global_pool():
@@ -22,3 +22,10 @@ def test_global_pool():
     assert out.size() == (2, 4)
     assert out[0].tolist() == x[:4].max(dim=0)[0].tolist()
     assert out[1].tolist() == x[4:].max(dim=0)[0].tolist()
+
+    out = global_readout(x, batch)
+    assert out.size() == (2, 8)
+    assert out[0].tolist() == torch.cat((x[:4].mean(dim=0), x[:4].max(dim=0)[0]), dim=-1).tolist()
+    assert out[1].tolist() == torch.cat((x[4:].mean(dim=0), x[4:].max(dim=0)[0]), dim=-1).tolist()
+
+test_global_pool()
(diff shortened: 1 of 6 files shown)


rusty1s:
Thank you for your contribution. I'm wondering if we really need our own operator for this kind of readout function. In the end, it can be implemented as a one-liner in a neural network architecture:
```python
x = torch.cat([global_mean_pool(x, batch), global_max_pool(x, batch], dim=1)
```
In addition, this operator is kind of arbitrary. If adding this, we could also add any combination of other readout functions. If you like to proceed pushing this to PyG, then please
1. Use `global_mean_pool` and `global_max_pool` calls in its implementation
2. Change the name to something more meaningful, *e.g.* `global_mean_and_max_pool`

ekagra-ranjan:
I agree that there can many combination of such readouts but I feel this combination might become a standard one as sparse pooling papers like [Towards Sparse Hierarchical Graph Classifiers](https://arxiv.org/abs/1811.01287) and [Self-Attention Graph Pooling](https://arxiv.org/pdf/1904.08082.pdf) have used during their benchmarking.

ekagra-ranjan:
At the moment I am only aware of this combination. Are you aware of any other combination of readout layer in GNN literature? If so then please do let me know :)

If there arent, I was thinking that maybe we can let it be named as `readout` until a new popular version comes to literature after which we can make readout a generic function which can generate a combination based on arguments passed to this `readout` function. What do you think about this?

rusty1s:
Well there are a dozen of readout layers in literature (`mean`, `add`, `max`, `set2set`, `sort_pool`, `perm-order`, any variant of global attention, ...). That's why I do not like its current generic name `global_readout`. In addition, I wouldn't call it a standard. Yes, it is used in the Top-k pooling paper (with SAGPooling just being a minor variant of it), but nowhere else. In addition, it is far from a novel contribution in these papers.

I generally have no problem with this function or adding more readout functions to PyG. In my own experience, I found it often benefical to combine multiple readout functions. It is just that I think that adding this function to PyG has no real benefit (because it is a one-liner anyway) and it is too specific (*e.g.* what about a combination of global attention and sum pooling?). In the end, we provide all necessary utilities to implement these combined readout functions. I do not think it is necessary to add a function to PyG which basically just concatenates two feature vectors. What do you think?

ekagra-ranjan:
You are right about the it. I haven't played with much of readout variants so I wasn't able to gauge the uniqueness of the proposed variant. 

Anyways, in your experience did you find any pattern when to use what type of combination? How do they depend on the dataset? Please share any pointers or intuition you have gained while experimenting with multiple readout function :)

rusty1s:
This highly depends on the dataset. While I trained the benchmark models for the PyG paper, I noticed that the results from the reported results of top-k pooling are mostly due to using a combination of global mean and global max pooling, and not actually due to the top-k pooling operator (results are much worse when only using mean pooling, which makes comparison to other approaches quite unfair IMO).

I generally test a number of readout functions and evaluate their effect, *e.g.*, if one needs to model injective functions one should use sum pooling (like mentioned in the GIN paper), if one wants to attend to only specific nodes one should use (soft)max pooling, etc. It is therefore not surprising that a combination of these readout functions often may improve results due to injecting different inductive biases into the model.

ekagra-ranjan:
Thanks for your review and insights. I will be closing this PR then.

PR #483, opened 2019-07-04
Allowing for permuted batch references for global pool operations

fdiehl:
At the moment, all `global_x_pool` layers assume (if no explicit `size` is given) that the last entry in `batch` + 1 is the number of all batches. While holds for the examples, it might not hold in a general case (in particular, pooling operations might not follow that convention).

See the new `test_permuted_global_pool` test in `test/nn/glob/test_glob.py` for an example.

The disadvantage of this is that we have to compute an additional `max` operation on the batch; however, the actual scatter operation greatly dominates anyway.

Note: The new test uses `pytest.approx` for checking the computation; since the order of operation varies, mean and add pooling run into float issues otherwise. I've used the same for max pooling for symmetry.

### DIFF
--- a/test/nn/glob/test_glob.py
+++ b/test/nn/glob/test_glob.py
@@ -1,4 +1,5 @@
 import torch
+import pytest
 from torch_geometric.nn import (global_add_pool, global_mean_pool,
                                 global_max_pool)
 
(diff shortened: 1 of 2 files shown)


rusty1s:
Yes, we should do this to be safe. Thank you :)

PR #484, opened 2019-07-04
Synthetic COLORS and TRIANGLES datasets and Threshold-based Top-K and SAG pooling

bknyaz:
Hi Matthias,

In this PR I'm adding: 
1. An extension of Top-K and SAG pooling that we used in our ICLR Workshop paper "Understanding Attention and Generalization in Graph Neural Networks". I feel that Top-K and SAG should be combined into a single class, that will have some neural function to predict attention scores as an argument.  Similarly what you have for GINConv. I decided not to create a separate class for our pooling method because it's an extension of both.
2. Synthetic dataset class that is inherited from TUDataset. The datasets are from our paper. I prepared them in the same format as the TU datasets and uploaded to my github. If I could upload it to the TU server, then it would eliminate the need to create this Synthetic dataset class at all.
I had to change the tu.py to support reading node attention.
3. Examples to train on our COLORS and TRIANGLES datasets.
4. Information about our paper to README.
5. A fix for a small bug in tu.py when reading edge_labels.

Hope you can merge it.
Cheers,
Boris


rusty1s:
Thank you very much for this awesome contribution.
1. I agree with you, but we can always do this later. Feel free to make this change in another PR. Otherwise, I can implement it.
2. If you do not mind, I can ask @chrsmrrs to upload the files to the TU server. I do not understand the need for the specific node attention file though. You can rename your `{}_node_attention.txt` to `{}_node_labels.txt` to save them as a one hot vector to `data.x`, or to `{}_node_attributes.txt` to save them as a float values to `data.x`. The normalization and renaming can still happen in a custom transform, *e.g.*:
```python
class HandleNodeAttention(object):
    def __call__(self, data):
        data.node_attention = torch.softmax(data.x, dim=0)
        return data
```
I think this approach is a little bit cleaner than actually modifying the TU format. WDYT?
5. Ups and thank you :)

I will try to merge this PR at the weekend. Thank you very much again.

bknyaz:
Thank you for a great suggestion regarding handling attention. I didn't realize that I could do this using a custom transform class. So, I think using `{}_node_attributes.txt` to keep both node features and node attention is the best for now. I hope to commit these update to my PR today.

For `COLORS` you are generally correct, however, in the last most challenging test subset we use mixtures of colors, so features are not one-hot, i.e. they can be like `[1 0 1 0]`. So I cannot use  `{}_node_labels.txt` :(

I wanted to introduce the `{}_node_attention.txt` file hoping that in future there will be more datasets with some form of attention labels for nodes/edges. But I guess now it's not worth it to add this new file to the TU format. Also I it would be nice to see the `Node attn` column in the table at https://ls11-www.cs.tu-dortmund.de/staff/morris/graphkerneldatasets. I used the statistics in that table a lot to decide what task and what dataset to try. So I thought that once people see `Node attn`, then they will have more information to decide. These are just my thoughts, it's up to you to decide :)

bknyaz:
Ok, so I made another couple of commits to this PR adding HandleNodeAttention.

PR #486, opened 2019-07-05
test_permuted_global_pool: Fixed seldom-occurring (~1% of cases) assert failure

fdiehl:
The test I added yesterday failed sometimes due to numerical issues in the order of operations between the permuted and non-permuted data - I should have checked that more. Your new test fails less, but still does so in ~1% of the cases.

I've therefore updated the test to use the same order for both the global_pool operations and the operation directly on the tensor. This makes the test somewhat more ugly (since it mostly replicated the same tests as `test_global_pool`, contrary to your changes in 14b628e18fdefc250915ddc7b7b39f7880dad6cf); however, the asserts will not fail (tested with 100k iterations; my first test failed in ~10% of cases; your updated one in ~1%; this one doesn't).


rusty1s:
Cool :)

PR #509, opened 2019-07-10
Implementation of GMMConv2 layer

arcus78:
Implementation of GMMConv2 layer with separate GMMs for every pair of input and output (CNN style)


rusty1s:
Thank you :)

PR #536, opened 2019-07-17
support property of num_class

sw-gong:
To make data and dataset consistent


rusty1s:
I am actually not a big fan of this because that should only work for node classification tasks. Given a single class for a label, this can result in unexpected errors.

sw-gong:
Yes, that's the case. Okay. It's fine. Let's close it.

PR #558, opened 2019-07-25
Fix for ChebConv normalisation

Stannislav:
As discussed in https://github.com/rusty1s/pytorch_geometric/issues/530 I've implemented the proper normalisation for the graph Laplacian in the `ChebConv`.

- The computation of the graph Laplacian was moved to `utils`.
- A transformer named `LaplacianLambdaMax` was added to compute the maximal eigenvalue of the graph Laplacian.
- The eigenvalue is stored in the attribute `lambda_max` of the `Data` object.


rusty1s:
Awesome! I will have a closer look during the weekend and will merge ASAP.

rusty1s:
No problem and thank you once again. I cleaned up the code a bit and also allowed different normalizations in the `ChebConv`. For eigenvalue computation, I just use the `eigs` method for any normalization, since I have noticed that `eigsh` can result in large fluctations for `lambda_max`.

Stannislav:
> No problem and thank you once again. I cleaned up the code a bit and also allowed different normalizations in the `ChebConv`. For eigenvalue computation, I just use the `eigs` method for any normalization, since I have noticed that `eigsh` can result in large fluctations for `lambda_max` (although I am not sure why).

The only reason to use `eigsh` would be the performance since it's much faster than `eigs` because it's specialised to hermitian / symmetric matrices, while `eigs` works for any square matrix. The result should be very similar either way. Regarding the fluctuations, are you sure you tested on symmetric matrices?

rusty1s:
Yes, I am fairly sure. `eigsh` yields largest eigenvalues of `1.86`, `2.35` and `2.35` on three runs on the matrix:
```
 1 -0.5  0
-1  1   -0.5
 0 -1    1
```

rusty1s:
You were right, I calculated it with directed edge weights, leading to a non-symmetric laplacian, uff. Fixed in master!

Stannislav:
> You were right, I calculated the symmetric Laplacian with directed edge weights, leading to a non-symmetric matrix, uff. Fixed in master!

Just got to checking the changes in `laplacian_lambda_max.py`, and I'm not sure it'll always be doing the correct thing.

So, `eigsh` will only work when the Laplacian is a symmetric matrix, which is the case when **all** of the following are true:
1. The adjacency matrix `A` (represented by `edge_index`, `edge_weight`) is symmetric, i.e. `A^T = A`. This is equivalent to the graph being undirected.
2. The normalisation is either
    1. `None`, then `L^T = A^T -  D^T = A - D = L`, or 
    2. `sym`, then `L^T = 1^T - (D^{-1/2}AD^{-1/2})^T = 1 - D^{-1/2}AD^{-1/2} = L`.

Right now it seems it's only checked if the normalisation is of type `sym`.

rusty1s:
Yes, you are right. I wonder if it is really necessary to check (1) because the Laplacian is only defined for undirected graphs, isn‘t it? This check might be even more expensive than just computing the eigenvalue via `eigh`.

Stannislav:
I think mathematically it's true that graph Laplacians only make sense for undirected graphs, but from experience I know that spectral convolutions actually do work fine on dircted graphs too, so I wouldn't exclude that other's try that as well...

But I agree that checking if `A` is symmetric might be quite expensive, and sometime take longer thatn `eigs`. On the other hand, for large graphs `eigs` can be quite slow, while `eigsh` would finish in reasonable time.

Another option might be to have yet another parameter `is_undirected=False`, which makes the function default to `eigs` and works for all cases. In cases where the user knows that the graph is undirected they could set `is_directed=True`.

rusty1s:
That is a good idea. I will implement it ASAP.

rusty1s:
Fixed in master. Thank you for your valuable input :)

PR #565, opened 2019-07-27
node2vec

b4rthm:
node2vec example and model

### DIFF
--- a/examples/node2vec.py
+++ b/examples/node2vec.py
@@ -1,0 +1,116 @@
+import math
+import os.path as osp
+import torch
+import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


vivald99:
Hi @b4rthm and @rusty1s . I'm interested in the structural similarity between nodes (for example, capturing roles that nodes play in a social network graph). Is there some plan to integrate other node embeddings methods to PyG, such as struc2vec (https://arxiv.org/pdf/1704.03165.pdf)?

rusty1s:
Not yet, but please feel free to request that feature in a new issue so we can keep track of it there.

PR #571, opened 2019-07-29
Add file existence check before download.

wlhgtc:
Some dataset(like Reddit) maybe too large. For some reason, the speed will be low.
Add the condition check before download the dataset, so that we could download it before.


rusty1s:
Looks good to me. Thank you :)

wlhgtc:
my pleasure :)

PR #573, opened 2019-07-29
ENH: Add initial unet.

### LINKED ISSUE #37: Graph U-Net Implementation
Sorry for incomplete suggestion. 
Is it possible to include gUnpool (graph un-pooling layer) described in graph U-Net paper (https://openreview.net/pdf?id=HJePRoAct7)

YannDubs:
Initial Graph Unet (closes #37). Notes:

* I currently unpack 5 arguments from the pooling layer, as this is the number of arguments I get from the last version on pip. The documentation and code on master shows that there should be 6. 

* Any concolution layer can be given as long as they take `x, edge_index, edge_attr` as argument. This includes many layers but not GIN and GAT, the call should be around a try-catch-rerun  without `edge_attr` such that these layers can be used.

* `is_double_conv` and `Normalization` are not used in the graph unet paper but they are in the CNN Unet, I kept them for comparison, but maybe the final code should remove those (depends on the benchmarks). 

* The paper also doesn't use `max_nchannels`, `factor_chan`. I think the first argument should be kept as the number of channels grows exponentially with the layers.

* `_get_in_out_channels` is very dirty and hopefully there's a clean but still a general way of computing the number of channels.

* I opened [an issue](https://github.com/HongyangGao/gunet/issues/7) in the official code to understand how exactly they use the adjacency augmentation. I think the way I did is the most natural. Another possibility would be to combine the stored adjacency matrix A^l and the current one A^{l+1} when unpooling.

* sorry for the lack of documentation, tests and examples, I'm procrastinating my master thesis ;). Feel free to ask if something is not clear.

Thanks again for the awesome library 💯


rusty1s:
Ok, I included an example which can reach 83% accuracy on Cora (with a bit of luck :P).

I also cleaned up the model implementation, and we can add some functionalities back in later. I think the current model makes it easier to test the current implementation:
1. Allow different `pool_ratios` for different depths.
2. Allow different conv layers than `GCNConv`.

I removed the `normalization` argument, `is_double_conv`, and `max_nchannels` argument. The number of channels does not really grow exponentially as long as your output channels stay the same.
I think the other flag will become obsolete as soon as we allow custom convolutions.

rusty1s:
Added documentation and tests. I think we can merge this now!

torch_geometric/nn/models/graph_unet.py
```diff
@@ -0,0 +1,113 @@
...
+
+class GraphUNet(torch.nn.Module):
+    r"""The Graph U-Net model from the `"Graph U-Nets"
+    <https://arxiv.org/abs/1905.05178>`_ paper which implements a U-Net graph
+    architecture with graph pooling and unpooling operations.
+
+    Args:
+        channels (int): Size of each sample.
+        depth (int): The depth of the U-Net architecture.
+        kernel_size (int or [int]): Size of the convolving kernel.
```
YannDubs: Am I missing something ? This argument doesn't seem to be used

YannDubs: Maybe due to a previous copy paste of docs ?

YannDubs:
Nice! I was very focused on the computer vision Unet which explains all the additional arguments.
I removed the `kernel_size` argument that was not used (probably a copy paste mistake? @rusty1s).

Out of interest, when you said in the open issue that you were not able to reproduce the results (although they are still not as good), was the only missing component the squared adjacency augmentation instead of adding two hops ? 

Concerning the results, it [seems](https://github.com/HongyangGao/gunet/issues/6) that you cannot reproduce the results even using the official hyperparameters. The author says "[...] select the best average performance for 10-fold experiments.". The use of the word **best** in this context is confusing, does it mean that they ran multiple 10 fold experiments and cherry-picked? I hope I'm misunderstanding something here. In any case, I still think it's a useful model :) 

Thanks for answering and doing all of that so quickly!

torch_geometric/nn/models/graph_unet.py
```diff
@@ -0,0 +1,112 @@
...
+            batch = edge_index.new_zeros(x.size(0))
+        edge_weight = x.new_ones(edge_index.size(1))
+
+        xs = [x]
+        edge_indices = [edge_index]
+        edge_weights = [edge_weight]
+        perms = []
+
+        for i in range(self.depth):
+            edge_index, edge_weight = self.augment_adj(edge_index, edge_weight,
```
YannDubs: why are you augmenting and pooling before the first convolution ? I don't think this is desirable. That's nto what the paper does right ?

torch_geometric/nn/models/graph_unet.py
```diff
@@ -0,0 +1,112 @@
...
+        self.depth = depth
+        self.pool_ratios = repeat(pool_ratios, depth)
+
+        self.act = act
+        self.sum_res = sum_res
+
+        self.pools = torch.nn.ModuleList()
+        self.down_convs = torch.nn.ModuleList()
+        for i in range(depth):
+            self.pools.append(TopKPooling(channels, self.pool_ratios[i]))
```
YannDubs: Number of pooling should be depth-1, I think the issue is that you are ppolign before the first convolution.

torch_geometric/nn/models/graph_unet.py
```diff
@@ -0,0 +1,112 @@
...
+            res = xs[j]
+            edge_index = edge_indices[j]
+            edge_weight = edge_weights[j]
+            perm = perms[j]
+
+            up = torch.zeros_like(res)
+            up[perm] = x
+            x = res + up if self.sum_res else torch.cat((res, up), dim=-1)
+
+            if i < self.depth - 1:
```
YannDubs: I don't think this is correct: you need to finish on a convolution! Right now if you are concatenaing, the output will be of dim*2 and not dim.

YannDubs:
In all the reviews, I was thinking of a Unet as an end to end model. From your example, I now realize that you see it as an intermediate layer that requires an initial and final convolution. 
I think both ways of doing are fine, but personally I feel that `GraphUNet` should be the whole unet  as it is called so (when I use such a module I naturally think that I get something similar to figure in the paper).

What are the reasons you decided to start and finish with pool / unpool ? Any advantages I have not though of ? If it's due to the different number of input / output channels, I think it makes more sense to either:
* not care about it, and the user will use a linear layer for changing the feature dimensions
* give `in_channels`, `in_channels`, `hid_channels` as parameters and take care of this internally 
* let `channels` be a list, in which case the user could give the different channels. Athough it would be annoying as the user would type `[in_channels, hid_channels, ..., hid_channels, out_channels ]`

torch_geometric/nn/models/graph_unet.py
rusty1s: Yes, thank you :)

torch_geometric/nn/models/graph_unet.py
rusty1s: It is: "Dropout is applied to both adjacency matrices and feature matrices with keep rates of 0.8 and 0.08, respectively."

rusty1s:
I see the point. The reason why I implemented it this way is because authors talk about embedding and final prediction layers independent from the model. In the end, this does not really matter too much. We can change this back using `in_channels, hidden_channels, out_channels` as input arguments.

eurunuela:
Thank you @YannDubs and @rusty1s for this PR. It has been quite the life-saver for me as I first feared I would have to implement the whole thing myself.

However, I do not understand why https://github.com/pyg-team/pytorch_geometric/pull/573/commits/c4c787ba6a03ab9b6d9341ea7d854c6f90238ceb made this change here: https://github.com/pyg-team/pytorch_geometric/blob/9b794b600d41802edaad26aac18bff958fc8f642/torch_geometric/nn/models/graph_unet.py#L64

Effectively, you made the number of channels constant for all the layers, when one would probably want the number of channels to increase in the encoder and decrease in the decoder. I see that was the initial implementation in this PR until pretty much the last commit.

Could you please explain why the change was made? Thanks!

PR #584, opened 2019-07-31
Added_obj_file read functionality

pekopekoda:
I would like to add a feature to load *.obj files along other formats like *.ply for example


rusty1s:
Looks great. Thank you :)

PR #587, opened 2019-08-01
Add an example of dgcnnV2 segmentation with multiple GPUs

cy69855522:
I made a simple test, 4 batch size, 30 epochs, Mean Class IOU was 77.57, the original paper was 82.3375, the setting of the original paper is: "We use Adam with learning rate 0.001 belong means by 2 every 20 epochs. The decay rate for batch normalization is initially 0.5 and 0.99 ﬁnally. The batch size is 32 and The momentum is 0.9. ". I guess the gap is caused by batch size and insufficient epochs, but it is within a reasonable range.


rusty1s:
Thank you. This looks good so far. However, I am currently on vacation and can't test the implementation, so the merge will be delayed a bit. Sorry!

cy69855522:
Sorry to disturb your holiday, I am doing further test, which still needs several days. So don't worry, have a good time~

cy69855522:
BN→GN，batchsize32→4，instance iou 85.2→83，epoch ？→100. Emmmmmm, there is still a gap of about 2%. The GPU memory is not enough to fully reproduce.

cy69855522:
Hi, I find “FixedPoints” function may produce same points. This approach is different from the original paper and may have bad effects, so I want to add an optional parameter.

PR #596, opened 2019-08-05
aggr argument in EdgeConv

flgw:
The parameter can't be passed because it is hard-coded in the call to __init__ of the base class.


rusty1s:
Perfect, thank you!

PR #605, opened 2019-08-07
accelarate geodesic distance computation in the case of whole mesh

sw-gong:
Related to https://github.com/rusty1s/pytorch_geometric/issues/593#issuecomment-518317120


rusty1s:
Perfect, thank you.

PR #622, opened 2019-08-14
Create negative_sampling utils

dongkwan-kim:
- Caught errors in negative_sampling, negative_triangle_sampling
- Create a batch version of negative_sampling
- Put all three in one utils/negative_sampling module.
- Add tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ If you are interested in contributing to PyTorch Geometric, your contributions w
 1. You want to implement a new feature:
     - In general, we accept any features as long as they fit the scope of this package. If you are unsure about this or need help on the design/implementation of your feature, post about it in an issue.
 2. You want to fix a bug:
-    - Feel free to send a Pull Request any time you encounter a bug. Please provide a clear and concise decription of what the bug was. If you are unsure about if this is a bug at all or how to fix, post about it in an issue.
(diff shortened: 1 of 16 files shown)


rusty1s:
Looks really good, will merge ASAP.

rusty1s:
I think this PR is now ready for merge. I removed the `is_directed` flag because in PyG we implement undirected graphs by encoding both directions in `edge_index`. WDYT?

dongkwan-kim:
> I think this PR is now ready for merge. I removed the `is_directed` flag because in PyG we implement undirected graphs by encoding both directions in `edge_index`. WDYT?

Oh, I see. Then, I am ok with that. Thank you for nice refactoring.

But, I found a mistake in line 29 of this commit (https://github.com/rusty1s/pytorch_geometric/pull/622/commits/6021aee3f3845627d8a817cb8b4784fa67bfc367#diff-ac3d66c32da6eab12a976416c535a12e). I think that edge_index.size(0) should be edge_index.size(1), since edge_index is a tensor the shape of which is [2, num_edges]. Could you look at this?

rusty1s:
Yes, thank you!

dongkwan-kim:
Hi, @rusty1s 

A few months ago, we agreed to remove `is_directed` for the input edge_index.
However, sometimes, we might need an output of undirected negative samples.

What do you think about adding `force_undirected` (bool, optional) to input attributes, like `dropout_adj`?

rusty1s:
Yes, I think we can do this by only working on the upper triangle adjacency matrix.

dongkwan-kim:
OK. I will send a pull request about this soon.

PR #626, opened 2019-08-15
fix docs typo

cuent:


rusty1s:
Thank you, but I do not think this is correct. D_ii is a scalar, so it should not be printed in bold.

cuent:
Ohh yeah, I will fix that. Just a sec.

rusty1s:
Thanks :)

PR #677, opened 2019-09-08
fix edge pool

zizai:
make num_nodes explicit for the edge score method. otherwise, it infers the wrong num_nodes when there are singletons.


rusty1s:
Thank you!

PR #724, opened 2019-10-07
Fixed bug in distance transform

OliverBScott:
PR fixes a bug in the distance transform when using norm=True and the max_value argument


rusty1s:
Great. Thank you!

PR #732, opened 2019-10-14
Implements CGCNNConv from Xie and Grossman (2018)

abhshkdz:
Implements the Crystal Graph Convolutional operator from https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.120.145301


rusty1s:
Thank you very much :)

PR #760, opened 2019-10-25
Add graph datasets without isomorphic copies.

nd7141:
This change adds additional graph data sets, where for each data set there are no isomorphic copies of graphs available in the original data set. 

The original data sets come from TU-Dortmund repository (https://ls11-www.cs.tu-dortmund.de/staff/morris/graphkerneldatasets). New, cleaned data sets come from this repository (https://github.com/nd7141/graph_datasets), the format of which is exactly the same as from the original repository, so the change is minimal for Pytorch-Geometric to adopt these data sets in addition to the original as a separate class, which is called `NonisoDataset` (name is up to modification). In the future, one can easily export a graph data set by using conventional commands: 

    root = osp.join('/', 'tmp', str(random.randrange(sys.maxsize)))
    dataset = NonisoDataset(root, 'ENZYMES')

Two tests, `test_noniso_enzymes.py` and `test_noniso_enzymes.py` are added to the repo to verify correct computation of graphs from the data sets.


rusty1s:
Hi and thank you very much for your valuable work.

Actually, I am not a big fan of creating a new dataset class. Instead, it would be more reasonable to add a `cleaned` flag to the `TUDataset`, and if checked, download the required data from `https://github.com/nd7141/graph_datasets`. What do you think?

nd7141:
Hey, 

Yes, indeed it would be cleaner and take less code. 

The only thing that could go wrong in the future is that if I will add more clean data sets to https://github.com/nd7141/graph_datasets which will not exist in the original data sets. But in that case either we can ask to duplicate them to the original list or handle such data sets in pytorch-geometric manually. 

I will update review.

rusty1s:
Sounds like a good solution. Thank you very much.

nd7141:
I updated the code and ran the tests (OK). It now has a single class TUDataset with 2 urls inside. It works fine. 

However, I have one concern. When we provide download a new data set, it works fine. 

For example: 

    root = osp.join('/', 'tmp', str(random.randrange(sys.maxsize)))
    dataset = TUDataset(root, 'ENZYMES')
    >>> ENZYMES(600) # correct

However, if we add after dataset with clean version with the same root, it will not download the clean version of dataset but instead will use the previous one. So: 

    dataset = TUDataset(root, 'ENZYMES', cleaned=True).
    >>> ENZYMES(600) # but should be ENZYMES(595)

In fact, this is a bigger problem, which can cause confusion to new users. For example consider: 

    root = './'
    dataset = TUDataset(root, 'ENZYMES')
    print(dataset)
    >>> ENZYMES(600) # correct

    dataset = TUDataset(root, 'MUTAG')
    print(dataset)
    >>> MUTAG(600) # wrong, reused previous data set 

From what I understood, this is due to `__init__` method in `Dataset` class: 
    self.raw_dir = osp.join(self.root, 'raw')

which should be 
    self.raw_dir = osp.join(self.root + self.name, 'raw')

to store each data set under its name. However, I need to double check. 

What do you think? Do you think it's the reason why the example with MUTAG above fails?

rusty1s:
Hi, I added the ability to change the `raw_dir` and `processed_dir` attributes. This way, the `TUDataset` now creates either a `raw` directory or a `raw_cleaned` directory, depending on the state of `cleaned`. This way
```python
TUDataset(..., cleaned=False)
TUDataset(..., cleaned=True)
```
will load the correct files. Thanks once again :)

Issue #767, opened 2019-10-28
GINConv lacks edge_index


Guilherme26:
## 🐛 Bug

GINConv requires an edge_index to perform the forward operations with the network received as a parameter by the GINConv constructor

## To Reproduce

1. Instantiate a GINConv object passing any PyTorch Geometric neural model;
2. Try to perform a **forward** operation;

## Environment

* OS: Ubuntu 18.04.02 LTS
* Python version: Python 3.6.8
* PyTorch version: 1.2.0
* CUDA/cuDNN version: No CUDA
* GCC version: GCC 7.4.0
* Any other relevant information: *

Guilherme26:
Solved in #768 (waiting for acceptance)

PR #768, opened 2019-10-28
Adding edge_index to GIN's forward function

Guilherme26:
Solving problem reported in #767


rusty1s:
Hi,

I do not really understand this fix. `self.nn` just works on the node features by definition (and should be implemented via an MLP). I see no reason why one wants to pass a GNN model inside the GIN operator. Can you clarify?

PR #787, opened 2019-11-06
Fix typos in create_dataset.rst

kamilest:
Adjust indentation to four spaces
Change `ops.join` to `osp.join`
Add matching parenthesis


rusty1s:
Great. Thank you!

PR #820, opened 2019-11-21
Add GDC

klicperajo:
This PR implements the full pipeline used by Graph Diffusion Convolution (GDC) from "Diffusion Improves Graph Learning" (NeurIPS 2019) as a preprocessing step (data transformation).

GDC is monolithically implemented in a single class. Not sure if it would be smarter to split up the separate steps and then just call them in a wrapping GDC function. But I think this is a good (and complete) start.

I've also added a fast and scalable implementation of PPR for this. This implementation requires Numba, so I've added Numba to the required packages. I'm open to better ways of handling this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ alpha/
 *.egg-info/
 .coverage
 .coverage.*
+.vscode
 
 !torch_geometric/data/
 !test/data/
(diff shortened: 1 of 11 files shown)


rusty1s:
Thank you very much! This already looks really good. I will try to merge this ASAP.

klicperajo:
Great! By the way, is there an easy way provided by PyG to obtain the largest connected component (LCC) of a graph? That would be very useful.

klicperajo:
Should I just extend `gcn.py` with GDC or write an additional file, based on `gcn.py`?

rusty1s:
I believe the easiest way is to convert your graph to `networkx`. We do not have our own implementation for this.

I think adding an argument to `gcn.py` with `argparse` should be sufficient, but this is up to you :)

klicperajo:
Done. I've also fixed the GCN normalization while I was at it.

klicperajo:
Should the documentation or something point to where this example is?

rusty1s:
Sure, we could add a link to it to let users know about it :)

klicperajo:
Not sure if this will work, but I've added a link.

rusty1s:
Works perfectly! i just fixed some linting problems. Thank you once again!

Oliver-0423:
is there an easy way provided by PyG to obtain the largest connected component (LCC) of a graph now?

PR #842, opened 2019-12-02
Remove compression metadata from ModelNet10 data.

SergioRAgostinho:
My dataset folder was looking like this:
```
$ ls modelnet10/
__MACOSX  processed  raw
```

Now the compression artefacts are removed after extraction. The folder only exists for ModelNet10.

### DIFF
--- a/torch_geometric/datasets/modelnet.py
+++ b/torch_geometric/datasets/modelnet.py
@@ -78,6 +78,11 @@ class ModelNet(InMemoryDataset):
         shutil.rmtree(self.raw_dir)
         os.rename(folder, self.raw_dir)
 
+        # Delete osx metadata generated during compression of ModelNet10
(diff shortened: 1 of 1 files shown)


rusty1s:
Looks good. I can merge it, but I think it would be much nicer if this is done in the `extract` methods automatically so that it removes such artifacts for any dataset. One could recursively search for those artifacts and remove them. WDYT?

SergioRAgostinho:
The metadata generated by osx compression tools as far as I know only exists at the root level of your archive. Spending resources recursively going through all folders of a dataset tree might not be worth it in this case.

rusty1s:
You're right! Checking for the root folder should be sufficient.

SergioRAgostinho:
You merged this but I hope we're still proceeding with suggestion you mentioned. :)

rusty1s:
Yeah sure, just merged this as a temporary fix :)

PR #845, opened 2019-12-04
GPU device inconsistency fix for subgraph

jzhou316:
Fix the case when `edge_index` is on GPU device and `relabel_nodes` is `True`: should return the new edge_index on the same device for consistency.


rusty1s:
Thank you!

PR #870, opened 2019-12-23
add aggregate method to message passing

ferrine:
I'm exploring exotic aggregation methods and this additional abstraction will save me from copy-pasting code from your repo.


rusty1s:
Great! Thank you.

ferrine:
BTW, what to do with edge weight? You never use it in aggregation explicitly, rather put in in the most optimized place. I will need it also to be present in aggregation, but it interacts with the rest of the library. How do you suggest to solve that?

Going to create a new PR since this one is merged

rusty1s:
Well, it could follow the same principles as `message`. Each argument you request in `aggregate` should be added to its argument list.

PR #871, opened 2019-12-23
create arglist for aggregate

ferrine:
I followed your suggestion and added dynamic arglist creation. I should note, that I have not yet verified it works as expected, but hopefully it should as I do the same as in "update"

### DIFF
--- a/torch_geometric/nn/conv/message_passing.py
+++ b/torch_geometric/nn/conv/message_passing.py
@@ -4,11 +4,20 @@ import inspect
 import torch
 from torch_geometric.utils import scatter_
 
-special_args = [
-    'edge_index', 'edge_index_i', 'edge_index_j', 'size', 'size_i', 'size_j'
+message_special_args = [
+    "edge_index",
+    "edge_index_i",
+    "edge_index_j",
+    "size",
+    "size_i",
+    "size_j",
 ]
-__size_error_msg__ = ('All tensors which should get mapped to the same source '
-                      'or target nodes must be of same size in dimension 0.')
+aggregate_special_args = ["out", "index", "dim", "dim_size"]
+update_special_args = ["aggr_out"]
+__size_error_msg__ = (
+    "All tensors which should get mapped to the same source "
+    "or target nodes must be of same size in dimension 0."
+)
 
 is_python2 = sys.version_info[0] < 3
 getargspec = inspect.getargspec if is_python2 else inspect.getfullargspec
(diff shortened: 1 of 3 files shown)


ferrine:
I can verify this worked for me as expected. The thing I found inconvenient is the specification of default arguments:

```python
    def propagate(self, edge_index, size=None, edge_weight=None, **kwargs):
        return super().propagate(edge_index, size=size, edge_weight=edge_weight, **kwargs)

    def aggregate(self, out, index, dim=0, dim_size=None, edge_weight=None):
        ...
```
I believe the support of default arguments may be way smarter and elegant

rusty1s:
You do not need to override `propagate`, just pass `edge_weight` to it so `aggregate` can find it. We can not really make use of default arguments here, since we want our message passing scheme to be dynamic. It may depend on node features, edge attributes, edge weights, spatial relations, node positions, node normals, .... We can not add all of these optional tensors as optional arguments.

ferrine:
While I expect `edge_weight` to be None, it does not show up in `**kwargs` causing error here https://github.com/rusty1s/pytorch_geometric/pull/871/files#diff-e6277df2f08483a87d6d9b92a1c2c2ebR133

I did not use `get` here

rusty1s:
Can you show me your code? Does
```python
def forward(self, x, edge_index, edge_weight):
    return self.propagate(edge_index, x=x, edge_weight=edge_weight)

def aggregate(self, out, index, dim, dim_size, edge_weight):
    pass
```
not work?

ferrine:
Here you requested edge weight explicitly. User would pass None if not using it. That is what i intended to avoid overriding propagate.

rusty1s:
Ah, I see! You should then make use of `get` here, similar to https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/nn/conv/message_passing.py#L115.

ferrine:
The point is that `edge_weight` is optional in my case and used within aggregate natively. Cluttering User Facing API with passing None is weird in this case. I tried to ser it to None as default, but got KeyError above. The only way to fix that with minimal pain was to override propagate

rusty1s:
Mh, sorry I am misunderstanding something, but why can't you simply do:
```python
update_args = [kwargs.get(arg, None) for arg in self.__update_args__]
aggregate_args = [kwargs.get(arg, None) for arg in self.__aggregate_args__]
```

ferrine:
Yes, that could be the case, but it leads to default values mismatch. Say my default was a boolean and this solution implicitly sets it to none

rusty1s:
I do not think this is so much of a problem. You can set `edge_weight` to a boolean in `propagate` if you like. `message`, `update` and `aggregate` do not really work with default arguments, and I do not see a reason why they should.

rusty1s:
Thank you! Regarding your concerns of default arguments: I now see more clearly the potential use-cases of this. If you know of a cool way to implement this, please let me know :)

ferrine:
Why is travis unhappy?

About the default args. I know how to setup this for python 3, but not sure if it works for python 2. Do you plan to drop python 2 support?

rusty1s:
No need to consider Python2 anymore :)

ferrine:
Hm, it appeared to be more complicated than I expected. I think setting defaults to None would be fine most of the time.

ferrine:
Don't merge. I'm trying to make it simpler

ferrine:
Is it ok that in case a required argument is not passed, it is set to None?

As I read the code it was ok for message args, but not for update args. That is strange

ferrine:
In the proposed solution unused arguments to propagate will be ignored. It is the same behavior as before

ferrine:
@rusty1s do you think we should check for unused arguments? Personally, I see it as not necessary ad propagate method is usually called within class

rusty1s:
I finally got to check your code. It works brilliantly. Thank you! In addition, I cleaned up the whole interface so that it is now more extendable and understandable. WDYT?

torch_geometric/nn/conv/message_passing.py
```diff
@@ -48,171 +47,131 @@ class MessagePassing(torch.nn.Module):
...
-                "missing required arguments".format(
-                    set(aggregate_special_args)
-                    - set(self.__aggregate_signature__.parameters)
-                )
-            )
+        self.__msg_sign__ = inspect.signature(self.message).parameters
 
-    def propagate(self, edge_index, size=None, **kwargs):
-        r"""The initial call to start propagating messages.
+        self.__aggr_sign__ = inspect.signature(self.aggregate).parameters
```
ferrine: It is no more signature, but parameters. And why do you pop the first parameter?

rusty1s: You are right. I think it is unnecessary to treat the outputs/inputs as special arguments with fixed naming convention and instead inject them as first argument.

ferrine: So what does pop stands for?

rusty1s: What do you mean? It just removes the first parameter. This way the output of message and aggregate are not treated as special arguments.

ferrine: Oh, i got you

ferrine: Then it looks fine

PR #880, opened 2019-12-28
TUDataset node attributes loader bug

gear:
As I mentioned in [issue 879](https://github.com/rusty1s/pytorch_geometric/issues/879), TUDataset might not load some datasets properly (e.g. BZR). This commit fixes the bug by using `torch.unique` instead of sum to find out the node attribute index. 

This fix passes all current tests case but would fail when node attributes itself is in {0,1} (which is unlikely).


rusty1s:
Thanks! Looks good to me :) I replaced the `unique` call with a simple `0/1` check. That should run much faster.

gear:
Awesome!

PR #884, opened 2019-12-31
Shapenet with normals and validation

nicolas-chaulet:
This is a proposal for adding ShapeNet with normals (as opposed to the raw Shapenet). This dataset is used in a few papers including PN++ for benchmarks. It also contains a validation set option.


rusty1s:
Thanks! Looks good to me :) I will merge this after a closer look in the upcoming week.

PR #902, opened 2020-01-15
'add grid_sampling for pointcloud'

tchaton:
This PR implements grid_sampling for pointcloud.

Given an subsampling_param and num_classes.
It fit the pointcloud to a voxel grid, select winner class for each voxel, get mean features for the other item within data and return new data.

### DIFF
--- a/test/transforms/test_grid_sampling.py
+++ b/test/transforms/test_grid_sampling.py
@@ -1,4 +1,5 @@
 import numpy as np
+import numpy.testing as npt
 import torch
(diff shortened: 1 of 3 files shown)


rusty1s:
Looks good to me. How do you want to treat edges and edge features?

tchaton:
Hey @rusty1s, 

It was built for pointcloud only. So we don't have any edges or edge_features.
It could be extended to support graphs with edge and edges_features.

Best,
Thomas Chaton.

PR #907, opened 2020-01-16
PANConv

YuGuangWang:
I have added file "pan_conv.py" into nn/conv/
and file "pan_conv_test.py" into example/

Reference:
Zheng Ma, Ming Li, Yu Guang Wang. PAN: Path Integral Based Convolution for Deep Graph Neural Networks. ICML 2019 Workshop on Learning and Reasoning with Graph-Structured Representations.


rusty1s:
Thank you for this PR. The one issue I have with this PR is that `A^n` is explicitly computed via `sparse->dense->sparse` conversions. I do not think `A^n` needs to be computed at all if one re-uses the output of `A^{n-1}XW` for computing `A^{n}XW`. WDYT?

PR #911, opened 2020-01-16
Add MinCUT pooling (Bianchi et al., 2019, arxiv.org/abs/1907.00481)

danielegrattarola:
This PR adds the MinCUT pooling layer from 

Mincut pooling in Graph Neural Networks
F. M. Bianchi, D. Grattarola, C. Alippi
https://arxiv.org/abs/1907.00481

Changes include the implementation of MinCUT pooling as a dense pooling layer, an example on Enzymes, docs, and tests.


rusty1s:
Thank you very much. I‘m a big fan of mincut pooling and very happy to see this PR. I will have a closer look tomorrow and will merge ASAP. Overall, this looks super good already :)

danielegrattarola:
Thanks!! 
I've been meaning to do this PR for a while, let me know if anything needs to be changed.

rusty1s:
Hi and thanks again!
I changed a few things:
* Moved the `ENZYMES` example to a `PROTEINS` example (because I continuously fail to achieve good results on enzymes and I don't know why)
* Added a `DenseGraphConv` that adds a skip connection and performs sum aggregation
* Removed the `DenseDataLoader` and perform sparse to dense conversion on the fly after the first convolution layer
* Aligned the network architecture with the one presented in the paper (skip connections, no SAGE conv, ReLU activation, hidden feature size)
* Linting

This achieves about 76% on PROTEINS. I am ready to merge. WDYT?

danielegrattarola:
I had the same issues with `ENZYMES` as well, weird.  76% on `PROTEINS' looks right, thanks for taking the time to match the architecture to the paper.

I am happy to see MinCUT in PyG!

jlevy44:
This looks awesome :) Thanks!

Issue #923, opened 2020-01-23
Documenting Singularity Build


fdiehl:
#922 reminded me of the work necessary to install and use pytorch-geometric on a singularity image. From https://github.com/rusty1s/pytorch_scatter/issues/12, I know that there are other people that also want to use singularity (not surprising, given the academic environment).

I am, however, unsure where to place this: Would that be as a documentation page, as a singularity file in the repo, or somewhere external?

### Singularity Definition

For now, I'm simply placing the code here, allowing people to find and refer to it. Note the following:

- This uses cuda10.1; your host needs at least the same cuda version.
- Uses Python 3.7.2, therefore also showing how to install custom python versions.
- I needed to compile `pytorch-geometric` and related libraries.
- This allows multiple different GPU types (#922); change the line `export TORCH_CUDA_ARCH_LIST="5.0 6.1"` to add your own.
- It requires a `requirements.txt` file with the installed packages; alternatively, just specify `pip install torch` instead of `pip install -r /requirements.txt`

```
Bootstrap: docker

From: nvidia/cuda:10.1-cudnn7-devel-ubuntu16.04

%post
CURDIR=`pwd`

apt-get update
apt-get install -y tmux nano git wget

apt-get install -y --no-install-recommends make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev


export PYENV_ROOT=/opt/pyenv
export PATH="/opt/pyenv/bin:$PATH"
curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
pyenv install 3.7.2
echo 'export PATH=/opt/pyenv/versions/3.7.2/bin/:$PATH' >> $SINGULARITY_ENVIRONMENT
export PATH=/opt/pyenv/versions/3.7.2/bin/:$PATH

pip install torch==1.3.0

mkdir -p $SINGULARITY_ROOTFS/tmp/sing_build_cuda
cd $SINGULARITY_ROOTFS/tmp/sing_build_cuda

export TORCH_CUDA_ARCH_LIST="5.0 6.1"

git clone https://github.com/rusty1s/pytorch_scatter.git && \
    cd ./pytorch_scatter && \
    git checkout 1.4.0 && \
    python3 ./setup.py install && \
    cd ..

[...]

PR #924, opened 2020-01-23
Added singularity file for pytorch-geometric

fdiehl:
Following #923. I've reworked the example file a bit (for example removing that tmp-directory issue), and removed a few extraneous things.


rusty1s:
Thank you very much :)

PR #933, opened 2020-02-01
small bug fix for reddit.py

jzhou316:
The neighbor sampler only accepts torch.LongTensor instead of bool masks.

### DIFF
--- a/examples/reddit.py
+++ b/examples/reddit.py
@@ -65,7 +65,7 @@ def train():
     model.train()
 
     total_loss = 0
-    for data_flow in loader(data.train_mask):
+    for data_flow in loader(data.train_mask.nonzero().squeeze()):
         optimizer.zero_grad()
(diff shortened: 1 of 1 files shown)


rusty1s:
Thank you! The NeighborSampler already supports bool tensors, see [here](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/data/sampler.py#L148). Does this not work for you?

jzhou316:
Oh I just read through that line! It works. My bad. Thanks for the quick response, I think it's my version (1.3.1) was before the update.

By the way, is there any quick way to grab all the neighbors of a node instead of sampling, by using the current functions from torch_cluster?

rusty1s:
Well, you can set the size to `1.0` to sample *all* neighbors.

jzhou316:
Thanks!!

Issue #945, opened 2020-02-06
Support GraphSAGE concatenation step


kkonevets:
## ❓ Questions & Help

Looking at the code I could not find an evidence that SAGEConv layer concatenates aggregated neighbor features with current node features, namely step 5 in the original paper, it only averages them by adding self loops. How easy is to add concatenation?

Concatenation is crucial as it preserves target node signal and does not vanish it by averaging. For example, social network needs to account current person information along with his/her neighbors. It is useful for person classification.

![Screenshot from 2020-02-06 13-05-52](https://user-images.githubusercontent.com/16817116/73927782-137da980-48e3-11ea-81f2-d187773e7268.png)

rusty1s:
We currently only have an implementation of the `mean` aggregator of GraphSAGE, which replaces lines 4 and 5 of the pseudocode with Eq. (2).

kkonevets:
Can I try to make a pull request to implement the feature?

rusty1s:
Sure, contributions are very welcome! Feel free to ask any questions :)

yinanzhangepfl:
Hi,

I have just started working on PyG and I am confused about the concatenation step of SAGEConv.

1) If we set concat=False, then only aggr_out is used for updating node features. Let's say we have node i, and its neighbors are node k and node j. In this case,  aggr_out =mean{f_k, f_j} or aggr_out = mean {f_i, f_k, f_j}? Which one is correct? 

2) If we set concat=True, then is aggr_out = mean(f_k, f_j)?

I am not sure if we need add self-loops to graphs in the first case and  
remove self-loops in the second case. 
Actually, I have checked your example Reddit.py, where you used a NeighborSampler and add self loops without concatenation. 

Many thanks:)

kkonevets:
It is confusing, I agree.
If we want to use concatenation then self loops addition must be done before `SageConv` initialization, e.g. in `NeighborSampler` that will produce `res_n_id` for us. Otherwise it is optional. If someone will try to use concatenation without self loops `res_n_id` will be `None`.
If we have self loops then `aggr_out = mean {f_i, f_k, f_j}` otherwise `mean(f_k, f_j)`. So concatenation is independent of aggregation.

yinanzhangepfl:
2. It is always mean(f_i, f_k, f_j). Seems like it is not the same as neighborhood function in original paper, because it should be mean(f_k, f_j). It is done this way only to get res_n_id. And as I said, adding self loops improves results.

[...]

PR #974, opened 2020-02-19
Enable GraphSAGE current node features concatenation step

kkonevets:
This pull request is related to #945

Note:
1. Aggregation is done before weight multiplication (as in original GraphSage paper).
2. Edge features are not used. Don't know how to use as of now - any advice is appreciated.

Results:
I have tested it on examles/reddit.py

1. If I do not use concatenation best test score is 0.9300 and layers [602, 16] [16,41] . If I increase hidden dimension [602, 32] [32,41] it becomes 0.9331 having 20576 number of weights total.
2. Best test score with `concatenation=True` is 0.9412. Layers dimensions are [1204, 16] [32, 41]. Total number of weights is also 20576.

I also tested it on my own big data set (social network), results are also better then before.


rusty1s:
Thank you! This looks already super good. It seems like in its current form `concatenate` can only be used for `DataFlow` objects. It would be good to also allow the `res_n_id=None` and `concatenate=True` case.

kkonevets:
The only possibility I see to accomplish this is through explicitly passing features we want to concatenate to `forward(..., x_concat=...)`. Is that what you mean?

Issue #975, opened 2020-02-19
Suggest to replace cora_full with citation_full


cshjin:
## Motivation
The class `CoraFull` basically download one of the datasets from the `graph2gauss` paper. 
However, there are other datasets including `citeseer`, `cora_ml`, `dblp`, `pubmed`.

More contents from the repo [graph2gauss](https://github.com/abojchevski/graph2gauss/tree/master/data)

cshjin:
Problem resolve by PR #976

PR #976, opened 2020-02-19
replace cora_full with citation_full

cshjin:
Try to resolve the issue #975 

* rename the file from `cora_full` to `citation_full`
* rename the class_name `CoraFull` to `CitationFull`

### DIFF
--- a/torch_geometric/datasets/__init__.py
+++ b/torch_geometric/datasets/__init__.py
@@ -1,7 +1,7 @@
 from .karate import KarateClub
 from .tu_dataset import TUDataset
 from .planetoid import Planetoid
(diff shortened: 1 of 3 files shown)


rusty1s:
Looks great! Thank you.

rusty1s:
Some minor request if you have time. It would be good to still provide an alias for `CoraFull` to ensure backwards compatibility. Something like:
```python
class CoraFull(CitationFull):
    def __init__(self, root, transform=None, pre_transform=None): 
        super(CoraFull, self).__init__(root, 'cora', transform=None, pre_transform=None)
```

rusty1s:
Awesome!

PR #994, opened 2020-02-26
Create train_test_split_edges utils & implement undirected negative_sampling

dongkwan-kim:
1. Separate split_edges in GAE to the individual util (train_test_split_edges). This can be globally used for the link prediction task. I also add a link prediction example (link_preds.py).
2. Exclude self-loops in sampling negative edges for GAE.
3. Implement negative_sampling with force_undirected. This is a draft, and we have to discuss some design choices.
    - I first set num_neg_samples to half of the num_neg_samples given [[1]](https://github.com/dongkwan-kim/pytorch_geometric/blob/master/torch_geometric/utils/negative_sampling.py#L34), then make them undirected at the last line of function [[2]](https://github.com/dongkwan-kim/pytorch_geometric/blob/master/torch_geometric/utils/negative_sampling.py#L55). Since `edge_index.size(1)` is the common way to get the number of edges in PYG,  I think it is natural for end-users to set `num_neg_samples` to the real size of `edge_index`. But in our current implementation, `perm` contains duplicated indices and self-loop indices. So if we apply to_undirected, we might get negative edges the size of which is smaller than `num_neg_samples`. What do you think about this?

### DIFF
--- a/examples/autoencoder.py
+++ b/examples/autoencoder.py
@@ -6,6 +6,7 @@ import torch.nn.functional as F
 from torch_geometric.datasets import Planetoid
 import torch_geometric.transforms as T
 from torch_geometric.nn import GCNConv, GAE, VGAE
+from torch_geometric.utils import train_test_split_edges
 
 torch.manual_seed(12345)
 
@@ -46,7 +47,7 @@ channels = 16
 dev = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
 model = kwargs[args.model](Encoder(dataset.num_features, channels)).to(dev)
 data.train_mask = data.val_mask = data.test_mask = data.y = None
-data = model.split_edges(data)
+data = train_test_split_edges(data)
 x, train_pos_edge_index = data.x.to(dev), data.train_pos_edge_index.to(dev)
 optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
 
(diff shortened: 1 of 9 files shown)


rusty1s:
Wow, this looks super good overall! Thank you very much! You are right that we should improve the negative sampling method. The problem I see with your implementation is that you basically create a dense tensor of shape `[num_nodes, num_nodes]` which just might not be feasible for large graphs. Eventually, we can come up with an implementation to fix that.

dongkwan-kim:
OK. I will try to fix the dense tensor problem.

dongkwan-kim:
I was thinking to replace 35-36 lines with [`numpy.triu_indices`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.triu_indices.html) at first, but implemented with python primitives.

rusty1s:
Hi, as far as I can tell, this will result in the same dense memory requirements. What about just sampling a number between `[0, N^2-N]` and converting this index to the right edge in the upper triangle?

dongkwan-kim:
Oh, I see. I was not aware that `random.sample(range(N), k)` is really efficient. I add index encoding/decoding of the upper triangle matrix at the front/back of the sampling procedure, respectively.

Also, my implementation samples between `[0, N(N+1)/2]`, since I could not come up with a method of sampling between `[0, N^2-N]`, which you mentioned.

Thanks!

rusty1s:
Yeah, sorry, I meant `[0, N(N+1)/2]` instead of `[0, N^2-N]`. Sorry! This looks quite good now overall. We should fix Travis complaints and then we can merge :) Thank you!

Issue #1005, opened 2020-03-04
Better utils.to_networkx()?


tjiagoM:
## 🚀 Feature

Hi, I have noticed two limitations from `utils.to_networkx()` which I was using in my project:

1. It can only return a directed DiGraph, whereas my original graph is actually undirected
2. It will always return the self loops for each node but I have them just for the sake of the message passing mechanism.

So, I was thinking about doing a PR with the changes I have made, but wanted to check with you first if you think this is something you'd like to include in pytorch geometric and whether you think I should make some changes first.

So far, the code I have is the following:
```python
def to_networkx(data, node_attrs=None, edge_attrs=None, to_undirected=False, remove_self_loops=False):
    r"""Converts a :class:`torch_geometric.data.Data` instance to a
    :obj:`networkx.DiGraph` if :attr:`to_undirected` is set to :obj:`True`, or an undirected
    :obj:`networkx.Graph` otherwise.

    Args:
        data (torch_geometric.data.Data): The data object.
        node_attrs (iterable of str, optional): The node attributes to be
            copied. (default: :obj:`None`)
        edge_attrs (iterable of str, optional): The edge attributes to be
            copied. (default: :obj:`None`)
        to_undirected (bool, optional): If set to :obj:`True`, it will return a :obj:`networkx.Graph`.
            In practice, the undirected graph will correspond to the upper triangle of the
            corresponding adjacency matrix. (default: :obj:`False`)
        remove_self_loops (bool, optional): If set to :obj:`True`, it will not include self loops in
            the returning graph.  (default: :obj:`False`)
    """

    if to_undirected:
        G = nx.Graph()
    else:
        G = nx.DiGraph()
    G.add_nodes_from(range(data.num_nodes))

    values = {key: data[key].squeeze().tolist() for key in data.keys}

    for i, (u, v) in enumerate(data.edge_index.t().tolist()):
        if to_undirected and v > u:
            continue
        if u == v and remove_self_loops:
            continue
        G.add_edge(u, v)
        for key in edge_attrs if edge_attrs is not None else []:
            G[u][v][key] = values[key][i]

    for key in node_attrs if node_attrs is not None else []:
        for i, feat_dict in G.nodes(data=True):
            feat_dict.update({key: values[key][i]})

    return G
```

[...]

PR #1007, opened 2020-03-05
Small tweak in utils.to_networkx

tjiagoM:
PR from discussion in #1005.

I basically added option to return an undirected Graph, and the option to remove self loops, with documentation.

I have added a few tests:
1. Changed previous test to include with and without removing self loops
2. An extra test for undirected graphs

For the test with the undirected graphs there is the issue that one edge attribute is lost, but that is something that cannot be avoided (I think) if the person starts already with a graph that is not exactly undirected in its attributes. In any case, I've added the assert as in the documentation I wrote "the undirected graph will correspond to the upper triangle of the corresponding adjacency matrix".

### DIFF
--- a/test/utils/test_convert.py
+++ b/test/utils/test_convert.py
@@ -46,14 +46,16 @@ def test_to_networkx():
     edge_attr = torch.Tensor([1, 2, 3])
     data = Data(x=x, pos=pos, edge_index=edge_index, weight=edge_attr)
 
-    for condition_sl in [True, False]:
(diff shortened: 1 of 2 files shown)


tjiagoM:
@rusty1s I would appreciate any help or comments overall. I've had a look to travis logs and I think the error occurring now is not related to this PR changes I believe?

rusty1s:
Thanks, this looks super good. The error from Travis is indeed not related. I will look at it.

PR #1014, opened 2020-03-08
Modify dockerfile to support cuda 10.1 and cudnn7

shengwenLeong:


rusty1s:
Thank you very much :) Looks great!

PR #1030, opened 2020-03-16
Fixed bug in read_obj

mwussow:
`read_obj` did not extract mesh faces correctly from Wavefront `.obj` files because only the first index per triangle was processed.

**Example:**
Face in .obj file: `f 1/2/3 4/1/6 3/2/4`
Processed output of current `read_obj`: `[0,3,2]`
Correct output: `[0,1,2,3,0,5,2,1,3]`


rusty1s:
Thanks!

Issue #775, opened 2019-10-31
error related to topk


ThyrixYang:
## 🐛 Bug

I just encountered the same error mention in

https://github.com/inyeoplee77/SAGPool/issues/7

I found that the max index output by topk is larger than score.shape[0], so I think this maybe a bug about topk function.

## To Reproduce

To reproduce, 

1. clone the SAGPool code on https://github.com/inyeoplee77/SAGPool

2. run with "python main.py --dataset PROTEINS"

## Environment

* OS: ubuntu 19.04
* Python version: 3.7.4
* PyTorch version: 1.2
* CUDA/cuDNN version: CUDA 10.0
* GCC version: 7.4.0
* Any other relevant information:

rusty1s:
This is not a PyG error, see my comment in inyeoplee77/SAGPool#7

Issue #982, opened 2020-02-20
Runtime Error on Training Custom Dataset


ilteralp:
## ❓ Questions & Help

Hi everyone, I am trying to train a simple GNN with using my dataset. I created the dataset with following [Creating Your Own Datasets](https://pytorch-geometric.readthedocs.io/en/latest/notes/create_dataset.html) tutorial. In my case nodes with the same label are connected to each other. To create the GNN, I used [enzymes_topk_pool.py](https://github.com/rusty1s/pytorch_geometric/blob/master/examples/enzymes_topk_pool.py) example. Here is the error I get, 

`RuntimeError: CUDA error: device-side assert triggered
The above operation failed in interpreter.
Traceback (most recent call last):
  File "/home/rog/anaconda3/envs/py36_pytorch/lib/python3.6/site-packages/torch_scatter/scatter.py", line 39
        else:
            size[dim] = int(index.max()) + 1
        out = torch.zeros(size, dtype=src.dtype, device=src.device)
              ~~~~~~~~~~~ <--- HERE
        return out.scatter_add_(dim, index, src)
    else:`

I also tried to train with `device = torch.device('cpu')`, it produces the error below, 

`RuntimeError: index out of range: Tried to access index 11 out of table with 3 rows. at /opt/conda/conda-bld/pytorch_1579022034529/work/aten/src/TH/generic/THTensorEvenMoreMath.cpp:418`

Here is what I have tried,
```
import json
import torch
from torch_geometric.data import InMemoryDataset, Data
from torch_geometric.data import DataLoader
from torch_geometric.nn import GraphConv, TopKPooling
from torch_geometric.nn import global_mean_pool as gap, global_max_pool as gmp
import torch.nn.functional as F
import os.path as osp


NUM_CLASSES = 15
ROOT_PATH = "./data/"

################################ Create Dataset ################################
class GraphInputDataset(InMemoryDataset):
    def __init__(self, root, train, transform=None, pre_transform=None):
        
        super(GraphInputDataset, self).__init__(root, transform, pre_transform)
        path = self.processed_paths[0] if train else self.processed_paths[1]
        self.data, self.slices = torch.load(path)
        
    # The name of the files to find in the self.raw_dir folder in order to skip the download.
    @property
    def raw_file_names(self):
        return ['sub_{}_edges.txt'.format(file_name) for file_name in ['train', 'test']]
    
    # A list of files in the processed_dir which needs to be found in order to skip the processing.
    @property
[...]

PR #1051, opened 2020-03-22
Fix out of index issue in topk.

nineye:
In topk, it makes dense and fixed node tensor according to x.shape and initialize all values to -2.
So, a graph which is smaller than biggest graph in batch has dummy nodes.
And then, x is copied to dense tensor and ordered by score.
If some scores in a graph which is smaller than biggest graph are smaller than -2,
they are moved to out of node size of original graph and dummy nodes are filled to those position.
So, returned perm has dummy node that have out of index.

Actually, this error was occurred in my custom dataset using SAGPool and I fixed the problem by this code.


nineye:
I think this change will resolve #775 and #982.

rusty1s:
Looks good. Thank you!

PR #1075, opened 2020-03-27
Fixed bipartite graph mini-batching example

jkuck:
Running the previous example produced an error for me.  This dimensions change fixes the problem.


rusty1s:
The broadcasting works for me, but I think its more understandable this way. Thanks.

PR #1096, opened 2020-04-03
use multiprocessing for faster computation

jannessm:


rusty1s:
Super! I also added a `num_workers` argument to the function.

jannessm:
great!

PR #1099, opened 2020-04-06
add node clustering as argva example

gsoosk:
Hi, 
I added an example of node clustering using the ARGVA method. I thought due to [this thread](https://github.com/rusty1s/pytorch_geometric/issues/203) it could be useful.

### DIFF
--- a/examples/argva_node_clustering.py
+++ b/examples/argva_node_clustering.py
@@ -1,0 +1,135 @@
+import os.path as osp
+
+from torch_geometric.datasets import Planetoid
+import torch_geometric.transforms as T
+from torch_geometric.utils import train_test_split_edges
+from torch_geometric.nn import GCNConv, ARGVA
+import torch.nn.functional as F
+import torch
+from sklearn.cluster import KMeans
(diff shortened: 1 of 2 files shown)


rusty1s:
Awesome. Thank you! I will have a closer look and merge ASAP.

rusty1s:
Sorry it took my so long to review this one. I have nothing to complain :) Thank you a lot!

examples/argva_node_clustering.py
```diff
@@ -84,52 +75,53 @@ def train():
     loss = model.recon_loss(z, data.train_pos_edge_index)
```
Wenjian-Ma, line 75: missing  loss = loss + model.reg_loss(z) ?

rusty1s: This is added in the line below, right?

rusty1s: This is fixed in master, see https://github.com/pyg-team/pytorch_geometric/blob/master/examples/argva_node_clustering.py#L77.

PR #1126, opened 2020-04-14
Move the nested collate function into a separate callable object.

coloss:
The original nested functions `collate` and `dense_collate` cause problems on Windows because subprocess (on Win) uses pickle for serialization when multiprocessing. Lambda and nested functions cannot be pickled on Windows. Therefore, `dense_collate` is moved out of `DenseDataLoader`  and `collate` is turned into a callable object `Collater` because it needs a member.

### DIFF
--- a/torch_geometric/data/dataloader.py
+++ b/torch_geometric/data/dataloader.py
@@ -5,6 +5,37 @@ from torch_geometric.data import Data, Batch
 from torch._six import container_abcs, string_classes, int_classes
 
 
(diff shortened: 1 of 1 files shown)


rusty1s:
Great. Thank you!

coloss:
My pleasure @rusty1s . Btw, I expect to be using pytorch geometric on Win in the next days/weeks/months so there might be more stuff I come across eventually.

PR #1127, opened 2020-04-15
Add installation of dependencies

ykumthekar4929:
Added installation of apt-transport-https ca-certificates at the top of the file, lack of these dependencies were throwing The method driver /usr/lib/apt/methods/https could not be found


rusty1s:
Looks good. Thank you!

PR #1142, opened 2020-04-21
Use 1-D convolution after sort_pool

muhanzhang:
Hi Matthias, 

Thank you for implementing sort_pool in PyG. I noticed that sort_pool is not correctly used in graph classification benchmarking experiments. After sort_pool, we get a sequence of node representations, where 1-D convolution is supposed to be applied to the sequence to learn meaningful information (like using 1-D convolution on time series), instead of concatenating them and applying MLP.

In this PR, I modify `sort_pool.py` a bit to add a simple 1-D convolution layer after sort_pool, for people referring to this repository to get a good sense of how to use sort_pool. The graph classification performance improves moderately. 

| Accuracy     | MUTAG        | PROTEINS           | COLLAB  |   IMDB-BINARY  | REDDIT-BINARY |
| ------------- |-------------|-------------|-------------|-------------|-------------|
| Before      |  77.3 ± 8.9 | 72.4 ± 4.1 | 77.7 ± 3.1 | 72.4 ± 3.8 | 74.9 ± 6.7 |
| After        |  81.4 ± 7.1 | 74.2 ± 3.8 |  78.3 ± 1.1 | 72.2 ± 5.7 | 81.5 ± 3.3 | 

There are also other differences from how I used sort_pool in DGCNN. I tried to reimplement DGCNN in PyG, and can get an 80.6 accuacy on COLLAB and 89.5 accuracy on REDDIT-BINARY. But I think it is not the purpose here. 

Thank you in advance for your time!

### DIFF
--- a/benchmark/kernel/sort_pool.py
+++ b/benchmark/kernel/sort_pool.py
@@ -1,24 +1,26 @@
 import torch
 import torch.nn.functional as F
-from torch.nn import Linear
+from torch.nn import Linear, Conv1d
 from torch_geometric.nn import SAGEConv, global_sort_pool
 
 
 class SortPool(torch.nn.Module):
     def __init__(self, dataset, num_layers, hidden):
         super(SortPool, self).__init__()
-        self.k = 10
+        self.k = 30
         self.conv1 = SAGEConv(dataset.num_features, hidden)
         self.convs = torch.nn.ModuleList()
         for i in range(num_layers - 1):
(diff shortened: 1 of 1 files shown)


rusty1s:
Hi Muhan, thank you very much! This looks super good. 1-D convolution makes indeed more sense on sequential data. Please let me know how your DGCNN architecture looks like. We can also add an example if you like.

PR #1143, opened 2020-04-21
Changed files_exist to return False if files list is empty

itsikad:
Hi,

I suggest this fix since Python's all() returns True for empty input list.
In case processed_dir doesn't exist/empty before Dataset init call, _process() will return prior to creating processed_dir and calling self.process.

Best regards,
Itsik

### DIFF
--- a/torch_geometric/data/dataset.py
+++ b/torch_geometric/data/dataset.py
@@ -16,7 +16,7 @@ def to_list(x):
 
 
 def files_exist(files):
-    return all([osp.exists(f) for f in files])
+    return len(files) != 0 and all([osp.exists(f) for f in files])
 
 
 def __repr__(obj):


rusty1s:
What's the reasoning behind this fixed. If there are no files to process, the `process` method should be skipped, right?

itsikad:
Thanks for you fast response.
I might be missing something, but according to the documentation, the self.process() method processes the raw files located in self.raw_dir and saves the output into self.processed_dir. 
When self._process() is call during init, up to that point, the processed dir should be empty/not exist. 
Calling files_exist(self.processed_path) should return true if all the files to be processed already exists in processed_dir so one can skip the process step.

rusty1s:
Yes, I agree. However, it is not fully specified what the intended behaviour should be in case `processed_files` is an empty list:
1. All the files needed for loading the data already exist (since there are none), so skip processing
2. Process everytime since the user doesn't want to save their processed data.
I agree that your solution is a bit more intuitive.

itsikad:
Thanks.
The most frequent usage is processing the downloaded data I believe, in this case `processed_files` is correctly empty. This also holds for the second scenario you suggested.
The first scenario basically means that the raw files can be loaded directly in `get` method (I ignore the option of empty dataset). In this case, one can easily set `processed_file_names` to return `self.raw_file_name'.

PR #1157, opened 2020-04-26
Add pytorch 1.5.0 to installation doc

Rishi8991:
Hi, I thought it might be worth updating the install to include the links for pytorch 1.5 since it is a later version


rusty1s:
Thanks, I missed that :)

PR #1186, opened 2020-05-05
Fix train_test_split_edges for nonfeaturized graph

laifi:
an exception error would be raised in `train_test_split_edges` in case of no featurized graph ( `data.x=None `) because after the line `data.edge_index = None` and when data.x is already None then no way to get the number of nodes, very simple fix would be to replace this line `num_nodes = data.num_nodes` above `data.edge_index = None`


rusty1s:
Sounds good, thanks :)

PR #1187, opened 2020-05-05
test/train/val pos masks

laifi:
the current implementation of [train_test_split_edges.py ](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/utils/train_test_split_edges.py)returns only list of the edges splitted into positive and negative train/val/test edges.
But, if we have as an example list of features (`data.x`) or other data information such as additional non-predefined list `data.edge_type` that we need in the model forward call, we have somehow find a way how to get the corresponding information/attributes of the train_pos_edge_index /  test_pos_edge_index / val_pos_edge_index during the call.
for this reason , I changed the implementation of the function to save in the data object the masks of the positive test/train/val edge indices, then it would be easier to extract for example the corresponding features of the train_pos_edges and so on... from :
- data.train_pos_mask
- data.test_pos_mask
- data.val_pos_mask

### DIFF
--- a/torch_geometric/utils/train_test_split_edges.py
+++ b/torch_geometric/utils/train_test_split_edges.py
@@ -22,9 +22,9 @@ def train_test_split_edges(data, val_ratio=0.05, test_ratio=0.1):
     """
 
     assert 'batch' not in data  # No batch-mode.
-
     num_nodes = data.num_nodes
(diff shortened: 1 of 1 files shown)


laifi:
Some issues to fix...

Issue #1227, opened 2020-05-15
What does each attention weight refer to in GATConv?


douglasrizzo:
## ❓ Questions & Help

Hello there. I saw that, in the master branch, GATConv has an option to return the attention weights after a call to `forward()`. The number of attention weights is equal to the number of edges + added self loops.

How do I know which attention weight belongs to which edge? Are they returned in the same order as `edge_index`? Where are the weights regarding self loops added? Thanks in advance.

---
I made an example with 4 nodes and 7 edges, which returned 11 weights, but I can't figure to which edge each weight refers to.

```python
import torch
import torch_geometric as pyg

torch.manual_seed(0)
l = pyg.nn.GATConv(6,6)
x = torch.distributions.normal.Normal(0,1).sample((4,6))
ei = torch.tensor([[1,2,2,0,3,0,1],
                   [0,0,1,2,2,3,3]],dtype=torch.long)

x, alpha = l(x,ei,return_attention_weights=True)

print(alpha)
```
```
tensor([[0.2542],
        [0.2742],
        [0.5898],
        [0.4758],
        [0.2371],
        [0.6427],
        [0.1914],
        [0.4715],
        [0.4102],
        [0.2871],
        [0.1659]], grad_fn=<DivBackward0>)
```

douglasrizzo:
I think I'm finding the answer myself. By calling:

```python
edge_index = torch.tensor([[1,2,2,0,3,0,1],
                   [0,0,1,2,2,3,3]],dtype=torch.long)

edge_index, _ = pyg.utils.loop.remove_self_loops(edge_index)
edge_index, _ = pyg.utils.loop.add_self_loops(edge_index)
```

just like GATConv does, I realized the self loops are added at the end of `edge_index`. So I'd guess the weights from `alpha` refer to the edges in `edge_index` with self loops added to the end, in order.

rusty1s:
That's correct. I agree that this is not really elegant. What do you think about removing the attention scores of self-loops before returning them?

douglasrizzo:
I'm consulting the GAT paper. They never mention adding self loops to the input graph before doing a forward pass, i.e. all operations are done in `N(i)` and not `N(i) U {i}`. Maybe GATConv should only generate and use attention weights for the loops (and self loops) that are declared in the `edge_index` that is provided to `forward()`. This way, if/when GATConv returns `alpha`, the outside code can use the existing `edge_index` to know where each weight comes from.

[...]

PR #1234, opened 2020-05-17
Store GAT attention and edge_index

douglasrizzo:
This is what I was thinking about. I believe it saves an unnecessary forward pass if someone wants both the output and the attention weights.

If there is a problem with storing these values inside GATConv to return later, I can change forward to return `x, edge_index, alpha` like commented on #1227. Although this will still store the state of alpha, as was done previously.

### DIFF
--- a/test/nn/conv/test_gat_conv.py
+++ b/test/nn/conv/test_gat_conv.py
@@ -13,7 +13,7 @@ def test_gat_conv():
     assert conv(x, edge_index).size() == (num_nodes, 2 * out_channels)
     assert conv((x, None), edge_index).size() == (num_nodes, 2 * out_channels)
 
-    out = conv.get_attention_weight(x, edge_index)
+    out = conv.get_attention_weight()
     assert out[0].size() == (2, edge_index.size(1) + num_nodes)
     assert out[1].size() == (edge_index.size(1) + num_nodes, conv.heads)
 
@@ -21,6 +21,6 @@ def test_gat_conv():
     assert conv(x, edge_index).size() == (num_nodes, out_channels)
     assert conv((x, None), edge_index).size() == (num_nodes, out_channels)
 
-    out = conv.get_attention_weight(x, edge_index)
+    out = conv.get_attention_weight()
     assert out[0].size() == (2, edge_index.size(1) + num_nodes)
     assert out[1].size() == (edge_index.size(1) + num_nodes, conv.heads)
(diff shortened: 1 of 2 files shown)


rusty1s:
Thanks! I put a bit more thought into this and I am currenlty in favor of returning to the old behaviour, and (as you suggested) return `(edge_index, alpha)` as an additional tuple. I think this is more in line with the PyTorch API, *e.g.* how the allow optional return values in [`torch.unique`](https://pytorch.org/docs/stable/torch.html#torch.unique).

lgray:
Hey guys - sorry to butt in on this one, but using class member variables (that are not buffers or parameters) to define how you store state is not compatible with all the jitting stuff. You have to pass it to forward.

Luckily the way that @rusty1s set things up with `get_attention_weights` as a separate function makes the interface remain the same more or less.

lgray:
If I'm just missing something with jit-script please feel free to point it out, I do like it when it's possible to have state in classes. :-)

rusty1s:
That's true. The current interface has nice jit support. As far as I can tell, optional return parameters are currently not supported by the jit-interface. @lgray Do you have an idea how `torch.unique` supports this?

douglasrizzo:
I don't know anything about TorchScript, but that would mean that even the old implementation of how the attention weights are returned would not be compatible.

Just for clarification, the current GATConv class uses member variables such as `self.out_channels`, `self.heads`, `self.concat`, `self.negative_slope`, `self.dropout` in the `message()` and `update()` methods (but not `forward()`). Is there a problem with those?

douglasrizzo:
I implemented what @rusty1s mentioned above, just in case.

douglasrizzo:
I just had another idea. Maybe we could use the Python `with` statement to set a flag inside the GATConv object, which would inform whether forward should return attention weights or not.

```python
out = conv(x, edge_index)
with conv.return_attention():
    out, (alpha, edge_index) = conv(x, edge_index)
```

And inside GATConv, we'd need to add the __enter__() and __exit__() methods, which are part of the `with` statement "contract":

```python
def __enter__():
    self.__att__ = True

def __exit__():
    self.__att__ = False
```

rusty1s:
Yeah, that's true. I'm in favor of the current one though because the interface is more clear to users by just looking at the function signature from `forward`.

One additional remark: IMO, we should return `(edge_index, alpha)` because this is more in line with how we treat sparse matrices in PyG. For example, `to_undirected` also returns a tuple of type `(edge_index, edge_attr)`.

rusty1s:
Awesome!

PR #1248, opened 2020-05-20
add Molecular Fingerprint conv

rhsimplex:
Hi, thanks for the great library!

I needed the [Molecular Fingerprint convolution](https://papers.nips.cc/paper/5954-convolutional-networks-on-graphs-for-learning-molecular-fingerprints) and as far as I could tell it is not implemented yet. This is using some improvements from the deepchem library (@rbharath if you are interested).

If there is interest in merging this, I can perhaps add an example and expand the test a bit.

Thanks!


rusty1s:
Thanks! This looks quite good already. Will try to merge ASAP.

rusty1s:
I made a little bit of cleanup (like replacing ModuleDicts with ModuleLists) and renamed arguments to be more in line with other conv operators. Additionally, I removed the `min_degree` argument as I feel that it is really unnecessary.

PR #1256, opened 2020-05-23
MessagePassing.jittable() implementation and tests

lgray:
Factorization of #1191 into reasonable sub-parts.

This PR contains the interface and simple tests of that interface.


rusty1s:
Made a pass and cleaned up a lot of things. IMO, it is now much more readable. From my side, this is ready to merge. We can always clean up more later :)

lgray:
@rusty1s Amazing job, this looks great. Did you test it against #1258 to make sure it covers all the cases?

rusty1s:
I added a lots of tests that (most) of #1258 should work. For now, I focused that there are no breaking changes, and will resolve any issues regarding #1258 once I get there.

There are still a few issues I would like to resolve in the upcoming weeks:
* `@torch.jit._overload_method` does not really work yet and I have no clue why
* `SparseTensor` type needs to be added to the propagate type
* Node features and tuple of node features do not play nicely together in JIT mode

I added some `pytest.raises` tests that identify these errors.

rusty1s:
Let me know if that is okay for you, so we can merge :)

lgray:
Far better than what I had in mind. Please go ahead and merge!

lgray:
For `@torch.jit._overload_method` we can ask some of the FB guys to take a look, they're very interested in this development, I'm sure we can figure it out expediently with them.

lgray:
As for "Node features and tuple of node features do not play nicely together in JIT mode", yeah I saw this too and chose to just work it all out for node features rather than tuple-of.

rusty1s:
It should work though once I get `overload` working. I will take a closer look at this in the next week.

PR #1258, opened 2020-05-23
MessagePassing.jittable() convolutional operators

lgray:
Factorization of #1191 into reasonable pieces.

This PR contains the changes needed to make nearly all convolutional operators jittable, including basic tests.
The few that are remaining are either a bit tricky or need the CPU version of knn to be made jittable before we can test them.
I have made one commit per convolutional op so it's easy to review.

Right now this has #1256 and #1257 merged in so tests will pass. Will rebase later or hopefully github is smart enough to realize when things have been merged to master.

### DIFF
--- a/benchmark/kernel/gin.py
+++ b/benchmark/kernel/gin.py
@@ -7,25 +7,25 @@ from torch_geometric.nn import GINConv, global_mean_pool, JumpingKnowledge
 class GIN0(torch.nn.Module):
     def __init__(self, dataset, num_layers, hidden):
         super(GIN0, self).__init__()
-        self.conv1 = GINConv(Sequential(
-            Linear(dataset.num_features, hidden),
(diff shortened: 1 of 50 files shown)


rusty1s:
This is done from my side :) Ready to merge :)

rusty1s:
To follow-up: I returned to the old `RGCNConv` because I did not find an elegant way to handle optional node features yet. I enhanced the `GATConv` by overloads, so that it does now only return attention weights in case `return_attention_weights` is given. Everything stayed mostly the same, expect for some test clean ups.

lgray:
Looks good! I think the patterns of how to do this gracefully in the future are starting to emerge, which is good. Thanks for doing an excellent job of review and cleanup, and of course maintaining this whole package! 💯

PR #1259, opened 2020-05-23
MessagePassing.jittable() examples

lgray:
Factorization of #1191 into reasonable sub-parts.

This PR contains examples of using `torch.script.jit` on `.jittable()` conv-ops and full models where possible.

Right now this has #1256, #1257, and #1258  merged in so tests will pass. Will rebase later or hopefully github is smart enough to realize when things have been merged to master.


rusty1s:
Also done. I removed the MNIST examples for now until `torch.unique` JIT support lands in PyTorch 1.6.0.

lgray:
Sure that makes sense, thanks again!

On a more pedagogical point, wouldn't it be better to have an example file specifically about jitting stuff? The way it is now it's just sort of snuck into the gatconv example. Replicating everything with an `_jit.py` version of it is silly, of course. I'm just looking for a better way to expose the functionality so people find it easily.

rusty1s:
I added a `jit/` folder to `examples`.

rusty1s:
In addition, I plan to clean-up the `examples/` folder, e.g., sub-folders and better documentation.

lgray:
Ah, derp, sorry I missed that. Perfect!

PR #1275, opened 2020-05-29
request to add parameter 'desired_max_num_nodes'

shi27feng:
For functions 'to_dense_adj' and 'to_dense_batch' in utils package, sometimes it requires to extend the dimension of feature matrix or adjacency matrix to 'desired max number of nodes' but not equal to the max number of nodes in a batch.


rusty1s:
Thanks :)

PR #1304, opened 2020-06-07
Update cluster_gcn.py

CFF-Dream:
Using cluster gcn to process PPI dataset


rusty1s:
Thanks! Can you move that to its own example, e.g., cluster_gcn_ppi.py?

CFF-Dream:
I have modify the file, is it ok?

rusty1s:
Yes, thanks.

rusty1s:
I updated the example and it now reaches about 98% micro F1-score. The key is to add use batch/layer normalization after each layer. I think one can get even higher scores implementing the diagonal enhancement scheme of ClusterGCN.

rusty1s:
I would like to merge that if you are fine with the changes. Please let me know :)

CFF-Dream:
Of course! But when i run your codes in the [https://github.com/rusty1s/pytorch_geometric/pull/1304/file](url) "examples/cluster_gcn_ppi.py"
the loss can be converged correctly, but the f1_score only reaches about 40% after 500 epochs. Is there anything wrong? or i missing some other key operations?

rusty1s:
It reaches 98% f1_score for me. Not sure what's going on.

CFF-Dream:
Maybe you can merge the final version code(cluster_gcn_ppi) into the original code, and I will try again. Thanks!

CFF-Dream:
I think i have figured out what cause the low f1_score, you forgot to add a Linear Module behind the Conv Module. It should be 
```
class Net(torch.nn.Module):
    def __init__(self, in_channels, hidden_channels, out_channels, num_layers):
        super(Net, self).__init__()
        self.convs = torch.nn.ModuleList()
        self.lins = torch.nn.ModuleList()
        self.batch_norms = torch.nn.ModuleList()
        self.convs.append(SAGEConv(in_channels, hidden_channels))
        self.lins.append(nn.Linear(in_channels, hidden_channels))
        self.batch_norms.append(BatchNorm(hidden_channels))
        for _ in range(num_layers - 2):
            self.convs.append(SAGEConv(hidden_channels, hidden_channels))
            self.lins.append(nn.Linear(hidden_channels, hidden_channels))
            self.batch_norms.append(BatchNorm(hidden_channels))
        self.convs.append(SAGEConv(hidden_channels, out_channels))
        self.lins.append(nn.Linear(hidden_channels, out_channels))

        for conv, lin, batch_norm in zip(self.convs[:-1], self.lins[:-1], self.batch_norms):
            x = conv(x, edge_index) + lin(x)
            x = batch_norm(x)
            x = F.relu(x)
            x = F.dropout(x, p=0.2, training=self.training)
        return self.convs[-1](x, edge_index)
```

rusty1s:
The `SAGEConv` has a skip-connection already implemented internally.

rusty1s:
Do you use PyG 1.5.0?

CFF-Dream:
Python3.7.4; 
Pytorch1.5.0;
torch-cluster1.5.4; 
torch-geometric1.4.3;
torch-scatter2.0.4;
torch-sparse0.6.4;
torch-spline-conv1.2.0;
CUDA10.2;
win10;

rusty1s:
I see. The `SAGEConv` changed slightly in PyG 1.5.0 (it now has `concat=True` by default), so upgrading PyG or setting `SAGEConv(..., concat=True)` should increase the performance (and is equivalent to your modification).

PR #1318, opened 2020-06-13
fixed bug in torch_geometric/nn/conv/hypergraph_conv.py

shrey-bansal:
changed the original aggregate function in message_passing.py as **dim_size does not match the index range.**

 index range = [0, size(hyperedge_index)] whereas  dim_size = size(x). 

Hence scatter function displayed index out of range error in the case when size(hyperedge_index) > size(x).

### DIFF
--- a/torch_geometric/nn/conv/hypergraph_conv.py
+++ b/torch_geometric/nn/conv/hypergraph_conv.py
@@ -108,6 +108,15 @@ class HypergraphConv(MessagePassing):
             out = alpha.view(-1, self.heads, 1) * out
         return out
 
+    def aggregate(self, inputs, index, ptr=None, dim_size=None):
+        if ptr is not None:
+            for _ in range(self.node_dim):
+                ptr = ptr.unsqueeze(0)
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/conv/hypergraph_conv.py
```diff
@@ -108,6 +108,15 @@ def message(self, x_j, edge_index_i, norm, alpha):
             out = alpha.view(-1, self.heads, 1) * out
         return out
 
+    def aggregate(self, inputs, index, ptr=None, dim_size=None):
+        if ptr is not None:
+            for _ in range(self.node_dim):
+                ptr = ptr.unsqueeze(0)
+            return segment_csr(inputs, ptr, reduce=self.aggr)
+        else:
+            return scatter(inputs, index, dim=self.node_dim, reduce=self.aggr)
```
shrey-bansal, line 117: Original : return scatter(inputs, index, dim=self.node_dim, dim_size=dim_size, reduce=self.aggr)

Issue #1343, opened 2020-06-19
Weighted edge sampler in GraphSAINT


KiddoZhu:
## ❓ Questions & Help

Hi! I found that `GraphSAINTEdgeSampler` performs uniformly sampling, while the original paper says the probability is proportional to the reciprocal of degrees.

I checked the original repo according to the comment, but I feels like the original implementation is weighted. In this [line](https://github.com/GraphSAINT/GraphSAINT/blob/fdbc901268b79d7868b97d91d9ed465c46c3d37b/graphsaint/graph_samplers.py#L54-L55), the CDF is `self.edge_prob_tri.data.cumsum()`, which is the symmetrically normalized weights from `preproc()`.

Shall I make a pull request to fix that?

rusty1s:
Oh, it seems that you are right. The `Edge` sampler does perform uniform sampling, not `Edge2`. I feel that we should support both versions. If you are willing to contribute, please do so :)

KiddoZhu:
Sure. I will update here once finished.

KiddoZhu:
I feels like both `Edge` and `Edge2` perform weighted sampling, as they both take an edge probability distribution. If I understand the paper correctly, there seems to be no reason to opt for a uniform sampling, except it is more parallelization friendly?

PR #1353, opened 2020-06-21
Weighted edge sampler for GraphSAINT

KiddoZhu:
As discussed in #1343


rusty1s:
That is awesome! Thank you so much :)

PR #1356, opened 2020-06-22
Fix GDC with self-loops

klicperajo:
to_dense_adj does not consider duplicate entries in edge_index. We therefore need to coalesce before calling it.


klicperajo:
This could also be solved more systematically by using `scatter_add` in `to_dense_adj`.

rusty1s:
Yes, you are right. I will try to add that to `to_dense_adj`.

rusty1s:
Done :)

klicperajo:
Nice, thank you!

PR #1368, opened 2020-06-24
On the computation of KL divergence for graph VAE

GuillaumeSalha:
Dear @rusty1s, 

First of all, I would like to congratulate you and your collaborators for this amazing work.

I recently noticed that your computation of the Kullback-Leibler divergence for the ELBO objective of graph VAE models slightly differs from Thomas Kipf's [original TF implementation](https://github.com/tkipf/gae/blob/master/gae/optimizer.py). More precisely, a square and a x2 terms are missing:
- Current KL: `-0.5 * torch.mean(torch.sum(1 + logvar - mu**2 - logvar.exp(), dim=1))`
- Pytorch equivalent of original KL:`-0.5 * torch.mean(torch.sum(1 + 2*logvar - mu**2 - logvar.exp()**2, dim=1))`

I think that the original version is the correct one (see e.g. [here](https://stats.stackexchange.com/questions/7440/kl-divergence-between-two-univariate-gaussians) for the computation in 1 dimension ; taking mu_2 = 0 and sigma_2 = 1 in the final formula), and I therefore open this PR for an update.

Besides, I also propose to replace the `logvar` naming by `logstd`, as in the original implementation. Such notation would be more consistent with the mathematical terms actually used in the KL (the first one should be the log of the standard deviation ; the second one - with the square - is the variance). It would also be more consistent w.r.t. the "reparameterization trick" (see line 146 of `torch_geometric/nn/models/autoencoder.py`) where the N(0,I) random samples need to be multiplied by the standard deviation, not by the variance. I therefore updated all occurences of `logvar` by `logstd` in the project.

Best,

Guillaume

### DIFF
--- a/examples/argva_node_clustering.py
+++ b/examples/argva_node_clustering.py
@@ -26,11 +26,11 @@ class Encoder(torch.nn.Module):
         super(Encoder, self).__init__()
         self.conv1 = GCNConv(in_channels, hidden_channels, cached=True)
         self.conv_mu = GCNConv(hidden_channels, out_channels, cached=True)
-        self.conv_logvar = GCNConv(hidden_channels, out_channels, cached=True)
+        self.conv_logstd = GCNConv(hidden_channels, out_channels, cached=True)
 
     def forward(self, x, edge_index):
         x = F.relu(self.conv1(x, edge_index))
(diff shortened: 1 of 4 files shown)


rusty1s:
That's great. Thank you :)

PR #1386, opened 2020-06-30
Add PNA to pytorch_geometric

lukecavabarrett:


rusty1s:
I made a pass to the PR and cleaned up a little bit. The only major difference is that I removed the `MLP` and `FCLayer` classes. In addition, I added a ZINC example since that is what you use in your paper. It reaches about 0.222 MAE, so it's not yet fully-reproducing your results. Do you have any idea what might be missing?

gcorso:
Thanks, great! We have often seen the patience wasn't enough, could you try running it with a patience of 20 for example?

rusty1s:
Yes, that works :) It now reaches 0.1788 MAE. I'm merging this now. Thank you for contributing and once again, congratz to your work!

gcorso:
Thank you for your work and congrats on your latest model which I heard gets even better results!

PR #1411, opened 2020-07-08
add neighbor max pooling

rhsimplex:
Another addition from Deepchem:

https://github.com/deepchem/deepchem/blob/2.3.0/deepchem/models/layers.py#L133

Adds a pooling operation that replaces each feature with the max from itself and all neighbors.


rusty1s:
Hi and thank you. I am not sure I understand the intuition behind this PR though. Isn‘t that already supported by `MessagePasssing` with `max` aggregation? What‘s the use-case of this function?

rhsimplex:
I think you mean like:

```python
from torch_geometric.nn import MessagePassing

class NeighborMaxPool(MessagePassing):
    def __init__(self):
        super().__init__(aggr='max')
        edge_index, _ = add_self_loops(edge_index) 
        return self.propagate(x=x, edge_index=edge_index)
```

Indeed this gives the same result, though marginally slower on CPU.

```python
>>> t = NeighborMaxPool()
>>> (t(data.x, data.edge_index) == neighbor_max_pool(data).x).all()
tensor(True)
```

The use case is reproducing the deepchem architecture, namely: https://github.com/deepchem/deepchem/blob/master/deepchem/models/graph_models.py#L550 This architecture is used e.g. [in this paper](https://www.mdpi.com/1420-3049/25/1/44). Note the `GraphPool` layers are the subject of this PR. 

I realize this is a pretty specific functionality, so no hard feelings if you pass on this one =)

rusty1s:
Thanks for the clarification. I'm fine with adding it, will merge ASAP.

torch_geometric/nn/pool/neighbor_max_pool.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    x, edge_index = data.x, data.edge_index
+    if flow == 'source_to_target':
+        edge_index = add_self_loops(edge_index)[0]
+        x = scatter(x[edge_index[1]], edge_index[0], reduce='max', dim=0)
+    elif flow == 'target_to_source':
+        edge_index = add_self_loops(edge_index)[1]
+        x = scatter(x[edge_index[0]], edge_index[1], reduce='max', dim=0)
+
+    data.x = x
```
rhsimplex, line 22: maybe this shouldn't be an assignment here

rusty1s:
Made some small changes:
* Renamed to `max_pool_neighbor_x`
* Added `avg_pool_neighbor_x`
* Moved code to `max_pool.py` and `avg_pool.py`

Thanks for this PR :)

PR #1431, opened 2020-07-16
Make sure all Linear layers respect bias argument

rhsimplex:
Hi @rusty1s, 

When I made the #1248 PR, I had originally proposed both root and neighbor transforms have an additive bias. I'm not sure if this was intentionally hardcoded as `False` but I think it's more intuitive if both transforms respect the bias argument.

Thanks!

### DIFF
--- a/torch_geometric/nn/conv/mf_conv.py
+++ b/torch_geometric/nn/conv/mf_conv.py
@@ -44,7 +44,7 @@ class MFConv(MessagePassing):
             in_channels = (in_channels, in_channels)
 
         self.lins_l = ModuleList([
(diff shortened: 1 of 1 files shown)


rusty1s:
You're spotting is on point :P I'm not sure that's really needed though. IMO, `lin_l(h) + lin_r(x)` does not need two bias vectors, since either `lin_l` or `lin_r` can learn the overall bias added to the output.

rusty1s:
@rhsimplex What's your take on that? Why are two bias values needed here?

rhsimplex:
I think you are right, sorry!

PR #1444, opened 2020-07-21
Fix a bug in gcn_norm

jeffhj:


rusty1s:
Can you clarify? Why is that a bug?

jeffhj:
> Can you clarify? Why is that a bug?

Hi @rusty1s, from the implementation of SparseTensor on https://github.com/rusty1s/pytorch_sparse/blob/master/torch_sparse/tensor.py, I found that fill_value_() is an in-place operation, but fill_value() not.

rusty1s:
That's true. But I don't think we want to modify the `SparseTensor` in-place, since that would result in unexpected behaviour outside this operator.

rusty1s:
Oh, but it doesn't get assigned to a new tensor. You're right. Thats a bug!

jeffhj:
> Oh, but it doesn't get assigned to a new tensor. You're right. Thats a bug!

Yes, if we don't use in-place operation, we should use `adj_t = adj_t.fill_value(1., dtype=dtype)` instead.

rusty1s:
Thanks :)

PR #1453, opened 2020-07-23
gat sampling example using multi-gpu training

maqy1995:


rusty1s:
Awesome :) Sorry that it took me so long to merge.

PR #1471, opened 2020-07-27
Layer Norm

aluo-x:
A layer norm layer with jit support


aluo-x:
Compared to the code in the post:

* Fixed the lack of eps. 
* Added type annotations so it works with jit.script (I don't think InstanceNorm supports jit)
* Added if condition on the affine. Looking at the code in pytorch cyclegan, I don't think the original approach of disabling gradients on the parameter would have worked. Since they toggle gradients in feedforward passes for all parameters

PR #1479, opened 2020-07-28
Performance improvements to line_graph transform

murnanedaniel:
Howdy,

I mentioned this upgrade a while ago - I need to convert "doublet" graphs into "triplet" (line) graphs very quickly, and have found that doing this with sparse matrices to be much faster:

image

Doing this on CPU, the fastest way is with Scipy, and on the GPU the fastest way is with SparseTensor. I added a check in the method for CPU/GPU,
```
if row.device.type == 'cuda': 
```
but I'm not sure if there's a more elegant way to do this. FYI, doing it with SparseTensor on CPU is _much_ slower, hence the check. For now, this PR is just a direct conversion of the original implementation. In future, I'd love to add some flexibility to what the line graph node features are. E.g. Concatenations of the original node features along each edge, additions, etc. 

P.S. I changed the functionality of the undirected graph to match the directed graph, so now both can refer directly back to the original edge list. See the test to see how this works. I think this is more useful than relabelling the edges as in the original implementation.


rusty1s:
Thank you! I will have a closer look around next week and will merge ASAP :)

PR #1514, opened 2020-08-07
Add GraphSAGE example

yuanx749:
An unsupervised GraphSAGE example using random walk sampling.


examples/unsup-sage-rw.py
```diff
@@ -0,0 +1,152 @@
...
+        data.edge_index, node_idx=rw_idx,
+        sizes=num_samples, batch_size=batch_size * walk_length, shuffle=False,
+        num_workers=0)
+
+    neg_loader = NeighborSampler(
+        data.edge_index, node_idx=node_idx,
+        sizes=num_samples, batch_size=batch_size, shuffle=True,
+        num_workers=0)
+
+    for (batch_size_, u_id, adjs_u), (_, v_id, adjs_v), (_, vn_id, adjs_vn) in\
```
wohlbier: I am running this script on my own data. I am getting node id's that are out of bounds. I'm not sure how to track this down. Any suggestions?

examples/unsup-sage-rw.py
```diff
@@ -0,0 +1,152 @@
...
+    pbar.set_description(f'Epoch {epoch:02d}')
+
+    total_loss = 0
+
+    node_idx = torch.randperm(data.num_nodes)  # shuffle all nodes
+    train_loader = NeighborSampler(
+        data.edge_index, node_idx=node_idx,
+        sizes=num_samples, batch_size=batch_size, shuffle=False,
+        num_workers=0)
+
```
wohlbier: I'm getting values in `rw_idx` that are greater than `num_nodes`. I added `data.num_nodes` to my call to `random_walk`. The values look to me to be the result of arrays reading memory they shouldn't be.

yuanx749: Weird, it seems to be a pytorch issue. Could you test ```random_walk``` seperately? What is the dtype of ```data.edge_index```?

wohlbier: I'll have to think about how to test `random_walk` separately. The type of `data.edge_index` is `<class 'torch.Tensor'>`.

I commented about this on a closed bug https://github.com/rusty1s/pytorch_cluster/issues/81#issuecomment-696927434. @rusty1s suggests that it has to do with self loops, but I have not had time to try it out yet.

PR #1519, opened 2020-08-08
[WIP] TGN: Temporal Graph Networks

rusty1s:


benelot:
@rusty1s @emalgorithm Hello guys! Thanks for your great work on integrating TGNs into pytorch geometric. It will certainly improve the outreach of such a great model!

I am particularly interested in dynamic node classification, for which only recently an example was written on the original tgn repository. Is there an example for this here as well? If not, is it on the roadmap somewhere? Thanks!

PR #1535, opened 2020-08-16
train_test_split_edges uses single random source

kessido:
current implementation of train_test_split_edges uses multiple randomness sources which harden the ability to reproduce experiment results. Given the name "torch_geometric" one would expect having set torch randomness state, that the produced split would be the same -> but this is not the case!

As such I suggest using only torch.randperm as the randomness source for the function.

Ido

### DIFF
--- a/torch_geometric/utils/train_test_split_edges.py
+++ b/torch_geometric/utils/train_test_split_edges.py
@@ -1,5 +1,4 @@
 import math
-import random
 import torch
 from torch_geometric.utils import to_undirected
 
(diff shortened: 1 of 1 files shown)


rusty1s:
I fully agree. The previous sampling technique has the advantage that it does not have to materialize `row.size(0)` entries, and hence is a little bit more memory-friendly.

kessido:
Given that `row` itself already exist, materializing a vector of similar size seem reasonable, but it is your call.
Ido

rusty1s:
That sounds good to me :) Thanks!

PR #1555, opened 2020-08-20
Parallel approximate PPR

klicperajo:
This PR parallelizes the approximate PPR used by GDC.

It also fixes a bug in the indptr, which was missing its last element (no idea why this didn't show up earlier).


rusty1s:
Awesome, thank you :)

PR #1603, opened 2020-09-04
Fix gnn_explainer bug

yuanx749:
It seems ```mean``` should be applied to feature mask here. The code below is from the original repo of [gnnexplainer](https://github.com/RexYing/gnn-model-explainer/blob/master/explainer/explain.py#L763).
```        feat_size_loss = self.coeffs["feat_size"] * torch.mean(feat_mask)```

If ```sum``` is used, this loss term becomes very large compared to other loss terms. I use heatmap to visualize the output ```node_feat_mask```, the importance of features are indistinguishable. After it is changed to ```mean```, important features could be identified.

### DIFF
--- a/torch_geometric/nn/models/gnn_explainer.py
+++ b/torch_geometric/nn/models/gnn_explainer.py
@@ -123,7 +123,7 @@ class GNNExplainer(torch.nn.Module):
         loss = loss + self.coeffs['edge_ent'] * ent.mean()
 
         m = self.node_feat_mask.sigmoid()
-        loss = loss + self.coeffs['node_feat_size'] * m.sum()
+        loss = loss + self.coeffs['node_feat_size'] * m.mean()
         ent = -m * torch.log(m + EPS) - (1 - m) * torch.log(1 - m + EPS)
         loss = loss + self.coeffs['node_feat_ent'] * ent.mean()
 
--- a/torch_geometric/nn/models/gnn_explainer.py
+++ b/torch_geometric/nn/models/gnn_explainer.py
@@ -44,7 +44,9 @@ class GNNExplainer(torch.nn.Module):
 
     coeffs = {
         'edge_size': 0.005,
+        'edge_reduction': 'sum',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by danilonumeroso** [comment]
Actually, it depends on the domains. There might be cases where ```torch.mean``` would result in a too low value compared to the other terms. As a rule of thumb, you should fix the importance of loss terms by tuning the associated hyper-parameter (i.e, ```self.coeffs['feat_size']``` in this case). 

If you really want to define an alternative loss function (needed in some cases) you can just extend the class and override the method. 
Substituting ```sum``` for ```mean``` is not likely to solve the issue for **every** situation.

Btw I'm not maintaining this repository, that's only my opinion :D

**Comment by rusty1s** [comment]
Thanks for the PR @yuanx749 and thanks for the clarification @danilonumeroso . It indeed looks like the official GNNExplainer code uses `mean` by default, and I feel like we should, too. How about we add this behaviour as an argument to `GNNExplainer`, similar to what PyTorch is doing for its loss functions, *e.g.*, [`nll_loss`](https://pytorch.org/docs/stable/nn.functional.html?highlight=nll_loss#torch.nn.functional.nll_loss) via `reduction="mean"`?

**Comment by yuanx749** [comment]
@danilonumeroso Thank you. I notice that the coefficient of ```edge_size``` is 0.005 while ```node_feat_size``` is 1.0, the same as the official code. I think that is why they use ```sum``` in the edge mask size loss term but ```mean``` in the node feature size loss term.

So maybe these ```coeffs``` could be added as arguments for user to tune, the default values following the official setting? @rusty1s

**Comment by rusty1s** [comment]
You should be able to tune them via
```
GNNExplainer.coeffs['node_feat_size'] = 1.0
```

yuanx749:
> You should be able to tune them via
> 
> ```
> GNNExplainer.coeffs['node_feat_size'] = 1.0
> ```

Yep, of course I know. Just an opinion.

PR #1636, opened 2020-09-14
Update negativesampling.py

hvplus:
Solved a bug which leads to negative alpha when num_neg_samples and size is equal


rusty1s:
Thanks!

PR #1652, opened 2020-09-20
Fix integer division of tensors

howardjp:
Python 3.8 changes how integer divison of tensors works.  This
patch is necessary to run the RENet example on Colab.


rusty1s:
Awesome, thank you!

PR #1656, opened 2020-09-21
Linear Graph AE and VAE

GuillaumeSalha:
Dear @rusty1s, 

Once again, congratulations to you and your collaborators for this amazing work!

I am the lead author of the article [Keep It Simple: Graph Autoencoders Without Graph Convolutional Networks](https://arxiv.org/abs/1910.00942), presented at the NeurIPS 2019 workshop on Graph Representation Learning. In this article, as well as in an [extended conference version](https://arxiv.org/abs/2001.07614) recently accepted at ECML-PKDD 2020, we studied simpler versions of Graph Autoencoders (AE) and Graph Variational Autoencoders (VAE), referred to as **Linear Graph AE and VAE**.

In a nutshell, we replaced the multi-layer GCN encoders from standard models (Kipf and Welling 2016) by simpler _linear models_ w.r.t. the adjacency matrix of the graph (`Z = AW`, or `Z = AXW` in case of node-level features `X`), involving a unique weight matrix to tune and no activation function. We showed that they often appear as effective and simpler alternatives to GCN-based models, for various graphs including the widely used Cora, Citeseer and Pubmed citation networks, achieving similar or even better performances on downstream tasks such as link prediction or node clustering. In our work, we also tried to provide insights on _when_ to use multi-layer GCN encoders and when, on the contrary, these encoders might actually be unnecessarily complex.

We released a [TensorFlow implementation](https://github.com/deezer/linear_graph_autoencoders) of Linear Graph AE and VAE a few months ago. However, some users recently asked for a PyTorch one, and I think that your `pytorch-geometric` project is the best place to include it.

Therefore, I am opening this PR to propose a modification of `examples/autoencoder.py`, aiming at including a working example of linear graph AE and VAE with application to link prediction, building upon the existing GAE and VGAE classes. With the proposed implementation, I manage to reach better results than standard GAE/VGAE on the provided Cora, Citeseer and Pubmed datasets from the example file, while being faster. So, I sincerely think this can be of interest for some future users. I also updated the README file accordingly.

Best,

Guillaume


examples/autoencoder.py
```diff
@@ -13,20 +13,22 @@
 parser = argparse.ArgumentParser()
 parser.add_argument('--model', type=str, default='GAE')
 parser.add_argument('--dataset', type=str, default='Cora')
+parser.add_argument('--epochs', type=int, default=400)
```
GuillaumeSalha, line 16: By the way, I added an argument for the number of training iterations, instead of the previous hardcoded value.

examples/autoencoder.py
```diff
@@ -42,10 +44,31 @@ def forward(self, x, edge_index):
         elif args.model in ['VGAE']:
             return self.conv_mu(x, edge_index), self.conv_logstd(x, edge_index)
 
+          
+# Linear encoder from linear graph AE and VAE models: https://arxiv.org/abs/1910.00942
```
GuillaumeSalha, line 48: At first, I considered putting all models in the existing (only) `Encoder` class, but quickly realized that I would end up with several if conditions (e.g. as the layer sizes differ), and that creating two classes improves the overall readability.

rusty1s:
Thank you so much for this PR. This looks awesome. I will take a closer look after ICLR deadline and will merge ASAP.

rusty1s:
Thank you :)

PR #1682, opened 2020-09-29
Update data_parallel.py

max-yue:


rusty1s:
Not really sure why this needs to be fixed. Can you elaborate?

PR #1687, opened 2020-09-30
Fix the indices when calculating angles

zetayue:
As described in the original paper (Figure 1), the angles are calculated between m_ji and m_kj. However, the implementation here calculates the angle between m_ji and m_ki mistakenly. 
The same error happens in the code provided by the authors. But it has been corrected already: https://github.com/klicperajo/dimenet/blob/master/dimenet/model/dimenet.py#L89

### DIFF
--- a/torch_geometric/nn/models/dimenet.py
+++ b/torch_geometric/nn/models/dimenet.py
@@ -444,9 +444,9 @@ class DimeNet(torch.nn.Module):
 
         # Calculate angles.
         pos_i = pos[idx_i]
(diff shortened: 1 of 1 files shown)


rusty1s:
Yeah, I also noticed that, but by fixing it it we no longer can make use of the pre-trained models.

zetayue:
Sorry for my delayed response, I will see if I can prepare the pre-trained models with the correct indices.

rick433 (commented):
If I am not wrong, you need to declare pos_j as pos[j] in your revision.

PR #1707, opened 2020-10-08
Update softmax.py

Book1996:
I propose a new version of softmax function in Utils module. The version can reduce  a little running time by 3 times, ^ _ ^.

```python
from torch import Tensor
from torch_scatter import scatter
import torch
from time import time

# set
N = 4000
src = torch.rand(N*(N+1)//2)
index = []
for i in range(N):
    tmp = [i]*(i+1)
    index = index+tmp
index = torch.tensor(index,dtype=torch.long )

# norm
t1 = time()
out = src
out = out.exp()
out_sum = scatter(out, index, dim=0, dim_size=N, reduce='sum')[index]
t2 = time()

# local  max
local_out = src - scatter(src, index, dim=0, dim_size=N, reduce='max')[index]
local_out = local_out.exp()
local_out_sum = scatter(local_out, index, dim=0, dim_size=N, reduce='sum')[index]
t3 = time()

# global  max
global_out = src - src.max()
global_out = global_out.exp()
global_out_sum = scatter(global_out, index, dim=0, dim_size=N, reduce='sum')[index]
t4 = time()
print('norm', t2 - t1)
print('local softmax', t3 - t2)
print('global softmax', t4 - t3)
a = out / (out_sum + 1e-16)
b = local_out / (local_out_sum + 1e-16)
c =  global_out / (global_out_sum + 1e-16)
eps = 0.0000001
print((torch.abs(a-c)>eps).sum())
print((torch.abs(a-b)>eps).sum())
# =============================================================================
# result
# norm 0.04488039016723633
# local softmax 0.1874983310699463
# global softmax 0.06482648849487305
# tensor(0)
# tensor(0)
# =============================================================================


```


rusty1s:
Thanks :)

PR #1711, opened 2020-10-10
Fix typo in total graphs in ZINC subset

vijaydwivedi75:
The ZINC subset used in [Benchmarking GNNs](http://arxiv.org/abs/2003.00982) has 12,000 molecular graphs - 10,000 train/1,000 val/1,000 test.

PR for this minor correction.


rusty1s:
Yeah, thanks :)

PR #1717, opened 2020-10-12
Accept custom aggregate method for convolution layers

m30m:
Set the aggr method only if it is not already set by the user

### DIFF
--- a/torch_geometric/nn/conv/agnn_conv.py
+++ b/torch_geometric/nn/conv/agnn_conv.py
@@ -37,7 +37,8 @@ class AGNNConv(MessagePassing):
     """
     def __init__(self, requires_grad: bool = True, add_self_loops: bool = True,
                  **kwargs):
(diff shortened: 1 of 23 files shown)


rusty1s:
Thanks for the PR. While I agree that the `aggr` keyword is missing for a few operators, I feel like allowing the user to modify the aggregation type for any kind of operator is not particular useful, *e.g.*, for `GCNConv` or `GATConv` which are defined to perform weighted mean normalization.

m30m:
Thanks for the quick reply. Yes, I was also not sure which operators need this and if this is valid for each operator or not.
That's why I didn't include it in the function signature directly, so the users don't see this option by default but the expert users who know such options exist can override the default `aggr`.
Still, I agree that it may make no sense for some operators.
For `GCNConv`, the user can disable the normalization by setting `normalize=False` and maybe it can be helpful in that scenario.

rusty1s:
I see. That makes sense, thanks a lot :)

PR #1731, opened 2020-10-16
Use indices in shortest_path() in seal_link_pred.py

muhanzhang:
Previously the __shortest_path()__ calculates shortest path distances between all pairwise nodes, but we only need the shortest path distances from all nodes to the src and dst nodes. To avoid redundant computation, I specify _indices_ in __shortest_path()__ directly. This brings 3 times speed up on Cora.


rusty1s:
Thanks :)

Issue #1699, opened 2020-10-06
Understanding examples/link_pred.py


reshinthadithyan:
## ❓ Questions & Help

I have a question regarding the Link Prediction Example provided. There are three Edge Indices involved in the process.  

> 1) train_pos_edge_index 
> 2) pos_edge_index
> 3) neg_edge_index

My understanding is that _train_pos_edge_index_ is the actual edge indices of the Graph. _pos_edge_index_ corresponds to the invisible links which aren't present in _train_pos_edge_index_ . _neg_edge_index_ are the negative edges which shouldn't be constructed.
### what happens when deploying?
Now when deploying I'll have something like _test_pos_edge_index_ which constructs the actual Graph. Now what about _pos_edge_index_ ? Will that be all possible edges excluding _test_pos_edge_index_ . My trained model will assign a probability for all the edges and I'll end up with having my new links right?

rusty1s:
When performing inference, one does not need to make use of `pos_edge_index` and `neg_edge_index` since those tensors are only used for training the model. During inference, one just needs to input the actual graph, and find all missing positive edges by looking at the output probabilistic adjacency matrix.

reshinthadithyan:
Then, from link_pred.py it should look something like this?
 
    def forward(self, data):

        x = F.relu(self.conv1(data.x, data.infer_pos_edge_index))
        x = self.conv2(x, data.infer_pos_edge_index)
        return x
The Feature Extractor is fine, How would I create such probabilistic adjacency matrix? Can you show me a small pseudo code?

tchaton:
Hey @reshinthadithyan,

I am trying to do such things there: https://github.com/tchaton/lightning-geometric.
Any help would be appreciated.

Best regards,
T.C

reshinthadithyan:
@rusty1s  Still not able to figure it.

rusty1s:
You can use the [`Decoder`](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/nn/models/autoencoder.py#L12-L45) to reconstruct such probabilistic adjacency matrix.

PR #1733, opened 2020-10-16
Addition of Inference Block in examples/link_pred.py

reshinthadithyan:
The existing version of examples/link_pred.py doesn't have a proper inference example. Refer #1699, added a inference block with `InnerProductDecoder.forward_all()` to give out a probability adjacency matrix. This would give a cleaner view on usage in inference stage.


rusty1s:
Thanks a lot! I modified it a bit. In particular, I moved the encoding and decoding part to the model.

PR #1757, opened 2020-10-24
Adding batching of arbitrary length paths

HekpoMaH:
### Adding batching of arbitrary length paths and their features.

For a project of mine I needed support of arbitrarily length paths. Current behaviour was concatenating these paths into a larger list, e.g. `[[paths of graph 1], [paths of graph 2], ...]` whereas I wanted it to be `[paths of graphs 1, paths of graph 2, ...]`, i.e. similar to how `edge_index` batching was done. For further examples see `test_batch.py` (I also provide some simple tests). I also didn't want to process again the data, after it was batched.

I am open to recommendations for a different ways to implement it.

Hope anyone finds it helpful!

**EDIT1**: I decided to switch to padding, due to performance issues when using separate threads (`num_workers`>0) for loading. Currently, paths for a single datapoint (graph) should be represented as a padded tensor (|P|x|L| tensor, where, P is the set of paths, |L| is the longest path in P). Note that when padding, a padding value which is larger than the number of nodes in the batch is expected.

When batching several graphs, the padded paths for each graph are 'uppaded', so that their length matches the length of the longest path in the whole batch. Padding value is always `batch.num_nodes+1`.

### DIFF
--- a/test/data/test_batch.py
+++ b/test/data/test_batch.py
@@ -5,31 +5,32 @@ from torch_geometric.data import Data, Batch
 
 def test_batch():
     torch_geometric.set_debug(True)
+    # import pdb; pdb.set_trace()
 
     x1 = torch.tensor([1, 2, 3], dtype=torch.float)
     e1 = torch.tensor([[0, 1, 1, 2], [1, 0, 2, 1]])
-    p1 = [torch.tensor([1, 0, 2]), torch.tensor([0, 2])]
+    p1 = torch.tensor([[1, 0, 2], [0, 2, 1e9]], dtype=torch.long)
     pf1 = [torch.tensor([False]), torch.tensor([True])]
     s1 = '1'
     x2 = torch.tensor([1, 2], dtype=torch.float)
     e2 = torch.tensor([[0, 1], [1, 0]])
-    p2 = [torch.tensor([1, 0])]
+    p2 = torch.tensor([[1, 0]])
     pf2 = [torch.tensor([True])]
     s2 = '2'
 
     data = Batch.from_data_list([
-        Data(x1, e1, s=s1, paths_index=p1, paths_features=pf1),
-        Data(x2, e2, s=s2, paths_index=p2, paths_features=pf2)
+        Data(x1, e1, s=s1, paths=p1, paths_features=pf1),
+        Data(x2, e2, s=s2, paths=p2, paths_features=pf2)
     ])
 
     assert data.__repr__() == (
-        'Batch(batch=[5], edge_index=[2, 6], paths_features=[3], '
-        'paths_index=[3], s=[2], x=[5])')
+        'Batch(batch=[5], edge_index=[2, 6], paths=[3, 3]'
+        ', paths_features=[3], s=[2], x=[5])'
+    )
     assert len(data) == 6
     assert data.x.tolist() == [1, 2, 3, 1, 2]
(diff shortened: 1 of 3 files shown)


rusty1s:
Hi and thanks for this issue. This seems really tight to a specific use-case though, and I'm not sure what's the benefit of adding it directly to PyG. I hope you can clarify this :)

Furthermore, I feel like this can be already achieved in PyG, *e.g.*:
```python
class MyData(Data):
    __cat_dim__(self, key, value):
        if key == 'paths':
            return 0
       return super(Data, self).__cat_dim__(key, value)

    __inc__(self, key, value):
        if key == 'paths':
            return self.num_nodes
       return super(Data, self).__inc__(key, value)

data1 = Data(num_nodes=4, paths=...)
data2 = Data(num_nodes=5, paths=...)
```
This should work just fine, but requires all paths to have the same length.

HekpoMaH:
Hi,

1) I do agree it is indeed specific - currently, only RouteNet [1] (think MPNN, but for routes) would require such feature, but other works are likely to stem from it, e.g. I'm working on a different topic related to networking, where I need to apply learning over the routes and (perhaps simultaneously) on the graph structure.

2) The requirement of all paths to have same length is not really feasible - yes, we could pad them to the maximum length, but this maximum length can vary between graphs. E.g. G1 has padded paths (0 is padding) `[[1,2,0][1,2,3]]`, while G2 has `[[1,2,3,4,5]]`. The padding has to be increased.

Further note, as we are trying to apply NNs to a real-time system, iterating a second time over the batch, to fix the padding may not be desirable.

An option would be to allow to override the collate function of `torch_geometric.data.DataLoader`; then this PR would be obsolete, as I can just override only the collate function.

```
class DataLoader(torch.utils.data.DataLoader):
    r"""Data loader which merges data objects from a
    :class:`torch_geometric.data.dataset` to a mini-batch.

    Args:
        dataset (Dataset): The dataset from which to load the data.
        batch_size (int, optional): How many samples per batch to load.
            (default: :obj:`1`)
        shuffle (bool, optional): If set to :obj:`True`, the data will be
            reshuffled at every epoch. (default: :obj:`False`)
        follow_batch (list or tuple, optional): Creates assignment batch
            vectors for each key in the list. (default: :obj:`[]`)
    """

    def __init__(self, dataset, batch_size=1, shuffle=False, follow_batch=[],
                 **kwargs):
        super(DataLoader,
              self).__init__(dataset, batch_size, shuffle,
                             collate_fn=Collater(follow_batch), **kwargs)  # Change here
```

I'm not claiming what I am adding is going to be fantastic or widely used feature but I guess if someone wants to use/reproduce our work, I'd have to either refer them to the forked repo to install a modified version of PyG, or copy paste some good amount of torch_geometric's data handling (which was kind of ugly in my personal opinion). 

Lastly, I'm open to suggestions how to handle my 'issue' in a nice way.

[1] Rusek, K., Jose, S. V., Mestres, A., Pere, B. R., & Albert, C. A. (2019). Unveiling the potential of Graph Neural Networks for network modeling and optimization in SDN. SOSR 2019 - Proceedings of the 2019 ACM Symposium on SDN Research, 140–151. https://doi.org/10.1145/3314148.3314357

rusty1s:
Thanks for your explanation. Note that you can already overwrite `collate_fn` by making use of the original `torch.utils.data.DataLoader`. In the end, this is exactly what we are already doing for `torch_geometric.data.DataLoader`.

PR #1784, opened 2020-11-01
[examples] Rename logits to log_probs

desktable:
This is a minor fix for better code readability.

In examples/cora.py, the model outputs log probabilities, rather than logits (logarithm of odds ratios). This PR changes the variable name from `logits` to `log_probs`.


rusty1s:
Awesome :)

PR #1785, opened 2020-11-01
Added option to directly specify k for topk_pool and some pooling methods.

mrmotallebi:
Small modification to directly specify the number of nodes for Top-K Pooling. Instead of looking at the ratio of pooling, directly specify the number of nodes to keep, overriding the ratio.
Put K as the last parameter in order to not break other functions.

### DIFF
--- a/test/nn/pool/test_asap.py
+++ b/test/nn/pool/test_asap.py
@@ -24,8 +24,7 @@ def test_asap():
         assert out[0].size() == (num_nodes // 2, in_channels)
         assert out[1].size() == (2, 4)
 
-        pool = ASAPooling(in_channels, ratio=2, GNN=GNN,
-                          add_self_loops=False)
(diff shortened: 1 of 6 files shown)


rusty1s:
Thank you very much! This looks good :)

One minor thing: Instead of adding an additional argument `k`, how about we allow `ratio` to represent both float values (in interval [0, 1]) and integer values.

mrmotallebi:
That actually sounds like a good idea! My only concern would be that the term "ratio" would not be easily understandable as the number of nodes to maintain (k), but I guess the current usecase (ratio of total number of nodes) is more common and if we add the explanation for the other one in the docs, it should be good.
I'll update the code.

mrmotallebi:
Sorry for the delay. Updated the code and docs, and added the tests. Please have a look.

rusty1s:
Thank you very much! I've made a pass and also fixed the `k > num_nodes` error.

Issue #1847, opened 2020-11-21
Convolution in digraphs


vymao:
## 🚀 Feature

[This](https://arxiv.org/pdf/2004.13970.pdf) is a paper that describes the convolutional operator applied to directed graphs, and [this ](https://proceedings.neurips.cc/paper/2020/file/cffb6e2288a630c2a787a64ccc67097c-Paper.pdf) is a better implementation. The results seem good for directed graphs. Would be great if these could be implemented; it seems the latter has already been done, but it would be great to make it official. 

Do any layers here handle address spectral-based convolution in digraphs?

rusty1s:
This looks super interesting. I added it to the PyG [roadmap](https://github.com/rusty1s/pytorch_geometric/issues/1465). If you are interested, feel free to contribute :)

reshinthadithyan:
Hello @rusty1s . I have a rough implementation intact with few changes from [DiGCN Github](https://github.com/flyingtango/DiGCN) . Message Passing Class is fine, Where should the pre-processing wrt the Receptive Fields sit within PyG?_(Should it be in torch_geometric.data like SAINTSampler?)_ I can share some rough construction wrt my understanding, please elaborate.

rusty1s:
I think it's fine to add them as additional functions to the same file, just like we do in [`GCNConv`](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/nn/conv/gcn_conv.py#L30).

PR #1850, opened 2020-11-22
Addition of DiGCN and DiGCN_Conv

reshinthadithyan:
Hello. As talked in #1847 . I've added digcn_conv.py(message passing) and digcn.py(model) respectively. Have a look and tell if the structure is fine, documentation needs to be done. I thought I'll show you a look onto the structure first.
### Need for a Model
The reason for having an encoder style model is that the major parts of the paper are:
1) Addition of proximity matrices 
2) Fusing the hidden states in each proximity,(refer nn\model\digcn.py)
Major parts of the code is taken from [DiGCN Github](https://github.com/flyingtango/DiGCN) .
If the structure works for PyG, I'll add documentation and sanitize the arguments. Thanks.


rusty1s:
Looks good to me :)

reshinthadithyan:
Good then, I'll add documentation and will update. Thanks.

PR #1854, opened 2020-11-23
Basic GGCN edge types support

mike-grayhat:
This is a basic implementation of edge type support for GGCN. At the moment it doesn't support fused operation since message_and_aggregate accepts lifted adjacency matrix, but in case of multiple edge types it should accept list of sparse tensors
```python
    def message_and_aggregate(self, adj_t: SparseTensor, x: Tensor, edge_attr: OptTensor) -> Tensor:
        return matmul(adj_t, x, reduce=self.aggr)
```


rusty1s:
Thank you very much! I will have a look ASAP.

PR #1856, opened 2020-11-24
Addition of DiGCN and DiGCN_Conv

reshinthadithyan:
Hello. As talked in #1847 . I've added digcn_conv.py(message passing) and digcn.py(model) respectively.  Thanks.


rusty1s:
Thank you very much. I will have a closer look ASAP.

PR #1895, opened 2020-12-05
FiLM conv layer

ldv1:
The author of the paper reported 0.992 on the PPI dataset, I haven't been able to even get decent results on that dataset (after including self loops). I do not know what kind of non-linearity he used, and how the skip connection (if any) looked like.


ldv1:
I am preparing a new version. A simple test on PPI gives 0.97 wo/ hyperparameter tuning.

rusty1s:
Ok, thanks for the update. Let me know how it goes.

ldv1:
I will make another PR.

PR #1910, opened 2020-12-09
Add WN18RR Dataset

minhtriet:
Reason: WN18 Dataset suffers from test set leakage
[Source](https://arxiv.org/abs/1707.01476): 
> In addition, it has been noted that the WN18 andFB15k datasets suffer from test set leakage, due to inverse relations from the training set being present in the test set – however, the extent of this issue has so far not been quantified.


rusty1s:
Thank you! Sorry it took me so long to merge.

PR #1912, opened 2020-12-10
Fix 2vec examples

declanmillar:
Ran into a multiprocessing bug when running the node2vec and metapath2vec examples on CPU on a MacBook.
These run fine if wrapped in a main function, so just submitting a pull request with these simple changes.
Also, wrapped the model parameters generator in a list for SparseAdam in metapath2vec.


rusty1s:
Super! Thanks a lot!

PR #1918, opened 2020-12-12
feature: add input and message dependent beta to the transformer conv

ldv1:
Hi,

1/ With the option `beta` set to `True`, the model parameter `beta` is input and message dependent, as in the paper.
2/ I extended the documentation if the dimension of edge features is passed to the constructor.
3/ with `root_weight`, it is possible to ignore everything but the message; I took `root_weight` from another conv (for consistency across convs) though I found it a bit misleading; I would prefer `target_weight` or `skip_connection`; It would be nice if every conv had this argument, so that we could easily plug the message into a RNN (as in GeniePath) without having to re-code a conv to avoid updating the target node.

regards
Laurent


rusty1s:
Thanks a lot! I indeed completely missed that `beta` is trainable. Thanks for cleaning up :)

PR #1921, opened 2020-12-14
Fix the wrong indices when filtering out molecules in QM9

zetayue:
The indices stored in the 'uncharacterized.txt' file are the numbers in the molecule names. Those indices start from 1 instead of 0. 
For example, when directly using index '58' to filter out the uncharacterized molecules in QM9, the molecule with name 'gdb_59' with be wrongly removed instead of 'gdb_58'. So that we have to minus 1 from the indices to correctly filter out the molecules.


rusty1s:
Oh shit, you are right. I just re-checked it. Thanks so much for spotting!

PR #1923, opened 2020-12-14
Feature: FiLM conv

ldv1:
This is the FiLM conv after [GNN-FiLM: Graph Neural Networks with Feature-wise Linear Modulation](https://arxiv.org/abs/1906.12192). Without hyperparameter tuning, I get 0.98 on PPI with slightly less than 0.5 Mio parameters (compared to GAT). I added examples/film.py after examples/ppi.py.


rusty1s:
I updated `FiLMConv` to support multiple relational types and added tests. To replicate the experiments in the paper, I needed to add BatchNorm to the example script. It now reaches 98.96 micro-F1 score, which is really close to the results reported. Thank you!

ldv1:
Thanks for your changes.
I noted that you hard-coded the function g as a linear function; that's what was done in the paper, and the author reported that indeed a simple linear function would perform well; I think maybe we should leave it to the user to decide what function g is best for the task at hand, e.g. a non-linear bottleneck architecture could potentially be more efficient and still does not necessitate too much parameters. Maybe we could have g as a linear function per default (covers most cases), otherwise it is user-defined.

rusty1s:
I agree. I added it back in.

PR #1955, opened 2020-12-31
Fixed from_networkx to support empty edges list

shakedbr:
In the ` from_networkx() ` function, in case the networkx graph has no edges, the operation
```
edge_index = torch.tensor(list(G.edges)).t().contiguous()
```
creates a `torch.float32` tensor which leads to an assertion failure in `nn/conv/message_passing.py`:
```
    def __check_input__(self, edge_index, size):
        the_size: List[Optional[int]] = [None, None]

            assert edge_index.dtype == torch.long
```

The fix:
Creating `LongTensor` explicitly:
```
edge_index = torch.LongTensor(list(G.edges)).t().contiguous()
```


rusty1s:
Awesome, thanks!

PR #1959, opened 2021-01-01
small change to support "softmax" of 3D tensor or batched graphs (same adjacency matrix)

shi27feng:
as the title


rusty1s:
Thanks. I additionally added the `dim` attribute to the `softmax` call to make its usage more intuitive.

PR #1971, opened 2021-01-04
Fixed Data attribute check in GCNNorm

shakedbr:
The original check `data.edge_index is not None`  raises an `AttributeError: 'Data' object has no attribute 'edge_index'` in case it's missing from the object.


rusty1s:
Thanks!

PR #2000, opened 2021-01-12
Fix typos in torchscript tests.

dongkwan-kim:
Many conv/ tests do not pass jittable instances into the corresponding assert lines.

If I missed something or made a mistake, I would appreciate it if you let me know.

Thank you!


rusty1s:
Ooops, thank you so much :) Awesome!

PR #2023, opened 2021-01-17
Mutliple graph sampling from Batch

ita9naiwa:
Hi. I'm wondering about implementing sampling k subgraphs from the batch.

#### Use case
- want to perform graph classification with n graphs but each graph is relatively small (<= 100 nodes)

I wrote some base code, but am not sure 

- This feature is considered to be useful for other people
- It is okay to implement things like the code I've already written. If it isn't I'd like to want help or some guidelines.


rusty1s:
Thanks for the PR. I'm just wondering if this functionality is really needed, since you can achieve the same via:
```python
Batch.from_data_list(batch[idx])
```

ita9naiwa:
> Thanks for the PR. I'm just wondering if this functionality is really needed, since you can already achieve the same via:
> 
> ```python
> Batch.from_data_list(batch[idx])
> ```

I didn't come up with this usage of Batch. You're right this functionality is redundant..

PR #2024, opened 2021-01-18
Add SuperGATConv

dongkwan-kim:
I added SuperGAT, which will be presented at ICLR 2021. ([Kim and Oh, How to Find Your Friendly Neighborhood: Graph Attention Design with Self-Supervision](https://openreview.net/forum?id=Wi5KUNlqWty))

One thing I want to consult is TorchScript support. The current implementation contains methods that do not support TorchScript (such as `negative_sampling`, `dropout_adj`), thus I separate and wrap them with `@torch.jit.ignore`. This is my first time using TorchScript, and I am not sure this is appropriate in PyTorch Geometric.

Thanks in advance.


rusty1s:
Thanks and congratulations to your paper acceptance :)
This looks super good already. Please don't worry about TorchScript support, I will make it work (probably after ICML deadline, sorry for the delay).

dongkwan-kim:
Thank you! It couldn't be done without PyTorch Geometric.

And also good luck with your submission.

rusty1s:
Thanks once again for this PR! It's always a pleasure to see PyG utilized so well :)

I really like `SuperGAT` and the idea of guiding attention via self-supervision is really clever. Results are pretty good, too. Congratz!

I made a pass through the PR and modified the following things:
* Moved `SuperGAT` into its own example.
* Reformulated attention loss computation: This method does now return the attention loss for a single layer. I think this is more intuitive.
* Made same things more efficient (e.g., separting `att` into `att_l` and `att_r` to avoid the usage of `torch.cat`).

I will look into JIT compatibility later, as this is not critically for merging.

dongkwan-kim:
Thank you, it really looks good to me. 
I always learn something when I make a pull request to PyG.

Also, I made some small improvements in `examples/super_gat.py` with a new pull request.
- made `Net` use `x` and `edge_index` in a local scope (like `examples/gat.py`)
- changed a hyperparameter.

Issue #1801, opened 2020-11-06
When num_hyperedges is greater than num_nodes, HypergraphConv cannot work


liulixinkerry:
## 🐛 Bug

When num_hyperedges`M` is greater than num_nodes`N`, HypergraphConv cannot work.

## To Reproduce

When I add hyperedge `4` in https://github.com/rusty1s/pytorch_geometric/blob/master/test/nn/conv/test_hypergraph_conv.py , the program crashes.
```python
  
import torch
from torch_geometric.nn import HypergraphConv

def test_hypergraph_conv():
    in_channels, out_channels = (16, 32)
    hyperedge_index = torch.tensor([[0, 0, 1, 1, 2, 3, 3, 3, 2, 1 ,2], [0, 1, 2, 1, 2, 1, 0, 3, 3, 4, 4]])
    num_nodes = hyperedge_index[0].max().item() + 1
    x = torch.randn((num_nodes, in_channels))

    conv = HypergraphConv(in_channels, out_channels)
    assert conv.__repr__() == 'HypergraphConv(16, 32)'
    out = conv(x, hyperedge_index)
    assert out.size() == (num_nodes, out_channels)


    conv = HypergraphConv(in_channels, out_channels, use_attention=True,
                          heads=2)
    out = conv(x, hyperedge_index)
    assert out.size() == (num_nodes, 2 * out_channels)

    print("finish")

if __name__ == "__main__":
    test_hypergraph_conv()
```

log:
```bash
Traceback (most recent call last):
  File "test.py", line 24, in <module>
    test_hypergraph_conv()
  File "test.py", line 12, in test_hypergraph_conv
    out = conv(x, hyperedge_index)
  File "/home/user/anaconda3/lib/python3.8/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/home/user/anaconda3/lib/python3.8/site-packages/torch_geometric/nn/conv/hypergraph_conv.py", line 126, in forward
    out = self.propagate(hyperedge_index, x=x, norm=B, alpha=alpha)
  File "/home/user/anaconda3/lib/python3.8/site-packages/torch_geometric/nn/conv/message_passing.py", line 252, in propagate
    out = self.aggregate(out, **aggr_kwargs)
  File "/home/user/anaconda3/lib/python3.8/site-packages/torch_geometric/nn/conv/message_passing.py", line 286, in aggregate
[...]

PR #2027, opened 2021-01-19
Fixed issue for num_edges > num_nodes in hypergraph_conv

THinnerichs:
PR for issue #1801. Sorry for the delay.

I am insecure whether this works for multiple heads, too, but it works for plain hyperconvolution.

### DIFF
--- a/test/nn/conv/test_hypergraph_conv.py
+++ b/test/nn/conv/test_hypergraph_conv.py
@@ -2,10 +2,10 @@ import torch
 from torch_geometric.nn import HypergraphConv
 
 
-def test_hypergraph_conv():
+def test_hypergraph_conv_with_more_nodes_than_edges():
     in_channels, out_channels = (16, 32)
     hyperedge_index = torch.tensor([[0, 0, 1, 1, 2, 3], [0, 1, 0, 1, 0, 1]])
-    hyperedge_weight = torch.tensor([1, 0.5])
+    hyperedge_weight = torch.tensor([1.0, 0.5])
     num_nodes = hyperedge_index[0].max().item() + 1
     x = torch.randn((num_nodes, in_channels))
 
@@ -25,5 +25,21 @@ def test_hypergraph_conv():
 
     conv = HypergraphConv(in_channels, out_channels, use_attention=True,
                           heads=2, concat=False, dropout=0.5)
+    out = conv(x, hyperedge_index, hyperedge_weight)
+    assert out.size() == (num_nodes, out_channels)
+
+
+def test_hypergraph_conv_with_more_edges_than_nodes():
+    in_channels, out_channels = (16, 32)
+    hyperedge_index = torch.tensor([[0, 0, 1, 1, 2, 3, 3, 3, 2, 1, 2],
+                                    [0, 1, 2, 1, 2, 1, 0, 3, 3, 4, 4]])
+    hyperedge_weight = torch.tensor([1.0, 0.5, 0.8, 0.2, 0.7])
+    num_nodes = hyperedge_index[0].max().item() + 1
+    x = torch.randn((num_nodes, in_channels))
+
+    conv = HypergraphConv(in_channels, out_channels)
+    assert conv.__repr__() == 'HypergraphConv(16, 32)'
+    out = conv(x, hyperedge_index)
+    assert out.size() == (num_nodes, out_channels)
     out = conv(x, hyperedge_index, hyperedge_weight)
     assert out.size() == (num_nodes, out_channels)
--- a/torch_geometric/nn/conv/hypergraph_conv.py
+++ b/torch_geometric/nn/conv/hypergraph_conv.py
@@ -123,10 +123,15 @@ class HypergraphConv(MessagePassing):
         if hyperedge_weight is not None:
             B = B * hyperedge_weight
 
+        dif = max([num_nodes, num_edges]) - num_nodes # get size of padding 
+        x_help = F.pad(x, (0, dif)) # create dif many nodes
+
         self.flow = 'source_to_target'
-        out = self.propagate(hyperedge_index, x=x, norm=B, alpha=alpha)
+        out = self.propagate(hyperedge_index, x=x_help, norm=B, alpha=alpha)
         self.flow = 'target_to_source'
         out = self.propagate(hyperedge_index, x=out, norm=D, alpha=alpha)
+
+        out = out[:num_nodes] # prune back to original x.size()
 
         if self.concat is True:
             out = out.view(-1, self.heads * self.out_channels)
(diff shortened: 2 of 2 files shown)


THinnerichs:
And I spotted the my first mistake, as ```num_nodes``` is not present in ```forward()```, which should be substituted by ```x.size(0)```

xnhp:
Just ran into the same scenario. Thanks a lot figuring out the issue and providing a fix.

I am not immediately sure however, whether that fixes the issue. On my run (with `use_attention=False`), I get the following stacktrace:

```
  File "C:\Users\Ben\PycharmProjects\thesis-playground\models\gcn.py", line 33, in forward
    h = self.conv1(x, edge_index)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch\nn\modules\module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\hypergraph_conv.py", line 131, in forward
    out = self.propagate(hyperedge_index, x=x_help, norm=B, alpha=alpha)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\message_passing.py", line 237, in propagate
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\hypergraph_conv.py", line 84, in message
    -1, self.heads, self.out_channels)
RuntimeError: shape '[-1, 1, 4]' is invalid for input of size 125
```

There's a good chance that's my fault but wanted to comment nevertheless. I'll investigate further.

xnhp:
Just ran into the same situation. Thanks a lot for investigating the issue and providing a patch.

I'm not immediately sure this really fixes the issue. On my test run (with our patch applied, and `use_attention=False`, I still get the following error stacktrace:

```
  File "C:\Users\Ben\PycharmProjects\thesis-playground\data\test_biomodels.py", line 37, in test_m_greater_n
    self.do_one_pass("small_model-more-edges.xml")
  File "C:\Users\Ben\PycharmProjects\thesis-playground\data\test_biomodels.py", line 25, in do_one_pass
    h = mod_hyper(data.x, data.edge_index)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch\nn\modules\module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\Users\Ben\PycharmProjects\thesis-playground\models\gcn.py", line 33, in forward
    h = self.conv1(x, edge_index)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch\nn\modules\module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\hypergraph_conv.py", line 131, in forward
    out = self.propagate(hyperedge_index, x=x_help, norm=B, alpha=alpha)
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\message_passing.py", line 237, in propagate
  File "C:\Users\Ben\.conda\envs\thesis-playground\lib\site-packages\torch_geometric\nn\conv\hypergraph_conv.py", line 84, in message
    -1, self.heads, self.out_channels)
RuntimeError: shape '[-1, 1, 4]' is invalid for input of size 125
```

THinnerichs:
Hey there, 
the error lies within the torch padding as it pads the 1st dimension and not the 2nd one. I will push my stuff tomorrow around lunchtime :)

xnhp:
Heyo,

thanks for the quick reply. I just realised the same.

That's great, thanks a lot!

THinnerichs:
Fixed it :)

xnhp:
@THinnerichs I can imagine your time is limited, but if you'd please, could you give me a few sentences on what the actual problem is? I happened upon the same scenario and tried finding the bug but wasn't successful. I read your bug report but couldn't really understand what the actual issue is. Would appreciate a more elaborate explanation so I can learn.

THinnerichs:
Hey @xnhp ,
You are welcome! I will try my best. :) 

In general in PyG edges are represented in the following way:
```
[[0,0,0,1,2,3],
[1,2,3,0,0,0]]
```
denotes a star-like structure with 4 nodes, one (0) connected to the other three, and vice versa. Basically, we have an ordered starting point list and an ordered target point list. If we know apply our convolution all messages are transferred from the each starting node towards its target. 

In the hypergraph scenario $G=(V,E)$ with $V=\{0,1,2,3\}$ and $E=\{\{0,1,2\}, \{1,2,3\}\}$ with undirected hyperedges, edges are represented as:
```
[[0,1,2,1,2,3],
[0,0,0,1,1,1]]
```
denoting that ```{0,1,2,3]``` belongs to the 0th edge, and ```{1,2,3]``` is the 1st edge, and so on for more edges. While this representation is rather counter-intuitive and clunky, it has one crucial upside: we can just the original message passing scheme. As mentioned, in ```hypergraph_conv``` we do

```
        self.flow = 'source_to_target'
        out = self.propagate(hyperedge_index, x=x, norm=B, alpha=alpha)
        self.flow = 'target_to_source'
        out = self.propagate(hyperedge_index, x=out, norm=D, alpha=alpha)
```
which is basically moving all messages forward and then backwards. W.r.t. to the hyperedge representation, all messages are transferred from their original node e.g. ```{0,1,2}``` in the first edge towards their target node, i.e. node 0, by propagating from source to target, and afterwards distributed back from this node, by propagating from target to source. For the hyperedge we propagate from ```{1,2,3}``` to ```1``` and the propagate backwards. In the default case, hypergraph convolution is basically just averaging among all edges. Note further, that this "intermediate" node 0 is just the original node 0 of our example.

What happens now if we have more edges than nodes? Well, we need an intermediate node for each edge, representing its accumulation of information for each edge to eventually transfer it back. This is now where the problem lies: If there are not enough original nodes to pass the messages to, the procedure will throw an error. However, we can simply add $dif = num_edges - num_nodes$ nodes temporarily, by padding the feature matrix to fix this issue. We then remove these added nodes afterwards by pruning the matrix back to the original size of $x$.

xnhp:
@THinnerichs thank's a lot for the elaborate explanation, that means a lot to me. It's clear to me now.

mjkmoynihan:
@THinnerichs very nice explanation! 
Would it be possible to add this clarification on the [edge_index] composition to the docs for [HyperGraphConv](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.HypergraphConv)?

rusty1s:
Thanks a lot for the PR and really sorry for the delay. I modified the code to make use of the bipartite `self.propagate` functionality (via `size`) instead of via padding. This should be very readable now :)

THinnerichs:
Thanks alot!

linlang1837:
When I use released torch_geometric 1.7.0, I still have the same error. Should I uninstall the torch_geometric 1.7.0 and install from the master?

rusty1s:
You need to install from master :)

PR #2045, opened 2021-01-25
Fix downloading error caused by file name suffix "?dl=1"

Les1ie:
Fix file writing error while downloading data from "https://www.dropbox.com/s/1bnz8r7mofx0osf/net_aminer.zip?dl=1", caused by the suffix "?dl=1".


rusty1s:
Awesome!

PR #2053, opened 2021-01-28
Fix num_classes to be integer

Xiuyu-Li:
Sometimes `dataset.num_classes` is not an integer, and this may cause issues in the code. Here is an example: 
```python
from torch_geometric.datasets import WikipediaNetwork
dataset = WikipediaNetwork(root='/tmp/Chameleon', name='Chameleon')

conv1 = GCNConv(dataset.num_features, 16)

# here dataset.num_classes=5.0, which will make GCNConv throw an error
conv2 = GCNConv(16, dataset.num_classes)
```
A simple fix is to cast the `num_classes` property to be an interger, since it seems like we never expect this property to be a non-integer.


rusty1s:
Thanks!

PR #2054, opened 2021-01-28
Fix error in DenseGraphConv triggered when aggr is max

quqixun:
In original implementation of DenseGraphConv, dimention mismatch error occured when aggr is max.

```python

# keep dimention while using max
# out = out.max(dim=-1)[0]
out = out.max(dim=-1, keepdim=True)[0]

# change the way to merge output of lin_rel and lin_root
# out += self.lin_root(x)
out = out + self.lin_root(x)

```

### DIFF
--- a/test/nn/dense/test_dense_graph_conv.py
+++ b/test/nn/dense/test_dense_graph_conv.py
@@ -1,11 +1,14 @@
+import pytest
+
 import torch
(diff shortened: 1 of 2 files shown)


rusty1s:
Thank you very much! Indeed, `DenseGraphConv` with `max` aggregation was really buggy. This is now fixed.

Issue #1981, opened 2021-01-07
Memory-Based Graph Networks (GMN)


wsad1:
## 🚀 Feature
[This](https://arxiv.org/abs/2002.09518) papers proposes a new graph convolution/coarsening algorithm. The convolution layer doesn't use any kind of gnn, i.e it doesn't explicitly use the graph structure, instead the graph structure is 'embedded' into the feature vector. 

## Motivation
The results on some graph classification datasets look significantly better than other graph convolution methods. Refer to the last 2 rows in the below image.
![image](https://user-images.githubusercontent.com/13963626/103901620-65483880-511f-11eb-902c-e3f1e1b13b70.png)

In addition to this it might be useful to check performance on the ogb-dataset too.

ldv1:
This is an interesting paper.

Two remarks:
* I am happy that you put this figure: SAGPool improves upon TopKPool (see [Self-Attention Graph Pooling](https://arxiv.org/pdf/1904.08082.pdf)) and so it is curious to see that in the table above TopKPool beats SAGPool by a large margin.
* GMN, the ones with the best performance, embeds the topology into the feature vector using random walk with restart. So, it seems that it  only works in the transductive setting.

wsad1:
> * SAGPool improves upon TopKPool (see [Self-Attention Graph Pooling](https://arxiv.org/pdf/1904.08082.pdf)) and so it is curious to see that in the table above TopKPool beats SAGPool by a large margin.

Thanks for bringing this up.  According to the SAGpool paper. SAG pool is better than Diffpool and topkpool(graph unet). However the GMN paper shows the opposite. One thing to note is that the SAGpool results are same in both the GMN paper and the SAGpool paper. Not sure whats causing this difference. I'll try to dig deeper into their training settings, or code(if available) to find some answers.
![image](https://user-images.githubusercontent.com/13963626/103901517-421d8900-511f-11eb-8b3c-8c7555d4974d.png)



> * GMN, the ones with the best performance, embeds the topology into the feature vector using random walk with restart. So, it seems that it  only works in the transductive setting.

For an unseen graph(inductive setting) S(difussion matrix from random walk) would have to be computed again but the learnt parameters W0 and W1 could be used as is. So it could work in the inductive setting too . Attaching the GMN equation(first layer) for reference.
![image](https://user-images.githubusercontent.com/13963626/103900160-7c862680-511d-11eb-9cc8-e5985aab77c9.png)

[...]

PR #2061, opened 2021-01-30
Mempool

wsad1:
Implemented graph memory network from #1981. The class has a function computeKl to get the additional loss term, this way a user can compute the loss if needed.

PS. this is my first time contributing, so please let me know if changes are needed to anything.


wsad1:
Will reopen with some fixes.

PR #2062, opened 2021-01-30
Mempool

wsad1:
Implemented graph memory network from #1981. 
1. The class is named `MemPool`. I thought that name made sense as the paper calls it a Memory-based GNN. Let me know if you think of a better name.
2. `kl_loss` returns the additional loss term, this way a user can compute the loss if needed.


wsad1:
Marking as draft as Iam cleaning up the code.

rusty1s:
Awesome :) Thank you very much for your contributions!

torch_geometric/nn/pool/mem_pool.py
```diff
@@ -0,0 +1,100 @@
...
+        .. math::
+            P_{i,j} &= \frac{S_{i,j}^2 / \sum_{n=1}^N S_{n,j}}{\sum_{k=1}^K
+            S_{i,k}^2 / \sum_{n=1}^N S_{n,k}}
+
+            \mathcal{L}_{\textrm{KL}} &= \textrm{KLDiv}(\mathbf{P} \Vert
+            \mathbf{S})
+        """
+        S_2 = S**2
+        P = S_2 / S.sum(dim=1, keepdim=True)
+        P = P / (S_2.sum(dim=2, keepdim=True) / S.sum(dim=1, keepdim=True))
```
wsad1, line 71: @rusty1s . shouldn't this line be
```
denom=P.sum(2,keepdim=True)
P=P/denom
```
image
In the denominator the inner summation is inside the outer summation.  The equation is probably confusing as they didn't add bracket after the outer summation.

wsad1, line 71:
```suggestion
P=P/P.sum(dim=2,keepdim=True)
```

PR #2076, opened 2021-02-03
Fix off-by-one error in TemporalData train/test split

yulonglin:
Currently, the index starting the validation (and test) set is displaced by `+1`

This is because `int((self.t <= val_time).sum())` already adds `1` to the index since the tensor is 0-indexed.

As a simplifying example, using just a train-val split (no test set) of ratio 0.75-0.25,
```
import numpy as np
t = [0,0,0,0,1]
val_time = np.quantile(t, 0.75)
val_idx = int((t <= val_time).sum()) + 1
t[:val_idx], t[val_idx:]
>>> ([0, 0, 0, 0, 1], [])
```

The desired split should be `([0, 0, 0, 0], [1])` instead, which we can rectify by removing the `+ 1` to `val_idx` and `test_idx`.

### DIFF
--- a/torch_geometric/data/temporal.py
+++ b/torch_geometric/data/temporal.py
@@ -98,8 +98,8 @@ class TemporalData(object):
             self.t.cpu().numpy(),
             [1. - val_ratio - test_ratio, 1. - test_ratio])
 
-        val_idx = int((self.t <= val_time).sum()) + 1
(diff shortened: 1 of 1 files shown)


rusty1s:
Yes, you are absolutely right. Thanks for spotting!

PR #2098, opened 2021-02-09
Number of classes in datasets with no y

Byambaa0325:
I noticed, while exploring around on Epinions1 dataset of SNAPDataset, that it raises an error when you try to see the number of classes. In general, num_classes doesn't handle datasets with Nonetype of y. It seemed logical that if there is no y then it should return 0 instead of an error.

P.S: I'm new to the library. Please, nevermind this pull request if the way it works is intentional.


rusty1s:
I agree. Thank you!

PR #2105, opened 2021-02-10
Add AddTrainValTestMask transform.

dongkwan-kim:
Add `AddTrainValTestMask` transform for node classification tasks.


rusty1s:
Awesome, I really like it. This should be really useful!

Issue #2092, opened 2021-02-07
Number of edges for an undirected graph is double counted


amirziai:
## 🐛 Bug

Number of edges for an undirected graph is double counted.

## To Reproduce

[this line](https://github.com/rusty1s/pytorch_geometric/blob/master/torch_geometric/data/data.py#L233) returns the number of items in the adjacency list, which is going to double count 

## Expected behavior

i think the number of edges should be divided by 2 when the graph is undirected. here's an example:
```
from torch_geometric.datasets import TUDataset
d = TUDataset('./enzymes', 'ENZYMES')[0]
assert d.is_undirected()
assert d.num_edges == 168

edges = set()
for u, v in zip(d.edge_index[0], d.edge_index[1]):
    u, v = u.item(), v.item()
    edges.add(tuple(sorted([u, v])))  # u=> v and v => u the same

assert len(edges) == (d.num_edges // 2)
```

rusty1s:
This is actually intended, as we do not distinguish between directed and undirected graphs. If you want your graph to be undirected, you will need to add edges for both directions.

amirziai:
👍

open to adding a comment clarifying this in the docstring? i can create a PR if you think it'd be useful. it can be a bit confusing.

rusty1s:
Sure, feel free to create a PR.

PR #2145, opened 2021-02-21
`num_edges` clarification

amirziai:
clarifying what is meant by `num_edges`, especially important for undirected graphs where it can be a bit confusing.

see #2092


rusty1s:
Awesome, thank you!

PR #2158, opened 2021-02-24
fix s3dis.py error in windows system

renyusang:
In windows environment os.sep is not '/', so
 `name = self.url.split(os.sep)[-1].split('.')[0]`
`os.rename(osp.join(self.root, name), self.raw_dir)`
 cause OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect.


rusty1s:
Thanks!

PR #2162, opened 2021-02-24
The gated graph conv after Bresson and Laurent

ldv1:
This is an implementation of [Residual Gated Graph ConNets](https://arxiv.org/pdf/1711.07553v2.pdf). Unfortunately, there is another [GatedGraphConv](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.GatedGraphConv). So, I chose a very original name: `GatedGraphConv2`...
The original paper uses the sum operator, which is inappropriate here because the gating function is the sigmoid. Indeed, I get better results using the mean operator. Nevertheless, the default is as in the paper (sum operator).


rusty1s:
How about we name that one `ResGatedGraphConv`? Would that work for you?

ldv1:
I thought about it, but in the paper "residual" refers to the "full" layer, i.e. x_i = x_i + conv() and not only to the convolution;  but it would be ok nevertheless (and reminds to where it comes from); GatedGraphConv2 is really ugly.

rusty1s:
I see. I added the `residual` argument to the `conv` to justify the re-naming :) I also added tests and allow for `PairTensor` as input. Thank you very much!

rusty1s:
Eq (12) in the paper does not mention any activation function for the residual term. Nonetheless, I removed the `residual` flag.

ldv1:
In the paper, the gated graph convnet, eq. (11), reads as follows:
`x_i^(l+1)= = ReLU( ResGatedGraphConv(x_i^l) )`
as far as I can see, the residual gated graph convnet, eq. (12), is
`x_i^(l+1)= = ReLU( ResGatedGraphConv(x_i^l) ) + x_i^l`
This is how I understand it, but I might err.

rusty1s:
Yeah, you are right!

PR #2164, opened 2021-02-25
Diffnorm

wsad1:
#2113. Added differentiable group norm. Key functions in the DiffNorm class contains

1. forward(): which implements the differentiable group norm
2. groupDistanceRatio(): which computes an over-smoothing metric given node embeddings.


wsad1:
Converting to draft for 2 reasons.

1. To add some more comments.
2. To add an example script.

Do review the main code in the meanwhile and let me know if changes are needed.

rusty1s:
This looks super good already. I have just one concern: The current implementation seems to only operate on a single graph, and does not take operating on multiple graphs in a mini-batch scenario into account. Are there any modifications necessary to allow that?

wsad1:
@rusty1s , good question. 
I don't think this needs to be modified to work on multiple graphs. I say this because diff_norm works exactly like batchnorm, with an additional grouping step. And since batchnorm takes mean of features across all nodes , even when the nodes belong to different graphs, diff_norm should do the same. 
Let me know if that's not clear, or if you have other comments.

rusty1s:
Thank you very much!

Issue #2189, opened 2021-03-02
Any import from `torch_geometric` loads `matplotlib` (and a lot more)


Linux-cpp-lisp:
## 🐛 Bug

Because `torch_geometric` imports `*` from `.nn`, which imports `*` from `.models`, which imports all of the models, a lot of unnecessary dependencies get loaded whenever _anything_ is imported from `torch_geometric`.

I came across this behavior because it was filling my logging with `matplotlib` messages, but even with logging disabled it would be good for performance if heavy libraries that do file I/O like `matplotlib` are only loaded if they are needed. This can be particularly important for start-up performance on clusters with networked filesystems.

## To Reproduce

```python
import logging
logging.basicConfig(level=logging.DEBUG)

import torch
from torch_geometric.data import Data

d = Data(x=torch.zeros(10, 3))
print(d)
```
Gives many pages of `matplotlib` debug logging, followed by the expected
```
Data(x=[10, 3])
```

## Expected behavior
That `matplotlib` (and other libraries that aren't necessary for basic functions) won't be loaded for any import from `torch_geometric`, giving an output of just
```
Data(x=[10, 3])
```

## Environment

* OS: Linux
* Python version: 3.8.5 / Anaconda
* PyTorch version: 1.7.1
* CUDA/cuDNN version: N/A
* GCC version: N/A
* Any other relevant information: N/A

rusty1s:
You are right! This should be fixed in master. Thanks!

Linux-cpp-lisp:
Thanks @rusty1s !

When I run my test script, I no longer get `matplotlib` but still get some output (`h5py` from the datasets, I believe):
```
DEBUG:h5py._conv:Creating converter from 7 to 5
DEBUG:h5py._conv:Creating converter from 5 to 7
DEBUG:h5py._conv:Creating converter from 7 to 5
DEBUG:h5py._conv:Creating converter from 5 to 7
Data(x=[10, 3])
```

I realize it would be a breaking change and so would need to be considered quite carefully, but it seems to me that it might be worth not importing `datasets` and `nn.models` automatically, since many many programs that use `torch_geometric` to train their own models on their own datasets do not need the wide variety of models, datasets, and their dependencies to be loaded.

It may also be possible to make `datasets` and `nn.models` lazy-loaded, preventing a breaking change but preventing unnecessary loads — see, for example:

- https://modutil.readthedocs.io/en/latest/#lazy_import
[...]

PR #2199, opened 2021-03-03
Lazy loading for `datasets` and `nn.models`

Linux-cpp-lisp:
See discussion in #2189 .

This PR uses the `LazyLoader` class from TensorFlow to lazy-load `torch_geometric.datasets` and `torch_geometric.nn.models`.

The file from TensorFlow is under an Apache 2.0 license, so it should be fine to include in an MIT license project? See:

- https://softwareengineering.stackexchange.com/questions/297217/using-code-under-apache-2-0-license-in-my-program-using-mit-license
- https://opensource.stackexchange.com/questions/8135/can-i-use-mit-isc-apache-2-0-in-a-project-licensed-as-mit-or-apache-2-0
- https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project


Linux-cpp-lisp:
The last commit is needed to lazy load all of `nn` instead of `nn.models`. Unfortunately, because `torch_geometric/nn/__init__.py` had 
```
from .models import *
```
there doesn't seem to be any way to lazy load those individual classes. 

This lazy loading setup (from TensorFlow) only supports modules. I tried to use `importlib.utils.LazyLoader` from the stdlib, but was unable to get it to work without circular import errors.

rusty1s:
I further polished and achieved faster loading by fixing some imports in `torch_geometric.transforms` which directly imported from `torch_geometric.nn`.

Linux-cpp-lisp:
Looks good, thanks @rusty1s . I'm not sure why the CI tests are failing, though.

rusty1s:
CI fails because of https://github.com/pytorch/vision/issues/3504, which is independent of this PR.

PR #2204, opened 2021-03-04
Local cartesian: make normalization optional

ldv1:
This is a small PR:
* LocalCartesian: The normalization is now optional
* LocalCartesian: I changed `cart = pos[col] - pos[row]` into `cart = pos[row] - pos[col]` because it is usually x_j - x_i (position of neighbor - position of central point)
* GMMConv: Add a note that points to LocalCartesian


rusty1s:
Thanks a lot. Note that I removed the `normalization` argument since this behavior is already achievable via `Cartesian()`.

ldv1:
I stumble upon `LocalCartesian`, and did not see `Cartesian`. `Cartesian` was what I needed. Sorry. Maybe both methods could hold a cross-"See also...". I think `GMMConv` should reference Cartesian (and not `LocalCartesian`, as I initially proposed).

PR #2207, opened 2021-03-05
avg_pool_neighbor_x now returns only the pooled neighbors' feature vectors

ldv1:
This is a small PR.

`avg_pool_neighbor_x` takes in `data` and modifies `data.x`; sometimes we only want to average pool the neighbors to get information as a patch, without overwriting `data.x`; so, now we can write either
`data.x = avg_pool_neighbor_x(data.x, data.edge_index, data.num_nodes)`
or
`patch_x = avg_pool_neighbor_x(data.x, data.edge_index, data.num_nodes)`

### DIFF
--- a/torch_geometric/nn/pool/avg_pool.py
+++ b/torch_geometric/nn/pool/avg_pool.py
@@ -1,8 +1,10 @@
 from typing import Optional
 
+from torch import Tensor
 from torch_scatter import scatter
 from torch_geometric.data import Batch
-from torch_geometric.utils import add_self_loops
+from torch_geometric.utils import add_self_loops as utils_add_self_loops
+from torch_geometric.utils.num_nodes import maybe_num_nodes
 
 from .consecutive import consecutive_cluster
 from .pool import pool_edge, pool_batch, pool_pos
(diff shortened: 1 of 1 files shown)


rusty1s:
Thanks for the PR. I'm not a big fan of `avg_pool_neighbor_x` to be honest, and will probably remove it at some point in time.

The proposed implementation is already quite similar to a normal `scatter` call:
```python
path_x = scatter(data.x, data.edge_index[0], dim_size=data.num_nodes, reduce='mean')
```
and so it's hard for me to think of a use-case where this might be useful.

ldv1:
You can think of it as a "local average pooling" as opposed to "global average pooling". I concatenate this information to the feature vector. It basically characterizes how much a node stands out among its neighbors (the information of a second-order derivative as a matter of fact). That's useful.

rusty1s:
Yes, I definitely agree that this is useful. It's just that this functionality is already equivalent to a normal `scatter` call.

rusty1s:
On a side node. What about we integrate that functionality in a non-trainable `MessagePassing` module?

PR #2209, opened 2021-03-06
Added a flag to gnn_explainer.

wsad1:
#2203. 

 Made changes to.
- gnn_explainer.py
- - Added a flag called `retrun_type` to gnn_explainer. This flag indicates if the model will output raw scores, probabilities or log of probability.
- - Within the `explian_node` module model outputs are converted to log of probability based on the flag. `to_log_proba` is used to convert outputs.
- test_gnn_explainer.py: Add a new test function to test `to_log_proba`

### DIFF
--- a/examples/gnn_explainer.py
+++ b/examples/gnn_explainer.py
@@ -40,7 +40,7 @@ for epoch in range(1, 201):
     loss.backward()
     optimizer.step()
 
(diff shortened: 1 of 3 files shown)


rusty1s:
Awesome! Thanks a lot :)

PR #2210, opened 2021-03-06
Mixhop synthetic dataset

ldv1:
This is the mixhop synthetic dataset. 
10 graphs and each graph has a specific homophily. Homophily ranges from 0 to 0.9.

To illustrate:

mixhop_0 9

With GATConv, I achieve ~ 0.98 accuracy when the homophily is 0.9

mixhop_0 1

The accuracy with GATConv is as low as  ~ 0.4  when the homophily is 0.1


ldv1:
Why is this dataset of tremendous importance?
Many models rely on the assumption that neighboring nodes are likely to share the same class label. However, this assumption does not hold in many real world graphs as relations between nodes in these graphs could be complicated. The synthetic dataset constructed by Abu-El-Haija et al. is meant to highlight precisely this bias.

rusty1s:
Thank you very much. I'm having a closer look at the moment. Do you mind removing the example? This example is already really similar to our `gat.py` example (just with another dataset).

ldv1:
You can delete the example, it is fine for me. The idea is, if someone wants to use a particular dataset, it is easier and faster to get an example from the directory  ./examples that loads that dataset and then build upon it. The example readily says how to handle the dataset. Maybe you could remove GAT and simply use an MLP. That's what I did at the onset. As in the paper, it tops at ~0.4 no matter the homophily (as expected).

PR #2238, opened 2021-03-10
measure for homophily

ldv1:
This is a measure of homophily as proposed in the paper [Geom-GCN: Geometric Graph Convolutional Networks](https://arxiv.org/abs/2002.05287).

### DIFF
--- a/torch_geometric/utils/__init__.py
+++ b/torch_geometric/utils/__init__.py
@@ -8,7 +8,7 @@ from .loop import (contains_self_loops, remove_self_loops,
                    add_remaining_self_loops)
 from .isolated import contains_isolated_nodes, remove_isolated_nodes
 from .subgraph import subgraph, k_hop_subgraph
-from .homophily import homophily, homophily_ratio
+from .homophily import homophily
 from .get_laplacian import get_laplacian
 from .to_dense_batch import to_dense_batch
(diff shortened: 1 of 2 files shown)


rusty1s:
Thank you. The existing implementation corresponds to the definition in [Beyond Homophily in Graph Neural Networks: Current Limitations and Effective Designs](https://arxiv.org/abs/2006.11468). Can you explain the differences to your proposed implementation?

ldv1:
I observed that `homophily_ratio` and `homophily` do not give the exact same answer. For some very small graphs (e.g. synthetic ones), answers can be radically different, for others (like the ones from the homophily dataset), both provide answers that are very close.
I cannot tell the differences. One is sure: Some authors use this definition of homophily and it would be nice to have it. Furthermore, this definition is really intuitive. At least for me. And my code is documented.

ldv1:
1/ There is now a single interface `homophily` to compute both kinds of homophily.
2/ I documented the homophily ratio.

rusty1s:
Awesome :) Really appreciate your contributions!

ldv1:
In the "edge" formula, the denominator must be \vert \mathcal{E} \vert instead of \vert \mathcal{V} \vert.
That's a mistake of mine, sorry.

rusty1s:
You are right. Fixed it.

PR #2275, opened 2021-03-19
More informative repr in NNConv

JohnFinn:
Simple adjustment making current `NNConv.__repr__` much more useful.


rusty1s:
Thanks!

PR #2284, opened 2021-03-20
Fagcn

wsad1:
Fagcn as suggested in #2225.
Three things to note.

1. forward expects `x` to be a tensor, like [GCNConv](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.GCNConv)
2. forward expects `edge_index` to be a tensor, like [SuperGATConv](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.SuperGATConv).
3. The class is named FAConv. Let me if you think of a better name.


rusty1s:
Awesome. I also added `SparseTensor` support. Thanks a lot!

wsad1:
@rusty1s thanks a lot for reviewing and merging.
One questions.
I noticed you removed `self.lin1` which transforms initial features to the current layers feature dimension and `self.lin2` which performs a linear transformation to the output. In the below image from the paper `self.lin1` denotes W1 and `self.lin2` denotes W2. 
image

Since you removed these linear transforms, are we expecting the user to perform these 2 operations.

rusty1s:
Yes, I removed it since those transformations are not really part of the GNN operator, and should be moved to the model part. This in line with how we implement `GCN2`, see [here](https://github.com/rusty1s/pytorch_geometric/blob/master/examples/gcn2_cora.py#L38-L46).

PR #2289, opened 2021-03-21
Faconv edge weights

ldv1:
There is an error if `edge_weight` is not given.
I took the opportunity to add the option `root_weight` because I treat the target node differently.

### DIFF
--- a/torch_geometric/nn/conv/fa_conv.py
+++ b/torch_geometric/nn/conv/fa_conv.py
@@ -40,8 +40,9 @@ class FAConv(MessagePassing):
             will use the cached version for further executions.
             This parameter should only be set to :obj:`True` in transductive
             learning scenarios. (default: :obj:`False`)
-        add_self_loops (bool, optional): If set to :obj:`False`, will not add
-            self-loops to the input graph. (default: :obj:`True`)
+        root_weight (bool, optional): If set to :obj:`False`, the layer will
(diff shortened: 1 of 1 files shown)


rusty1s:
`edge_weight` needs to be in fact always be given. It is either computed on-the-fly (with `normalize=True`), or passed manually (with `normalize=False`). I added a check to ensure that.

I'm not a big fan of the additional `root_weight` argument, as this can be simply achieved via `eps=0`. I replaced the check with
```python
if self.eps != 0:
     out += self.eps * x_0
```
Thanks for digging into this!

ldv1:
ok, here we can set `eps=0.`, you are right, but more generally I would prefer to always have a `root_weight` argument. This is inline with message passing where we form the message, and then combine it with the target node. In most papers, this is done in one step. But I believe it is good practice to break it down in two steps as it is useful in situations where you want to combine the message and the target node differently, as in e.g. GeniePath with a RNN.

rusty1s:
I fully understand you. It's just that a `root_weight` option is kind of confusing in this operator:
(1) Similar to `GCNConv`, it will not treat root nodes differently than neighboring nodes
(2) IMO, the `root_weight` option might be confusing to users, as we do not additionally transform the central node features, but instead add the *initial* node features to the output.

Therefore, I'm in favor of dropping the `root_weight` argument from this operator.

PR #2292, opened 2021-03-22
attentive fp

thegodone:
this is the basic version of AttentiveFP using the full power of PyG (message passing)


rusty1s:
Thanks! Can you upload your training script as well?

thegodone:
sure, I add esol.csv / Atom-Bond-featurer / test jupyper notebook (all in nn folder)

thegodone:
Last missing part is the GAT "heads" (based on original code heads = max number of neighbours ) in the atom embedding part. Can you please first focus on this point ?
so we need to use the include something like this similar to gat_conv.py where H = 7 and C = fpdim ?
x = self.lin(x).view(-1, H, C)
alpha= (x * self.att).sum(dim=-1)

rusty1s:
Hi, I re-factored quite a bit.

* I re-use the GATConv for implementing both atom and molecule embeddings
* I implemented a `GATEConv` which makes of use edge features for computing both attention coefficients and node features
* I moved all code of `AttentiveFP` to the actual module implementation.
* I think I fixed a crucial part of the `AttentiveFP` pipeline: In the molecule module, output features are computed based on *all* previous acquired node features, not just the last ones

Please have a look at tell me what you think. I will clean up the example afterwards.

thegodone:
Yes, good idea to reuse GATconv & promote edge (aka GATEconv). Also speed is there & model converge in a more accurate way than my submission. I glad you find & fix my bug at the molecule level. great!

thegodone:
I don't understand why you do the sum on att_i + att_j:  

image

rusty1s:
I updated the example code (re-use `MoleculeNet` data). However, it cannot yet reach the performance reported in the paper. It would be great to reach that. In particular, it is unclear to me whether the hyperparameters match with those used in the paper.

thegodone:
I saw that In the original implementation they use only one GRU for all T loops for molecule embedding cause it’s a static object not like for the atoms. Can you add an option to decide on one / per loop GRU please ? 

Envoyé de mon iPhone

thegodone:
What are our performance so far ? It may be a seed optimal solution in the paper. On Esol 1127 pts and not 1128 (last paper 2021) they have 0.587 rmse for 50 runs where SVM give 0.569 and XGBoost 0.582 see table 9 page 16 jiang j cheminfo using DGL implementation. While it was reported in 2019 with original code 0.503 in Xiong j med chem paper. I never got this results 0.503 with 1128 molecules with both original or dgl implementations. And dgl is generally worse than original for performance. 

Envoyé de mon iPhone

rusty1s:
I made some further changes. The current implementation should now exactly match with the original one. It looks like the authors indeed apply the readout layer only based on the latest node representations (although the paper states the opposite), and they are indeed only using a single `conv` and `gru` layer and apply it recursively.

The performance is around 0.5-0.6 and depends heavily on the given split. I think we can merge this if it looks correct to you.

thegodone:
You can merge it's look good to me. thanks

PR #2309, opened 2021-03-26
fixed bug in kl_loss.

wsad1:
Changed this line `P = P / (S_2.sum(dim=2, keepdim=True) / S.sum(dim=1, keepdim=True))`, to 
`P=P/P.sum(2,keepdim=True)` in kl_loss in mem_pool.py.
Posting the equation from the paper. In the denominator the inner summation is inside the outer summation.  The equation is probably confusing as  bracket are missing after the outer summation.

image

Also pasting an image from the original [paper](https://arxiv.org/pdf/1511.06335.pdf) that proposed this metric. Here the equation is clearer.
image


rusty1s:
Indeed. Thanks!

PR #2313, opened 2021-03-27
Homophilyfix.

wsad1:
An error shows up while measuring homophily on the arxiv dataset. Check code and error below.
```
dataset=PygNodePropPredDataset(name='ogbn-arxiv',root='../node_dataset/')
split_idx = dataset.get_idx_split()
data = dataset[0]
data.y.shape # [169343, 1]
homophily(data.edge_index,data.y,method="node")
```
```
<ipython-input-121-3e87edc6610c> in homophily(edge_index, y, method)
     65     else:
     66         out = torch.zeros_like(row, dtype=float)
---> 67         out[(y[row] == y[col])] = 1.
     68         out = scatter_mean(out, col, 0, dim_size=y.size(0))
     69         return float(out.mean())

IndexError: too many indices for tensor of dimension 1
```
This PR has 2 changes.
1. fixes the error .
2. Adds a test for homophily.

### DIFF
--- a/test/utils/test_homophily.py
+++ b/test/utils/test_homophily.py
@@ -1,0 +1,9 @@
+import torch
+from torch_geometric.utils import homophily
+
+
+def test_homophily():
+    edge_index = torch.tensor([[0, 1, 2], [1, 2, 0]])
+    y = torch.tensor([0, 0, 0])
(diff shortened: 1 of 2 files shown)


rusty1s:
Yes, this method failed when `y` is not one-dimensional. Thanks.

Issue #2403, opened 2021-04-13
Implementation of DimeNet++


kryczko:
From now on, we recommend using our discussion forum (https://github.com/rusty1s/pytorch_geometric/discussions) for proposing new features.

## 🚀 Feature
A implementation of the DimeNet++ model. 
## Motivation
The DimeNet++ model has shown to have better accuracy and is more efficient than it's V1 counterpart.

## Additional context
The DimeNet++ pretrained model is already being hosted on Github.

thegodone:
Cautions guys there are already 3 bugs reported in the original tensorflow code. I would suggest to fix them first. I have DimeNet++ model with PyG I was able to fix one bug but not the 2 others.

rusty1s:
There are plans to integrate DimeNet++. @thegodone Which bugs are you referring to?

thegodone:
![image](https://user-images.githubusercontent.com/1186658/114666582-b52a3300-9cfe-11eb-9da1-9ccb539f54f7.png)

kryczko:
I believe @rusty1s has already commented on these bugs in another issue, although I can't remember the issue itself. The tensorflow issue is fine, since pytorch is being used in this repo.

thegodone:
Thanks kryczko but, I would appreciate to have a clear answer of Matthias especially because the last bug just pop up. It's always better to be sure than to speculate.

rusty1s:
I think [this is the issue](https://github.com/rusty1s/pytorch_geometric/pull/1687) @kryczko has mentioned. Basically, we need to inherit the errors of `DimeNet(++)` if we want to make use of the pre-trained models, and want to provide the "fixed" version in case one wants to train a model from scratch.

thegodone:
correct for the first issue, what about 2 and 5 ? that would be nice to have a switch to use "alpha-pretrain-dimenet" or "fix-dimenet" as you said.

rusty1s:
Yes, we can automatically set this variable in case the user makes use of `from_qm9_pretrained` or not.

thegodone:
my changes to get pp blocks:
`class InteractionBlockPP(torch.nn.Module):
    def __init__(self, hidden_channels, int_emb_size, basis_emb_size, num_spherical,
                 num_radial, num_before_skip, num_after_skip, act=swish):
        super(InteractionBlockPP, self).__init__()

        self.act = act
        self.basis_emb_size = basis_emb_size
        self.hidden_channels = hidden_channels
[...]

PR #2421, opened 2021-04-19
`DimeNet++`

thegodone:
This is the code and example on MoleculeNet ESOL loader and Feature generator.

Comments part are here to related the fix in Distance / RBF Envelopp / ...  

this is to fix issue #2403


rusty1s:
Hi @thegodone. Thanks for this PR. Is it possible for you to implement a `DimeNet++` model directly in PyG (at best by re-using some functionality from the `DimeNet` model?

thegodone:
I can but it means that I have to add 4,5 modifications in the current PyG Dimenet part. Should I set a "fix" parameter of a "2.0" version parameter in those functions/classes ?

rusty1s:
Can we make that an internal flag, i.e., `self._pretrained_version`, which just get sets to `True` in case we are calling `from_pretrained`?

thegodone:
Yes we can do that, it makes sense. For existing Dimenet pretrained model is it per target or a multitask model stored in TF and reload in PyG ?

rusty1s:
It is per target.

thegodone:
I implement the code in the Dimenet PyG code and Add a fix bool parameter

PR #2425, opened 2021-04-19
Add Efficient Graph Convolution (EGC)

shyam196:
This is an implementation for EGC. Included are tests (including that the layer can be jitted) and an example on OGB-MolHIV. Apologies for the delay in getting this upstreamed :-)

This differs from the reference implementation by taking advantage of batched matrix multiplications, and does not use the `_AggLayer`s. I did try to make some sensible optimizations to avoid recomputing the same things repeatedly, although this has made the code a little bit uglier.

The likely source of bugs will be in forward where the symmetric normalization or self-loops are calculated. I have double checked myself but would appreciate another pair of eyes! Other sources may be `aggregate` or `message_and_aggregate`.


rusty1s:
Awesome. Thank you so much :)

PR #2440, opened 2021-04-21
Changes lambda collate_fn to static method for DataListLoader

zimonitrome:
Since lambda functions can't be pickled on Windows (for multiprocessing), instead users can provide their own collate function.
This would be a preferable solution. Alternatively, one could add a functional identity function:

```python
def identity(x):
    return x
```

and set `collate_fn=identity`.


rusty1s:
I prefer your second solution as I don't think a `collate_fn` argument in `DataListLoader` makes a low of sense :)

zimonitrome:
@rusty1s Thanks for the quick response. Please see the revised commit on the same branch.

I am not sure where in the codebase an `identity()` function would best be suited. I thought you might want the class self-contained so I created a static `collate_fn` method instead.

rusty1s:
Awesome. Thanks :)

PR #2466, opened 2021-04-25
Add adversrially regularization loss for ARGVA exmaple

Curt-Park:
Currently, `examples/argva_node_clustering.py` doesn't consider the adversrial regularization loss.

### [Major Fix] Add model.reg_loss(z)
Without `model.reg_loss(z)`, it runs same as VGAE.

### [Minor Fix] Remove discriminator.train()
model.train() executes train methods of its children which include discriminator.

### Consol Log
[log.txt](https://github.com/rusty1s/pytorch_geometric/files/6371943/log.txt)

### Plot
plot


rusty1s:
Thanks :)

PR #2488, opened 2021-04-28
add mnist superpixel visualization

Somedaywilldo:
Dear Matthias,

I believe this is what everybody wants when they are trying to visualize the mnist_superpixel dataset. 

image

Actually it took me quite some time to make it looks perfect since I found the default axis direction of networkx.draw() is wried, the y-axis of data.pos needs to be flipped over,  also the index of networkx graph is not in order, so it requires some re-indexing operations.

As mentioned in https://github.com/rusty1s/pytorch_geometric/issues/320, your old script [here](https://github.com/rusty1s/pytorch_geometric/blob/da78a9bc6074913713e0fffd908a4365dfe20ef9/torch_geometric/visualization/mnist_graph.py) is no longer working with latest torch_geometric. And I think my solution is very easy to read and not hacky at all : ), so I'd love to share it with everybody.

### DIFF
--- a/examples/mnist_visualization.py
+++ b/examples/mnist_visualization.py
@@ -1,0 +1,60 @@
+import os.path as osp
+
+import numpy as np
+import pandas as pd
+from PIL import ImageColor
+import seaborn as sns
+import networkx as nx
+from matplotlib import pyplot as plt
+
+from torchvision.datasets import MNIST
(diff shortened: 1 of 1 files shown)


RanQChi:
change from
x, edge_index = data.x, data.edge_index
to 
x, edge_index = np.array(data.x), np.array(data.edge_index)

or it seems to meet mistake~

PR #2553, opened 2021-05-10
Add the UPFD dataset

YingtongDou:
As mentioned in [this](https://github.com/rusty1s/pytorch_geometric/discussions/2508) discussion thread, I add the UPFD dataset and an example code.

### DIFF
--- a/examples/upfd.py
+++ b/examples/upfd.py
@@ -1,0 +1,242 @@
+import argparse
+import time
+from tqdm import tqdm
+from sklearn.metrics import f1_score, accuracy_score, recall_score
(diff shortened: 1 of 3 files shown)


rusty1s:
Thanks, this looks great. Will have a more in-depth look ASAP.

rusty1s:
Thanks for this wonderful PR :) I cleaned up the example a little bit, and modified the dataset so that it uses a `split` argument for splitting (as it is the case for other PyG datasets). Hope this is okay :)

YingtongDou:
Hi Matthias,

Thanks for cleaning up the example and dataset code, that makes the code neat and easy to start with.

PR #2576, opened 2021-05-15
`PGExplainer`

### LINKED ISSUE #2069: Implementation for PGExplainer (GNN interpretation method)?
## 🚀 Feature

GNN interpretation is becoming more and more important and would be pretty great to have in this framework. GNNExplainer is good but recently a new interpretation model emerged in NuerIPS 2020 called PGExplainer. It has much better performance and is parametrized by the nodes meaning it's a broader method than GNNExplainer. The user just needs to train the explainer once for the whole graph and can produce an explanation for any given node without the need for any significant overhead. The paper is linked here https://arxiv.org/pdf/2011.04573

wsad1:
Closes #2069.
Implementation of PGexplainer as described in #2069.

This implementation is based off a pytorch implementation in [openreview](https://openreview.net/forum?id=8JHrucviUf), which is based on the authors [original implementation](https://github.com/flyingdoog/PGExplainer). The authors original implementation has some differences from the original paper.


rusty1s:
@RBendias is trying to map the existing `PGExplainer` to inherit from the base `Explainer` class. @wsad1 Please let me know if that aligns with your thinking as well.

RBendias:
@wsad1 and @rusty1s I set the class to the current state of the GNNExplainer; for regression cases, the PGExplainer should also work now. We could still adjust the example according to the GNNExplainer example (using Cora instead of Mutag or the other way around). Thus the attribution values of GNNExplainer/PGExplainer and to_captum would be comparable.

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,270 @@
...
+
+                    explainer_in = self._create_explainer_input(
+                        edge_index_n, z_n, mapping).detach()
+                    self.edge_mask = self._compute_edge_mask(
+                        self.explainer_model(explainer_in), t, bias=bias)
+                    # Should this be edge_index_n or edge_index?
+                    set_masks(self.model, self.edge_mask, edge_index_n)
+                    out = self.model(x=x_n, edge_index=edge_index_n,
+                                     **kwargs_n)
+                    loss += self.get_loss(out, prediction, mapping)
```
wsad1, line 219: One of my key concerns while implementing this was, the loss for the node classification task  is a sum across nodes. I am not sure how stable this would be if a large number of nodes are used to train `explainer_model`. @RBendias  any thoughts on this from your experience with explainers.

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,270 @@
...
+        size_loss = mask.sum() * self.coeffs['edge_size']
+        mask_ent_reg = -mask * mask.log() - (1 - mask) * (1 - mask).log()
+        mask_ent_loss = mask_ent_reg.mean() if batch is None else scatter_mean(
+            mask_ent_reg, batch[edge_index[0]]).sum()
+        mask_ent_loss *= self.coeffs['edge_ent']
+
+        return loss + size_loss + mask_ent_loss
+
+    def train_explainer(self, x: Tensor, z: Tensor, edge_index: Tensor,
+                        node_idxs: Optional[int] = None, batch: Tensor = None,
```
wsad1, line 134:
```suggestion
                        node_idxs: Optional[LongTensor] = None, batch: Tensor = None,
```

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,270 @@
...
+            (default: :obj:`30`)
+        lr (float, optional): The learning rate to apply.
+            (default: :obj:`0.003`)
+        num_hops (int, optional): The number of hops the :obj:`model` is
+            aggregating information from.
+            If set to :obj:`None`, will automatically try to detect this
+            information based on the number of
+            :class:`~torch_geometric.nn.conv.message_passing.MessagePassing`
+            layers inside :obj:`model`. (default: :obj:`None`)
+        task (str): Denotes the type of task that needs explanation. Valid
```
wsad1, line 41: This is probably something we can address in another PR.
`PGExplainer` and `GNNExplainer` have a different way of choosing graph and node tasks. `GNNExplainer` uses different functions and `PGExplianer` uses an if condition.

wsad1 (commented):
Yes I think `PGExplainer` inheriting from `Explainer` is a good approach.
@RBendias  I like the idea of using the same example across explainers, it'll give users some way of comparing explainers.
I have some comments specific to `PGExplainer`.

torch_geometric/nn/models/pg_explainer.py
RBendias: I guess that a single function `explain` (and `train_explainer`) is more practical. However, in this case, we would have to change the API for the GNNExplainer class. What do you think is the better option here? I also see you already considered this when introducing the graph explanation method for `GNNExplainer`.

torch_geometric/nn/models/pg_explainer.py
RBendias: This is the way they do it in the original [code](https://github.com/flyingdoog/PGExplainer/blob/ec57ea2db3b196da7eb93179acffc64ec6958828/BA-shapes.ipynb), right!? In the paper they use a rather small number of training examples (e.g. 5-30), maybe we could add that as a comment to the `node_idxs` argument. Does this concern the graph classification as well, since we use the sum in the loss function?

wsad1: Mentioning this in the comments is a good idea. And yeah this applies to both tasks.

torch_geometric/nn/models/pg_explainer.py
wsad1: Yeah two functions `explain` and `train_explainer` (GNNExplainer wouldn't have this) seems cleaner to meet too. Although this would be a breaking change to `GNNExplainer`. @rusty1s  WDYT?

test/nn/models/test_pg_explainer.py
```diff
@@ -0,0 +1,104 @@
+import pytest
+import torch
+import torch.nn.functional as F
+from torch.nn import Linear
+
+from torch_geometric.nn import (GATConv, GCNConv, MessagePassing, PGExplainer,
+                                global_add_pool)
+
+
+def assert_edgemask_clear(model):
```
wsad1, line 10:
```suggestion
def assert_mask_clear(model):
```

torch_geometric/nn/models/pg_explainer.py
wsad1: Ah nice idea @rusty1s .

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,270 @@
...
+    def _compute_edge_mask(self, edge_weight, temperature=1.0, bias=0.0,
+                           training=True):
+
+        if training:  # noise is added to edge_weight.
+            bias += 0.0001
+            eps = (bias -
+                   (1 - bias)) * torch.rand(edge_weight.size()) + (1 - bias)
+            eps = eps.to(edge_weight.device)
+            return (eps.log() -
+                    (1 - eps).log() + edge_weight).squeeze() / temperature
```
SensorLock, line 104: I think the squeeze() here and on line 107 need to be limited to certain dims. I hit a 2 node / 1 edge subgraph while exploring this code, and the squeeze reduced the edge_weight to a scalar tensor, causing IndexErrors later.  I tried it with squeeze(dim=1) in these two places and that worked, but I'm not familiar enough with the pg_explainer code to know if this edge_weight can ever have more than 2 dimensions.

SensorLock: This may also be needed in the squeeze on line 143.

simongraham:
Hi!

Thanks for the great effort with this PR. Is there any update on getting it verified and merged?

rusty1s:
@wsad1 @RBendias Anything to add before merging?

wsad1:
@rusty1s, we need to resolve a few things before merging. Will look at it this week and get back here.

torch_geometric/nn/models/pg_explainer.py
wsad1: I added a test for a 2 node 1 edge graph, seems to work fine
```
    # test with 2 node 1 edge graph:
    explainer = PGExplainer(model, out_channels=z.shape[1], task='node',
                            log=False, return_type=return_type)
    explainer.train_explainer(x[:2], z[:2], edge_index[:, :1],
                              node_idxs=torch.arange(0, 1))
    assert_mask_clear(model)
    mask = explainer.explain(x[:2], z[:2], edge_index[:, :1], node_idx=1)
    assert mask.shape[0] == 1
```
Does this test cover the scenario you mentioned?

wsad1: The test works even when the `dim=1` argument is not used in `squeeze`.

wsad1:
> Hi!
> 
> Thanks for the great effort with this PR. Is there any update on getting it verified and merged?
> 
> Also, in the `GNNExplainer` class there is an option to select the feature mask type - i.e `feature` `individual_feature` and `scalar`. This is really useful. Could something similar be used in the `PGExplainer` class?

@simongraham this is a good idea. However, the `PGExplainer` paper only talks about a parametric approach to generating subgraphs(edge masks) that explain a prediction. I guess this can be extended to learn feature masks too. Are you familiar with some paper that does this, it'll be good to work off that.

torch_geometric/nn/models/pg_explainer.py
RBendias: Sounds good! The explanation AUC for mutagenicity remains the same when taking the mean instead of the sum. Also, I tried some data sets where 100 training samples worked up to 4x better than 30. So maybe your first guess was correct, and we should switch e.g. to mean.

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,275 @@
+from typing import Optional
+
```
RBendias, line 2:
```suggestion
import warnings
```
See comment below.

test/nn/models/test_pg_explainer.py
```diff
@@ -0,0 +1,124 @@
...
+        f', {z.shape[1]}, epochs=30, lr=0.003, task=node)')
+    explainer.train_explainer(x, z, edge_index, node_idxs=torch.arange(0, 3))
+    assert_mask_clear(model)
+
+    mask = explainer.explain(x, z, edge_index, node_idx=1)
+    assert mask.shape[0] == edge_index.shape[1]
+    assert mask.max() <= 1 and mask.min() >= 0
+
+    # test with 2 node 1 edge graph:
+    explainer = PGExplainer(model, out_channels=z.shape[1], task='node',
```
RBendias, line 96: This test no longer works once the loss warning and skipping are included. Because the only node used for training is an isolated node.

torch_geometric/nn/models/pg_explainer.py
RBendias: I got the same error when using the second node as a training instance, instead of the first. 

`explainer.train_explainer(x[:2], z[:2], edge_index[:, :1],
                              node_idxs=torch.arange(1, 2))`

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,275 @@
...
+                    explainer_in = self._create_explainer_input(
+                        edge_index_n, z_n, mapping).detach()
+                    self.edge_mask = self._compute_edge_mask(
+                        self.explainer_model(explainer_in), t, bias=bias)
+                    # Should this be edge_index_n or edge_index?
+                    set_masks(self.model, self.edge_mask, edge_index_n)
+                    out = self.model(x=x_n, edge_index=edge_index_n,
+                                     **kwargs_n)
+                    loss += self.get_loss(out, prediction, mapping)
```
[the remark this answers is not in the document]
wsad1: @RBendias , nice catch on the bug. I updated the code to handle this error, by removing isolated nodes from `node_idxs`.

torch_geometric/nn/models/pg_explainer.py
```diff
@@ -0,0 +1,287 @@
...
+        mask_ent_loss *= self.coeffs['edge_ent']
+
+        return loss + size_loss + mask_ent_loss
+
+    @staticmethod
+    def _drop_isolated_nodes(node_idxs: Tensor, edge_index: Tensor) -> Tensor:
+
+        mask = torch.isin(node_idxs, edge_index[1])
+        if mask.sum() != node_idxs.size(-1):
+            warnings.warn(f"Dropping nodes {node_idxs[~mask]} from train set."
```
RBendias, line 142:
```suggestion
            warnings.warn(f"Dropping nodes {node_idxs[~mask]} from train set. "
```

wsad1:
@rusty1s , this is ready to merge. please take a look.

rusty1s:
Super, will take a look ASAP.

torch_geometric/nn/models/pg_explainer.py
RBendias: I think the problem still exists when nodes with only self-connections are considered.

wsad1: Thanks @RBendias  and @SensorLock for bringing this up. Both issues 1. node with 1 incoming edge and 2. node with only self loop are the same and fixed by setting `dim=1` as @SensorLock  suggested.

rusty1s:
Closing this in favor of https://github.com/pyg-team/pytorch_geometric/pull/6204

PR #2611, opened 2021-05-20
Index select type issue

josephenguehard:
When selecting an index from a dataset, the original repo requires an integer, slice, tuple, tensor or a list:

```python
import numpy as np
tu_dataset = TUDataset(root='~/notebooks/tmp/ENZYMES', name='ENZYMES')

idx = 0
tu_dataset[idx]
>>> Data(edge_index=[2, 168], x=[37, 3], y=[1])
```

However, when using numpy integers like `np.int32`, it breaks:

```python
idx = np.int32(0)
tu_dataset[idx]
>>>IndexError: Only integers, slices (`:`), list, tuples, and long or bool tensors are valid indices (got int32).
```

This PR fixes it:

```python
idx = np.int32(0)
tu_dataset[idx]
>>> Data(edge_index=[2, 168], x=[37, 3], y=[1])
```

It is also now possible to pass numpy arrays as indices:

```python
idx = np.arange(10)
tu_dataset[idx]
>>>ENZYMES(10)
```

### DIFF
--- a/torch_geometric/data/dataset.py
+++ b/torch_geometric/data/dataset.py
@@ -204,7 +204,7 @@ class Dataset(torch.utils.data.Dataset):
             elif idx.dtype == torch.bool or idx.dtype == torch.uint8:
                 return self.index_select(
                     idx.nonzero(as_tuple=False).flatten().tolist())
-        elif isinstance(idx, list) or isinstance(idx, tuple):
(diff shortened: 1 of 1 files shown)


rusty1s:
Super, I like it. Thank you very much.

PR #2625, opened 2021-05-22
Add PyTorch dataset and dataloader classes for TemporalData

yulonglin:
## What
This adds native PyTorch `DataLoader` support for `TemporalData`, which is used for dynamic graph datasets. This is based on the idea mentioned here: #2567 

## Why
By allowing dynamic graph events in `TemporalData` format to be loaded via the new `TemporalDataLoader`, we can inherit the functionality of PyTorch `DataLoader`. For example, this is useful for:
- Interfacing with other libraries such as PyTorch Lightning, where a PyTorch `DataLoader` is expected
- Shuffling data loading using the `shuffle` argument when defining static baselines where time information is not important
- Parallelising data loading using the `num_workers` argument, in particular when there is preprocessing

Note: Currently, `TemporalData` has a `seq_batches()` method that generates slices of a tensor, batch by batch. In the case where we have a small dataset that we just need to load sequentially without preprocessing and don't need a `DataLoader`-like API, `seq_batches()` is faster. This is because using PyTorch `DataLoader` means splitting the data into individual training samples before collating them, but `seq_batches` skips the splitting and collating steps. This is similar to PyTorch and PyTorch Geometric `DataLoader` classes.

## How
A `TemporalDataset` is created that wraps around `TemporalData` to create a PyTorch `Dataset` object. Individual samples are collated by the `TemporalDataLoader` using a special collater for `TemporalData` objects.  

## Testing
Tests were implemented to check for the shapes and values of inputs and outputs to `TemporalDataLoader`. The TGN-attn implementation in `examples/` was modified by replacing the `seq_batches()` method with `TemporalDataLoader` and it produced similar results.

### DIFF
--- a/examples/tgn.py
+++ b/examples/tgn.py
@@ -17,7 +17,6 @@ import torch
 from torch.nn import Linear
 from sklearn.metrics import average_precision_score, roc_auc_score
 
-from torch_geometric.data import TemporalDataset, TemporalDataLoader
 from torch_geometric.datasets import JODIEDataset
 from torch_geometric.nn import TGNMemory, TransformerConv
 from torch_geometric.nn.models.tgn import (LastNeighborLoader, IdentityMessage,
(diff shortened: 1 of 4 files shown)


Padarn:
hey @yulonglin and @rusty1s  - I was looking at the requirement for this from https://github.com/pyg-team/pytorch_geometric/issues/3230, what are your plans for this PR?

rusty1s:
I'm in favor of re-using the `torch_geometric.loader.DataLoader` for mini-batching of temporal graphs as well. This should be (hopefully) supported out-of-the-box once `TemporalData` inherits from `BaseData`.

Padarn:
Yeah that makes a lot of sense to me. Will revisit if anything is needed once that is addressed from https://github.com/pyg-team/pytorch_geometric/issues/3230

rusty1s:
Closing this in favor of https://github.com/pyg-team/pytorch_geometric/pull/3985. Thank you!

PR #2633, opened 2021-05-25
Update x_conv.py

Bear-kai:
fix the wrong size of matmul()


rusty1s:
LGTM :) Thank you very much!

PR #2640, opened 2021-05-26
Add more NumPy indexing and allow negative indices in Batch

josephenguehard:
This PR has three different parts:

1. Extend NumPy arrays indexing for Dataset. It is now possible to use ``bool`` NumPy arrays, similarly to tensors:

```python
import numpy as np

dataset = TUDataset(root='~/notebooks/tmp/ENZYMES', name='ENZYMES')
idx = np.arange(5).astype('bool')
dataset[idx]
>>> ENZYMES(4)
```

2. Add NumPy arrays indexing to Batch, similarly to Dataset:

```python
 from torch_geometric.data import Batch

idx = np.arange(5)
Batch.from_data_list([dataset[i] for i in range(10)])[idx]
>>> [Data(edge_index=[2, 168], x=[37, 3], y=[1]),
...     Data(edge_index=[2, 102], x=[23, 3], y=[1]),
...     Data(edge_index=[2, 92], x=[25, 3], y=[1]),
...     Data(edge_index=[2, 90], x=[24, 3], y=[1]),
...     Data(edge_index=[2, 90], x=[23, 3], y=[1])]
```

3. Allow negative indices for Batch:

```python
batch = Batch.from_data_list([dataset[i] for i in range(10)])
batch[-1]
>>> Data(edge_index=[2, 106], x=[32, 3], y=[1])
```

The range of allowed indices is [-n, n-1] for a Batch of size n, similarly to tensors.

### DIFF
--- a/torch_geometric/data/batch.py
+++ b/torch_geometric/data/batch.py
@@ -1,5 +1,6 @@
 from typing import List
 
+import numpy as np
 import torch
 from torch import Tensor
 from torch_sparse import SparseTensor, cat
@@ -174,6 +175,7 @@ class Batch(Data):
 
         for key in self.__slices__.keys():
             item = self[key]
+            idx = idx % self.num_graphs if -self.num_graphs < idx < 0 else idx
             if self.__cat_dims__[key] is None:
                 # The item was concatenated along a new batch dimension,
                 # so just index in that dimension:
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
Looks great to me. Thank you very much!

PR #2657, opened 2021-05-29
Fix to `test` in proteins_mincut_pool.py 

wsad1:
Bug: The `test` function in [proteins_mincut_pool.py](https://github.com/rusty1s/pytorch_geometric/blob/a17f81217f043bb9d97eca83e1b671b5f48e0ae6/examples/proteins_mincut_pool.py#L91)  returns loss of the last batch. 
Fix: `test` now returns average loss across all batches.


rusty1s:
Thanks!

PR #2666, opened 2021-06-01
Error when computing num_graphs

josephenguehard:
Num graphs inconsistent in Batch:

```python
from torch_geometric.data import Batch

dataset = TUDataset(root='~/notebooks/tmp/ENZYMES', name='ENZYMES')
batch = Batch.from_data_list([dataset[i] for i in range(10)])

batch.__num_graphs__
>>> 10

batch.ptr.numel() + 1
>>> 12

int(batch.batch.max()) + 1
>>> 10
```

This PR replaces ``batch.ptr.numel() + 1`` with ``batch.ptr.numel() - 1``

### DIFF
--- a/torch_geometric/data/batch.py
+++ b/torch_geometric/data/batch.py
@@ -271,7 +271,7 @@ class Batch(Data):
         if self.__num_graphs__ is not None:
             return self.__num_graphs__
         elif self.ptr is not None:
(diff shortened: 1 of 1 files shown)


rusty1s:
Oh, wow. Thanks for noticing!

PR #2691, opened 2021-06-04
Fixed dilated KNN to have fixed number of unique `K`s

### LINKED ISSUE #2684: PointCNN dilated KNN might select less than `K` neighbours
## 🐛 Bug

In the current X-Conv implementation, I can see the following:

https://github.com/rusty1s/pytorch_geometric/blob/e8e004439e3204a7b888a21e508c45d166c5817c/torch_geometric/nn/conv/x_conv.py#L130-L137

As, `torch.randint` would sample with replacement, there's a chance that the number of actual neighbours can be less than `K`. I am thinking we can fix this using something like:

```
indices = torch.randperm(K * dil)[:K]
```

mayur-ag:
- [x] Fixes #2684

### DIFF
--- a/torch_geometric/nn/conv/x_conv.py
+++ b/torch_geometric/nn/conv/x_conv.py
@@ -125,17 +125,11 @@ class XConv(torch.nn.Module):
         edge_index = knn_graph(pos, K * self.dilation, batch, loop=True,
                                flow='target_to_source',
                                num_workers=self.num_workers)
-        row, col = edge_index[0], edge_index[1]
 
         if self.dilation > 1:
-            dil = self.dilation
-            randnum = torch.randperm(K * dil, dtype=torch.long,
-                                     device=row.device)[:K]
-            edge_index = edge_index.view(2, -1, K * dil)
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/conv/x_conv.py
```diff
@@ -129,12 +129,13 @@ def forward(self, x: Tensor, pos: Tensor, batch: Optional[Tensor] = None):
 
         if self.dilation > 1:
             dil = self.dilation
-            index = torch.randint(K * dil, (N, K), dtype=torch.long,
-                                  device=row.device)
-            arange = torch.arange(N, dtype=torch.long, device=row.device)
-            arange = arange * (K * dil)
-            index = (index + arange.view(-1, 1)).view(-1)
-            row, col = row[index], col[index]
+            randnum = torch.randperm(K * dil, dtype=torch.long,
```
rusty1s, line 132: It looks like the same indices are selected for all nodes. Can we prevent that, e.g. via:
```python
index = torch.randperm(N * K * dil).view(N, K * dil)[N, :K] % (K * dil)
```

PR #2692, opened 2021-06-04
Add split functionality for edge attributes

saiden89:
Added functionality for edge attribute splitting for positive edge indexes only, as discussed in #2669. Negative edge attributes are not created because highly dependant on individual use cases.
Now, after calling `train_test_split_edges` the returned `Data` object will also carry `train_pos_edge_attr`, `val_pos_edge_attr` and `test_pos_edge_attr` as attributes.


rusty1s:
Super. Thanks a lot!

PR #2693, opened 2021-06-05
Changes to Mempooling.

wsad1:
Issue:
  Currently it's not straightforward to stack `Mempooling` layers. See code below, the call to mpool2 will throw an error as out1 has shape `B*N*F` but `forward` expects `Nmax*F`. 
  ```
  mpool1 = MemPooling(4, 8, heads=3, num_clusters=2)
  mpool2 = MemPooling(8, 4, heads=2, num_clusters=1)
  out1,_=mpool1(x,batch)
  out2,_=mpool2(out1) # this will throw an error.
  ```
Fix:
This PR fixes that by adding a line to allow dense inputs to forward. Another way to allow stacking would have been to create a `to_sprase_batch` function that does the reverse of `to_dense_batch`. I could create a PR for that if its useful.

### DIFF
--- a/test/nn/pool/test_mem_pool.py
+++ b/test/nn/pool/test_mem_pool.py
@@ -10,10 +10,11 @@ def test_mem_pool():
 
     x = torch.randn(17, 4)
     batch = torch.tensor([0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 4, 4, 4])
-    _, mask = to_dense_batch(batch, batch)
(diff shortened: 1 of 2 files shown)


rusty1s (commented):
Thanks. I'm wondering whether we need a `mask` as input argument as well? Currently, in case the input format is of type "dense", the mask is set to `True` for all values by default.

wsad1:
Added mask to `forward`.
Thanks for the good suggestion.

rusty1s:
Thanks a lot :)

PR #2727, opened 2021-06-11
SuperGATConv bug?

anniekmyatt:
I think there might be a typo in the SuperGATConvlayer? I think it should be using the positive supervision edges in the attention calculation, rather than the message passing edges (just like in the calculation of the attention for the negative supervision edges).

### DIFF
--- a/torch_geometric/nn/conv/supergat_conv.py
+++ b/torch_geometric/nn/conv/supergat_conv.py
@@ -187,7 +187,7 @@ class SuperGATConv(MessagePassing):
             pos_edge_index = self.positive_sampling(edge_index)
 
             pos_att = self.get_attention(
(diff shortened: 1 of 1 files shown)


rusty1s:
Thanks. This is not necessarily a bug, since `pos_edge_index` is a subset of `edge_index`, but it's definitely a typo :) Thanks for fixing!

anniekmyatt:
Glad it's helpful! 

I thought that for link prediction we tend to want our supervision edges and message passing edges to be disjoint, so we don't end up just memorising the supervision edges? Very happy to be corrected, I am still working on getting my head around a lot of these things!

I appreciate that it doesn't matter for the PR and it's more for my understanding.

PR #2734, opened 2021-06-14
rgcn link prediction 

moritzblum:
implemented rgcn link prediction from Schlichtkrull et al. as examle, the required FB15k237 dataset and the evaluation


moritzblum:
test scores:
```
{
   'MRR': 0.24043533205986023, 
   'H@1': 0.16512753069400787, 
   'H@3': 0.26759013533592224, 
   'H@10': 0.3896218240261078
}
```

examples/rgcn_link_prediction.py
```diff
@@ -0,0 +1,232 @@
...
+        return score
+
+
+def negative_sampling(edge_index, edge_type, num_nodes, num_edges, omega=1.0, corrupt_subject=True,
+                      corrupt_predicate=False, corrupt_object=True):
+    negative_samples_count = int(omega * edge_type.size(0))
+    permutation = np.random.choice(np.arange(edge_type.size(0)), negative_samples_count)
+    s = edge_index[0][permutation]
+    r = edge_type[permutation]
+    o = edge_index[0][permutation]
```
anniekmyatt, line 54: Hope it's ok if I comment. Should it be `o = edge_index[1][permutation]`?

moritzblum: Oh yeah, you're right. Thanks!

moritzblum:
I am not sure what the problem is. Any suggestions?

rusty1s:
The tests seem to fail independent of the PR, so don't worry about that.

Nonetheless, thanks for reminding me of this PR. It seems that I have forgotten to submit my review :(

torch_geometric/datasets/link_prediction.py
```diff
@@ -0,0 +1,117 @@
+import os
+import urllib.request
+import torch
+from torch_geometric.data import InMemoryDataset, Data
+
+
+class LinkPrediction(InMemoryDataset):
```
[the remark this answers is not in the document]
moritzblum: The class is designed that datasets of similar format can be implemented in the future easily, e.g. WN18. Maybe we choose something more general, e.g. RelationalLinkPredictionDatasets?

torch_geometric/datasets/link_prediction.py
```diff
@@ -0,0 +1,117 @@
...
+
+        assert name in ['FB15k-237']
+        self.name = name
+        super(LinkPrediction, self).__init__(root, transform, pre_transform)
+        self.data, self.slices = torch.load(self.processed_paths[0])
+        self.data.num_relations = self.data.num_relations.tolist()
+
+    def download(self):
+        if self.name == 'FB15k-237':
+            urllib.request.urlretrieve("https://raw.githubusercontent.com/MichSchli/RelationPrediction/master/data/FB-Toutanova/README.txt", f"{self.raw_dir}/README.txt")
```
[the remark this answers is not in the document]
moritzblum: Makes sense. I will adjust the code. :)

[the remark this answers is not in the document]
rusty1s: Thanks, will wait on it before merging.

rusty1s:
I converted the example to a single script and re-wrote the evaluation metric. Please let me know if the applied changes work for you.

moritzblum (commented):
Looks great. Just a few remarks.

torch_geometric/datasets/rel_link_pred_dataset.py
```diff
@@ -85,7 +85,15 @@ def process(self):
                 kwargs[f'{split}_edge_index'] = torch.tensor([src, dst])
                 kwargs[f'{split}_edge_type'] = torch.tensor(rel)
 
-        data = Data(num_nodes=len(entities_dict), **kwargs)
+        # For message passing, we add reverse edges and types to the graph:
```
moritzblum, line 88: If this is done in the dataset before the sampling, the negative samples have no reverse edges and are easy to recognize. This should be applied to the negative samples as well. Furthermore, it could be more memory efficient if this is done only on the current batch.

examples/rgcn_link_pred.py
```diff
@@ -0,0 +1,164 @@
...
+    z = model.encode(data.edge_index, data.edge_type)
+
+    valid_mrr = test_split(z, data.valid_edge_index, data.valid_edge_type)
+    test_mrr = test_split(z, data.valid_edge_index, data.valid_edge_type)
+
+    return valid_mrr, test_mrr
+
+
+@torch.no_grad()
+def test_split(z, edge_index, edge_type):
```
moritzblum, line 112: This function name sounds strange. What about test_scores()?

moritzblum (commented):
Some technical aspects that were lost in your simplification. Thanks for the adjustments, I learned a lot through them.

examples/rgcn_link_pred.py
```diff
@@ -0,0 +1,164 @@
...
+
+model = GAE(
+    RGCNEncoder(data.num_nodes, hidden_channels=500,
+                num_relations=dataset.num_relations),
+    DistMultDecoder(dataset.num_relations // 2, hidden_channels=500),
+)
+optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+
+
+def negative_sampling(edge_index, num_nodes):
```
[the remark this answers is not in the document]
rusty1s: Yeah, this is indeed a simplification as checking for duplicates is real pain. If you want to, please feel free to add it back in.

torch_geometric/datasets/rel_link_pred_dataset.py
rusty1s: Not sure I understand that. While message passing makes use of reverse edges, the decoder does not.

moritzblum: Ok, I didn't see that the decoder doesn't get the reverse edges. In my implementation, both (encoder & decoder) got the reverse edges, as in most implementations I saw. The original paper does not look at this in detail. I don't know if we can achieve the same results this way.

rusty1s: I think we only need to decode "one direction" as the reverse direction is given by it as well. IMO, reverse edges are only useful to enhance the message passing flow, and shouldn't play any role in the prediction part.

moritzblum: Makes sense.

rusty1s:
Thanks for your comments and for the PR. I'm merging this now. Please feel free to address your concerns, if any, in a follow-up PR.

PR #2737, opened 2021-06-16
avoid the 'inf'

Ethanzjp:
# What problem does this PR solve?
when replace 'add_self_loop' using 'replace_self_loop', it occurs 'inf'.


rusty1s:
Thanks :)

PR #2739, opened 2021-06-17
fix relu activation function

Ethanzjp:


rusty1s:
Thank you, but I don't think this needs to be fixed, as the `GAT` paper explicitly mentions using ELU activation.

PR #2743, opened 2021-06-17
The Wikipedia datases were introduced in the MUSAE paper not in Geom-GCN

benedekrozemberczki:
Dear Matthias,

These datasets were introduced in the Multi-scale Attributed Node Embedding paper. If you look in the Geom-GCN paper they cited our paper.

https://arxiv.org/abs/1909.13021

Benedek


rusty1s:
Thanks and sorry :)

PR #2748, opened 2021-06-17
Adding the Facebok Page-Page dataset

benedekrozemberczki:
Another widely used benchmark dataset from:

https://arxiv.org/abs/1909.13021

Also up on:

http://snap.stanford.edu/data/index.html


rusty1s:
Thanks a lot :)

PR #2752, opened 2021-06-18
Added the FEATHER datasets

benedekrozemberczki:
Added two node classification datasets from this paper:

Characteristic Functions on Graphs: Birds of a Feather, from Statistical Descriptors to Parametric Models

https://arxiv.org/abs/2005.07959

The datasets are also hosted on SNAP.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,55 @@
+name: Documentation
+
+on: [push, pull_request]
+
+jobs:
+
+  make_html:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.8]
+        torch-version: [1.9.0]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Update pip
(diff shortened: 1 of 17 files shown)


benedekrozemberczki:
Besides the FEATHER datasets I added the following modifications:

1. The Geom-GCN people dropped certain node features and truncated the target variable for the Wiki dataset which is not in the original paper.
2. One of the Wiki datasets was left out for no reason.
3. Added the datasets from the GEMSEC paper - these are widely used for multi-label multi-class node classification

rusty1s:
Thank you very much! You are that the Wiki dataset used in Geom-GCN differs from the original dataset. I was not aware of this. One concern that I have is backward compatibility/supporting the Geom-GCN pre-processing strategy as well.

Do you think it is useful to add some `geom_gcn_preprocess` argument that converts the original dataset into the GeomGCN variant (e.g., adding splits and casting the dataset to node classification task)?

benedekrozemberczki:
Hi there,

Yes, I can add a pre-processing step and add the classification targets to the NumPy pickle and load the appropriate target for the users.

Bests,

Benedek

benedekrozemberczki:
1. Reuploaded the dataset with the labels and masks on graphmining.ai.
2. Created a set of masks (10) and classification labels for the third dataset.
2. Added a preprocessing parameter for the loader - now it supports the masks and both types of targets (classification/regression).

rusty1s:
Thank you very much for supporting this :) This is great!

PR #2777, opened 2021-06-28
Transpose the stored transformation matrix in LinearTransformation

beneisner:
so that the matrix multiplication matches the expected convention of a linear transformation. See #2772 for discussion.

Tested by running:
```
pytest --no-cov test/transforms/test_linear_transformation.py
```


rusty1s:
Super :)

Issue #1574, opened 2020-08-25
Reading tetrahedral meshes (.msh/.vtk file formats) and feed to graph Nets


isurusuranga:
## ❓ Questions & Help

I'm new to pytorch-geometric and I am trying to feed 3D volumetric meshes (i.e. .msh) into graph neural nets. I have gone though the documentation and I have found that pytorch-geometric supports trimesh. But I'm not sure that trimesh can load .msh files properly means may be some data get lost on reload.
I have produced volumetric meshes using gmsh and now I need to feed these data into graph neural nets using the pytorch-geometric. Is there a way to feed this type of data ? I really appreciate your help in this matter. I have attached a sample .msh file for your further reference. 

[ellipsoid_0.zip](https://github.com/rusty1s/pytorch_geometric/files/5125649/ellipsoid_0.zip)

rusty1s:
Hi and thanks for this issue. Loading `*.msh` files into PyG is currently not supported, I am sorry! You may need to write your own processing procedure to convert these types of meshes into a graph-structured form. If you are interested in doing so, please consider sending a PR. I would be happy to help you out :)

isurusuranga:
I just implemented two functions as below using the meshio package (pip3 install meshio). This is something very similar to the trimesh functionalities in the util package in pytorch geometric.

```
def from_meshio(mesh):
    r"""Converts a :.msh file to a
    :class:`torch_geometric.data.Data` instance.

    Args:
        mesh (meshio.read): A :obj:`meshio` mesh.
    """

    if meshio is None:
        raise ImportError('Package `meshio` could not be found.')

    pos = torch.from_numpy(mesh.points).to(torch.float)
    face = torch.from_numpy(mesh.cells[0].data).to(torch.long).t().contiguous()
    tetra = torch.from_numpy(mesh.cells[1].data)

    return Data(pos=pos, face=face, tetra=tetra)

def to_meshio(data):
    r"""Converts a :class:`torch_geometric.data.Data` instance to a
    :obj:`.msh format`.

    Args:
        data (torch_geometric.data.Data): The data object.
    """

    if meshio is None:
        raise ImportError('Package `meshio` could not be found.')
    
    points = data.pos.detach().cpu().numpy()
    faces = data.face.detach().t().cpu().numpy()
    tetra = data.tetra.detach().cpu().numpy()
    
    cells = [("triangle", faces), ("tetra", tetra)]

    return meshio.Mesh(points, cells)
```

[...]

PR #2808, opened 2021-07-03
read and convert volumetric meshes for feeding to graphNets #1574

isurusuranga:
https://github.com/rusty1s/pytorch_geometric/issues/1574

Reading tetrahedral mesh in .msh/.vtk format for feeding to graph Nets


javirk:
Looking at the code and using my own examples, I wonder if the line 
`tetra = torch.from_numpy(mesh.cells[0].data).to(torch.long).t().contiguous()` (line 235 in `/utils/convert.py`)
is correct. In my case mesh.cells[0]  doesn't have "tetra" information but vertex, wouldn't it be better to use
`tetra = torch.from_numpy(mesh.cells_dict['tetra']).to(torch.long).t().contiguous()`
instead?

In my case `(mesh.cells[1].data == mesh.cells_dict['tetra']).all()` returns `True`, so I guess they contain the same information, but I haven't found any documentation.

isurusuranga:
Thank you for bringing this to our attention. I use a.vtk file to test this approach, and it contains a cells property in the 0th element, as seen below. In order to extract tetra, I utilised ref.cells[0].data.

[CellBlock(type='tetra', data=array([[172, 275, 279, 574],
        [640, 700, 581, 767],
        [459, 521, 479, 557],
        ...,
        [787, 805, 719, 812],
        [776, 787, 678, 812],
        [776, 805, 787, 812]]))]

Anyway I didn't realize that it had an additional attribute called cells_dict['tetra']. Thank you for bringing this to my attention once more. I just checked it, and you're right, they have the same information. I'll apply the modifications to my pull request as soon as possible. 

Finally, I can confirm that it is consistent with tetrahedral meshes in.msh and.vtk formats after this modification. :)

rusty1s:
Since `meshio` is an optional dependency, you need to install it manually [here](https://github.com/rusty1s/pytorch_geometric/blob/master/.github/workflows/testing.yml#L58).

PR #2837, opened 2021-07-09
Improve terminology

MarkusZopf:
I think the previous version of the document uses the term 'symmetric' incorrectly. A symmetric matrix is a square matrix that is is equal to its transpose (https://en.wikipedia.org/wiki/Symmetric_matrix). However, the text is only talking about the shape of the matrix, not its content. Hence, 'square (matrix)' would be the correct term to use.


rusty1s:
Thanks :)

PR #2838, opened 2021-07-09
Add batch_size input to to_dense_batch

jimmiebtlr:
to_dense_batch doesn't work quite as expected if batches can also have 0 elements.  Add a param to handle this case.

Also adds a note about batch needing to be ordered.


rusty1s:
Super. Thanks a lot!

PR #2843, opened 2021-07-11
MD17

M-R-Schaefer:
Added the [MD17 ](http://quantum-machine.org/gdml/#datasets) datasets.
These are essentially the MNIST of Neural Network force fields just as QM9 is the MNIST of molecular property prediction.


rusty1s:
Super cool :) Thanks a lot!

PR #2917, opened 2021-07-28
Numerical error in LayerNorm leading to NaN in gradients

fbragman:
Similar to this issue: https://github.com/rusty1s/pytorch_geometric/issues/2559

Added same fix to `layer_norm`, changing

```
out = x / (var.sqrt()[batch] + self.eps)
```

to 

```
out = x / (var + self.eps).sqrt()[batch]
```


rusty1s:
Thanks!

PR #2923, opened 2021-07-30
Update from_networkx

PabloAMC:
Changes discussed in 
https://github.com/rusty1s/pytorch_geometric/issues/2920

### DIFF
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -105,16 +105,23 @@ def to_networkx(data, node_attrs=None, edge_attrs=None, to_undirected=False,
     return G
 
 
-def from_networkx(G, group_node_attrs: List[str] or all=None, group_edge_attrs: List[str] or all=None):
+def from_networkx(G, group_node_attrs: Optional[List[str]] = None,
+                  group_edge_attrs: Optional[List[str]] = None):
     r"""Converts a :obj:`networkx.Graph` or :obj:`networkx.DiGraph` to a
     :class:`torch_geometric.data.Data` instance.
 
     Args:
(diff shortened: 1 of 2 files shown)


torch_geometric/utils/convert.py
```diff
@@ -119,12 +123,23 @@ def from_networkx(G):
...
+    node_attributes = feat_dict.keys()
+    if group_node_attrs is all:
+        group_node_attrs = node_attributes
 
     for i, (_, feat_dict) in enumerate(G.nodes(data=True)):
+        if set(feat_dict.keys()) != set(node_attributes): raise ValueError(f'Not all nodes contain the same attributes')
         for key, value in feat_dict.items():
             data[str(key)] = [value] if i == 0 else data[str(key)] + [value]
 
+    _, _, feat_dict = list(G.edges(data=True))[0]
```
PabloAMC, line 136: I am not sure why this line is failing in the pytest tests, I can run it locally without fail...

rusty1s: No worries about linting, I will fix that later :)

rusty1s: This may be due to different `networkx` versions as they are changing their API frequently.

PabloAMC: Actually, it was just for a test of a graph with no edges, and I was looking in the wrong direction. So fortunately that's solved now!

PabloAMC:
I'm not so sure about how to correct the remaining Linting CI fail, but the rest seems to work

rusty1s:
Thanks. Will have a look in the upcoming week.

rusty1s:
Thank you!

PR #2924, opened 2021-07-30
Transform Base Class

CCInc:
As per https://github.com/rusty1s/pytorch_geometric/discussions/2903:

Enables typechecking/linting on Transform objects by inheriting from a common class other than `object`. 

```python
# current:
from typing import Callable

def run(transform: Callable):
	T.Compose([transform])
	
run(abs) # callable, but not a transform

# proposed:
from torch_geometric.transforms.base_transform import BaseTransform

def run(transform: BaseTransform): # all transforms inherit from common class
	T.Compose([transform])

run(T.RandomScale()) # okay
run(abs) # throws linter error
```

Would you want me to update typing references for `Callable` to `BaseTransform` as well?

### DIFF
--- a/torch_geometric/transforms/__init__.py
+++ b/torch_geometric/transforms/__init__.py
@@ -1,3 +1,4 @@
+from .base_transform import BaseTransform
 from .compose import Compose
 from .to_sparse_tensor import ToSparseTensor
 from .to_undirected import ToUndirected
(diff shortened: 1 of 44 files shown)


rusty1s:
Thanks. LGTM. I don't think there is a need to update `Callable` typing references since a pure function is still a valid transform.

PR #2932, opened 2021-08-01
Correct gini coefficient mathcal formula

TimDaub:
/cc @rhsimplex @Jostarndt

Current formula
Screen Shot 2021-08-01 at 22 00 47

PR suggestion:
Screen Shot 2021-08-01 at 22 00 28

For a set {1,0}, my napkin math results in an expected Gini coefficient of 1 for the proposed formula.


rhsimplex:
Thanks for the correction @TimDaub. I had indeed moved the parenthesis accidentally.  With your correction, the documentation now matches the implementation. Of course, we could factor out an n to clean it up a bit

PR #2950, opened 2021-08-06
Make informational `print`s go to stderr

Linux-cpp-lisp:
Currently, `Dataset` prints out
```
Processing...
Done!
```
when processing a dataset. This is useful feedback for the user (usually), but goes to `stdout`, corrupting the output of scripts that depend on it. (I am trying to build a UNIX-style script where messages go to `stderr` and only the machine readable output goes to `stdout` for easy piping into a file or other tool.)

This PR switches those `print` statements to print to stderr instead. Two other options are:
 1. Use Python's built-in `logging` to give the user full control of these messages
 2. Do not emit these updates at all

Searching, this seems to be the only place in the code that uses `print`.

### DIFF
--- a/torch_geometric/data/dataset.py
+++ b/torch_geometric/data/dataset.py
@@ -1,5 +1,6 @@
 from typing import List, Optional, Callable, Union, Any, Tuple
 
+import sys
 import re
 import copy
 import warnings
@@ -159,7 +160,7 @@ class Dataset(torch.utils.data.Dataset):
         if files_exist(self.processed_paths):  # pragma: no cover
             return
 
-        print('Processing...')
+        print('Processing...', file=sys.stderr)
 
         makedirs(self.processed_dir)
         self.process()
@@ -169,7 +170,7 @@ class Dataset(torch.utils.data.Dataset):
         path = osp.join(self.processed_dir, 'pre_filter.pt')
         torch.save(_repr(self.pre_filter), path)
 
-        print('Done!')
+        print('Done!', file=sys.stderr)
 
     def __len__(self) -> int:
         r"""The number of examples in the dataset."""


rusty1s:
Thanks! I think I'm in favor of doing this via `logging`, as outputing to `stderr` may also have undesired side-effects.

Linux-cpp-lisp:
@rusty1s updated the PR, agree that `logging` is the most proper way to do it

rusty1s:
While we are at it, feel free to also make adjustments to `data/extract.py` and `data/download.py`.

Linux-cpp-lisp:
^

rusty1s:
Thanks. I actually wasn't aware that `logger.info` does not print out anything by default (only with `logging.basicConfig(level=logging.DEBUG)`). Is it possible to print out logging information by default, and only silence logging in case the user wants to (instead of the other way round?)

Linux-cpp-lisp:
I don't _think_ so, since my understanding is that the whole point of the architecture is to prevent libraries from circumventing the users settings on the root logger, giving the host application full control...

Obviously, changing the root logger's settings from a library would be very bad practice.

You could of course circumvent `logging` entirely and have your own `torch_geometric.logging_on` option of some kind, but that seems like a lot of code infrastructure for a couple of mostly unimportant informational messages.

rusty1s:
Yes, I understand. Changing root logger settings is definitely something we do not want. As I think the print information is quite useful for the average user, it might be a bit weird to disable them by default. If you do not mind and it fits your use-case, I would very much appreciate it if you can change the behavior to `print(..., file=sys.stderr)` once again. Sorry for being so unexperienced with the `logging` module and for stealing your time.

Linux-cpp-lisp:
@rusty1s no worries, and as always we appreciate your responsiveness when little issues come up!

I reverted (a bit messily, couldn't figure out the git commands) to the `print` statements and changed all others I found in the codebase to go to `stderr`.

While I do still think it would be best not to put any unexpected output _anywhere_ by using `logging`, I agree with your assessment that this might be surprising for the majority of users of the code. Most people probably use `torch_geometric` in a "Jupyter notebooks and hacked together data science scripts" context, which wants responsive updates on what's happening, and not "tools with well defined output" like we are. At least by using `stderr` we avoid corrupting `stdout` data outputs :+1:

PR #2965, opened 2021-08-12
support `del data["key"]` in `Data`

Linux-cpp-lisp:
Currently the only way to remove a key from a `Data` object is by calling `delattr`, which somewhat breaks the abstraction of `Data` being `dict`-like and only using the `*attr` methods as an internal implementation detail.

This PR adds a `__delitem__` method to `Data` completing the `dict`-like interface.


rusty1s:
Super. Thanks!

Issue #2979, opened 2021-08-16
The inconsistent feature dimension size of WikipediaNetwork dataset


ZhuYun97:
## 🐛 Bug

The feature dimension size of Chameleon and Squirrel dataset is 2325 and 2089 in pyg 1.7.0. But in the latest pyg version(1.7.2), the feature size both are 128 for Chameleon and Squirrel dataset which is not the same as reported in [ref1](https://arxiv.org/abs/2002.05287).
![image](https://user-images.githubusercontent.com/37068560/129560098-00def02d-00f0-498e-acaa-d85efe5cc1ff.png)
[ref1](https://arxiv.org/abs/2002.05287)

## Environment

* PyTorch Geometric version: 1.7.2

## Additional context
If I want to use the original dimension size(2325 and 2089), is it the only way to degrade pyg version?

rusty1s:
Thanks for reporting, I fixed this issue in master, see https://github.com/rusty1s/pytorch_geometric/commit/665c5dad47884eeb5738478db99d75dec85b1816. If you install PyG from master
```python
pip install git+https://github.com/rusty1s/pytorch_geometric.git
```
you should now be able to see the correct feature sizes.

PR #3065, opened 2021-09-02
Fix:Replace to_undirected() with coalesce().

GitEventhandler:
The issue #2979 fixed in commit https://github.com/rusty1s/pytorch_geometric/commit/665c5dad47884eeb5738478db99d75dec85b1816 still cause different benchmark result on dataset chameleon or squirrel. This problem is caused by changing function coalesce() to to_undirected(). It may be needed to reset to_undirected() to coalesce() in order to keep the API layer compatibility.


rusty1s:
Thanks. Sounds good :)

PR #3066, opened 2021-09-02
fix namedtuple become tuple when move Data to GPU

WuliangHuang:
Some tensor like objects are actually `namedtuple`, such as `torch.nn.utils.rnn.PackedSequence`.
A related discussion is [here](https://github.com/pytorch/pytorch/issues/44009). We modify this function first to support PackedSequence, and then we also hope to be compatible with other `namedtuple `types.
It is not easy to check if the item is an instance of a `namedtuple `directly, so we check attribute `_asdict` and `_fields`.

However, parameter `device `in `PackedSequence.cuda` function in `torch.nn.utils.rnn.py` has a default value, which is `cuda`.
It can be seen since [here](https://github.com/pytorch/pytorch/commit/527b10c2d1b605828df3b4844436ef0636d2c5e6#).
Therefore, we also modify `.cuda` function for supporting `PackedSequence `as well.


rusty1s:
Thanks :)

PR #3090, opened 2021-09-06
Update GCN example code

monk1337:
The example code is not using the parent class in super class and giving error while running the code.


rusty1s:
Thanks :)

PR #3101, opened 2021-09-08
Add feature decomposition method

ZhouAo-ZA:
This is a PR for  adding the feature decomposition method to the `message_passing `module.

### DIFF
--- a/torch_geometric/nn/conv/message_passing.py
+++ b/torch_geometric/nn/conv/message_passing.py
@@ -197,7 +197,7 @@ class MessagePassing(torch.nn.Module):
 
         return out
 
-    def propagate(self, edge_index: Adj, size: Size = None, decomposed_layer=1, **kwargs):
+    def propagate(self, edge_index: Adj, size: Size = None, decomposed_layer = 1, **kwargs):
         r"""The initial call to start propagating messages.
 
         Args:
--- a/torch_geometric/nn/conv/message_passing.py
+++ b/torch_geometric/nn/conv/message_passing.py
@@ -197,10 +197,11 @@ class MessagePassing(torch.nn.Module):
 
         return out
 
-    def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
+    def propagate(self, edge_index: Adj, size: Size = None, decomposed_layer=2, **kwargs):
         r"""The initial call to start propagating messages.
 
         Args:
+
             edge_index (Tensor or SparseTensor): A :obj:`torch.LongTensor` or a
                 :obj:`torch_sparse.SparseTensor` that defines the underlying
                 graph connectivity/message passing flow.
@@ -224,6 +225,12 @@ class MessagePassing(torch.nn.Module):
                 and assumed to be quadratic.
                 This argument is ignored in case :obj:`edge_index` is a
                 :obj:`torch_sparse.SparseTensor`. (default: :obj:`None`)
+            decomposed_layer (integer, optional): The number of feature decomposition layers,
+                the default value is 1 (that is, the feature decomposition method is not used).
+                For the gather-scatter mode calculated on the CPU side, the feature decomposition
+                method can effectively reduce the peak memory usage and increase the calculation speed.
+                When the average node degree is higher, it is more effective to use this method
+                (for example, Reddit dataset). More details in this paper: https://arxiv.org/abs/2104.03058 .
             **kwargs: Any additional data which is needed to construct and
                 aggregate messages, and to update node embeddings.
         """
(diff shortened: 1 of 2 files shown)


rusty1s:
Super. Thanks a lot :) Will look into it ASAP.

ZhouAo-ZA:
The PR result shows that some checks were not successful, but I don’t know why, because my code does not change its logic, so it should be consistent with all tests.

rusty1s:
Yes, no worries. The tests are failing due to an internal error in `torch-sparse`. This is not related to your PR :)

torch_geometric/nn/conv/message_passing.py
```diff
@@ -197,10 +197,11 @@ def __collect__(self, args, edge_index, size, kwargs):
 
         return out
 
-    def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
+    def propagate(self, edge_index: Adj, size: Size = None, decomposed_layer = 1, **kwargs):
```
wsad1, line 200: I think the `decomposed_layer` argument should be passed to the `__init__` function in the message passing class. If its an argument to `propogate` then users will have to modify the call to `propogate` in the `forward` of conv layers like `GCNConv` to use feature decomposition. But if its an argument to `__init__` of message passing then the feature decomposition can be used as `GCNConv(16, 32, decomposed_layer =2)`.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -224,6 +225,12 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
                 and assumed to be quadratic.
                 This argument is ignored in case :obj:`edge_index` is a
                 :obj:`torch_sparse.SparseTensor`. (default: :obj:`None`)
+            decomposed_layer (integer, optional): The number of feature decomposition layers,
+                the default value is 1 (that is, the feature decomposition method is not used).
+                For the gather-scatter mode calculated on the CPU side, the feature decomposition
+                method can effectively reduce the peak memory usage and increase the calculation speed.
+                When the average node degree is higher, it is more effective to use this method
+                (for example, use Reddit dataset, achieves 2x - 3x speedup). More details in this paper: https://arxiv.org/abs/2104.03058 .
```
wsad1, line 233: Correct me if I am wrong here, but it looks like decomposition won't work on layers like [TransformerConv](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/transformer_conv.html#TransformerConv) or [AGNNConv](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/agnn_conv.html#AGNNConv), as the attention is computed based on `x_i` and `x_j` in `message` and decomposition sends a decomposed `x_i` and `x_j`. It'll be good to mention here or as a note the scenarios where decomposition won't work.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -258,54 +265,68 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
...
-            # message passing computation scheme.
-            if self.__explain__:
-                edge_mask = self.__edge_mask__.sigmoid()
-                # Some ops add self-loops to `edge_index`. We need to do the
-                # same for `edge_mask` (but do not train those).
-                if out.size(self.node_dim) != edge_mask.size(0):
-                    edge_mask = edge_mask[self.__loop_mask__]
-                    loop = edge_mask.new_ones(size[0])
-                    edge_mask = torch.cat([edge_mask, loop], dim=0)
-                assert out.size(self.node_dim) == edge_mask.size(0)
```
wsad1, line 286: edge_mask needs to computed only once, when `i=0`.

rusty1s: Yes. It's probably a good idea to set `decomposed_layers=1` in case self.__explain__=True`.

torch_geometric/nn/conv/message_passing.py
ZhouAo-ZA: Yes，you are right. My method is not effective for models ( Like: GAT, TransformerConv.....) based on attention mechanisms. My method mainly improves basic message passing models like GCN, GraphSage, GIN, etc.

ZhouAo-ZA: Yes，you are right. My method is not effective for models based on attention mechanism (such as GAT, TransformerConv ). My method mainly improves basic messaging models like GCN, GraphSage, GIN, etc.

torch_geometric/nn/conv/message_passing.py
ZhouAo-ZA: @rusty1s @wsad1 I think your idea is very good, but I don’t know how to do it. Could you please help me modify it, or tell me how to do it.

wsad1: So you could add these lines in the `__init__`
```
if self.__explain__=True:
    self.decomposed_layers=1
```
And mention in the init docstring that this happens.

ZhouAo-ZA: @wsad1 Thank you for your help, I have modified it, do I resubmit the PR?

wsad1: You could make the modification to your forked [branch](https://github.com/ZhouAo-ZA/pytorch_geometric) and they will show up on this PR. I don't think we need to create a new PR for this.

ZhouAo-ZA: @wsad1 I have committed the code, thanks for the help.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -48,6 +48,19 @@ class MessagePassing(torch.nn.Module):
...
         node_dim (int, optional): The axis along which to propagate.
             (default: :obj:`-2`)
+        decomposed_layer (integer, optional): The number of feature decomposition layers.
+            (default:: int: 1). The feature decomposition method reduces the peak memory
+            usage by slicing the feature dimensions into separated feature decomposition
+            layers during GNN Aggregation. This method is disabled by default.This method
+            works well with accelerating GNN executions on CPU platforms, and can obtain
+            better results when operating on denser graphs (e.g. 2-3x speedup on Reddit dataset).
+            It can provide acceleration for common GNN models such as GCN, GIN, GraphSAGE, etc.
+            However, this method is not applicable to GNNs with attention mechanism. The selection
```
wsad1, line 58: Correct me if I am wrong, i think decomposition will work for attention based GNNs like [GATConv](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.GATConv) where parts of the attention are computed outside `message`. So to to be specific here we could say it's not applicable to GNNs where things like attention are computed based on `x_i` or `x_j` in the `message` function.

ZhouAo-ZA: I'm very sorry, my previous statement is indeed wrong, thank you for pointing out, I have revised it.

ZhouAo-ZA:
@rusty1s @wsad1 Do I need to submit additional supplement materials or make any further modifications?

wsad1:
@ZhouAo-ZA , thanks for the great work till now. It'll be good to add some tests for this feature to existing tests in pytorch_geometric/test/nn/conv/. For example, you could add some lines to test_gcn_conv.py.
```
    conv = GCNConv(16, 32)
    conv2 = GCNConv(16, 32, decomoposed_layers = 2 )
    out1 = conv(x, edge_index)
    assert torch.allclose(conv2(x, edge_index), out1, atol=1e-6) # output is same with and without decomposition.

    t = '(Tensor, Tensor, OptTensor) -> Tensor'
    jit = torch.jit.script(conv2.jittable(t)) # jittable works with decmoposition.
    assert jit(x, edge_index).tolist() == out1.tolist()
```

ZhouAo-ZA:
@wsad1 Thanks for your guidance, I will add some tests for this feature. Correct me if I am wrong here, but I think the test code you provided should be incorrect, different GCNConv will output different results because their initial argument is not the same. As follows:

image

rusty1s:
You can deep-copy the model and set `decomposed_layers` after-wards:
```python
conv = GCNConv(16, 32)
conv2 = copy.deepcopy(conv)
conv2.decomposed_layers = 3
```

torch_geometric/nn/conv/message_passing.py
```diff
@@ -227,6 +242,8 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
             **kwargs: Any additional data which is needed to construct and
                 aggregate messages, and to update node embeddings.
         """
+        if self.__explain__:
+            self.decomposed_layers = 1  # assuming decomposed layers is passed to init
```
wsad1, line 246:
```suggestion
        decomposed_layers = 1 if self.__explain__ else self.decomposed_layers 
```

torch_geometric/nn/conv/message_passing.py
```diff
@@ -258,54 +275,68 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
...
-
-            aggr_kwargs = self.inspector.distribute('aggregate', coll_dict)
-            for hook in self._aggregate_forward_pre_hooks.values():
-                res = hook(self, (aggr_kwargs, ))
-                if res is not None:
-                    aggr_kwargs = res[0] if isinstance(res, tuple) else res
-            out = self.aggregate(out, **aggr_kwargs)
-            for hook in self._aggregate_forward_hooks.values():
-                res = hook(self, (aggr_kwargs, ), out)
+            if self.decomposed_layers > 1:
```
wsad1, line 279:
```suggestion
            if decomposed_layers > 1:
```

torch_geometric/nn/conv/message_passing.py
```diff
@@ -258,54 +275,68 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
...
+                x = kwargs['x']
+                split_x = torch.chunk(x, self.decomposed_layers, -1)
+                decomposed_out = []
+            for i in range(0, self.decomposed_layers):
+                if self.decomposed_layers > 1:
```
wsad1, line 284:
```suggestion
                if decomposed_layers > 1:
```

torch_geometric/nn/conv/message_passing.py
```diff
@@ -258,54 +275,68 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
...
+                    if res is not None:
+                        out = res
+
+                update_kwargs = self.inspector.distribute('update', coll_dict)
+                out = self.update(out, **update_kwargs)
+
+                if self.decomposed_layers > 1:
+                    decomposed_out.append(out)
+
+            if self.decomposed_layers > 1:
```
wsad1, line 331:
```suggestion
                if decomposed_layers > 1:
                    decomposed_out.append(out)

            if decomposed_layers > 1:
```

wsad1 (commented):
`self.decomposed_layers` is set to 1 if `self.__explain__` is `True`, I think changing an instance variable might not be a good idea. Instead setting another variable `decomposed_layers` to `self.decomposed_layers` or `1` depending on `self.__explain__` and using that all over `propogate ` might be better. Let me know what you think.

torch_geometric/nn/conv/message_passing.py
ZhouAo-ZA: I agree with you very much and I have modified it.

PR #3168, opened 2021-09-17
Added HEAT

Xiaoyu006:
Added HEAT, which is an extension to GAT to handle heterogeneous graphs with many kind of nodes, edges and edge attributes.

### DIFF
--- a/torch_geometric/nn/conv/heat_pyg.py
+++ b/torch_geometric/nn/conv/heat_pyg.py
@@ -1,0 +1,187 @@
+import torch
+import torch.nn as nn
+
+def retrieve_masks(node_or_edge_type, wanted_types=[]):
+    assert set(wanted_types) < set(node_or_edge_type)
+    return [True if x in wanted_types else False for x in node_or_edge_type]
+
+def flatten_a_list_of_list(list_of_list):
+    flat_list = []
+    for l in list_of_list:
+        flat_list += l
+    return flat_list
+
+class HEATlayer(nn.Module):
+    ''' 
+        1.  type-specific transformation for nodes of different types: ('n1', 'n2').
+            transform nodes from different vector space to the same vector space.
+        2.  edges are assumed to have different types but contains the same attributes.            
+    '''
+    def __init__(self,  node_type_set = ('n1', 'n2', 'n3'), edge_type_set = ('e1', 'e2', 'e3', 'e4'),
+                        in_channels_node=4, in_channels_edge_attr=2, in_channels_edge_type=2, 
+                        edge_attr_emb_size=5, edge_type_emb_size=5, node_emb_size=5, 
+                        out_channels=6, heads=3, concat=True):
+
+        super(HEATlayer, self).__init__()
+        ## Parameters
+        self.node_type_set = node_type_set
+        self.edge_type_set = edge_type_set
+        self.in_channels_node = in_channels_node # 32
+        self.in_channels_edge_attr = in_channels_edge_attr # 2
+        self.in_channels_edge_type = in_channels_edge_type # 2
+        self.out_channels = out_channels
+        self.heads = heads
+        self.concat = concat
+        self.edge_attr_emb_size = edge_attr_emb_size
+        self.edge_type_emb_size = edge_type_emb_size
+        self.node_emb_size = node_emb_size
+
+        #### Layers ####
+        ## Embeddings 
+        self.set_node_emb()
+        self.set_edge_emb()
+
+        ## Transform the concatenated edge_nbrs feature to out_channels to update the next node feature
+        self.node_update_emb = nn.Linear(self.edge_attr_emb_size + self.node_emb_size, self.out_channels, bias=False) 
+
+        ## Attention
+        self.attention_nn = nn.Linear(self.edge_attr_emb_size + self.edge_type_emb_size + 2*self.node_emb_size, 1*self.heads, bias=False)
+        self.leaky_relu = nn.LeakyReLU(0.2)
+        self.soft_max = nn.Softmax(dim=1)
+    
+    def set_node_emb(self):
+        ''' assume that different nodes have the same dimension, but different vector space. '''
+        for node_type in self.node_type_set:
+            exec('self.{}_node_feat_emb = nn.Linear(self.in_channels_node, self.node_emb_size, bias=False)'.format(node_type))
+    
+    def set_edge_emb(self):
+        ''' assume that different edges have different types but contains the same kind of attributes. '''
+        # for edge_type in self.edge_type_set:
+        #     exec('self.{}_edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) '.format(edge_type))
+        self.edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) 
+        self.edge_attr_emb = nn.Linear(self.in_channels_edge_attr, self.edge_attr_emb_size, bias=False) 
+
+    def embed_nodes(self, node_features, node_type_list):
+        ''' embed node attributes according to their types. '''
+        emb_node_features = torch.zeros(node_features.shape[0], self.node_emb_size).to(node_features.device)
+        for node_type in self.node_type_set:
+            node_mask = retrieve_masks(node_type_list, wanted_types=[node_type])
+            exec('emb_node_features[node_mask] = self.{}_node_feat_emb(node_features[node_mask])'.format(node_type))
+        return emb_node_features
+    
+    def embed_edges(self, edge_attrs, edge_types):
+        ''' embed edge attributes and edge types respectively. '''
+        emb_edge_attributes = self.leaky_relu(self.edge_attr_emb(edge_attrs))
+        emb_edge_types = self.leaky_relu(self.edge_type_emb(edge_types))
+        return emb_edge_attributes, emb_edge_types
+    
+    def forward(self, node_feats, edge_index, edge_attr, edge_type, node_type_list, edge_type_list):
+        """
+        Args:
+            node_f:[num_node, in_channels_node]
+            edge_index: [2, number_edge]
+            edge_attr: [number_edge, len_edge_attr]
+            edge_type: [number_edge, len_edge_type]
+            node_type_list: [ type of each node ]
+            edge_type_list: [ type of each edge ]
+        """
+        # Node-type specific transformation
+        node_emb = self.embed_nodes(node_feats, node_type_list)
+
+        # Edge attribute and type transformation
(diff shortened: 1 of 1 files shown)


Xiaoyu006:
Xiaoyu‘s First Pull Request!

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+    assert set(wanted_types) < set(node_or_edge_type)
+    return [True if x in wanted_types else False for x in node_or_edge_type]
+
+def flatten_a_list_of_list(list_of_list):
+    flat_list = []
+    for l in list_of_list:
+        flat_list += l
+    return flat_list
+
+class HEATlayer(nn.Module):
```
rusty1s, line 14:
```suggestion
class HEATConv(nn.Module):
```

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+        1.  type-specific transformation for nodes of different types: ('n1', 'n2').
+            transform nodes from different vector space to the same vector space.
+        2.  edges are assumed to have different types but contains the same attributes.            
+    '''
+    def __init__(self,  node_type_set = ('n1', 'n2', 'n3'), edge_type_set = ('e1', 'e2', 'e3', 'e4'),
+                        in_channels_node=4, in_channels_edge_attr=2, in_channels_edge_type=2, 
+                        edge_attr_emb_size=5, edge_type_emb_size=5, node_emb_size=5, 
+                        out_channels=6, heads=3, concat=True):
+
+        super(HEATlayer, self).__init__()
```
rusty1s, line 25:
```suggestion
        super().__init__()
```

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+            exec('self.{}_node_feat_emb = nn.Linear(self.in_channels_node, self.node_emb_size, bias=False)'.format(node_type))
+    
+    def set_edge_emb(self):
+        ''' assume that different edges have different types but contains the same kind of attributes. '''
+        # for edge_type in self.edge_type_set:
+        #     exec('self.{}_edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) '.format(edge_type))
+        self.edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) 
+        self.edge_attr_emb = nn.Linear(self.in_channels_edge_attr, self.edge_attr_emb_size, bias=False) 
+
+    def embed_nodes(self, node_features, node_type_list):
```
rusty1s, line 64: Can we make this a general module outside of `HEATConv`, *.e.g*, `HeteroLinear(torch.nn.Module)`? This should be helpful for all kinds of heterogeneous graph learning.

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+            exec('emb_node_features[node_mask] = self.{}_node_feat_emb(node_features[node_mask])'.format(node_type))
+        return emb_node_features
+    
+    def embed_edges(self, edge_attrs, edge_types):
+        ''' embed edge attributes and edge types respectively. '''
+        emb_edge_attributes = self.leaky_relu(self.edge_attr_emb(edge_attrs))
+        emb_edge_types = self.leaky_relu(self.edge_type_emb(edge_types))
+        return emb_edge_attributes, emb_edge_types
+    
+    def forward(self, node_feats, edge_index, edge_attr, edge_type, node_type_list, edge_type_list):
```
rusty1s, line 78:
```suggestion
    def forward(self, x, edge_index, edge_attr, node_type, edge_type):
```

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+        self.node_update_emb = nn.Linear(self.edge_attr_emb_size + self.node_emb_size, self.out_channels, bias=False) 
+
+        ## Attention
+        self.attention_nn = nn.Linear(self.edge_attr_emb_size + self.edge_type_emb_size + 2*self.node_emb_size, 1*self.heads, bias=False)
+        self.leaky_relu = nn.LeakyReLU(0.2)
+        self.soft_max = nn.Softmax(dim=1)
+    
+    def set_node_emb(self):
+        ''' assume that different nodes have the same dimension, but different vector space. '''
+        for node_type in self.node_type_set:
```
rusty1s, line 54: Use a `ModuleList` to store `Linear` layers.

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+            exec('self.{}_node_feat_emb = nn.Linear(self.in_channels_node, self.node_emb_size, bias=False)'.format(node_type))
+    
+    def set_edge_emb(self):
+        ''' assume that different edges have different types but contains the same kind of attributes. '''
+        # for edge_type in self.edge_type_set:
+        #     exec('self.{}_edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) '.format(edge_type))
+        self.edge_type_emb = nn.Linear(self.in_channels_edge_type, self.edge_type_emb_size, bias=False) 
+        self.edge_attr_emb = nn.Linear(self.in_channels_edge_attr, self.edge_attr_emb_size, bias=False)
```
rusty1s, line 62: Since `edge_type` is a PyTorch vector of dtype `torch.long`, we can embed edge types via `torch.nn.Embedding` rather than via `Linear`.

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+        # Node-type specific transformation
+        node_emb = self.embed_nodes(node_feats, node_type_list)
+
+        # Edge attribute and type transformation
+        edge_attr_emb, edge_type_emb = self.embed_edges(edge_attr, edge_type)
+
+        # Edge-enhanced masked attention
+        emb_edge_feat = torch.cat((edge_attr_emb, edge_type_emb), dim=1) # e_{i,j} in heat paper
+
+        src_node_emb = node_emb[edge_index[0]]
```
rusty1s, line 97: We should make use of the `MessagePassing` interface to implement this, i.e. in `self.propagate`.

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+        flat_list += l
+    return flat_list
+
+class HEATlayer(nn.Module):
+    ''' 
+        1.  type-specific transformation for nodes of different types: ('n1', 'n2').
+            transform nodes from different vector space to the same vector space.
+        2.  edges are assumed to have different types but contains the same attributes.            
+    '''
+    def __init__(self,  node_type_set = ('n1', 'n2', 'n3'), edge_type_set = ('e1', 'e2', 'e3', 'e4'),
```
rusty1s, line 20: I don't think we need the exact names of node/edge types here. `num_node_types` and `num_edge_types` seems to be sufficient.

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+    data_1 = Data(  x=node_features, edge_index=edge_index, 
+                    node_type_list=node_type_list, edge_type_list=edge_type_list,
+                    edge_type=edge_type, edge_attr=edge_attr )
+                    
+    node_features = torch.randn((3, 4))
+    
+
+    node_type_list = ['n1', 'n3', 'n2']
+    edge_type_list = ['e1', 'e3', 'e4', 'e2', 'e3']
+    edge_type = torch.randn((5, 2))
```
rusty1s, line 152: Why is `edge_type` a feature matrix?

torch_geometric/nn/conv/heat_pyg.py
```diff
@@ -0,0 +1,187 @@
...
+    return flat_list
+
+class HEATlayer(nn.Module):
+    ''' 
+        1.  type-specific transformation for nodes of different types: ('n1', 'n2').
+            transform nodes from different vector space to the same vector space.
+        2.  edges are assumed to have different types but contains the same attributes.            
+    '''
+    def __init__(self,  node_type_set = ('n1', 'n2', 'n3'), edge_type_set = ('e1', 'e2', 'e3', 'e4'),
+                        in_channels_node=4, in_channels_edge_attr=2, in_channels_edge_type=2,
```
rusty1s, line 21:
```suggestion
                        in_channels=4, edge_dim=2, 
```

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: In this `embed_nodes` function, we are embedding node features according to their types. so in this function, we need the predefined embedding functions (which you suggest implementing with ModuleList) that should be predefined for each task to initialize HEATConv. Do you mean we pass the ModuleList as an argument to the HeteroLinear (embed_nodes) function?

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: Yes! Thank you for this comment, I revised the code accordingly, and it works! thank you!

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: Thank you, ModuleList works very well!

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: Thank you! Changed accordingly.

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: Thank you! nn.Embedding works well!

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: edge_type does not need to be a feature matrix with floats. I changed edge_type to `edge_type = torch.tensor([[0, 2, 1, 1, 2]], dtype=torch.long)` something like this, indicating 5 edges falls into 3 types (0,1,2).

torch_geometric/nn/conv/heat_pyg.py
Xiaoyu006: Thank you! I changed accordingly.

Xiaoyu006:
Hi, @rusty1s, Thank you very much for your suggestions. I addressed all. Please check.

PR #3174, opened 2021-09-18
construct layer config to pass parameters from cfg to layers.

yaoyaowd:
This PR introduces intermediate layer config to pass parameters from cfg to layers.

Next step:
1. I will refactor general conv and combine generalconv.py and general_conv.py into one layer.
2. I will refactor some other configurations for graph gym so that configurations are more modularized.

### DIFF
--- a/torch_geometric/graphgym/models/layer.py
+++ b/torch_geometric/graphgym/models/layer.py
@@ -1,14 +1,62 @@
+import copy
+from dataclasses import dataclass
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 import torch_geometric as pyg
 
-from torch_geometric.graphgym.config import cfg
 from torch_geometric.graphgym.models.act import act_dict
 from torch_geometric.graphgym.contrib.layer.generalconv import (
     GeneralConvLayer, GeneralEdgeConvLayer)
 
 import torch_geometric.graphgym.register as register
+
+
+@dataclass
+class LayerConfig:
+    # batchnorm parameters.
+    has_batchnorm: bool = False
+    bn_eps: float = 1e-5
+    bn_mom: float = 0.1
+
+    # mem parameters.
+    mem_inplace: bool = False
+
+    # gnn parameters
+    dim_in: int = -1
+    dim_out: int = -1
+    edge_dim: int = -1
+    has_l2norm: bool = True
+    dropout: float = 0.0
+    has_act: bool = True
+    final_act: bool = True
+    act: str = 'relu'
+    has_bias: bool = True
+    keep_edge: float = 0.5
+    dim_inner: int = None
+    num_layers: int = 2
+
+
+def new_layer_config(dim_in, dim_out, num_layers, has_act, has_bias, cfg):
+    return LayerConfig(
+        has_batchnorm=cfg.gnn.batchnorm,
+        bn_eps=cfg.bn.eps,
+        bn_mom=cfg.bn.mom,
+        mem_inplace=cfg.mem.inplace,
+        dim_in=dim_in,
+        dim_out=dim_out,
+        edge_dim=cfg.dataset.edge_dim,
+        has_l2norm=cfg.gnn.l2norm,
+        dropout=cfg.gnn.dropout,
+        has_act=has_act,
+        final_act=True,
+        act=cfg.gnn.act,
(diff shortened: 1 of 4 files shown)


rusty1s:
Thanks @yaoyaowd. This now makes `GeneralConv` be a lazy operator by design, which might break GraphGym at the moment. Instead, I feel that we should make `edge_dim` an argument of the `GeneralEdgeConvLayer`.

yaoyaowd:
I tested with example bash and they worked fine. Currently there's a sequence of steps to create the GeneralEdgeConv:
`main->model_builder->gnn->layer->generalconv`. They all access the same cfg file and I think it is better to reduce the parameter we need from cfg and then refactor each piece slowly.

rusty1s:
I absolutely agree with you. I'm just of the opinion that lazy ops currently do not work in GraphGym, as the optimizer is initialized before the very first forward pass. This does result in an error, but it certainly leads to lazy `Linear` layers being untrained.

yaoyaowd:
I did a bigger refactor so that there's a intermediate layer_config between gnn and layers.

torch_geometric/graphgym/models/layer.py
```diff
@@ -1,16 +1,67 @@
...
 
-from torch_geometric.graphgym.config import cfg
 from torch_geometric.graphgym.models.act import act_dict
 from torch_geometric.graphgym.contrib.layer.generalconv import (
     GeneralConvLayer, GeneralEdgeConvLayer)
 
 import torch_geometric.graphgym.register as register
 
 
+@dataclass
```
rusty1s, line 16: Is there a difference between `@dataclass` and simply using `NamedTuple`s?

rusty1s:
This looks better IMO. Thank you. Let's see what @JiaxuanYou thinks.

torch_geometric/graphgym/models/layer.py
```diff
@@ -156,18 +223,21 @@ class MLP(nn.Module):
...
+        dim_inner = layer_config.dim_in \
+            if layer_config.dim_inner is None \
+            else layer_config.dim_inner
+        layer_config.has_bias = True
         layers = []
-        if num_layers > 1:
+        if layer_config.num_layers > 1:
             layers.append(
-                GeneralMultiLayer('linear', num_layers - 1, dim_in, dim_inner,
+                GeneralMultiLayer('linear', layer_config.num_layers - 1,
```
JiaxuanYou, line 235: E.g., user may not know there is a `layer_config.num_layers` in `layer_config`

torch_geometric/graphgym/models/layer.py
```diff
@@ -156,18 +223,21 @@ class MLP(nn.Module):
         num_layers (int): Number of layers in the stack
         **kwargs (optional): Additional args
     """
-    def __init__(self, dim_in, dim_out, bias=True, dim_inner=None,
-                 num_layers=2, **kwargs):
+    def __init__(self, layer_config: LayerConfig, **kwargs):
```
JiaxuanYou, line 226: E.g., has_batchnorm does not apply here, but a user may not know that, and still set `has_batchnorm` and expect it to work.

torch_geometric/graphgym/models/layer.py
yaoyaowd: Data class is a class to store data with attribute while named tuple is a dictionary under the hood.
https://www.geeksforgeeks.org/difference-between-dataclass-vs-namedtuple-vs-object-in-python/

yaoyaowd:
The main thing I want to address is to modularize parameters for different function areas, i.e dataset accesses parameters from data_config, model accesses parameters from model_config, layer accesses parameters from layer_config.

I think issues you described above are both hard problems for a general LayerConfig and the cfg. Where users need to be familiar with cfg and parameters in cfg may not impact the model if it is not config correctly. But I want to make a step forward so that we know what parameters in the cfg goes into layers and then we separate the problem into two smaller problems that can be solved in parallel.
- continue refactor cfg into individual module config.
- refactor layer config to reflect different config for different layers.

JiaxuanYou:
Thanks for explaining. This makes sense to me, and I agree with the whole picture of improving the structure of cfg.

yaoyaowd:
Thanks, I will assume it is a LGTM and merge the code :)

PR #3177, opened 2021-09-19
GNNExplainer support for multiple mask types, regression

gregorkrz:
This updates the `GNNExplainer` model with the following:
* Support for  `return_type='regression'`. The error term of loss is then calculated as Euclidean distance between the original and current prediction.
* Support for different types of node mask (node-level, feature-level and feature-level for each individual node)
* Some subgraph visualization parameters (`node_alpha`, `edge_y` and `plot_random_state` for reproducible node placement)

### DIFF
--- a/torch_geometric/nn/models/gnn_explainer.py
+++ b/torch_geometric/nn/models/gnn_explainer.py
@@ -51,13 +51,13 @@ class GNNExplainer(torch.nn.Module):
             to use individual feature-level mask for each node, and
             :obj:`'scalar'` to use a scalar (node-level) mask for each
             individual node. (default: :obj:`'feature'`)
-        allow_edge_mask (boolean, optional): if False, edge mask will not
+        allow_edge_mask (boolean, optional): if set to :obj:`False`, edge mask will not
             be optimized.
             (default: :obj:`True`)
-        coeffs (dict, optional): Dictionary of hyperparameters with which
-            to update the coefficients dictionary
         log (bool, optional): If set to :obj:`False`, will not log any learning
             progress. (default: :obj:`True`)
+        **kwargs (dict, optional): Additional hyperparameters with which
+            to update the coefficients dictionary
     """
 
     coeffs = {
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/models/gnn_explainer.py
```diff
@@ -125,13 +155,21 @@ def __subgraph__(self, node_idx, x, edge_index, **kwargs):
...
+        return x, edge_index, mapping, edge_mask, subset, kwargs
 
     def __loss__(self, node_idx, log_logits, pred_label):
         # node_idx is -1 for explaining graphs
-        loss = -log_logits[
-            node_idx, pred_label[node_idx]] if node_idx != -1 else -log_logits[
-                0, pred_label[0]]
+        if self.return_type == 'regression':
+            if node_idx != -1:
+                loss = torch.cdist(
```
rusty1s, line 164: Does this work better in practice compared to something like `MSELoss`?

rusty1s (approved):
This looks super good. I like it. I only have minor comments.

torch_geometric/nn/models/gnn_explainer.py
gregorkrz: In terms of performance, differences between both seem to be insignificant. But yes, if that looks better, we can also use something like
```
self.mse_loss = torch.nn.MSELoss(reduction='sum')
...
loss = torch.sqrt(self.mse_loss(log_logits[node_idx], pred_label[node_idx]))
```
What do you think?

torch_geometric/nn/models/gnn_explainer.py
```diff
@@ -57,24 +71,40 @@ class GNNExplainer(torch.nn.Module):
...
+            self.node_feat_mask = torch.nn.Parameter(torch.randn(N) * std)
+        else:
+            self.node_feat_mask = torch.nn.Parameter(torch.randn(F) * std)
 
         std = torch.nn.init.calculate_gain('relu') * sqrt(2.0 / (2 * N))
-        self.edge_mask = torch.nn.Parameter(torch.randn(E) * std)
+        if self.allow_edge_mask:
+            self.edge_mask = torch.nn.Parameter(torch.randn(E) * std)
+        else:
+            # here edge_mask is fixed; sigmoid will always return 1
```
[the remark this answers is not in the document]
gregorkrz: I think `edge_mask` needs to be a parameter, as the `MessagePassing` module only allows that:
`TypeError: cannot assign 'torch.FloatTensor' as parameter '__edge_mask__' (torch.nn.Parameter or None expected)`

rusty1s:
Super. Thanks a lot!

PR #3184, opened 2021-09-21
add hetero unsupervised graph representation learning method

Yonggie:
I wanted to just add `example/hetero/hetero_unsupervised_dblp.py` as an `hetero graph unsupervised method example` and leave everything else unchanged, but unfortunately it seems git detected many changes though I dont know where exactly...

So it's basically applied DGI method in heterogeneous graph.
The code might not be so neat, but it's the right framework.


rusty1s:
Thank you. I'm not sure what's the issue with GitHub is, but this needs to be resolved. Can you try to checkout a clean PyG version, apply your changes there, and try to submit a new PR?

PR #3189, opened 2021-09-21
Unsupervised Heterogeneous Graph Learning

Yonggie:
added unsupervised heterogeneous graph representation learning example.


examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+        return self.activation(self.conv(x, edge_index))
+
+
+class HeteroUnsupervised(torch.nn.Module):
+    def __init__(self, metadata, hidden_channels, out_channels, num_layers):
+        super().__init__()
+
+        # for discriminator
+        self.embed_size = out_channels
+        self.weight = Parameter(torch.Tensor(out_channels, out_channels))
```
rusty1s, line 45: `self.weight` is uninitialized.

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+
+        self.convs = torch.nn.ModuleList()
+        for _ in range(num_layers):
+            conv = HeteroConv({
+                edge_type: SAGEConv((-1, -1), hidden_channels)
+                for edge_type in metadata[1]
+            })
+            self.convs.append(conv)
+
+    def forward(self, xx_dict, edge_index_dict):
```
rusty1s, line 58:
```suggestion
    def forward(self, x_dict, edge_index_dict):
```

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+        x_dict = {}
+        for conv in self.convs:
+            x_dict = conv(xx_dict, edge_index_dict)
```
rusty1s, line 61:
```suggestion
            x_dict = conv(x_dict, edge_index_dict)
```
```suggestion
            x_dict = conv(xx_dict, edge_index_dict)
```

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+
+        neg_embed = self.encoder(corrupted_feat, target_edge_index)
+
+        summary = torch.mean(pos_embed, dim=0)
+
+        return pos_embed, neg_embed, summary
+
+    def convert_to_meta_path_edge_index(self, edge_index_a, edge_index_b):
+        paper2author_dict = \
+            {idx_a.item(): idx_b.item() for idx_a, idx_b in edge_index_b.T}
```
rusty1s, line 87: It seems like you are only holding one neighbor, and dropping all the remaining ones. Is this intended?

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+        target_edge_index = []
+        for idx_a, idx_b in edge_index_a.T:
```
rusty1s, line 89: Any chance we can omit the for-loop here?

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,187 @@
...
+
+@torch.no_grad()
+def test():
+    model.eval()
+    pos_embed, _, _ = model(data.x_dict, data.edge_index_dict)
+    mask = data['author']['val_mask']
+    labels = data['author'].y[mask]
+
+    # I dont get it why it dosen't work
+    valid_embed = pos_embed[mask]
```
rusty1s, line 160: What does not work?

examples/hetero/hetero_unsupervised_dblp.py
Yonggie: It is intended.
I just added one meta path `author-paper-author(APA)`, we can also add `author-paper-conference-paper-author(APCPA)`, etc. It requires more codes though.
This meta path method is actually from DMGI from AAAI2020, https://www.sciencedirect.com/science/article/abs/pii/S0950705120302240 . 

But it seems a little bit untidy as it requires pretty much preprocessing on data edge index, and I'm doing a simpler method paper method, but as mentioned before it's not finished yet.

examples/hetero/hetero_unsupervised_dblp.py
Yonggie: Yes. `paper2author_dict = dict(edge_index_b.T.cpu().numpy())`  and add `.to(device)` in the `return` line.

examples/hetero/hetero_unsupervised_dblp.py
Yonggie: Please ignore this annotation, I forgot to delete it lol....

examples/hetero/hetero_unsupervised_dblp.py
rusty1s: In case this is intended, you can create a `[num_authors, 2]` matrix and simply scatter paper indices pointing to a particular source/destination author nodes, e.g., something like:
```python
out = torch.empty((num_authors, 2), dtype=torch.long)

out[0].scatter_(edge_index_a[0], edge_index_a[1])
out[1].scatter_(edge_index_b[1], edge_index_b[0])
```

examples/hetero/hetero_unsupervised_dblp.py
rusty1s: Ok, just let me know once you feel this PR is ready for review :)

Yonggie:
@rusty1s  Hi, it's ready for review :)

examples/hetero/hetero_unsupervised_dblp.py
Yonggie: I particularly keep the code `xx_dict` because if I wrote `x_dict`, everything inside `x_dict` would change including `feature.shape`, and that is not what it ought to be.

benbusath: Isn't x_dict supposed to change after each convolution layer? Each convolution layer works on the output of the last convolution layer? In your current setup each layer works on the same unchanging input xx_dict

vincentvic:
Hi!
Is this code is working ? because I'am trying to train an unsupervised Heterogeneous Graph Learning to find a proximity between two entites in a graph for example (item and users) based on an embeddings with a cosinus distance.
Thank you for your help.
Victor

examples/hetero/hetero_unsupervised_dblp.py
```diff
@@ -0,0 +1,231 @@
+import math
+import os.path as osp
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+from sklearn.linear_model import LogisticRegression
+from sklearn.model_selection import train_test_split
+from torch.nn import Parameter
+
```
vincentvic, line 10:
```suggestion
import torch_geometric.transforms as T
```

vincentvic (commented):
Use the class AddMetaPaths for a cleaner implementation

Yonggie:
It's been quite a long time since I wrote this code. I may take a while to catch it up again.

Yonggie:
I implemented DMGI using torch_geometric, but it seems the performance is not as good as the paper said. Original implementation: https://github.com/Yonggie/TCM_DMGI/blob/master/models/DMGI.py

rusty1s:
Thanks for the updates @Yonggie, and sorry for the delay. We will try to get this merged in the upcoming week.

rusty1s:
Do you know of any differences between this implementation and the official one?

Yonggie:
I just found a small difference in the encoder and changed the code. but it still seems not good enough.  And another weird thing is valid score and test score are exactly the same.

rusty1s (approved):
Thank you :)

PR #3199, opened 2021-09-22
Add PointTransformerConv

QuanticDisaster:
Implementation of Point Transformer from Hengshuang Zhao and all (2020)  https://arxiv.org/abs/2012.09164 detailled in the 3.1, 3.2 and 3.3 sections and features an attention layer.

I did not manage to implement the test, I could not import the new convolution from there and failed to use jit due to errors and my incomprehension of how to use it. I am unsure about the names of the sub_layers and variables if we should call them by the greek letters used in the paper or by what they represent (position encoding, attention layers,...).


torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -0,0 +1,110 @@
...
+              self).__init__(aggr='add')  # "Add" aggregation
+
+        self.delta = Sequential(Linear(3, pos_mlp_channels), ReLU(),
+                                Linear(pos_mlp_channels, out_channels))
+        self.gamma = Sequential(
+            Linear(out_channels, out_channels * attn_mlp_channels), ReLU(),
+            Linear(out_channels * attn_mlp_channels, out_channels))
+        self.alpha = Linear(in_channels, out_channels)
+        self.phi = Linear(in_channels, out_channels)
+        self.psi = Linear(in_channels, out_channels)
```
wsad1, line 63: Do you think we could have an API similar to [PointNetCov](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.PointNetConv). So `gamma` and `delta` would become `gloabl_nn` and `local_nn` respectively, and they can be passed to `__init__` as `torch.nn.Module` . Their default values could probably be the current networks of `gamma` and `delta`.

torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -0,0 +1,110 @@
...
+        out_delta = self.delta((pos_j - pos_i))  # compute position encoding
+
+        out_phi_psi = self.phi(x_i) - self.psi(
+            x_j)  # compute point wise transformation
+        out_gamma = self.gamma(out_phi_psi +
+                               out_delta)  # compute importance factor
+
+        # normalize the result with softmax and multiply
+        # by features transformed by alpha
+        msg = softmax(out_gamma, 1) * (
```
wsad1, line 106: Correct me if Iam wrong. Currently the `softmax` operation is applied over all nodes. Shouldn't be applied only over neighboring nodes?
In which case, use `softmax` from `torch_geometric.utils.softmax`, like its used in [GATConv](https://github.com/pyg-team/pytorch_geometric/blob/764ceec99698120e8a667e30625825de14ee5126/torch_geometric/nn/conv/gat_conv.py#L191).

wsad1 (commented):
Thanks for all the work, the PR looks good. I have a few comments.

torch_geometric/nn/conv/point_transformer_conv.py
rusty1s: I think this observation is correct. The `softmax` also is currently applied in the feature dimension.

torch_geometric/nn/conv/point_transformer_conv.py
QuanticDisaster: I also thought about it, hesitating between putting the 5 in an API form or none at all, but doing it for gamma and delta only would be a good compromise indeed !

torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -0,0 +1,110 @@
...
+        self.add_self_loops = add_self_loops
+
+    def reset_parameters(self):
+        reset(self.delta)
+        reset(self.gamma)
+        reset(self.alpha)
+        reset(self.phi)
+        reset(self.psi)
+
+    def forward(self, x: Union[OptTensor, PairOptTensor],
```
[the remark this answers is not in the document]
QuanticDisaster: To be honest, I didn't fully grasp the architecture of functions, this is why I copy pasted this part from PointConv that I used as a reference. As rusty1 pointed out in this discussion https://github.com/pyg-team/pytorch_geometric/discussions/3169 we can possibly work without features with the `ToConstant(1)` transform trick. 
However I personaly think it would be better to make it the user's responsability however, so I agree with you

torch_geometric/nn/conv/point_transformer_conv.py
QuanticDisaster: My bad, I didn't understand correctly how x_i and x_j were related I will correct it

torch_geometric/nn/conv/point_transformer_conv.py
QuanticDisaster: @wsad1 I made it in the form of an API, however I am disatified with the current way I implemented it as "default". Indeed I need the out_channels parameters to define the nn so I can't define it in the arguments, but putting as default `None` and saying it is initialized just after bothers me. What would be the best practice here ?

torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -0,0 +1,144 @@
...
+            '  )\n'` (default: :obj:`None`)
+        global_nn : (torch.nn.Module, optional): A neural network
+            :math:`\gamma` which maps transformed node features of shape
+            :obj:`[-1, out_channels]` to shape :obj:`[-1, out_channels]`,
+            *e.g.*, defined by :class:`torch.nn.Sequential`. In case
+            :obj:`None` is given, is initialized with :obj:`Sequential(\n'
+            '  (0): Linear(in_features=32, out_features=64, bias=True)\n'
+            '  (1): ReLU()\n'
+            '  (2): Linear(in_features=64, out_features=32, bias=True)\n'
+            ')'` (default: :obj:`None`)
```
wsad1, line 55: @rusty1s , I suggested we set the default values of `local_nn` and `global_nn` to the network architecture in the paper, so users have access to that by default. Do you think thats the right way to go here, or should `local_nn` and `global_nn` be non-optional.

torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -0,0 +1,144 @@
...
+            case local_nn is not None (default: :obj:`64`)
+        attn_mlp_channels (int, optional) : factor size of intermediary
+            channels in transformed features mlp. Has no effect in
+            case global_nn is not None (default: :obj:`2`)
+        **kwargs (optional): Additional arguments of
+            :class:`torch_geometric.nn.conv.MessagePassing`.
+    """
+
+    def __init__(self,
+                 in_channels,
```
wsad1, line 69: Might have to add some code to handle the case where `in_channels` is a tuple of `int`. Refer to [tranformer_conv](https://github.com/pyg-team/pytorch_geometric/blob/b3618d2dc57ad31a420e400ee83e42a5200037e4/torch_geometric/nn/conv/transformer_conv.py#L114).

wsad1 (commented):
@QuanticDisaster, Thanks for all the updates. Here are my last set of comments :).

torch_geometric/nn/conv/point_transformer_conv.py
QuanticDisaster: Added. However I do not really understand how it can be used so I hope I did it right

torch_geometric/nn/conv/point_transformer_conv.py
QuanticDisaster: @wsad1 @rusty1s 
The main constraint is that local_nn maps from dim `[-1,3]` to `[-1, out_channels]` so it **must** be initialized. 
It can't be put as default argument as it depends on another argument. I had a look at Linear from torch_geometric which allows lazy initialization, however it can only do so with input channel.

I added a commit "no default nns" to see how it could be done if we leave local_nn and global_nn to the user. While it makes the code shorter so simpler, it leaves out the default network architectures of the paper. PointConv does it that way but it has an example for himself in this repo. But to me a user expects the default values to be at least acceptable (which is not the case if we leave the nns to the user charge), and should not have to look into the paper to define the corresponding `local_nn ` and `global_nn `.

With accessibility in mind, I think the best way to go is the first way I did it and as you say, with the networks being non optional. `(in_channels, out_channels, mlp_pos, attention_dim)`  would be the arguments and `local_nn` and `global_nn` would be defined in the function. An experienced user will still have the possibility to change it manually by redefining the attribute. Although it is ugly, I think it will only concern a minority of nitpicking users while making it more accessible to the majority.

What are your thoughts on it ?

wsad1 (approved):
Thanks a lot @QuanticDisaster. Barring some minor tweaks this PR looks good to merge. Maybe @rusty1s  can chime in on whether the MLPs should have default values or not.

rusty1s (approved):

rusty1s:
I like the default values of the NNs. This is something we should add to `PointNetConv` and `GINConv` as well, I think.

This is a great PR. Thank you so much @QuanticDisaster and @wsad1.

PR #3203, opened 2021-09-22
Fix initialization of GeneralMultiLayer for graphgym MLP layer

fjulian:
In graphgym MLP layer, a `GeneralMultiLayer` was not initialized correctly, using a `LayerConfig` object. This also led the `run_batch.sh` script to fail.

This PR fixes this. `run_batch.sh` runs successfully now.


yaoyaowd (approved):
Thanks

Issue #3061, opened 2021-09-01
Use `pathlib` instead of `os.path`


sumanthratna:
(very minor suggestion)

## 🚀 Feature

instead of using `os.path`, the high-level `pathlib` module can be used

## Motivation

I prefer `path / f'data_{ind}.pt'` over `os.path.join(path, f'data_{ind}.pt')`. 

Using pathlib internally would make it easier for users who subclass `torch_geometric.data.Dataset`, since relevant file path variables would be `pathlib.Path` instead of `str`. In other words, this would prevent users from needing to manually call `pathlib.Path(self.processed_dir)`

## Additional context

pytorch-geometric requires Python >=3.6
https://github.com/rusty1s/pytorch_geometric/blob/e425622d6efc6832b15e9fe577710a7119d76cef/setup.py#L35

and pathlib is available in Python 3.6, so this change can be made safely

rusty1s:
Thanks for this great suggestion. It seems that `pathlib` is indeed the preferred method when working with paths in Python. If you are interested, I'm happy to welcome a PR on this one.

PR #3207, opened 2021-09-23
pathlib.Path instead of os.path

rietesh:
As discussed in this issue [#3061](https://github.com/pyg-team/pytorch_geometric/issues/3061#issue-985437172) This PR implements the pathlib.Path keeping in mind the changes to pathlib library in python 3.8. The changes should support python 3.6 and python 3.9 for both torch 1.8.* and 1.9.*. The test and benchmark modules still use os.path which can be changed if required.


rusty1s:
Thanks a lot. I think a lot of files were re-formatted which makes reviewing this PR quite complicated. Is there any possibility to revert the format changes?

rietesh:
Thank you for the comment, the contribution guidelines had outlined the formatting requirements nevertheless I can revert the formatting commits I will push the changes as soon as possible.

rietesh:
I have reverted the Code format changes I hope it is now easier to review changes

rusty1s:
Thank you!

examples/cora.py
```diff
@@ -11,7 +11,7 @@
     T.RandomNodeSplit(num_val=500, num_test=500),
     T.TargetIndegree(),
 ])
-path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', dataset)
+path = Path.joinpath(Path(__file__).resolve().parent, '..', 'data', dataset)
```
cthoyt, line 14: why not replace the `..` with a second parent?

examples/dgcnn_classification.py
```diff
@@ -10,7 +10,7 @@
 
 from pointnet2_classification import MLP
 
-path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data/ModelNet10')
+path = Path.joinpath(Path(__file__).resolve().parent, '..', 'data/ModelNet10')
```
cthoyt, line 13: why not split ` 'data/ModelNet10'`

examples/cora.py
rietesh: Yes its possible `path = Path.joinpath(Path(__file__).resolve().parent.parent, 'data', dataset)` would give the same result

examples/dgcnn_classification.py
rietesh: Yes that would be easy to read

sumanthratna:
appreciate this PR @rietesh! Just wanted to bump this and see if there were any issues finalizing this PR?

rusty1s:
Yeah, this PR got a little bit under the radar. We would need to resolve conversations and conflicts, and ensure that CI runs through.

rietesh:
Raised a new PR https://github.com/pyg-team/pytorch_geometric/pull/4220

PR #3211, opened 2021-09-23
Fix memory updater in tgn.py.

cwh104504:
The behavior of `scatter` is non-deterministic when indices are not unique ([docs](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html)), resulting in `last_update` to be replaced with random timestamp. Modified `scatter` to `scatter_max` in torch_scatter.


rusty1s:
Thanks!

PR #3218, opened 2021-09-23
Improve negative sampling

WuliangHuang:
This PR does three things.
1. Allow `structured_negative_sampling` does not sample negative self-loops. 
2. As reject sampling will loop forever or raise an error if there are no possible negative edges. We add a function `is_neg_samp_feasible` for judging the feasibility of sampling. Its application utilizes `assert` which will only work in `debug` mode. We also make this function suitable for bipartite graphs and `force_undirected = True`.
3. Add related test unit cases.


WuliangHuang:
Please review in (#3219 ), I'm sorry for the confusion commit messages here.

PR #3219, opened 2021-09-23
Imporve negative sampling

WuliangHuang:
This PR does three things:
1. Allow `structured_negative_sampling` not to sample negative self-loops. 
2. As reject sampling will loop forever or raise an error if there are no possible negative edges. We add a function `is_neg_samp_feasible` for judging the feasibility of sampling. Its application utilizes `assert` which will only work in `debug` mode. We also make this function suitable for bipartite graphs and `force_undirected = True`.
3. Add related test unit cases.

### DIFF
--- a/torch_geometric/utils/negative_sampling.py
+++ b/torch_geometric/utils/negative_sampling.py
@@ -1,11 +1,12 @@
 from typing import Optional, Union, Tuple
 
 import random
+import warnings
 
 import torch
 import numpy as np
 from torch import Tensor
-from torch_geometric.utils import degree
+from torch_geometric.utils import degree, remove_self_loops, to_undirected
 
 from .num_nodes import maybe_num_nodes
 
@@ -46,6 +47,9 @@ def negative_sampling(edge_index: Tensor,
     size = maybe_num_nodes(edge_index) if size is None else size
     size = (size, size) if not bipartite else size
     force_undirected = False if bipartite else force_undirected
+
+    assert is_neg_samp_feasible(edge_index, 'common', num_nodes, bipartite,
+                                False, force_undirected)
 
     idx, population = edge_index_to_vector(edge_index, size, bipartite,
                                            force_undirected)
@@ -90,7 +94,8 @@ def negative_sampling(edge_index: Tensor,
     return vector_to_edge_index(neg_idx, size, bipartite, force_undirected)
 
 
-def structured_negative_sampling(edge_index, num_nodes=None):
+def structured_negative_sampling(edge_index, num_nodes=None,
+                                 contains_neg_self_loops=True):
     r"""Samples a negative edge :obj:`(i,k)` for every positive edge
     :obj:`(i,j)` in the graph given by :attr:`edge_index`, and returns it as a
     tuple of the form :obj:`(i,j,k)`.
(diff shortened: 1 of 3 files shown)


rusty1s:
Thank you. It seems like you have submitted two PRs with the same content. Shall we close one of them?

WuliangHuang:
Yeah, I closed the previous (#3218 ).

torch_geometric/utils/negative_sampling.py
```diff
@@ -99,23 +104,34 @@ def structured_negative_sampling(edge_index, num_nodes=None):
...
+
+    self_loops = torch.arange(num_nodes) * (num_nodes + 1)
     i, j = edge_index.to('cpu')
     idx_1 = i * num_nodes + j
 
-    k = torch.randint(num_nodes, (i.size(0), ), dtype=torch.long)
+    k = torch.randint(num_nodes, (i.size(0),), dtype=torch.long)
     idx_2 = i * num_nodes + k
 
     mask = torch.from_numpy(np.isin(idx_2, idx_1)).to(torch.bool)
```
wsad1, line 125: I think we need the to modify the `mask`  before the while loop too. Because lets say none of the nodes are true negatives then rest would be empty and the check for self loops wouldn't run.
```suggestion
    mask = torch.from_numpy(np.isin(idx_2, idx_1)).to(torch.bool)
    if not contains_neg_self_loops:
        self_loop_mask = torch.from_numpy(np.isin(idx_2, self_loops)).to(
                torch.bool)
        mask = torch.logical_or(mask, self_loop_mask)
```

torch_geometric/utils/negative_sampling.py
```diff
@@ -99,23 +104,34 @@ def structured_negative_sampling(edge_index, num_nodes=None):
...
     :rtype: (LongTensor, LongTensor, LongTensor)
     """
     num_nodes = maybe_num_nodes(edge_index, num_nodes)
 
+    assert is_neg_samp_feasible(edge_index, 'structure', num_nodes, False,
+                                contains_neg_self_loops)
+
+    self_loops = torch.arange(num_nodes) * (num_nodes + 1)
     i, j = edge_index.to('cpu')
     idx_1 = i * num_nodes + j
```
wsad1, line 120: What if self loops are added to idx_1 if contains_neg_self_loops is false. Do you think that would be more straight forward.

wsad1 (commented):
@WuliangHuang , thanks for the great work. This is a really useful feature, I have a few comments.

torch_geometric/utils/negative_sampling.py
wsad1, line 120: what do you think of modifying `idx_1` if `contains_neg_self_loops` is `false`. That way the if condition in the while loop will not be needed.
```suggestion
    self_loops = torch.arange(num_nodes) * (num_nodes + 1)
    i, j = edge_index.to('cpu')
    idx_1 = i * num_nodes + j
    idx_1 = idx_1 if contains_neg_self_loops else torch.cat([idx_1, self_loops])
```

torch_geometric/utils/negative_sampling.py
```diff
@@ -99,23 +104,34 @@ def structured_negative_sampling(edge_index, num_nodes=None):
...
     rest = mask.nonzero(as_tuple=False).view(-1)
     while rest.numel() > 0:  # pragma: no cover
-        tmp = torch.randint(num_nodes, (rest.numel(), ), dtype=torch.long)
+        tmp = torch.randint(num_nodes, (rest.numel(),), dtype=torch.long)
         idx_2 = i[rest] * num_nodes + tmp
         mask = torch.from_numpy(np.isin(idx_2, idx_1)).to(torch.bool)
+        if not contains_neg_self_loops:
+            self_loop_mask = torch.from_numpy(np.isin(idx_2, self_loops)).to(
+                torch.bool)
+            mask = torch.logical_or(mask, self_loop_mask)
```
wsad1, line 134:
```suggestion
```

wsad1 (commented):
**ignore this comment** Added it by mistake.

WuliangHuang:
Thank you for your comments, I will look into them : ).

WuliangHuang:
Thank @wsad1 for your comments, they are useful, I just commit all of them.

WuliangHuang:
I'm so thank you for all your work here. I think that dividing the function `negative_sampling_feasible` into two separate functions is a really good way for simplifying the codes.

PR #3251, opened 2021-09-28
Fixed an issue with parent state changing upon pickling

zepx:
There's currently an issue with `BaseStorage` pickling `parent` objects.

The state of `BaseStorage` is modified upon pickling resulting the properties reliant on `self._parent` throwing errors.
For example, the property `num_nodes` and `num_edges` which relies on the call `self._parent()` will return unexpected result before and after pickling.
- Before Pickling: `num_nodes` and `num_edges` will return values accordingly when `x`, `pos` or `batch` is given.
- After Pickling: `num_nodes` and `num_edges` will throw an Exception because `self._parent` is now `self._parent()` and not a `weakref`.

This PR attempts to resolve this issue by returning a copy of `self.__dict__` instead of modifying it directly.

### DIFF
--- a/test/data/test_batch.py
+++ b/test/data/test_batch.py
@@ -193,10 +193,17 @@ def test_batching_with_new_dimension():
 def test_pickling():
     data = Data(x=torch.randn(5, 16))
     batch = Batch.from_data_list([data, data, data, data])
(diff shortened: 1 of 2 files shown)


rusty1s:
Thank you so much for spotting. Added an additional test that catches this error.

PR #3278, opened 2021-10-03
chore: fix grammar error

Cho-Geonwoo:
I think `__delitem__` method in the Data class has a typo, so I fixed it.


rusty1s:
Thanks for catching this!

PR #3282, opened 2021-10-04
add missing module

ngoanpv:
The bug appeared when running the sample code on the session Quick Tour for New Users: 
 `NameError: name 'Tensor' is not defined`

Add this line to fix the problem: 
`from torch import Tensor`


rusty1s:
Thanks!

PR #3284, opened 2021-10-04
fix nameerror 

ngoanpv:
The bug is below:
`NameError: name 'data' is not defined`

The reason due to the missing variable `data`

How to fix:
Initialize variable `data`:  
`data = dataset.data`


rusty1s:
Thanks for spotting this :)

PR #3305, opened 2021-10-07
chore: change test code for multiprocessing

Cho-Geonwoo:
Multiprocessing worked with process spawning, so I changed the test code.


rusty1s:
Thank you!

PR #3317, opened 2021-10-09
`setup.py test` is deprecated

jamesmyatt:
See:

* https://github.com/pytest-dev/pytest/issues/5534
* https://github.com/pytest-dev/pytest-runner/blob/main/README.rst

Also updated some instructions to use `python -m pytest ...` as recommended by: https://docs.pytest.org/en/6.2.x/usage.html#calling-pytest-through-python-m-pytest

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ install_requires = [
     'PyYAML',
 ]
 setup_requires = []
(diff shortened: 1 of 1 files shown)


jamesmyatt:
No problem. Just wanted to understand.

PR #3319, opened 2021-10-11
Add airport datasets

GitEventhandler:
Add support for dataset Airports :   
The Airports dataset from the ["struc2vec: Learning Node Representations from Structural Identity"](https://arxiv.org/abs/1704.03165), where nodes denote airports and labels correspond to activity levels. The feature generating method (one-hot encoding) is from the ["GraLSP: Graph Neural Networks with Local Structural Patterns"](https://arxiv.org/abs/1911.07675).


rusty1s:
Thank you!

PR #3329, opened 2021-10-13
Fix neighbor loader device

guyguygang:
Fix neighbor_loader device, make sure features and edges are on the same device. In some cases, the sampled edges/adj_t are on CPU while data.x is on GPU.


guyguygang:
Cool. Thank you!

PR #3344, opened 2021-10-16
Point Transformer examples

QuanticDisaster:
Hello

This PR is an implementation of 2 examples using Point Transformer Conv for classification and part segmentation

I tried to follow the architecture described in the paper although some details are not provided. The author released an appendix a month ago, and affirmed the full details and implementations will be disclosed to the community soon, so the architecture may have to be updated later.
https://arxiv.org/abs/2012.09164

It performs roughly around 0.84 for classification (without normals) and 0.80 for part segmentation (against 0.93% with normals and 0.84 instance IoU in paper). The learning rates parameters are different as the results seemed worse when using the one from the paper but if you have any advice/remark, I would be glad to test.

I am unsure about the "elegance" of some parts of the code, especially the transition down and up block parts, with the mlps in argument and all. Same for the outputs in x, pos and batch that are saved in a list for skipping connections in segmentation

schema

edit : I just realized I am pushing from master, I can't change it now, so tell me if I need to open a new PR or if we go on with this branch

### DIFF
--- a/examples/point_transformer_classification.py
+++ b/examples/point_transformer_classification.py
@@ -190,15 +190,16 @@ def test(loader):
 
 
 if __name__ == '__main__':
-    
+
     device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
     model = Net(train_dataset.num_classes, 1, dim_model=[
                 32, 64, 128, 256, 512], k=16).to(device)
     # NB : learning parameters are different from the paper
     optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
-    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.5)
-    
-    
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer,
+                                                step_size=20,
+                                                gamma=0.5)
+
     for epoch in range(1, 201):
         loss = train()
         test_acc = test(test_loader)
--- a/examples/point_transformer_classification.py
+++ b/examples/point_transformer_classification.py
@@ -95,6 +95,11 @@ class TransitionDown(torch.nn.Module):
         sub_pos, out = pos[id_clusters], x_out
         return out, sub_pos, sub_batch
 
+def MLP(channels):
+    return Seq(*[
+        Seq(Lin(channels[i - 1], channels[i]), BN(channels[i]), ReLU())
+        for i in range(1, len(channels))
+    ])
 
 class Net(torch.nn.Module):
     def __init__(self, in_channels, out_channels, dim_model, k=16):
@@ -105,11 +110,7 @@ class Net(torch.nn.Module):
         in_channels = max(in_channels, 1)
 
         # first block
-        self.mlp_input = Seq(
-            Lin(in_channels, dim_model[0]),
-            BN(dim_model[0]),
-            ReLU()
-        )
+        self.mlp_input = MLP([in_channels, dim_model[0]])
 
         self.transformer_input = TransformerBlock(in_channels=dim_model[0],
                                                   out_channels=dim_model[0])
--- a/examples/point_transformer_classification.py
+++ b/examples/point_transformer_classification.py
@@ -6,7 +6,8 @@ from torch.nn import Sequential as Seq, Linear as Lin, BatchNorm1d as BN, ReLU
 
 import torch_geometric.transforms as T
 from torch_geometric.datasets import ModelNet
-from torch_geometric.data import DataLoader, Data
+from torch_geometric.data import Data
+from torch_geometric.loader import DataLoader
 from torch_geometric.nn.pool import max_pool_neighbor_x, knn
 from torch_geometric.nn.conv import PointTransformerConv
 from torch_geometric.graphgym.models import global_mean_pool
@@ -91,13 +92,6 @@ def transition_down(x, pos, mlp, batch=None, k=16):
     return new_x, new_pos, sub_batch
 
 
-def MLP(channels, batch_norm=True):
-    return Seq(*[
-        Seq(Lin(channels[i - 1], channels[i]), ReLU(), BN(channels[i]))
-        for i in range(1, len(channels))
-    ])
-
-
 class Net(torch.nn.Module):
     def __init__(self, NUM_CLASSES, NUM_FEATURES, dim_model, k=16):
         super(Net, self).__init__()
@@ -168,15 +162,6 @@ class Net(torch.nn.Module):
         return F.log_softmax(out, dim=-1)
 
 
-device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-model = Net(train_dataset.num_classes, 1, dim_model=[
-            32, 64, 128, 256, 512], k=16).to(device)
-optimizer = torch.optim.SGD(
-    model.parameters(), lr=0.05, momentum=0.9, weight_decay=0.0001)
-scheduler = torch.optim.lr_scheduler.MultiStepLR(
-    optimizer, milestones=[120, 160], gamma=0.1)
-
-
 def train():
     model.train()
 
@@ -204,9 +189,19 @@ def test(loader):
     return correct / len(loader.dataset)
 
 
-for epoch in range(1, 201):
-    loss = train()
-    test_acc = test(test_loader)
-    print('Epoch {:03d}, Loss: {:.4f}, Test: {:.4f}'.format(
-        epoch, loss, test_acc))
-    scheduler.step()
+if __name__ == '__main__':
+    
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+    model = Net(train_dataset.num_classes, 1, dim_model=[
+                32, 64, 128, 256, 512], k=16).to(device)
+    # NB : learning parameters are different from the paper
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.5)
+    
+    
+    for epoch in range(1, 201):
+        loss = train()
+        test_acc = test(test_loader)
+        print('Epoch {:03d}, Loss: {:.4f}, Test: {:.4f}'.format(
+            epoch, loss, test_acc))
+        scheduler.step()
--- a/examples/point_transformer_classification.py
+++ b/examples/point_transformer_classification.py
@@ -16,8 +16,6 @@ from torch_cluster import fps
 from torch_scatter import scatter_max
 
 path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data/ModelNet10')
-# In its current state, PointTransformerConv does not support the absence of
-# features, so we create dummy nodes features with the 'Constant' Transform
 pre_transform, transform = T.NormalizeScale(), T.Compose(
     [T.SamplePoints(1024)])
 train_dataset = ModelNet(path, '10', True, transform, pre_transform)
@@ -27,9 +25,9 @@ test_loader = DataLoader(test_dataset, batch_size=32, shuffle=False)
 
 
 class TransformerBlock(torch.nn.Module):
-    def __init__(self, in_channels, out_channels, hidden_channels):
+    def __init__(self, in_channels, out_channels):
         super(TransformerBlock, self).__init__()
-        self.lin_in = Lin(in_channels, hidden_channels)
+        self.lin_in = Lin(in_channels, in_channels)
         self.lin_out = Lin(out_channels, out_channels)
 
         self.pos_nn = Seq(
@@ -45,7 +43,7 @@ class TransformerBlock(torch.nn.Module):
         )
 
         self.transformer = PointTransformerConv(
-            hidden_channels,
+            in_channels,
             out_channels,
             pos_nn=self.pos_nn,
             attn_nn=self.attn_nn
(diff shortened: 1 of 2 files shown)


examples/point_transformer_classification.py
```diff
@@ -0,0 +1,204 @@
...
+                BN(dim_model[i + 1]),
+                ReLU())
+            )
+
+        # end block
+        self.transformer = TransformerBlock(
+            in_channels=dim_model[-1],
+            out_channels=dim_model[-1],
+            hidden_channels=dim_model[-1]
+        )
```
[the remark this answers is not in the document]
QuanticDisaster: I reorganized the code in the "XXX visually" commits as well as fix the correct output to take (the one from point_transformer)

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,204 @@
...
+            self.transformers_down.append(
+                TransformerBlock(in_channels=dim_model[i],
+                                 out_channels=dim_model[i + 1],
+                                 hidden_channels=dim_model[i])
+            )
+            self.mlp_down.append(Seq(
+                Lin(dim_model[i + 1], dim_model[i + 1]),
+                BN(dim_model[i + 1]),
+                ReLU())
+            )
```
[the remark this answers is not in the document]
QuanticDisaster: This is the part I am still struggling with, I would be in favor of simply dropping the residual connection as for the moment I don't manage to make it as efficient as the net without (maybe I need to let it train much much more ?)

QuanticDisaster:
I committed the various fixes pointed out and additionnaly, added automatic dummy features. The original version was a pain to use for transfert learning as you had to add/remove the T.Constant(1) transform if you wanted to use it with independant scripts, this is no longer the case and is taken care of in the initilization and features automatically added in the forward function.

I also reformatted the code into classes for TransitionDown and TransitionUp, in the same fashion as [ POSTECH-CVLab /
point-transformer](https://github.com/POSTECH-CVLab/point-transformer) implementation, this is indded much more cleaner.

examples/point_transformer_classification.py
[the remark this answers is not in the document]
wsad1: I agree, its ok to drop the residual connection for now. Will review the rest of your updates this weekend.

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+    [T.SamplePoints(1024)])
+train_dataset = ModelNet(path, '10', True, transform, pre_transform)
+test_dataset = ModelNet(path, '10', False, transform, pre_transform)
+train_loader = DataLoader(train_dataset, batch_size=32, shuffle=True)
+test_loader = DataLoader(test_dataset, batch_size=32, shuffle=False)
+
+
+class TransformerBlock(torch.nn.Module):
+    def __init__(self, in_channels, out_channels):
+        super(TransformerBlock, self).__init__()
```
rusty1s, line 29:
```suggestion
        super().__init__()
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+            pos_nn=self.pos_nn,
+            attn_nn=self.attn_nn
+        )
+
+    def forward(self, x, pos, edge_index):
+        # x_skip = x.clone()
+        x = self.lin_in(x).relu()
+        x = self.transformer(x, pos, edge_index)
+        x = self.lin_out(x).relu()
+        # x = x + x_skip # negatively impacts training
```
rusty1s, line 57: Then let's remove it :)

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class TransitionDown(torch.nn.Module):
+    '''
+        Samples the input point cloud by a ratio percentage to reduce
+        cardinality and uses an mlp to augment features dimensionnality
+    '''
+
+    def __init__(self, in_channels, out_channels, ratio=0.25, k=16):
+        super(TransitionDown, self).__init__()
```
rusty1s, line 68:
```suggestion
        super().__init__()
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # Max pool onto each cluster the features from knn in points
+        x_out, _ = scatter_max(x[id_k_neighbor[1]],
+                               id_k_neighbor[0],
+                               dim_size=id_clusters.size(0),
+                               dim=0)
+
+        # keep only the clusters and their max-pooled features
+        new_pos, new_x = pos[id_clusters], x_out
+        return new_x, new_pos, sub_batch
```
rusty1s, line 99: Let's call the subset `pos` tensor `sub_pos` as well to make naming consistent.
```suggestion
        return out, sub_pos, sub_batch
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class Net(torch.nn.Module):
+    def __init__(self, NUM_CLASSES, NUM_FEATURES, dim_model, k=16):
+        super(Net, self).__init__()
```
rusty1s, line 104:
```suggestion
        super().__init__()
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+                               dim_size=id_clusters.size(0),
+                               dim=0)
+
+        # keep only the clusters and their max-pooled features
+        new_pos, new_x = pos[id_clusters], x_out
+        return new_x, new_pos, sub_batch
+
+
+class Net(torch.nn.Module):
+    def __init__(self, NUM_CLASSES, NUM_FEATURES, dim_model, k=16):
```
rusty1s, line 103:
```suggestion
    def __init__(self, in_channels, out_channels, dim_model, k=16):
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+        super(Net, self).__init__()
+        self.k = k
+
+        # dummy feature is created if there is none given
+        NUM_FEATURES = max(NUM_FEATURES, 1)
+
+        # first block
+        self.mlp_input = Seq(
```
rusty1s, line 111: Shall we add a simple `MLP` class on top which we can re-use? Alternatively, we can provide this in `torch_geometric.nn` as well.

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+            )
+
+        # class score computation
+        self.lin = Seq(Lin(dim_model[-1], 64),
+                       ReLU(),
+                       Lin(64, 64),
+                       ReLU(),
+                       Lin(64, NUM_CLASSES))
+
+    def forward(self, data):
```
rusty1s, line 143:
```suggestion
    def forward(self, x, pos, batch=None):
```
Let's make required arguments explicit.

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+        x, pos, batch = data.x, data.pos, data.batch
+
+        # add dummy features in case there is none
+        if x is None:
+            x = torch.ones((pos.shape[0], 1)).to(pos.get_device())
```
rusty1s, line 148: Instantiate directly on the device:
```suggestion
            x = torch.ones((pos.shape[0], 1), device=pos.get_device())
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+        optimizer.zero_grad()
+        out = model(data)
+        loss = F.nll_loss(out, data.y)
+        loss.backward()
+        total_loss += loss.item() * data.num_graphs
+        optimizer.step()
+    return total_loss / len(train_dataset)
+
+
+def test(loader):
```
rusty1s, line 186: Prefer decorator `@torch.no_grad()` instead of `with torch.no_grad()`.

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+        with torch.no_grad():
+            pred = model(data).max(dim=1)[1]
+        correct += pred.eq(data.y).sum().item()
+    return correct / len(loader.dataset)
+
+
+if __name__ == '__main__':
+
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+    model = Net(train_dataset.num_classes, 1, dim_model=[
```
rusty1s, line 201:
```suggestion
    model = Net(train_dataset.num_classes, 0, dim_model=[
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,213 @@
...
+                32, 64, 128, 256, 512], k=16).to(device)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer,
+                                                step_size=20,
+                                                gamma=0.5)
+
+    for epoch in range(1, 201):
+        loss = train()
+        test_acc = test(test_loader)
+        print('Epoch {:03d}, Loss: {:.4f}, Test: {:.4f}'.format(
```
rusty1s, line 211: Prefer `f`-string syntax:
```suggestion
        print(f'Epoch {epoch:03d}, Loss: {loss:.4f}, Test: {test_acc:.4f}')
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,250 @@
...
+        # transform low-res features and reduce the number of features
+        x_sub = self.mlp_sub(x_sub)
+
+        # interpolate low-res feats to high-res points
+        x_interpolated = knn_interpolate(
+            x_sub, pos_sub, pos, k=3, batch_x=batch_sub, batch_y=batch)
+
+        x = self.mlp(x) + x_interpolated
+
+        return x, pos
```
rusty1s, line 64: Do we return `pos` just for convenience (as it is simply passed through)?

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class TransitionUp(torch.nn.Module):
+    '''
+        Reduce features dimensionnality and interpolate back to higher
+        resolution and cardinality
+    '''
+
+    def __init__(self, in_channels, out_channels):
+        super(TransitionUp, self).__init__()
```
rusty1s, line 41:
```suggestion
        super().__init__()
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,250 @@
...
+            x_sub, pos_sub, pos, k=3, batch_x=batch_sub, batch_y=batch)
+
+        x = self.mlp(x) + x_interpolated
+
+        return x, pos
+
+
+class Net(torch.nn.Module):
+    def __init__(self, NUM_CLASSES, NUM_FEATURES, dim_model, k=16):
+        super(Net, self).__init__()
```
rusty1s, line 69:
```suggestion
        super().__init__()
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,250 @@
...
+        x_interpolated = knn_interpolate(
+            x_sub, pos_sub, pos, k=3, batch_x=batch_sub, batch_y=batch)
+
+        x = self.mlp(x) + x_interpolated
+
+        return x, pos
+
+
+class Net(torch.nn.Module):
+    def __init__(self, NUM_CLASSES, NUM_FEATURES, dim_model, k=16):
```
rusty1s, line 68: I prefer
```suggestion
    def __init__(self, in_channels, out_channels, dim_model, k=16):
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+
+        # class score computation
+        self.lin = Seq(Lin(dim_model[0], 64),
+                       ReLU(),
+                       Lin(64, 64),
+                       ReLU(),
+                       Lin(64, NUM_CLASSES))
+
+    def forward(self, data):
```
rusty1s, line 136:
```suggestion
    def forward(self, x, pos, batch=None):
```

rusty1s (approved):
This is absolutely great! Just have minor comments regarding coding style :)

examples/point_transformer_segmentation.py
QuanticDisaster: Indeed, I didn't realize it. I changed it and pos is not returned anymore along with x

examples/point_transformer_classification.py
QuanticDisaster: I feel like putting data would allow models to be more independant, like bricks that we can move from one code to another. Is there a good practice consensus on this, wether in PyG or in Pytorch ? I took as reference the example on PointNet in PyG : https://github.com/pyg-team/pytorch_geometric/blob/master/examples/pointnet2_segmentation.py

examples/point_transformer_classification.py
QuanticDisaster: Providing it in torch_geometric would be good indeed. While it could be considered as a detail, some may prefer BN before Relu and vice versa however

examples/point_transformer_classification.py
rusty1s: If you feel that you would leave the input args as `data` I'm fine with that as well. I know that we used `data` as input argument in previous examples extensively, but I consider it bad-practice for now, since it is not really clear which attributes of the `data` needs to be set.

An alternative way is to define `forward` as:
```python
def forward(self, x, pos, batch=None, *args, **kwargs):
    pass
```
and call it via
```python
model(**data.to_dict())
```

QuanticDisaster:
@rusty1s Done, I finally changed to explicit arguments as you recommended, so I think it is finished now. Maybe a disclaimer to point out the possible differences (in skip connection notably) and we are good to go !

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,211 @@
...
+    '''
+        Samples the input point cloud by a ratio percentage to reduce
+        cardinality and uses an mlp to augment features dimensionnality
+    '''
+
+    def __init__(self, in_channels, out_channels, ratio=0.25, k=16):
+        super().__init__()
+        self.k = k
+        self.ratio = ratio
+        self.mlp = Seq(
```
rusty1s, line 68:
```suggestion
        self.mlp = MLP([in_channels, out_channels])
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,211 @@
...
+test_loader = DataLoader(test_dataset, batch_size=32, shuffle=False)
+
+
+class TransformerBlock(torch.nn.Module):
+    def __init__(self, in_channels, out_channels):
+        super().__init__()
+        self.lin_in = Lin(in_channels, in_channels)
+        self.lin_out = Lin(out_channels, out_channels)
+
+        self.pos_nn = Seq(
```
rusty1s, line 32:
```suggestion
        self.pos_nn = MLP([3, 64, out_channels], batch_norm=False)
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,211 @@
...
+            Lin(3, 64),
+            ReLU(),
+            Lin(64, out_channels)
+        )
+
+        self.attn_nn = Seq(
```
rusty1s, line 38:
```suggestion
        self.attn_nn = MLP([out_channels, 64, out_channels], batch_norm=False)
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,211 @@
...
+                               dim=0)
+
+        # keep only the clusters and their max-pooled features
+        sub_pos, out = pos[id_clusters], x_out
+        return out, sub_pos, sub_batch
+
+
+def MLP(channels):
+    return Seq(*[
+        Seq(Lin(channels[i - 1], channels[i]), BN(channels[i]), ReLU())
```
rusty1s, line 101:
```suggestion
from torch.nn import Identity
        Seq(Lin(channels[i - 1], channels[i]), BN(channels[i]) if batch_norm else Identity(), ReLU())
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,246 @@
...
+
+class TransitionUp(torch.nn.Module):
+    '''
+        Reduce features dimensionnality and interpolate back to higher
+        resolution and cardinality
+    '''
+
+    def __init__(self, in_channels, out_channels):
+        super().__init__()
+        self.mlp_sub = Seq(
```
rusty1s, line 43:
```suggestion
        self.mlp_sub = MLP([in_channels, out_channels])
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,246 @@
...
+            Lin(in_channels, out_channels),
+            BN(out_channels),
+            ReLU()
+        )
+
+        self.mlp = Seq(
```
rusty1s, line 49:
```suggestion
        self.mlp = MLP([out_channels, out_channels])
```

examples/point_transformer_segmentation.py
```diff
@@ -0,0 +1,246 @@
...
+                             out_channels=dim_model[i])
+            )
+
+            self.transformers_up.append(
+                TransformerBlock(in_channels=dim_model[i],
+                                 out_channels=dim_model[i])
+            )
+
+        # summit layers
+        self.mlp_summit = Seq(
```
rusty1s, line 115:
```suggestion
        self.mlp_summit = MLP([dim_model[-1], dim_model[-1]], batch_norm=False)
```

examples/point_transformer_classification.py
```diff
@@ -0,0 +1,211 @@
...
+                               k=self.k)
+            )
+
+            self.transformers_down.append(
+                TransformerBlock(in_channels=dim_model[i + 1],
+                                 out_channels=dim_model[i + 1])
+            )
+
+        # class score computation
+        self.lin = Seq(Lin(dim_model[-1], 64),
```
rusty1s, line 137:
```suggestion
        self.mlp_output = Seq(Lin(dim_model[-1], 64),
```

rusty1s (approved):
Thanks! This is perfect. My only question would be if we want to use `MLP` more consistently across all models.

QuanticDisaster:
@rusty1s Woops I didn't noticed it could be replaced in all these parts, sorry for the additionnal workload. It should be corrected now !

rusty1s:
Yeah, you are right that it currently restricts the usage of `DataParallel` (which I'm not that happy to begin with), and even PyTorch recommends to use `DistributedDataParallel`. If the user wants to make use of `DataParallel` within this example, it's trivial to adjust the input arguments. As such, I'm happy to merge this PR. Thanks for your efforts!

PR #3350, opened 2021-10-19
An example of heterogeneous link prediction(classification) via RandomLinkSplit

### LINKED ISSUE #3217: An example of heterogeneous link prediction via `RandomLinkSplit`

anniekmyatt:
PR to resolve #3217.

This PR contains:
- The MovieLens dataset extracted into a dataset file to be able to cache the data.
- A link prediction example `hetero_link_pred.py` to the examples/hetero folder, based on an RMSE loss.

The PR is now ready for review :-). Thank you!

edit: I had another look and noticed that there was an issue in the reporting, the sqrt was missing for my rmse calculation. 🤦‍♀️ 
I fixed that but of course the performance looks worse now. I can get to an rmse of about 0.90 for my test set, which is not [state of the art](https://paperswithcode.com/sota/collaborative-filtering-on-movielens-1m), but my example is also a lot more simple than the published papers.

Please let me know if you would prefer I worked on this a bit further or if it suffices as a simple, illustrative example.


rusty1s:
Super cool :) I think we can try to frame this problem as a regression problem, and evaluate via RMSE. It seems that this is the preferred way on this dataset, see [here](https://paperswithcode.com/sota/collaborative-filtering-on-movielens-1m).

anniekmyatt:
Ah yes, good tip, thanks! I’ll give that a go. And I’ll do some tidying so the style checks etc. won’t fail anymore!

anniekmyatt:
I've changed it to use an MSE based loss but I'm still a bit unhappy with it. The final RMSE goes to about 0.9 for the test set, which isn't too far off the ~0.83 reported in the [link ](https://paperswithcode.com/sota/collaborative-filtering-on-movielens-1m) you provided but if you split the performance by label, you can see that the reason it performs well is that we have quite an imbalanced dataset, and for the labels with fewer examples it performs rather poorly. The picture is similar for accuracy.

I have tried different things, e.g. 
- resampling of the supervision edges to handle the unbalanced labels
- having different edge types, according to the associated rating, for the message passing, to possibly encode more information about the rating behaviour of a user (rather than just working with a pattern of "this user rated these movies")
- using a GNN set up with the `HeteroConv` wrapper so I can have batch-normalisation, and dropout to try to reduce overfitting
- using a bilinear decoder
- using minibatching

... and none of them seem to make a huge difference to the final performance.

I have added a very simplified example, `hetero_link_pred.py`, to the repo because I didn't want to clutter it with things that didn't help very much. I'm happy to add more complex examples if you're interested to have them.

My question is: how good does it need to be to be merged?  I appreciate that it is not the best performance but maybe it is just an illustration of you would set up a heterogeneous link prediction? I can add the slightly more complex version with the `HeteroConv` wrapper, so that people can see how to use that, if that is helpful? If it needs to be better, I'm happy to continue to work on it, I just wanted to find out what I'm aiming for :).


Another question: In the dataset file `datasets\movie_lense.py` you were importing the `SentenceTransformer` from huggingface but some tests are failing because this is not part of the requirements, and I can imagine you don't want to clutter your requirements file with all these extra dependencies. What is your preferred way of handling this?

anniekmyatt:
BTW, if I set `disjoint_train_ratio=0` in the `RandomLinkSplit` call, rather than having it at `disjoint_train_ratio=0.2`,  I can get a similar overall RMSE as reported in the papers with code [link ](https://paperswithcode.com/sota/collaborative-filtering-on-movielens-1m)(around 0.83) but the underlying issue of poor performance for some labels remains, and we would then have target leakage.

rusty1s:
Thank you so much for your hard work on this one. Please correct me if I'm wrong, but using `disjoint_train_ratio=0` should be okay to use as we are not using edge labels during inference anyway. I'm definitely okay with merging this one :)

Not yet sure how we should handle the `SentenceTransformer` errors. One way I could think of is to just store and load the pre-processed embeddings upon dataset creation. If you are okay with this, I can take care of that.

anniekmyatt:
I think it might be better to have a non-zero disjoint ratio for several reasons: 
- to highlight that you have to be extra careful to avoid target leakage in link prediction, and to raise awareness of the existence of the `disjoint_train_ratio` option in your code base.
- I think we are actually kind of using the edge labels during inference, as we are using the associated rating of a pair of nodes as a label to predict. It is a bit of an interesting example because we only have the edge types `('user', 'rates', 'movie')` and the reverse, so we're not predicting different edges of the type `('user', 'rates_movie_as_<n>', 'movie')` but I do think the information associated with an edge label goes in, so I think it is safer to have the disjoint option.
 
On the other hand I found this [paper](https://arxiv.org/abs/1706.02263) that looks like it's got a similar implementation to this one (although mine is simplified), and in the associated data pre-processing [code](https://github.com/riannevdberg/gc-mc/blob/master/gcmc/preprocessing.py#L166-L168y), it doesn't look like a disjoint split is performed.

I favour the option of having a  non-zero disjoint ratio for the reasons I mentioned above but I'm happy to take your lead.

Thanks for taking care of the `SentenceTransformer` issue! Does this mean that there is an option to download pre-processed embeddings rather calculating them yourself in the code? 
Having different features on the nodes would probably have an impact on performance, so if anything changes, I'm happy to take another look at this.

rusty1s:
How about we make `disjoint_train_ratio` as part of an `argparse` to let the user know about the differences in potential implementations? I still don't think there is a data leakage during encoding, which would only happen if you incorporate edge labels during message passing.

I can create and upload embeddings obtained from `SentenceTransformer` to https://data.pyg.org. We could make this optionable, and let the user decide if it wants to generate them from scratch by itself (e.g., by using a different model), or use the existing ones. CI should be fixable nonetheless by moving
```python
from sentence_transformers import SentenceTransformer
```
to the
```python
def process():
    from sentence_transformers import SentenceTransformer
```
method

anniekmyatt:
Yeah I see what you mean about the edge label not going in during the encoding with the GNN. I was worried about the end-to-end process, because there are also learnable weights in the decoder, where we do take into account the ratings info, but there we calculate the predicted rating as the expectation value based on all ratings options, and not just the ratings label . So I think you're right. Sorry, to me the issue of target leakage for link prediction is really not trivial and perhaps I tend to be overly cautious!

I like your idea of making it an option in `argparse`, I will do that. 

I can also move the import of the sentence transformers into the process function to try and fix the CI. Let me know when the the data is up on https://data.pyg.org/ and I can change the code in the `MovieLens` file to use it.

One question about making the model to create the node features optional for the users: would the user pass in a `SequenceEncoder` object, and if that input is empty it defaults to the one we are currently using? I'm sorry to ask but I'm not sure yet how much flexibility you would like to offer the user.

rusty1s:
I think the way we handle `SequenceEncoder` for now is not critical. I suggest to simply expect a `model_name` as input argument to the dataset. We can then change it later to load embeddings from an URL for models for which we have pre-computed them.

examples/hetero/hetero_link_pred.py
```diff
@@ -0,0 +1,219 @@
...
+data = dataset[0]
+
+data['user'].x = torch.eye(data['user'].num_nodes, dtype=data['movie'].x.dtype)
+
+# We can now convert `data` into an appropriate format for training a
+# graph-based machine learning model:
+
+# 1. Add a reverse ('movie', 'rev_rates', 'user') relation for message passing.
+data = ToUndirected()(data)
+del data['movie', 'rev_rates', 'user'].edge_label  # Remove "reverse" label.Fd
```
anniekmyatt, line 43: I'll remove the stray "Fd" at the end here but I don't want to keep changing the code until you've had a chance to take a look as well.

anniekmyatt:
Hello @rusty1s ! Just a quick follow-up to check if you had a chance to look at this PR?

I'm happy to spend more time on it but it would be very helpful to have some feedback on which areas I should focus on.

Thank you! :-)

rusty1s:
Super, will merge today or tomorrow :) Thank you!

QuanticDisaster:
> I've changed it to use an MSE based loss but I'm still a bit unhappy with it. The final RMSE goes to about 0.9 for the test set, which isn't too far off the ~0.83 reported in the link you provided but if you split the performance by label, you can see that the reason it performs well is that we have quite an imbalanced dataset, and for the labels with fewer examples it performs rather poorly. The picture is similar for accuracy.

@anniekmyatt I am just passing by so I am not exactly aware of what you are trying to do and if this fit your use case but with regards to your class imbalance problem, can't you use a weighted loss ? such as : https://discuss.pytorch.org/t/how-to-implement-weighted-mean-square-error/2547/2

A first approximation of weights of each label would be 1 / number of instances. Although it has the downside of having to iterate over all the dataset once at first to compute the number of instances of each label.

anniekmyatt:
@QuanticDisaster , thank you for the suggestion! I had tried resampling the classes with smaller numbers of samples but I hadn't tried this yet. I'll try this out and will report back :) .

anniekmyatt:
The weighted loss improves the performance for the lower frequency classes quite a bit but this comes at a small cost to the overall performance. I think it would be better to use the weighted loss, as without it, the predictions for the ratings 0 and 1 are very poor, there wasn't really any learning going on. Thanks for the suggestion, @QuanticDisaster!

I've just pushed the changes.

QuanticDisaster:
@anniekmyatt Great, I am glad it has been of help !

rusty1s (approved):

rusty1s:
Cleaned up a little bit :) I think using an MLP decoder makes more sense as `DistMult` in the scenario of regression (hope that is okay). I accidentally removed the weighted MSE loss formulation. Will add it back in.

anniekmyatt:
Thanks very much for tidying, it looks great! :)

Yes, using an MLP decoder makes sense, it took me a while to figure out how to turn the `DistMult' decoder into a regression version that backpropagates, it is not very intuitive.

One thing I noticed this morning is that for the weights in the weighted loss I actually got better results with something below, rather than using the inverse class label instances, as the zero ratings instances are so few that the weights for the higher ratings classes become very small, and that makes the training not very smooth (the rmse jitters up and down). I'm afraid I haven't got time to add it myself today, I just wanted to suggest this in case you want to use it.

```
weights = torch.tensor([3,3,1,1,1,2])
weights = weights/torch.sum(weights)
```

rusty1s:
Yeah, you need to pass in the inverse, i.e., fewer labels need to be weighted higher. I achieved this via:
```python
weight = torch.bincount(...)
weight = weight.max() / weight
```
The performance indeed decreases, so I made this weighted loss formulation optional.

Thank you very much for your contribution. I liked it a lot. This is now merged :)

anniekmyatt:
Excellent, so pleased that this is merged! :)

Ah yes, I realised that I made a mistake in my previous comment, in my code I did use `torch.sum(weights)/weights` not just the weights 🤦 , apologies for the confusion.

Anyway, it has been a pleasure to work with you on this!

PR #3386, opened 2021-10-25
Graph Explainer Experiments 

RBendias:
Graph Explainer Experiments according to [Ying et al.](https://arxiv.org/pdf/1903.03894.pdf) & [Luo et al.](https://arxiv.org/pdf/2011.04573.pdf)
In the original papers of both PGExplainer and GNNExplainer synthetic datasets were used to test the mentioned methods. We aim to reproduce the results in the examples here.


torch_geometric/datasets/__init__.py
```diff
@@ -62,8 +62,10 @@
 from .gemsec import GemsecDeezer
 from .twitch import Twitch
 from .airports import Airports
+from .ba_shapes import BAShapes
 
 __all__ = [
+    'BAShapes',
```
rusty1s, line 68: Please move to the end of the list.

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+        y = torch.LongTensor(y)
+        test_mask = np.zeros(num_nodes, dtype=bool)
+        test_mask[range(400, num_nodes, 5)] = True
+
+        data = Data(x=x, edge_index=edge_index, y=y, test_mask=test_mask,
+                    edge_label=edge_labels)
+
+        self.data, self.slices = self.collate([data])
+
+    def __repr__(self):
```
rusty1s, line 102: Can be dropped.

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+def house():
+    G = nx.house_graph()
+    labels = [1, 1, 2, 2, 3]
+    attr = {i: labels[i] for i in range(5)}
+    nx.set_node_attributes(G, attr, 'label')
+    nx.set_edge_attributes(G, 1, 'label')
+    return G
+
+
+def ba(num_nodes=300, m=5, seed=None):
```
rusty1s, line 18: There also exists [`barabasi_albert_graph`](https://pytorch-geometric.readthedocs.io/en/latest/modules/utils.html#torch_geometric.utils.barabasi_albert_graph) in PyG.

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+            version. The data object will be transformed before every access.
+            (default: :obj:`None`)
+    """
+    def __init__(
+        self,
+        seed: Optional[int] = None,
+        connection_distribution: str = 'random',
+        transform: Optional[Callable] = None,
+    ):
+        super().__init__('.', transform, None, None)
```
rusty1s, line 51:
```suggestion
        super().__init__('.', transform)
```

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+        assert connection_distribution in ['random', 'uniform']
+        if seed is not None:
+            np.random.seed(seed)
```
rusty1s, line 54: I think I would drop this line. If the user wants deterministic behavior, the user can simply set `torch_geometric.seed_everything` at the root-level.

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+        num_shapes = 80
+        if connection_distribution == 'random':
+            connecting_nodes = np.random.choice(num_nodes, num_shapes, False)
+        else:
+            step = num_nodes // num_shapes
+            connecting_nodes = np.arange(0, num_nodes, step=step)
+
+        # Connect shapes to basis graph
+        for i in range(num_shapes):
+            G_house = house()
```
rusty1s, line 71: I prefer to operate on the tensor-level, which is much more PyG-like:
```python
house_edge_index = house() + num_nodes
connection_edge_index = torch.tensor([[rand_int], [num_nodes]])
edge_index = torch.cat([edge_index, house_edge_index, connection_edge_index], dim=-1)
```

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+
+        num_nodes = len(G.nodes())
+        x = torch.ones((num_nodes, 10), dtype=torch.float)
+        edge_index = torch.LongTensor(list(G.edges)).t().contiguous()
+        y = list(nx.get_node_attributes(G, 'label').values())
+        y = torch.LongTensor(y)
+        test_mask = np.zeros(num_nodes, dtype=bool)
+        test_mask[range(400, num_nodes, 5)] = True
+
+        data = Data(x=x, edge_index=edge_index, y=y, test_mask=test_mask,
```
rusty1s, line 97: Don't we need a `train_mask` as well?

torch_geometric/datasets/ba_shapes.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class BAShapes(InMemoryDataset):
+    r"""The BA-Shapes dataset from the `"GNNExplainer: Generating Explanations
+    for Graph Neural Networks" <https://arxiv.org/pdf/1903.03894.pdf>`_ paper,
+    containing a Barabasi-Albert (BA) graph with 300 nodes and a set of 80
+    "house"-structured graphs connected to it.
+
+    Args:
+        seed (int, optional): Random seed used for :obj:`networkx` Barabási
```
rusty1s, line 32: Can we also add the option to load the `BAShapes` data pre-defined from `PGExplainer`?

rusty1s (approved):

PR #3411, opened 2021-10-30
I am creating a sample script for shadow sampler.

SubhajitDuttaChowdhury:
This is a simple implementation of shadow sampler.


shadow_sampler.py
```diff
@@ -0,0 +1,104 @@
...
+from torch_geometric.loader import GraphSAINTRandomWalkSampler
+from torch_geometric.loader import ShaDowKHopSampler
+from torch_geometric.nn import GraphConv
+from torch_geometric.utils import degree
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Flickr')
+dataset = Flickr(path)
+data = dataset[0]
+row, col = data.edge_index
+data.edge_weight = 1. / degree(col, data.num_nodes)[col]  # Norm by in-degree.
```
rusty1s, line 16: I don't think we need this normalization here, as `ShaDowKHopSampler` does not perform a re-weighting.

shadow_sampler.py
```diff
@@ -0,0 +1,104 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--use_normalization', action='store_true')
+args = parser.parse_args()
+
+
+indices = data.train_mask.nonzero()
+index = torch.flatten(indices)
+loader = ShaDowKHopSampler(data, depth=2, num_neighbors=5, node_idx=index)
```
rusty1s, line 25: It should work with `data.train_mask` as input as well. If not, please let me know.

shadow_sampler.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Net(torch.nn.Module):
```
rusty1s, line 28: Let's make direct use of `torch_geometric.nn.GraphSAGE` to reduce code size:, see https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.models.GraphSAGE.

shadow_sampler.py
```diff
@@ -0,0 +1,104 @@
...
+    model.set_aggr('add' if args.use_normalization else 'mean')
+
+    total_loss = total_examples = 0
+    for data in loader:
+        data = data.to(device)
+        optimizer.zero_grad()
+
+        if args.use_normalization:
+            edge_weight = data.edge_norm * data.edge_weight
+            out = model(data.x, data.edge_index, edge_weight)
```
rusty1s, line 71: It seems the model is missing the pooling step to go from subgraph-level embeddings to node-level embeddings.

rusty1s (changes requested):

shadow_sampler.py
SubhajitDuttaChowdhury: Yes, you are right. I have removed that part.

shadow_sampler.py
SubhajitDuttaChowdhury: Yes it works. I have changed it.

shadow_sampler.py
SubhajitDuttaChowdhury: I have added the pooling step.

shadow_sampler.py
SubhajitDuttaChowdhury: Okay. I have changed it to SAGEConv.

rusty1s (approved):

rusty1s:
Thank you!

PR #3412, opened 2021-10-31
fix args.linear option in autoencoder example

terencekwt:
swap options for VariationalGCNEncoder and VariationalLinearEncoder for args.linear option


rusty1s:
Thanks! :)

PR #3421, opened 2021-11-02
Add edge_attr support to GATv2Conv

Kenneth-Schroeder:
Adding simple implementation for edge_attr support in GATv2Conv, inspired by GATConv.

### DIFF
--- a/torch_geometric/nn/conv/gatv2_conv.py
+++ b/torch_geometric/nn/conv/gatv2_conv.py
@@ -37,10 +37,10 @@ class GATv2Conv(MessagePassing):
         \exp\left(\mathbf{a}^{\top}\mathrm{LeakyReLU}\left(\mathbf{\Theta}
         [\mathbf{x}_i \, \Vert \, \mathbf{x}_k]
         \right)\right)}.
-    
+
     If the graph has multi-dimensional edge features :math:`\mathbf{e}_{i,j}`,
     the attention coefficients :math:`\alpha_{i,j}` are computed as
-    
+
     .. math::
         \alpha_{i,j} =
         \frac{
@@ -115,7 +115,7 @@ class GATv2Conv(MessagePassing):
                                 weight_initializer='glorot')
 
         self.att = Parameter(torch.Tensor(1, heads, out_channels))
-        
+
         if edge_dim is not None:
             self.lin_edge = Linear(edge_dim, heads * out_channels, bias=False,
                                    weight_initializer='glorot')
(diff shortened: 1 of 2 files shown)


Kenneth-Schroeder:
Dear maintainers,
can I do anything else to get this merged? :)

Best,
Kenneth

rusty1s:
No, this looks good. Thank you! Will have a closer look by tomorrow and will merge ASAP.

torch_geometric/nn/conv/gatv2_conv.py
```diff
@@ -147,13 +193,31 @@ def forward(self, x: Union[Tensor, PairTensor], edge_index: Adj,
...
+                        "simultaneously is currently not yet supported for "
+                        "'edge_index' in a 'SparseTensor' form")
+
+        edge_attr_transformed: OptTensor = None
+        if edge_attr is not None:
+            if edge_attr.dim() == 1:
+                edge_attr = edge_attr.view(-1, 1)
+            assert self.lin_edge is not None
+            edge_attr_transformed = self.lin_edge(edge_attr)
+            edge_attr_transformed = edge_attr_transformed.view(-1, H, C)
```
wsad1, line 216: This is a minor nitpick. Do you think we can move this piece of code into the `message` function, this makes it more consistent with other layers that use `edge_attr` like [TransformerConv](https://github.com/pyg-team/pytorch_geometric/blob/958b1dd3e80e0ea670f1cb149e536ed1e98b55cc/torch_geometric/nn/conv/transformer_conv.py#L220) or [GatConv](https://github.com/pyg-team/pytorch_geometric/blob/958b1dd3e80e0ea670f1cb149e536ed1e98b55cc/torch_geometric/nn/conv/gat_conv.py#L256).

Kenneth-Schroeder: Thanks for the feedback!
I moved it to the message function for consistency :)
However, I found it a little easier to understand in the forward function, since the node features (x) are also usually transformed there.
Do you know why this is handled differently for the edge features? Does it have performance benefits?

wsad1: Given the message->aggregate->update scheme, ideally both the linear operation on node and linear on edge attributes should be computed in the `message` function since these operations are used to compute the message.

But running the linear operation in forward for the node features gives you performance benefits, as in message the operation is over x_i or x_j(E*F) and in forward its just over x(N*F). For edge attributes it where the linear operation is done doesn't make a difference.

Kenneth-Schroeder: I was a bit confused because forward and message are both only called once in the current setup and I thought there would be no performance differences. But now I realise that MessagePassing's propagate is encapsulating the rearrangement of the elements of X that are passed to the message function.
Thank you :)

wsad1 (approved):
@Kenneth-Schroeder , this looks good. Thanks.

zcaicaros:
Hi, thanks for the edge_attr update. The current message() API version does not consider edge features embedded in the deeper layer. Like the node embedding where it comes from shallow layers get updated in the deeper layer by passing them as arguments, it would be great if the edge_attr follows the same procedure. Currently, the message() API only returns node embedding but no edge embedding, making it impossible to pass the edge_attr from layer to layer. Any suggestions?

rusty1s:
We are working on an `update_edge` functionality directly in the `MessagePassing` interface, which should make updating edge features much more straightforward outside of `message`, see https://github.com/pyg-team/pytorch_geometric/pull/3450.

You are right with your observations regarding `add_self_loops`, but I don't have any good ideas on how to fix that. One option is to disable `add_self_loops` by default if we encounter bipartite message passing, but there are still some corner cases where we want to add self-loops even in a bipartite graph, e.g., in `NeighborSampler` scenarios (where the bipartite graph gradually increases the number of nodes, and `x_src[i]` and `x_dst[i]` still point to the same node).

PR #3444, opened 2021-11-07
Added HANConv and its respective tests

rishubhkhurana:
Added HANConv network based on the paper: https://arxiv.org/abs/1903.07293


wsad1:
@rishubhkhurana Thanks for adding the han layer. I made a few changes. The key change is to the  `group` function.

rusty1s:
Thanks @rishubhkhurana and @wsad1. Will review as well ASAP.

test/nn/conv/test_han_conv.py
```diff
@@ -0,0 +1,78 @@
...
+    in_channels = {'author': 16, 'paper': 12, 'term': 3}
+
+    conv = HANConv(in_channels, 16, metadata, heads=2)
+    assert str(conv) == 'HANConv(16, heads=2)'
+    out_dict1 = conv(x_dict, edge_index_dict)
+    assert len(out_dict1) == 3
+    assert out_dict1['author'].size() == (6, 16)
+    assert out_dict1['paper'].size() == (5, 16)
+    assert out_dict1['term'] is None
+    del x_dict['term']
```
rusty1s, line 40: Not sure why we need this.

torch_geometric/nn/conv/han_conv.py
```diff
@@ -0,0 +1,164 @@
...
+        attn = F.softmax(attn_score, dim=0)
+        out = torch.sum(attn.view(num_edge_types, 1, -1) * out, dim=0)
+        return out
+
+
+class HANConv(MessagePassing):
+    r"""
+    The Heterogenous Graph Attention Operator from the
+    `"Heterogenous Graph Attention Network"
+    <https://arxiv.org/pdf/1903.07293.pdf>`_ paper.
```
rusty1s, line 32: Shall we add a small example to show-case the usage of `HANConv` in combination with `T.AddMetaPaths()`?

torch_geometric/nn/conv/han_conv.py
```diff
@@ -0,0 +1,164 @@
...
+        reset(self.proj)
+        glorot(self.lin_src)
+        glorot(self.lin_dst)
+        self.k_lin.reset_parameters()
+        glorot(self.q)
+
+    def forward(
+        self,
+        x_dict: Dict[NodeType, Tensor],
+        edge_index_dict: Union[Dict[EdgeType, Tensor], Dict[EdgeType,
```
rusty1s, line 97:
```suggestion
        edge_index_dict: Dict[EdgeType, Adj]

```

rusty1s (approved):

test/nn/conv/test_han_conv.py
wsad1: To clarify. Are you referring to the node type `term` and the check `assert out_dict`['term'] is None`. This is a check to make sure that, nodes that don't receive any message have output none.

rusty1s: I was referring to the `del` statement.

wsad1: Ah I see. So few lines after this  `out_dict1` and `out_dict2`  are compared, 
```
    for out1, out2 in zip(out_dict1.values(), out_dict2.values()):
        assert torch.allclose(out1, out2, atol=1e-6)
```
And `torch.allclose` doesn't seem to compare NoneType.

rusty1s: Yeah, the `del x_dict` line is unnecessary nonetheless. But this is nitpicky.

wsad1: Thanks. Your right, the zip command would only contain as many elements as the smaller list. Removed that delete statement.

rusty1s: The `zip` command is a bit dangerous as well, as it requires that dictionary keys are sorted identically, which probably does not need to be the case:
```python
for node_type in out_dict1.keys():
    assert torch.allclose(out_dict1[node_type], out_dict2[node_type])
```

wsad1: Yeah good point.  Fixed that in the latest commit.

torch_geometric/nn/conv/han_conv.py
wsad1: Yeah, I think that's needed. Will create an issue for that.

rusty1s:
We have to thank you :)

Issue #813, opened 2019-11-20
How to update edges?


LeeHuan18:
Hi,
    Just like the nodes update ( and message passing ) by x_j = f(x_i), I want to also update the edges features e_j = f(node1, node2, e_j), with node1, node2 it's the two nodes edge e_j connect to. I don't want to use line_graph in my case. Is there anything like the MassagePassing  https://pytorch-geometric.readthedocs.io/en/latest/notes/create_gnn.html for the edges?

fdiehl:
To the best of my knowledge, the best way of doing this is using the [`MetaLayer`](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.meta.MetaLayer) (depending on your use-case, simply ignore the `GlobalModel` part).

For example, for my recent work, I used something close to the example from `MetaLayer`:

```python
class EdgeModel(torch.nn.Module):
    def __init__(self, n_features, n_edge_features, hiddens, n_targets, residuals):
        super().__init__()
        self.residuals = residuals
        self.edge_mlp = nn.Sequential(
            nn.Linear(2 * n_features + n_edge_features, hiddens),
            nn.ReLU(),
            nn.Linear(hiddens, n_targets),
        )

    def forward(self, src, dest, edge_attr, u=None, batch=None):
        out = torch.cat([src, dest, edge_attr], 1)
        out = self.edge_mlp(out)
        if self.residuals:
            out = out + edge_attr
        return out


class NodeModel(torch.nn.Module):
    def __init__(self, n_features, n_edge_features, hiddens, n_targets, residuals):
        super(NodeModel, self).__init__()
        self.residuals = residuals
        self.node_mlp_1 = nn.Sequential(
            nn.Linear(n_features + n_edge_features, hiddens),
            nn.ReLU(),
            nn.Linear(hiddens, n_targets),
        )
        self.node_mlp_2 = nn.Sequential(
            nn.Linear(hiddens + n_features, hiddens),
            nn.ReLU(),
            nn.Linear(hiddens, n_targets),
        )

    def forward(self, x, edge_index, edge_attr, u, batch):
        row, col = edge_index
        out = torch.cat([x[col], edge_attr], dim=1)
        out = self.node_mlp_1(out)
        out = scatter_mean(out, row, dim=0, dim_size=x.size(0))
        out = torch.cat([x, out], dim=1)
        out = self.node_mlp_2(out)
        if self.residuals:
            out = out + x
        return out


[...]

PR #3450, opened 2021-11-08
`edge_update` example

Padarn:
PR that addresses discussion in https://github.com/pyg-team/pytorch_geometric/issues/813#issuecomment-962912361


rusty1s:
Yes, I'm absolutely in favor of adding this. The interface looks good to me.

Padarn:
Sorry no time during the week to work on this. I'll cleanup the code on the backend and push to this PR.

Padarn:
I've added a basic implementation of this (obvious needs tests and documentation etc). 

Do you think it might be better to reuse the `propagate` function with some arguments e.g
```
x = self.propagate(...., edges=True)
```
Instead of splitting into the `self.edges_propagate`, I've added?

examples/edge_update.py
```diff
@@ -0,0 +1,32 @@
...
+])
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', dataset)
+dataset = Planetoid(path, dataset, transform=transform)
+data = dataset[0]
+
+class EdgeMessagePassingExample(MessagePassing):
+
+    def forward(self, data: Data):
+        x = self.propagate(data.edge_index, edge_attr=data.edge_attr, x=data.x)
+        edge_attr = self.edges_propagate(data.edge_index, edge_attr=data.edge_attr, x=data.x)
```
rusty1s, line 22: We do not really propagate information here. I would prefer the earlier naming of `update_edge` (or something similar).

torch_geometric/nn/conv/message_passing.py
```diff
@@ -115,8 +116,12 @@ def __init__(self, aggr: Optional[str] = "add",
         # Hooks:
         self._propagate_forward_pre_hooks = OrderedDict()
         self._propagate_forward_hooks = OrderedDict()
+        self._edges_propagate_forward_pre_hooks = OrderedDict()
```
rusty1s, line 119: Do you think we need two hooks here? In contrast to `propagate`, edge updates only call a single function.

rusty1s (commented):
I think this is great. I don't think we want to re-use `propagate` for this in order to prevent that our `propagate` functionality becomes too complicated (as it already is).

torch_geometric/nn/conv/message_passing.py
Padarn: Good point, this seems unneeded, I will change to one.

examples/edge_update.py
Padarn: I liked update_edge too, but I hate `update_edges` for the propagate and `edge_updates` for the message function.. which seemed a bit confusing.

Padarn:
Hey @rusty1s - I've made some changes and add some basic tests based on the ones that were already there for message passing. What else would you like to see before we get something merged: Happy to add more tests or examples.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -353,6 +356,59 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
 
         return out
 
+    def edge_update(self, edge_index: Adj, size: Size = None, **kwargs):
```
rusty1s, line 359: I think `size` can be dropped, as `size` is only useful for message aggregation.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -353,6 +356,59 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
...
+                for a in decomp_args
+            }
+            decomp_out = []
+
+        for i in range(decomposed_layers):
+            if decomposed_layers > 1:
+                for arg in decomp_args:
+                    kwargs[arg] = decomp_kwargs[arg][i]
+
+            coll_dict = self.__collect__(self.__user_args__, edge_index,
```
rusty1s, line 394: I think this is problematic, as `self.__collect__` will collect all arguments (even for those in `propagate` although we only need the arguments passed to `edge_message`. I think we need to disentangle `__user_args__` in arguments for `propagate` and arguments for `edge_message`.

rusty1s (commented):
Super, thanks for the update! Left some final comments before we can merge. One thing we need to add in a later PR is `torch.jit.script` support.

torch_geometric/nn/conv/message_passing.py
Padarn: makes sense I'll update

Padarn:
thanks @rusty1s - will update in the weekend (no time during the week usually)

torch_geometric/nn/conv/message_passing.py
Padarn: understood, I will figure out how to clearly do this

Padarn:
I've pushed changes for the discussed points.

But one thing I haven't wrapped my head around it how the jit stuff works, especially with the jinja templating: I'm a bit lost reading that code, any suggested reading to understand what the template is trying to achieve?

Padarn:
Hey @rusty1s. Anything pending on this one? I'm happy to work on the JIT support too but may need some pointers in the right direction

rusty1s:
Yes, let's merge this :) We can work on JIT support in a follow-up PR.

rusty1s (approved):

Padarn:
Cool thanks! I'll study the JIT implementation we currently have and see if I can get the gist of it.

rusty1s:
Yes, JIT is sadly troublesome to understand and to extend. If you feel like you are wasting too much time to understand, just let me know and I will try to integrate it by myself. Happy to answer any questions though :)

Padarn:
started a PR here https://github.com/pyg-team/pytorch_geometric/pull/3765 - I think I have it mostly working but its a bit messy and there are a few minor problems.

PR #3456, opened 2021-11-08
Change default dim_inner for GeneralMultiLayer

JiaxuanYou:


JiaxuanYou:
This allows `GeneralMultiLayer` to work even when ```dim_in = -1```

PR #3461, opened 2021-11-09
`mypy` and `pylint` GitHub Action setup

rusty1s:


.github/workflows/linting.yml
```diff
@@ -9,29 +9,58 @@ on:
...
+        with:
+          python-version: 3.9
+
+      - name: Install dependencies
+        run: |
+          pip install mypy
+
+      - name: Check package metadata
+        run: |
+          mypy -m torch_geometric
```
cthoyt, line 66: mypy needs some type stub packages installed to work properly, the `--install-types --non-interactive` flags in combination are the appropriate way to do this in a CI setting. Not sure if `--ignore-missing-imports` is necessary or not. Typically, I usually set up mypy to work on the source files itself instead of using `-m`, so I'm not sure what'll happen next

```suggestion
          mypy --install-types --non-interactive --ignore-missing-imports -m torch_geometric
```

cthoyt: btw are you guys celebrating Karneval all the way up in Dortmund? Kölle alaaf ;)

rusty1s: Thanks for digging into this. I'm still a `mypy` noob :(
I applied the suggested changes, let's see how it goes.

PS: We are not really celebrating Karneval here in Dortmund :D

cthoyt: Looks like it worked, but now the real difficult part of mypy begins. The way I applied this to some of my big projects like PyKEEN was to only apply it to sub modules incrementally. It took a long time, and probably will motivate some pretty major refactoring

rusty1s: Super, thanks a lot. Indeed, we have now arrived at the challenging part. Going one submodule at a time sounds like a good idea. It also seems that `mypy` cannot handle decorators such as `@torch.jit._overload`, so I'm not even sure if I can resolve all of its complaints.

torch_geometric/__init__.py
```diff
@@ -1,23 +1,23 @@
...
-from .seed import seed_everything
-from .debug import is_debug_enabled, debug, set_debug
-
 import torch_geometric.data
 import torch_geometric.loader
 import torch_geometric.transforms
 import torch_geometric.utils
 import torch_geometric.profile
 
+from .seed import seed_everything
```
cthoyt, line 10: you could consider using `isort` to reorganize imports automatically and add the `flake8-isort` plugin as an extra requirement for the flake8 action

cthoyt: I'll send a short PR demonstrating

PR #3463, opened 2021-11-09
doc fix for TransformerConv's beta coefficient

GYHHAHA:
Both [code](https://github.com/pyg-team/pytorch_geometric/blob/27e3e26ce73676f6d99377b4ea98cf62b2ac63aa/torch_geometric/nn/conv/transformer_conv.py#L187) and paper use `W_1 x_i` instead of `x_i`.

Inked2021-11-09_194117_LI


rusty1s:
Thanks!

PR #3474, opened 2021-11-11
Update quick start widget for PyTorch 1.10

kathyfan:
Also add CUDA 11.3 option.


rusty1s:
Thanks. I completely missed this one :)

kathyfan:
BTW I wasn't sure if we should also add a block like "'# PyTorch 1.10.* binaries do not support CUDA 10.1'" - Seems like maybe this is true if 1.9 didn't support it, but I wasn't able to verify this fact.

rusty1s:
No worries. You can check [here](https://data.pyg.org/whl/index.html) which PyTorch versions support which CUDA versions.

PR #3476, opened 2021-11-11
Update line_graph.py

ZIKEYUAN:
Convert the node features in the original graph into edge features in the line graph on the directed graph.

### DIFF
--- a/torch_geometric/transforms/line_graph.py
+++ b/torch_geometric/transforms/line_graph.py
@@ -39,6 +39,12 @@ class LineGraph(BaseTransform):
         (row, col), edge_attr = coalesce(edge_index, edge_attr, N, N)
 
         if self.force_directed or data.is_directed():
+# ==============change==================
+            print("It's a direct graph")
+            old_src, old_dst = data.edge_index #
+            line_graph_edge_attr = data.x[old_dst]#
+            (_,_),line_graph_edge_attr = coalesce(edge_index, line_graph_edge_attr, N, N)
+# ======================================
             i = torch.arange(row.size(0), dtype=torch.long, device=row.device)
 
             count = scatter_add(torch.ones_like(row), row, dim=0,
@@ -56,6 +62,10 @@ class LineGraph(BaseTransform):
             data.edge_index = torch.stack([row, col], dim=0)
             data.x = data.edge_attr
             data.num_nodes = edge_index.size(1)
+# ============change=====================
+            src, dst = data.edge_index
+            data.edge_attr = line_graph_edge_attr[src]
+# =======================================
 
         else:
             # Compute node indices.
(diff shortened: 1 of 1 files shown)


torch_geometric/transforms/line_graph.py
```diff
@@ -39,6 +39,12 @@ def __call__(self, data: Data) -> Data:
         (row, col), edge_attr = coalesce(edge_index, edge_attr, N, N)
 
         if self.force_directed or data.is_directed():
+# ==============change==================
+            print("It's a direct graph")
+            old_src, old_dst = data.edge_index #
+            line_graph_edge_attr = data.x[old_dst]#
+            (_,_),line_graph_edge_attr = coalesce(edge_index, line_graph_edge_attr, N, N)
+# ======================================
```
wsad1, line 47: The comments indicating parts of the code that changed can be remove, as github highlights changes automatically.
```suggestion
            old_src, old_dst = data.edge_index
            line_graph_edge_attr = data.x[old_dst]
            (_,_),line_graph_edge_attr = coalesce(edge_index, line_graph_edge_attr, N, N)
```

torch_geometric/transforms/line_graph.py
```diff
@@ -56,6 +62,10 @@ def __call__(self, data: Data) -> Data:
             data.edge_index = torch.stack([row, col], dim=0)
             data.x = data.edge_attr
             data.num_nodes = edge_index.size(1)
+# ============change=====================
+            src, dst = data.edge_index
+            data.edge_attr = line_graph_edge_attr[src]
```
wsad1, line 67: Will this work as expected?
lets say the graph originally had E edges and the line graph has E* edges. Then the shape of `line_graph_edge_attr` is (E,-1), and if E* is greater than E then wouldn't this throw an index error.

torch_geometric/transforms/line_graph.py
```diff
@@ -89,6 +99,7 @@ def generate_grid(x):
                 data.x = scatter_add(edge_attr, i, dim=0, dim_size=N)
             data.edge_index = joints
             data.num_nodes = edge_index.size(1) // 2
+            data.edge_attr = None
 
-        data.edge_attr = None
+#         data.edge_attr = None
```
wsad1, line 104:
```suggestion
```

wsad1 (changes requested):
Thanks for the PR. Here is my initial review.

torch_geometric/transforms/line_graph.py
ZIKEYUAN: I don't think it will happen, because the data.edge_index here has been converted to the edge_index in the linegraph, so it can run normally.

rusty1s: Can we add a test to `test/transforms/test_line_graph.py` to ensure that?

Issue #3349, opened 2021-10-18
Custom evaluation metric module


RemyLau:
I'm wondering would it be possible to implement a module for **custom evaluation metrics**. There are a few reasons why this is desirable. The most obvious reason is **performance**. Currently, all the evaluations are from ``scikit-learn``, rendering _extremely_ low GPU usage (I can provide measurements if needed). In the past, I've written custom metric functions in a more engineered way to massively speed up the computation.

On a separate but related note, would it be possible to have the option to disable evaluation logging during training? Specifically, disabling this block. https://github.com/pyg-team/pytorch_geometric/blob/ec952b48e2a80136a253eb1e5399cb55cac91e67/torch_geometric/graphgym/train.py#L23-L27

Having this option would further increase the GPU usage by disabling unnecessary transfer to the CPU (correct me if I'm wrong). Also, the ``ogbn-proteins`` example script has this desired behavior of being able to choose the evaluation logging step. https://github.com/snap-stanford/ogb/blob/9d7a3f5e89640938f93b52388cfe259ae8fe3855/examples/nodeproppred/proteins/gnn.py#L160

JiaxuanYou:
Thanks for your suggestion. In my daily use cases, this logger update takes negligible time. Could you describe your use case, where this could be a bottleneck?
Also, some metrics, such as AUC ROC, it not straightforward to implement on GPU, while `scikit-learn` provides convenient support.
I'm happy to hear your suggestions. Also, PR for more evaluation metrics is welcomed!

PR #3494, opened 2021-11-14
Add custom metrics and logging options

RemyLau:
Hi, I've been using GraphGym in PyG for a while now, mainly for large-scale evaluation on many datasets. Here are a few things that I think would be really helpful to have, which are the main changes in this PR.
1. **Option to disable model checkpoint**
2. **Option to disable evaluation during training epochs**
3. **Option for using customized metrics**

### DIFF
--- a/test/graphgym/example_node.yml
+++ b/test/graphgym/example_node.yml
@@ -14,7 +14,7 @@ train:
   eval_period: 2
   ckpt_period: 100
   enable_ckpt: false
-  skip_train_eval: True
+  skip_train_eval: true
   sampler: full_batch
 model:
   type: gnn
--- a/test/graphgym/example_node.yml
+++ b/test/graphgym/example_node.yml
@@ -11,9 +11,10 @@ dataset:
   edge_encoder_name: Bond
 train:
   batch_size: 128
-  eval_period: 1
+  eval_period: 2
   ckpt_period: 100
   enable_ckpt: False
+  skip_train_eval: True
   sampler: full_batch
 model:
   type: gnn
@@ -35,4 +36,4 @@ gnn:
 optim:
   optimizer: adam
   base_lr: 0.01
-  max_epoch: 3
+  max_epoch: 6
--- a/test/graphgym/example_node.yml
+++ b/test/graphgym/example_node.yml
@@ -13,6 +13,7 @@ train:
   batch_size: 128
   eval_period: 1
   ckpt_period: 100
+  enable_ckpt: False
   sampler: full_batch
 model:
   type: gnn
--- a/test/graphgym/example_node.yml
+++ b/test/graphgym/example_node.yml
@@ -13,7 +13,7 @@ train:
   batch_size: 128
   eval_period: 2
   ckpt_period: 100
-  enable_ckpt: False
+  enable_ckpt: false
   skip_train_eval: True
   sampler: full_batch
 model:
--- a/test/graphgym/test_graphgym.py
+++ b/test/graphgym/test_graphgym.py
@@ -1,10 +1,10 @@
-import itertools
 import os.path as osp
 import random
 import shutil
 import sys
 from collections import namedtuple
 
+import pytest
 import torch
 
 from torch_geometric import seed_everything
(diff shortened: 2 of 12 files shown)


torch_geometric/graphgym/config.py
```diff
@@ -229,9 +229,15 @@ def set_cfg(cfg):
     # Evaluate model on test data every eval period epochs
     cfg.train.eval_period = 10
 
+    # Option to skip training epoch evaluation
```
JiaxuanYou, line 232: I think adding a configuration that selects the customized statistics could be helpful. Of course, we can also assume that all the registered customized statistics will be used. Please let me know if what do you think.

test/graphgym/example_node.yml
```diff
@@ -11,8 +11,10 @@ dataset:
   edge_encoder_name: Bond
 train:
   batch_size: 128
-  eval_period: 1
+  eval_period: 2
```
JiaxuanYou, line 14: If this is change is not needed, please keep the default configs unchanged, so that users can reproduce the examples in older versions of PyG GraphGym.

JiaxuanYou (commented):
Thanks for your contribution! Please see my comments attached

torch_geometric/graphgym/train.py
```diff
@@ -69,13 +70,14 @@ def train(loggers, loaders, model, optimizer, scheduler):
     split_names = ['val', 'test']
     for cur_epoch in range(start_epoch, cfg.optim.max_epoch):
         train_epoch(loggers[0], loaders[0], model, optimizer, scheduler)
-        loggers[0].write_epoch(cur_epoch)
+        if is_train_eval_epoch(cur_epoch):
```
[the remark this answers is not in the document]
RemyLau: Thanks for your reply! I've been using GraphGym to evaluate some large datasets (node-level tasks) including the OGBN ``arxiv`` and ``proteins``. Due to a large number of nodes in these graphs, the computation of auroc is very expensive and I've observed a severe GPU usage bottleneck (near 0% usage) because of this. Skipping the training epoch evaluation could effectively remedy this issue. Furthermore, I don't see the evaluations for the training epoch being used anywhere in the downstream analysis, so I thought it might not hurt to disable it in favor of improving the GPU usage.

It has also been suggested to look into the [torchmetric](https://torchmetrics.readthedocs.io/en/latest/) in https://github.com/pyg-team/pytorch_geometric/issues/3349#issuecomment-946354517, the metric refactoring in this PR should also enable this in a future PR.

test/graphgym/example_node.yml
RemyLau: This is modified to see if the the ``cfg.train.skip_train_eval`` option is functioning correctly. If ``cfg.train.eval_period`` is set to 1, there's no difference regardless of whether ``skip_train_eval`` is ``True`` or ``False``.

I could also create a new copy of this yml and keep the ``eval_peirod`` in the original yml to be 1 for the purpose of reproducing older examples. Let me know if you prefer this option.

torch_geometric/graphgym/config.py
RemyLau: Yeah, I think it's a good idea to specify which custom metric group to use.

In the [current setup](https://github.com/RemyLau/pytorch_geometric/blob/bed926723b5ef7a4e261522e5ba56e019c7e9917/torch_geometric/graphgym/logger.py#L163-L168), I tried to follow the GraphGym code design language for loading custom functions (the example I had in mind was the [``torch_geometric.graphgym.loadder.py``](https://github.com/RemyLau/pytorch_geometric/blob/bed926723b5ef7a4e261522e5ba56e019c7e9917/torch_geometric/graphgym/loader.py#L164-L168)):
1. Look through all relevant registered custom functions and return immediately if a custom function returns nontrivial results
2. If the above failed, use the default graphgym functions

Going back to the point of specifying which custom metric group to use, maybe one possible strategy could be:
1. Specify the desired custom metric group to use as ``cfg.custom_metric`` (set to ``'none'`` by default)
2. Instead of looking through all registered custom metric functions, find the one with the matched key of ``cfg.custom_metric``. And maybe print warning message if no matched key for ``cfg.custom_metric`` was found and ``cfg.custom_metric != 'none'`` and fall back to using the default implemented metrics.

Let me know what you think

torch_geometric/graphgym/train.py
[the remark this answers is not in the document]
JiaxuanYou: Thanks for explaining! This makes sense. And indeed, future PRs on `torchmetric` is welcomed!

test/graphgym/example_node.yml
JiaxuanYou: I see. This makes sense. I think having two examples may not be necessary, so you can go ahead with this change

torch_geometric/graphgym/config.py
JiaxuanYou: I think this API sounds great! You could further make `cfg.custom_metric` to be a list, since a user may want to use multiple customized metrics. If the list is empty, then no customized metric is used.

test/graphgym/test_graphgym.py
```diff
@@ -75,3 +76,51 @@ def test_run_single_graphgym():
 
     shutil.rmtree(cfg.out_dir)
     shutil.rmtree(cfg.dataset.dir)
+
+
+def trivial_metric(true, pred, task_type):
+    return {'trivial': 1}
+
+
+def test_run_trivial_graphgym():
```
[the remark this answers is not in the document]
RemyLau: Hi @JiaxuanYou I looked a little bit more into the ``update_stats`` function and I don't think it will call the custom metrics directly. In my current setup, which is the same as that in the original setup, the ``task_stats`` (i.e. the metrics) are updated during the function call of ``write_epoch`` using ``self._train`` and ``self._pred`` as input, both of which are set during ``update_stats``.

So ``update_stats``does not directly call ``trivial`` but sets the ``self._train`` and ``self._pred`` instead. Let me know if I'm understanding this wrong or if you want me to handle the custom metric function calls differently.

rusty1s:
@RemyLau @JiaxuanYou Any status updates on this PR? Is this blocked by anything or ready to get merged?

RemyLau:
Hi @rusty1s, thanks for bringing this up again! The most recent commit should resolve a major comment by @JiaxuanYou about selecting a group of custom metrics to be used, so it should be ready to be reviewed.

Because it's been a long time since the last commit, there are multiple conflicts with the main branch. Should I do a rebase and force push to update this PR?

rusty1s:
Yes, that would be great :)

RemyLau:
I've just rebased the branch back to master so all conflicts are fixed now

rusty1s:
Thank you @RemyLau.

@JiaxuanYou Can you finish your review? I'm happy to merge it afterwards.

JiaxuanYou:
Sorry for missing the thread. Yes, the new functionality of enabling a list of customized metrics looks good to me!

JiaxuanYou (approved):

RemyLau:
Great! Thanks a lot!

test/graphgym/test_graphgym.py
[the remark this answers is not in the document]
rusty1s: I think what @JiaxuanYou means is that we never actually test if `trivial_metric` is executed. To fix, we might want to add a counter to `trivial_metric`, similar to [here](https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/conv/test_message_passing.py#L195-L218).

RemyLau:
Hi @rusty1s, I've attempted to make the changes you suggested: 1) test the execution of ``trivial_metric`` using a counter, 2) combine the two test functions. Let me know if this looks good to you

rusty1s (approved):
Thank you! I modified the tests a little further to make use of `pytest.mark.parametrize`.

PR #3501, opened 2021-11-14
Updated GEDDataset documentation

kunind27:
Added an extensive documentation about the inner working of the GED-Dataset class of PyG including the pre-processing steps to aid practitioners who might want to use similar pre-processing. Detailed explanation for each step of pre-processing from raw graphs to PyG graph format is given.


rusty1s:
This is great. Thank you!

PR #3502, opened 2021-11-14
Automatically sort imports with `isort`

cthoyt:
This PR does the following:

1. Introduces a `tox.ini` configuration file for use with `tox`. This can be installed with `pip install tox` and run from the shell with `tox`. It is similar to `make`, but more oriented towards Python. Configurations can be added with `[testenv:<your name]` then run with `tox -e <your name>`. If `tox` is run by itself, everything in the `envlist` gets run.
2. Ports the commands to install and run `flake8` into the `tox.ini` and update the GitHub actions to use this
3. Adds a configuration for running `isort` in the `tox.ini` file. This automatically sorts imports in Python files to ensure consistency, especially from a wide variety of contributors. Right now the configuration is only demonstrated on `collate.py` but should be updated to read `isort torch_geometric/ test/ setup.py` to run on all source files before accepting this PR
4. Add the `flake8-isort` plugin to the flake8 dependencies to have CI check that isort was run properly. Note, this means that CI will indeed fail until running isort on the rest. I just didn't do this yet since it would make a _massive_ diff. 

TL;DR: streamlined linting config/installation, demonstrated sorting imports with `isort`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,16 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
+[tox]
+envlist =
+    lint
+
+[testenv:lint]
+deps =
+    isort
+skip_install = true
+commands =
+    isort torch_geometric/data/collate.py
+description = Run linters.
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@
 [tox]
 envlist =
     lint
+    flake8
 
 [testenv:lint]
 deps =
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -0,0 +1,25 @@
...
+envlist =
+    lint
+    flake8
+
+[testenv:lint]
+deps =
+    isort
+skip_install = true
+commands =
+    isort torch_geometric/data/collate.py
```
cthoyt, line 16:
```suggestion
    isort torch_geometric/ test/ setup.py
```

rusty1s:
Thank you! We added `isort` support via `pre-commit`, see [here](https://github.com/pyg-team/pytorch_geometric/blob/master/.pre-commit-config.yaml).

PR #3531, opened 2021-11-18
[WIP] `LightningLite` example

rusty1s:


examples/pytorch_lightning/lightning_lite.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Lite(LightningLite):
+    def run(self, path: str, num_workers: int = 0):
+        dataset = Flickr(path)
+        train_loader = NeighborLoader(dataset[0], num_neighbors=[10] * 2,
+                                      shuffle=True, batch_size=1024,
+                                      input_nodes=dataset[0].train_mask,
+                                      num_workers=num_workers,
+                                      persistent_workers=num_workers > 0)
+        train_loader = self.setup_dataloaders(train_loader)
```
JiaxuanYou, line 22: Should `loader`, `model` and ` optimizer` part of the attribute of `Lite`? I notice that no `__init__` is declared

awaelchli: @JiaxuanYou No, there is no such requirement from the side of Lite. This is up to the user :)

PR #3533, opened 2021-11-18
Make batch argument optionnal in voxel_grid / update GridSampling documentation 

QuanticDisaster:
This PR does the following :

- adds documentation on return nature of GridSampling
- removes the mandatory argument 'batch' for voxel_grid, defaulting it to None


rusty1s:
This is great. Thank you!

PR #3539, opened 2021-11-20
Zip fix

wsad1:
This PR makes the following updates to tests.

1. removed the usage of zip form test_hgt_conv.
2. in test_data and test_hetero_data, check if the lenght of stores match before comparing them.


rusty1s (approved):
Thank you for fixing. LGTM.

Issue #3565, opened 2021-11-25
PyG's DataLoader hangs when I use it in a subprocess created by the "fork" start method


Enolerobotti:
## 🐛 Bug
There are 2 identical completely independent functions which use torch_geometric.loader.DataLoader. If we run function 1 in the main process and then function 2 in the subprocess created by the "fork" start method the program hangs between [these](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/data/collate.py#L103-L104) 2 lines. Note, that for standard torch.utils.data.DataLoader all works well. If we use another start method all works well too. Also, the program works If we run only function 2 in the subprocess created by any start method (w/o calling the function 1 first).  And, the most interesting, the program runs well if we move tensors to the GPU.
I attach the example code for the broken case above. 

## To Reproduce

```python
import queue
import torch
from torch_geometric.loader import DataLoader
from torch_geometric.data import Data
from multiprocessing import Queue, get_context


def worker_function(_queue: Queue = None):
    data = [Data(x=torch.rand(size=(100, 100)), edge_index=torch.randint(100, size=(0, 100))) for _ in range(10)]
    dataloader = DataLoader(data, num_workers=0, batch_size=len(data)//2)
    dataloader_iter = enumerate(dataloader)
    idx, batch = next(dataloader_iter)
    if _queue is not None:
        _queue.put(idx)
    else:
        return idx


def completely_independent_function(_queue: Queue = None):
    # Actually, it is a copy of the function above
    data = [Data(x=torch.rand(size=(100, 100)), edge_index=torch.randint(100, size=(0, 100))) for _ in range(10)]
    dataloader = DataLoader(data, num_workers=0, batch_size=len(data)//2)
    dataloader_iter = enumerate(dataloader)
    idx, batch = next(dataloader_iter)
    if _queue is not None:
        _queue.put(idx)
    else:
        return idx


def main(context: str):
    ctx = get_context(context)
    q = ctx.Queue()
    p = ctx.Process(target=worker_function, args=(q,))
    p.start()
    result = q.get(timeout=10)
    p.join()
    return result


if __name__ == '__main__':
    print("Start")
    i = completely_independent_function()
    print("Finished completely_independent_function in the main process")
    for method in ['fork', 'forkserver', 'spawn']:
        try:
            r = main(method)
            assert i == r
            print(f"The '{method}' method succeed")
        except queue.Empty:
            print(f"The '{method}' hangs due to unknown reason")
```

[...]

PR #3566, opened 2021-11-25
Fix `DataLoader` in forks

rusty1s:


Enolerobotti (approved):
It fixes my [issue](https://github.com/pyg-team/pytorch_geometric/issues/3565). Good point! Thanks!

Enolerobotti (approved):

PR #3577, opened 2021-11-26
added dropout argument to han_conv.

wsad1:
In the HAN paper, dropout is applied on attention while training on various datasets. 
This PR adds the dropout argument to HanConv.


rusty1s (approved):
Super. Thank you!

PR #3581, opened 2021-11-27
Han example

wsad1:
Added a example for HanConv.

Side Note:
This example uses the IMDB dataset, the model has  a test  F1-score of around 50 when around 10% of the data is used for training as reported in the paper. But the script only reports accuracy.


examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+from torch_geometric.nn import HANConv
+import torch_geometric.transforms as T
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '../../data/IMDB')
+metapaths = [[('movie', 'actor'), ('actor', 'movie')],
+             [('movie', 'director'), ('director', 'movie')]]
+transform = T.AddMetaPaths(metapaths=metapaths, drop_orig_edges=True)
+dataset = IMDB(path, transform=transform)
+data = dataset[0]
+del data['director'], data['actor']
```
rusty1s, line 18: We think we should add this functionality to `AddMetaPaths` directly. If `drop_orig_edges=True`, there is really no need to maintain node types that are no longer connected by any edge type. WDYT?

examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+print(data)
+
+
+class han(nn.Module):
```
rusty1s, line 22:
```suggestion
class HAN(nn.Module):
```

examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, in_channels: Union[int, Dict[str, int]],
+                 out_channels: int, hidden_channels=128, heads=8):
+        super().__init__()
+        self.han_conv = HANConv(in_channels, hidden_channels, heads=heads,
```
rusty1s, line 26: Does the paper only make use of a single layer as well?

examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+                                dropout=0.6, metadata=data.metadata())
+        self.lin = nn.Linear(hidden_channels, out_channels)
+
+    def forward(self, x_dict, edge_index_dict):
+
```
rusty1s, line 31: nit: Remove line 31 :)

examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+    out = model(data.x_dict, data.edge_index_dict)
+    mask = data['movie'].train_mask
+    loss = F.cross_entropy(out[mask], data['movie'].y[mask])
+    loss.backward()
+    optimizer.step()
+    return float(loss)
+
+
+@torch.no_grad()
+def test() -> List:
```
rusty1s, line 59:
```suggestion
def test() -> List[float]:
```

examples/hetero/han_imdb.py
```diff
@@ -0,0 +1,89 @@
...
+          f'Val: {val_acc:.4f}, Test: {test_acc:.4f}')
+
+    if best_val_acc <= val_acc:
+        patience = start_patience
+        best_val_acc = val_acc
+    else:
+        patience -= 1
+
+    if patience <= 0:
+        print('Stopping training as Val accuracy did not improve '
```
rusty1s, line 87:
```suggestion
        print('Stopping training as validation accuracy did not improve '
```

rusty1s (commented):
This is great. Thank you! What's the model performance? :)

examples/hetero/han_imdb.py
wsad1: It looks like that, but the paper isn't entirely clear on that. 
Also the F1 score I am seeing is very similar to the one reported in the paper.

examples/hetero/han_imdb.py
wsad1: Yeah I think that's a good idea.
So if `drop_orig_edges` is True and there are node types types that are not connected by any edge type we just drop them inside the transform. 
I'll update the implementation and docs of AddMetaPaths this way then.

examples/hetero/han_imdb.py
rusty1s: Super :)

examples/hetero/han_imdb.py
rusty1s: Super, thank you! Let me know once this is ready to be merged :)

wsad1: I implemented the dropping of nodes in a slightly different.
Add a new argument `drop_unconnected_nodes`, which if set to true remove node types not connected by any edges type.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -73,15 +74,18 @@ class AddMetaPaths(BaseTransform):
             by a list of lists of
             :obj:`(src_node_type, rel_type, dst_node_type)` tuples.
         drop_orig_edges (bool, optional): If set to :obj:`True`, existing edge
-            types will be dropped. (default: :obj:`False`)
+            types will be dropped.(default: :obj:`False`)
```
rusty1s, line 77:
```suggestion
            types will be dropped. (default: :obj:`False`)
```

torch_geometric/transforms/add_metapaths.py
```diff
@@ -73,15 +74,18 @@ class AddMetaPaths(BaseTransform):
...
         keep_same_node_type (bool, optional): If set to :obj:`True`, existing
             edge types between the same node type are not dropped even in case
             :obj:`drop_orig_edges` is set to :obj:`True`.
             (default: :obj:`False`)
+        drop_unoconnected_nodes (bool, optional): If set to :obj:`True` drop
```
rusty1s, line 82: Is this a typo?

rusty1s (commented):
Super, thanks for adding this functionality.

rusty1s (approved):

torch_geometric/transforms/add_metapaths.py
wsad1: Yes. Thanks for catching that :).

PR #3586, opened 2021-11-28
Add stochastic_blockmodel_dataset

dongkwan-kim:
This PR contains the implementation of `StochasticBlockModelDataset` and `RandomPartitionGraphDataset`. These are both synthetic graphs generated by `utils.stochastic_blockmodel_graph` and node features are created by `sklearn.datasets.make_classification`. I fix some important arguments for `sklearn.datasets.make_classification` at lines 56 -- 59, and It would be great to discuss and decide which one is appropriate as the default setting.


torch_geometric/datasets/stochastic_blockmodel_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+                      'n_redundant': 0,
+                      'flip_y': 0.0,
+                      'shuffle': False}
+        self.x_kws.update(x_kws or {})
+
+        super().__init__(root, transform, pre_transform)
+        self.data, self.slices = torch.load(self.processed_paths[0])
+
+    @property
+    def name(self):
```
rusty1s, line 66: How fast is the generation of SBMs? Do we really need to save them or can we simply generate them on-the-fly?

torch_geometric/datasets/stochastic_blockmodel_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+                n_samples=sum(self.block_sizes),
+                n_features=self.num_channels,
+                n_classes=len(self.block_sizes),
+                weights=torch.tensor(self.block_sizes) / sum(self.block_sizes),
+                **self.x_kws,
+            )
+            x = torch.tensor(x, dtype=torch.float)
+        else:
+            x = None
+        y = torch.cat([torch.full((b,), i)
```
rusty1s, line 98: I think we can replace this with:
```python
torch.arange(num_blocks).repeat_interleave(self.block_sizes)
```

torch_geometric/datasets/stochastic_blockmodel_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+        return os.path.join(self.root, self.__class__.__name__,
+                            self.name, 'raw')
+
+    @property
+    def processed_dir(self):
+        return os.path.join(self.root, self.__class__.__name__,
+                            self.name, 'processed')
+
+    @property
+    def processed_file_names(self):
```
rusty1s, line 81:
```suggestion
    def processed_file_names(self):
        return f'{self.name}.pt'
```

torch_geometric/datasets/stochastic_blockmodel_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+        self.data, self.slices = torch.load(self.processed_paths[0])
+
+    @property
+    def name(self):
+        return f'sbm-{self.num_channels}-{self.block_sizes.tolist()}' \
+               f'-{self.edge_probs.tolist()}'
+
+    @property
+    def raw_dir(self):
+        return os.path.join(self.root, self.__class__.__name__,
```
rusty1s, line 72:
```suggestion
        return os.path.join(self.root, self.__class__.__name__, 'raw')
```

rusty1s (commented):
Thank you. This is super cool. I left come (nitpicky) comments :)

I think the default args for `make_classification` are reasonable as long as they are always customizable by the user.

torch_geometric/datasets/__init__.py
```diff
@@ -68,6 +68,8 @@
 from .omdb import OMDB
 from .polblogs import PolBlogs
 from .email_eu_core import EmailEUCore
+from .stochastic_blockmodel_dataset import StochasticBlockModelDataset
```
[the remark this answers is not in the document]
dongkwan-kim: I just found that this line exceeds 79 characters and violates E501. Is it okay to use separate lines, or just import all datasets in a single line?

torch_geometric/datasets/stochastic_blockmodel_dataset.py
dongkwan-kim: It does not take a long time if the size is similar to the benchmark we often use. However, since the generation process relies on the stochastic process, I think saving processed files is more appropriate than using them on the fly.

rusty1s (approved):
Thanks for updating :)

dongkwan-kim:
Thank you very much. Now it looks more awesome :)

PR #3592, opened 2021-11-29
bug fix reading MalNet Tiny

rampasek:
1. The PyG's MalNetTiny class was wrongly reading `*.edgelist` files, when only first digit from source node and last digit from the destination node numbers were parsed. 
2. In MalNet `*.edgelist` files some nodes are not used at all, i.e. they are isolated if we consider the graph to have `edge_index.max().item() + 1` nodes. I used [remove_isolated_nodes](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/utils/isolated.html#remove_isolated_nodes) to remove these isolated nodes as well as those that only have a self-loop.


rusty1s (approved):
Super! Thanks for spotting.

Issue #3601, opened 2021-11-30
NeighborLoader bug when using DDP on heterogeneous graphs via PyTorch-Lightning


pbielak:
## 🐛 Bug

I was trying to utilize multiple GPUs to train a GNN model on a custom heterogeneous graph. I used PyTorch-Lightning's implementation of the DDP mode to parallelize the computation across GPUs. However, it fails completely due to a bug in the `NeighborLoader` implementation.

### To Reproduce

Just run the example `https://github.com/pyg-team/pytorch_geometric/blob/master/examples/pytorch_lightning/relational_gnn.py` with `gpus=2`. It will yield following exception:

```
Original Traceback (most recent call last):
  File "(...)/.venv/lib/python3.6/site-packages/torch/utils/data/_utils/worker.py", line 287, in _worker_loop
    data = fetcher.fetch(index)
  File "(...)/.venv/lib/python3.6/site-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
    return self.collate_fn(data)
  File "(...)/.venv/lib/python3.6/site-packages/pytorch_lightning/utilities/auto_restart.py", line 474, in _capture_metadata_collate
    data = default_collate(samples)
  File "(...)/.venv/lib/python3.6/site-packages/torch_geometric/loader/neighbor_loader.py", line 219, in hetero_sample
    self.directed,
IndexError: _Map_base::at
```

### Expected behavior

It should work fine, as in the case of simple (not heterogenous) graphs.

### Explanation

I debugged this issue and found the root cause:
- PyTorch-Lightning automatically replaces the sampler with a Distributed Sampler implementation, so that each GPU receives its own subset of node indices

See: `https://github.com/PyTorchLightning/pytorch-lightning/blob/e94aff1c5bed2c616143dc694343c71a2be0b1bd/pytorch_lightning/trainer/data_loading.py#L125`

- the actual replacement of the sampler happens here: `https://github.com/PyTorchLightning/pytorch-lightning/blob/e94aff1c5bed2c616143dc694343c71a2be0b1bd/pytorch_lightning/utilities/data.py#L180` - it reconstructs the class kwargs and creates another instance of the DataLoader (here: `NeighborLoader`) while attaching a new sampler implementation

- Now why does that fail in the case of heterogeneous graphs? It because of the following lines:

```python
# See: https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/loader/neighbor_loader.py#L168

(...)
[...]

PR #3602, opened 2021-11-30
Fix NeighborLoader bug with PyTorch-Lightning

pbielak:
When using PyTorch-Lightning's DDP on heterogeneous graphs, the NeighborLoader
implementation fails due to re-creation of the loader object. While Lightning
replaces the sampler to a distributed version, the `num_neighbors` parameter
is extracted from the current object and passed to the new one. At this point
it is already in the desired format, so we must not convert the dict keys again
using `edge_type_to_str()` function. This bug fix checks whether the `edge_type`
is already a string and does not apply the conversion in such case.

Resolves issue: https://github.com/pyg-team/pytorch_geometric/issues/3601

### DIFF
--- a/test/loader/test_loader_utils.py
+++ b/test/loader/test_loader_utils.py
@@ -1,0 +1,16 @@
+# This file cannot be named `test_utils` as there exists already such file
+# in `test/profile/test_utils.py` which would yield a name conflict.
+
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
Thank you for digging into this :)

PR #3631, opened 2021-12-04
[WIP] Point Transformer S3DIS (adapt released code)

QuanticDisaster:
Hi, this PR intents to do the following : 

- add S3DIS example of PointTransformer
- which is used to adapt original code of the author which was released by [POSTECH-CVLab](https://github.com/POSTECH-CVLab/point-transformer) The original code attains the mIoU declared in the paper, and was also confirmed by a user who also kindly provided training the training log [on this comment](https://github.com/POSTECH-CVLab/point-transformer/issues/17#issuecomment-984594772)

This is primarily a draft (and a bit messy moreover), so I am trying to reproduce the results before updating the different modules present in the classification and part segmentation examples. I replaced imports by the custom PyG code that we previously developed in one commit to make it easier to compare changes.

As it is a bit long to train, especially on my unique GPU, I mainly look at the first mIoUs at the end of the first epochs
I believe I have updated PyG code according to the original code, but the first IoU on tests epochs are around 0.32 on my side while being at 0.48 on the training log. It could come from my reduced batch_size or another element independant from the model but would be glad if someone could double check the model implementation (dimensions and all)


QuanticDisaster:
A small update on this PR :

I believe the difference in accuracy comes from the dataset used which is built differently in the POSTECH repo as the built-in S3DIS of PyG went from 0.38 to 0.57mIoU in a full run with the PyG implementation. This is why I successfully replaced their model in their code with the custom implementation above with some corrections (there was 1 transformer block in each decoder/decoder more than it should have, exploding the number of parameters to train).

I think the architecture is more or less right, the first losses are on the same magnitude order as the train log provided by a user. However, I am faced with a problem of slowness, which comes mainly from the computation of edge index which takes several seconds for some of the batch (which size varies from 30k to 80k), compared to their implementation for which a forward pass on a batch takes only 0.5s on average with my GPU (against around 10s for mine). The original implementation already takes much time on a single GPU like 4-5 days, but with this it goes to 2000h 

Altough exact reproducibility is not the goal, to reproduce it fairly in this state we would need to lighten/change the dataset or find a miracle optimization or make use of their knn query, or trust the mIoU on the few first epochs.

rusty1s:
Can you give some information about their `knn` query implementation? It looks like it also operates in a purely brute-force fashsion, see [here](https://github.com/POSTECH-CVLab/point-transformer/blob/master/lib/pointops/src/knnquery/knnquery_cuda_kernel.cu#L65-L108), so I'm wondering why it's so much slower.

QuanticDisaster:
It is hard to go into details, but with this snippet added in the first transition down module : 

```python
start = time.time()
id_k_neighbor = knn(pos, pos[id_clusters], k=self.k)#,
                    #batch_x=batch, batch_y=sub_batch)
logger.info("compute k neighbors in  TransitionDown :" + str(time.time() - start) +  " for " + str(pos.shape[0]) + " and " + str(pos[id_clusters].shape[0]) + " points")

start = time.time()
m = pos.shape[0]
nsample = 16
pointops_cuda.knnquery_cuda(m, nsample, pos, pos[id_clusters], torch.IntTensor(1), torch.IntTensor(1), torch.cuda.IntTensor(m, nsample).zero_(), torch.cuda.FloatTensor(m, nsample).zero_())
logger.info("compute knn_cuda in TransitionDown :" + str(time.time() - start) +  " for " + str(pos.shape[0]) + " and " + str(pos[id_clusters].shape[0]) + " points")
```

I obtain this result for the same batch of points : 
```python
[2021-12-13 10:33:53,750 INFO pointtransformer_seg.py line 270 258822] compute k neighbors in  TransitionDown :5.033841133117676 for 57157 and 14290 points
[2021-12-13 10:33:53,750 INFO pointtransformer_seg.py line 278 258822] compute knn_cuda in TransitionDown :7.2479248046875e-05 for 57157 and 14290 points
```

The difference is quite large. Maybe there is something interfering with the knn of PyG in their repo as if I do it simply in a console, the result is much better (0.05s for 57k and 14k points), but I used torch.rand for my console test so the uniform distribution makes it quite biaised, I will try to extract the tensor of the batch to do it independently

rusty1s:
Ah, I got it. The `pytorch-cluster` k-NN implementation aims to be fast in a mini-batch scenario (at the compromise of being super slow for single examples). The `pointops` library implements the single example k-NN call efficiently, but does not have a batch implementation (at least it looks like that).

We should probably integrate a similar routine for operating on single examples as well, and then call the respective function in case no `batch` tensor is given. WDYT? Are you interested in contributing?

QuanticDisaster:
It does have a batch implementation if I am not mistaken, as the original code uses multiple batch, it is implemented through the offset variable which is a list corresponding to the index where the batch  changes like [1,254,567] for a tensor of 567 rows with 2 batches, you can also see it in the query and group function here https://github.com/POSTECH-CVLab/point-transformer/blob/master/lib/pointops/functions/pointops.py 
So what could be the main bottleneck on PyG cluster a they both seem to implement batching ?

I could try something if we dont find any simpler mean but it would take me time as I am only lightly  acquainted with C++ and even  less with cuda programming, so i would prefer to leave the idea for later on my side

rusty1s:
Ah, thanks. Let me look into that tomorrow. I will keep you posted.

QuanticDisaster:
Thanks for your help. To make it easier, here is an example of a real call argument that is done on the queryandgroup function :
nsample : 16 is the number of neighbors
xyz is the pos tensor  (11753x3 for instance)
new_xyz is the sub_points tensor : 2938x3, same as the feat tensor : 2938x3
idx : none
offset : list containing the batch assignation of xyz : ([ 1533, 11753], dtype=torch.int32). There are 2 batches here for instance
new_offset : list containing the batch of new_xyz : tensor([ 383, 2938], dtype=torch.int32)

rusty1s:
Hi @QuanticDisaster,

I updated the `knn_cuda` and `radius_cuda` call to now parallelize correctly across each data point rather than each example :) This should bring speed close to the `pointops` library. You can check it out by installing from master:
```
pip install git+https://github.com/rusty1s/pytorch_cluster.git
```

rusty1s:
Interesting. The code is nearly identical now, not sure what explains the difference in runtimes. Are you sure you are benchmarking correctly? Keep in mind that you need to run `torch.cuda.synchronize()` every time before you start and end time measurements. This might also explain our other issues.

QuanticDisaster:
Hi @rusty1s 
I resumed after some weeks away, so to answer your question, I was not benchmarking correctly indeed and it answered my additional question. However there is still a factor 10 difference in speed between PyG knn and their query and group whole function.

If you are interested, here is the snippet of code I integrated in the forward function of their TransitionDown module to do the comparison

```python
torch.cuda.synchronize()
start = time.time()
x = pointops.queryandgroup(self.nsample, p, n_p, x, None, o, n_o, use_xyz=True)  # (m, 3+c, nsample)
torch.cuda.synchronize()
end = time.time()
logger.info("compute query and group in  TransitionDown :" + str(end - start) +  " for " + str(p.shape[0]) + " and " + str(n_p.shape[0]) + " points")


batch = torch.ones(p.shape[0], dtype=torch.long).cuda()
sub_batch = torch.ones(n_p.shape[0], dtype=torch.long).cuda()
torch.cuda.synchronize()
start = time.time()
knn(p, n_p, k=self.nsample,  batch_x=batch, batch_y=sub_batch)
torch.cuda.synchronize()
end = time.time()
logger.info("compute knn in  TransitionDown :" + str(end - start) +  " for " + str(p.shape[0]) + " and " + str(n_p.shape[0]) + " points")
```

and the result 

```
compute query and group in  TransitionDown :	0.0006411075592041016 for 1533 and 383 points
compute knn in  TransitionDown :	        0.006720781326293945 for 1533 and 383 points

compute query and group in  TransitionDown :	0.0004963874816894531 for 383 and 95 points
compute knn in  TransitionDown :		0.0017185211181640625 for 383 and 95 points

compute query and group in  TransitionDown :	0.0010342597961425781 for 95 and 23 points
compute knn in  TransitionDown :		0.0008969306945800781 for 95 and 23 points

compute query and group in  TransitionDown :	0.016150474548339844 for 46824 and 11706 points
compute knn in  TransitionDown :		0.10952234268188477 for 46824 and 11706 points

compute query and group in  TransitionDown :	0.0023033618927001953 for 11706 and 2926 points
compute knn in  TransitionDown :		0.015674352645874023 for 11706 and 2926 points

compute query and group in  TransitionDown :	0.0008792877197265625 for 2926 and 731 points
compute knn in  TransitionDown :		0.005333662033081055 for 2926 and 731 points

compute query and group in  TransitionDown :	0.00048279762268066406 for 731 and 182 points
compute knn in  TransitionDown :		0.0020308494567871094 for 731 and 182 points
```

rusty1s:
Thank you. This is interesting as the `knn` code is nearly identical now. Will try to benchmark it as well.

QuanticDisaster:
Hello, I am opening this PR to review as I believe that I am done with adapting the original code. 

As for the results, although I did not perform a full training again on it due to hardware limitations, I manually used the pretrained weights from a user and manually mapped all original layers to their equivalent in this implementation. By replacing the model in the original repo (so with the same test dataset and transformations), I was able to reach the same results on the small validation set.

As it goes, the classification reaches 0.89 but still oscillates a bit at the end of training. The original code for classification has yet to be published.
Part segmentation quickly reaches and stabilizes at 0.825
Finally, S3DIS ends around 0.565. However, the S3DIS dataset built-in PyG seems quite different from the one used in the original repo hence differences in results.
I may update the learning rates and optimizer to the original ones used if it shows a great improvement

The original code is available here https://github.com/POSTECH-CVLab/point-transformer/blob/master/model/pointtransformer/pointtransformer_seg.py

On a side note : 

I see there is a conflict between files but I do not understand why it only triggers small parts of 2 files amongst all the modifications.

@rusty1s I believe the binaries with the fast knn implementation are not published yet on linux as you have to install them manually through the pip command you gave me and neither on windows, on which compiling them didn't work for me. Would it be possible to add them ?

As for the hardware limitations, my implementation is also more memory consuming than the original one reducing the number of batches possible at the same time. 782Mb vs 462 Mb for a forward pass, tested with fil memory profiler https://pythonspeed.com/fil/ which pointed to the message function of the convolution as the bottleneck in my case.

rusty1s:
Thanks @QuanticDisaster. Yes, we will release a new version of `torch-cluster` once PyTorch 1.11 lands. Any chance you can explain yourself the difference in memory consumption? It looks like the official implementation needs to materialize messages as well, right?

rusty1s:
[`torch-cluster==1.6.0`](https://github.com/rusty1s/pytorch_cluster/releases/tag/1.6.0) finally released :)

CCInc:
FYI @QuanticDisaster , I'm working to integrate point transformer into tp3d using your PointTransformerConv module. I'll send you the code once I'm done so you can compare.

QuanticDisaster:
Hi @CCInc Thx for the information, that will allow to double check! Looking forward to see it work

CCInc:
FYI, my point transformer implementation achieved higher score on the s3dis dataset than minkowski and kpconv. It's still running (it's significantly slower, by an order of magnitude) but I will link the wandb projects when they are finished.

BTW, have you seen this project @QuanticDisaster ? I would be happy to help implement it with you if you're interested. It seems that it's faster and higher performing than point transformer. https://github.com/dvlab-research/Stratified-Transformer

CCInc:
I realized that my implementation is running on the slow version of torch-cluster, I'll try upgrading to the latest pyg & torchcluster and see how much faster it is.

rusty1s:
Any benchmarking highly appreciated :) Thank you!

QuanticDisaster:
Hi @CCInc
I am glad to see your runs give positive results, keep us informed on the benchmark results! Btw I don't see any on the tp3d repo, I think it would be a great addition as I supposed your models are all trained/tested.

I didn't know about this stratified transformer project, it seems interesting indeed. I will have a look but probably nothing more at the moment

I personally used this PointTransformer model on some datasets, mostly with very small training sets (between 1M and 3M points with around 4 to 8 classes) and got acceptable results on the condition of removing the concatenation of positions in the features (or replacing them by a constant dummy feature if none), as it seems it was harmful in most of my cases.

PR #3639, opened 2021-12-06
Bug Fix: Order of transforms in label_prop example

Riyer01:
This example doesn't work properly as currently written. The correct way is to reverse the order of transformations, such that we first convert the graph to undirected, then get the sparse representation.

Empirically, try running this file with the transformations listed both ways for 50 iterations, and it becomes clear from the results that the original implementation is incorrect. 

Submitted for extra credit in CS224W


rusty1s (approved):
You are right. Thanks for spotting!

PR #3646, opened 2021-12-07
Fix key output of networkx edge getter:

max-zipfl-fzi:
Removed the key output, which resulted in a three-dimensional tensor, when using MultiDiGraphs.
Three-dimensional edge tuples could not be converted into the required [2,-1] COO format.

Example:
Right:
` print(list(G.edges(keys=False)))`
 [(2, 1), (5, 4), (5, 4), (5, 4), (6, 1), (7, 1)]

Wrong:
`print(list(G.edges))`
[(2, 1, 0), (5, 4, 0), (5, 4, 1), (5, 4, 2), (6, 1, 0), (7, 1, 0)]

Wrong:
`print(torch.LongTensor(list(G.edges)))`
 tensor([[2, 1, 0],
        [5, 4, 0],
        [5, 4, 1],
        [5, 4, 2],
        [6, 1, 0],
        [7, 1, 0]])
Right:
`print(torch.LongTensor(list(G.edges(keys=False))).t().contiguous())`
 tensor([[2, 5, 5, 5, 6, 7],
        [1, 4, 4, 4, 1, 1]])`

### DIFF
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -133,11 +133,12 @@ def from_networkx(G, group_node_attrs: Optional[Union[List[str], all]] = None,
     G = nx.convert_node_labels_to_integers(G)
     G = G.to_directed() if not nx.is_directed(G) else G
 
-    if type(G) is nx.MultiGraph or type(G) is nx.MultiDiGraph:
-        edge_index = torch.LongTensor(
-            list(G.edges(keys=False))).t().contiguous()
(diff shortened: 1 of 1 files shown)


rusty1s (approved):
Super :)

max-zipfl-fzi:
I forget to check non multi-graphs:
Will add a check, whether a key should be removed or if it is not present in the first place

rusty1s:
Super. Thank you :)

PR #3684, opened 2021-12-11
`@register_*` decorator in GraphGym

rusty1s:
Adds decorator functionality to register GraphGym modules:
```python
@register_layer('linear')
class Linear(torch.nn.Module):
    pass
```


JiaxuanYou (approved):
This makes the implementation a lot cleaner. Thanks!

PR #3665, opened 2021-12-09
Add LightGCN layer and model

KathyFeiyang:
This pull request adds implementation of LightGCN neighborhood aggregation layer, `LightGCNConv`, and LightGCN model, `LightGCN`, as proposed in "LightGCN: Simplifying and Powering Graph Convolution Network for Recommendation" [[paper](https://arxiv.org/abs/2002.02126)] [[author code](https://github.com/gusye1234/LightGCN-PyTorch)]. This pull request is submitted as part of Stanford [CS224W: Machine Learning with Graphs](http://web.stanford.edu/class/cs224w/), Autumn 2021-2022.

PR #3685, opened 2021-12-11
Implementation of LightGCN Model + Convolutional Layer + Extensive Tests

LukasHaas:
Efficient implementation of the LightGCN model and Light Graph Convolution (LGC) operator from the paper [LightGCN: Simplifying and Powering Graph Convolution Network for Recommendation](https://arxiv.org/abs/2002.02126) including test cases.

The model offers two different prediction heads for ranking and link prediction tasks as well as a matching loss function which can be used depending on the objective:

- Bayesian Personalized Ranking loss for ranking objective (needed to be implemented from scratch)
- Binary Cross-Entropy loss for link prediction objective.

Test cases for the model and convolutional layer have been added.


rusty1s:
Thank you. We also received https://github.com/pyg-team/pytorch_geometric/pull/3665. Is there a possibility to collaborate?

LukasHaas:
Yes, this is seems like a good idea.

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def __repr__(self):
+        return (f'{self.__class__.__name__}('
+                f'num_layers={self.num_layers}, '
+                f'num_nodes={self.num_nodes}, '
+                f'embedding_dim={self.embedding_dim}, '
+                f'objective=\"{self.objective}\")')
+
+
+class BPRLoss(_Loss):
```
qcw171717, line 249: Would it be easier to move this as a bpr_loss method inside the model class?

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        if weights is not None:
+            assert (weights.shape == self.embeddings.shape), \
+                'Pre-computed embeddings weights must match shape of \
+                embedding layer.'
+            self.embeddings.weight.data.copy_(weights)
+
+        out = self._propagate_embeddings(edge_index)
+
+        nodes_fst = torch.index_select(out, 0, edge_label_index[0, :].long())
```
qcw171717, line 128: In the case of bipartite graph where all edges are edge_label_index, it seems that if we set edge_label_index=edge_index when no edge_label_index is passed in, we can achieve by default the behaviour of the original LightGCN implementation - please feel free to correct me here if I misunderstood the implementation. Perhaps adding a default value for edge_label_index is more intuitive for users who are familiar with the original paper but not so much with the feature added by this PR.

torch_geometric/nn/models/lightgcn.py
LukasHaas: Agree it might be easier but the advantage of a separate class it that it can inherit from PyTorch‘s general _Loss class and could in the future be replaced by a proper PyTorch implementation.

torch_geometric/nn/models/lightgcn.py
LukasHaas: Bipartite default option is implemented in latest commit :) Thank you.

torch_geometric/nn/conv/lg_conv.py
```diff
@@ -0,0 +1,48 @@
...
+
+from torch import Tensor, ones
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.utils import degree
+
+
+class LGConv(MessagePassing):
+    r"""The Light Graph Convolution (LGC) operator from the `"LightGCN:
+    Simplifying and Powering Graph Convolution Network for Recommendation"
+    <https://arxiv.org/abs/2002.02126>`_ paper.
```
KathyFeiyang, line 11: Since this implementation adds additional functionalities on top of the vanilla LightGCN, I think it's worth mentioning what those functionalities are and how to use them here (eg. switching normalization on/off), so that the users are aware.

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,311 @@
...
+        out = torch.sum(nodes_fst * nodes_sec, dim=-1)
+
+        if self.objective == LightGCN.objectives[1]:
+            out = torch.sigmoid(out)
+
+        return out
+
+    def predict(self, edge_index: Adj, edge_label_index: Adj = None,
+                prob: bool = False) -> Tensor:
+        """Predict links between nodes specified in :obj:`edge_label_index`.
```
KathyFeiyang, line 144: nit: Since this method works only for the link prediction objective, I think it's worth specifying so here, in addition to the assertion. Perhaps the method name can become `predict_link`.

torch_geometric/nn/conv/lg_conv.py
LukasHaas: Agree :) Would you mind editing the documentation?

torch_geometric/nn/models/lightgcn.py
LukasHaas: This is a good idea! Would you mind changing it with the documentation?

torch_geometric/nn/conv/lg_conv.py
KathyFeiyang: Done :)

torch_geometric/nn/models/lightgcn.py
KathyFeiyang: Done :)

LukasHaas:
@rusty1s All changes have been implemented. It looks like we are ready to merge now :)

torch_geometric/nn/conv/lg_conv.py
```diff
@@ -0,0 +1,49 @@
...
+    .. math::
+        \mathbf{x}^{\prime}_i = \sum_{j \in \mathcal{N}(i)}
+        \frac{1}{\sqrt{\deg(i)}\sqrt{\deg(j)}}
+        \mathbf{x}_j
+
+    Args:
+        aggr (string, optional): The aggregation scheme to use.
+            (:obj:`"add"`, :obj:`"mean"`, :obj:`"max"`).
+            (default: :obj:`"add"`)
+        normalize (bool, optional): If set to :obj:`True`, output features
```
rusty1s, line 23:
```suggestion
        normalize (bool, optional): If set to :obj:`False`, output features will not
```

torch_geometric/nn/models/__init__.py
```diff
@@ -48,6 +49,7 @@
     'AttentiveFP',
     'RECT_L',
     'LINKX',
+    'LightGCN'
```
rusty1s, line 52:
```suggestion
    'LightGCN',
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+    <https://arxiv.org/abs/2002.02126>`_ paper.
+
+    LightGCN learns embeddings by linearly propagating them on the underlying
+    graph, and uses the weighted sum of the embeddings learned at all layers as
+    the final embedding as follows:
+
+    .. math::
+        \textbf{x}_i = \sum_{l=0}^{L} \alpha_l \textbf{x}^{l}_i
+
+    Each layer's embeddings are computed as:
```
rusty1s, line 24:
```suggestion
    where ach layer's embedding is computed as:
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+    .. note::
+
+        This implementation works not only on bipartite graphs but on any
+        kind of connected graph. Embeddings are propagated according to the
+        graph connectivity specified by :obj:`edge_index` while rankings or
+        probabilities are only computed (the forward pass) according to the
+        edges specified by :obj:`edge_label_index`.
+
+    Args:
+        num_layers (int): The number of Light Graph Convolution
```
rusty1s, line 45:
```suggestion
        num_layers (int): The number of :class:`~torch_geometric.nn.conv.LGConv` layers
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+            (:obj:`LGConv`) layers.
+        num_nodes (int): The number of nodes in the graph.
+        embedding_dim (int, optional): The dimensionality of the node
+            embeddings (default: 32).
+        objective (str, optional): The objective of the model. The original
```
rusty1s, line 50: Shall we align this with `GAE` (https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/autoencoder.py#L26-L28), e.g.:
```suggestion
        sigmoid (bool, optional): ...
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+            (default: :obj:`"ranking"`).
+        lambda_reg (int, optional): The :math:`L_2` regularization strength of
+            the Bayesian Personalized Ranking (BPR) loss for the "ranking"
+            objective (default: 1e-4).
+        alpha (float, optional): The vector specifying the layer weights in the
+            final embeddings. If set to :obj:`None`, the original paper's
+            uniform initialization of :obj:`(1 / num_layers + 1)` is used
+            (default: :obj:`None`).
+        **kwargs (optional): Additional arguments of the underlying
+            :class:`torch_geometric.nn.conv.MessagePassing` layers.
```
rusty1s, line 64:
```suggestion
            :class:`torch_geometric.nn.conv.LGConv` layers.
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+
+    objectives = ['ranking', 'link_prediction']
+
+    def __init__(self, num_layers: int, num_nodes: int,
+                 embedding_dim: int = 32, objective: str = 'ranking',
+                 lambda_reg: float = 1e-4, alpha: Tensor = None, **kwargs):
+        super().__init__(**kwargs)
```
rusty1s, line 72:
```suggestion
        super().__init__()
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        self.num_nodes = num_nodes
+        self.embedding_dim = embedding_dim
+        self.embeddings = Embedding(num_nodes, embedding_dim)
+        self.objective = objective
+        self.lambda_reg = lambda_reg
+        self.alpha = alpha
+        self.convs = ModuleList()
+
+        for _ in range(num_layers):
+            self.convs.append(LGConv())
```
rusty1s, line 97:
```suggestion
            self.convs.append(LGConv(**kwargs))
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+
+    objectives = ['ranking', 'link_prediction']
+
+    def __init__(self, num_layers: int, num_nodes: int,
+                 embedding_dim: int = 32, objective: str = 'ranking',
+                 lambda_reg: float = 1e-4, alpha: Tensor = None, **kwargs):
+        super().__init__(**kwargs)
+
+        assert (num_layers >= 1), 'Number of layers must be larger than 0.'
```
rusty1s, line 74: I think I would remove those assertions. They are pretty basic :)

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        objective (str, optional): The objective of the model. The original
+            paper implements the :obj:`"ranking"` objective with a simple dot
+            product between embeddings as a prediction head. The option
+            :obj:`"link_prediction"` passes the ranking result through a
+            sigmoid layer (:obj:`"ranking"`, :obj:`"link_prediction"`)
+            (default: :obj:`"ranking"`).
+        lambda_reg (int, optional): The :math:`L_2` regularization strength of
+            the Bayesian Personalized Ranking (BPR) loss for the "ranking"
+            objective (default: 1e-4).
+        alpha (float, optional): The vector specifying the layer weights in the
```
rusty1s, line 59:
```suggestion
        alpha (Tensor, optional): The vector specifying the layer weights in the
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        super().__init__(**kwargs)
+
+        assert (num_layers >= 1), 'Number of layers must be larger than 0.'
+        assert (num_nodes >= 1), 'Number of nodes must be larger than 0.'
+        assert (embedding_dim >= 1), 'Embedding dimension must be larger ' \
+            'than 0.'
+        assert (objective in LightGCN.objectives), 'Objective must be one ' \
+            f'of {LightGCN.objectives}'
+
+        if alpha is not None:
```
rusty1s, line 81:
```suggestion
        if isinstance(alpha, Tensor):
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+            assert (alpha.size(0) == (num_layers + 1)), 'alpha must be of ' \
```
rusty1s, line 82: We need to wrap `alpha` into a `Buffer` so that it gets correctly moved to GPU upon `model.cuda()`.

torch_geometric/nn/conv/lg_conv.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    propagate_type = {'x': Tensor, 'norm': Tensor}
+
+    def __init__(self, aggr: str = 'add', normalize: bool = True, **kwargs):
+        super().__init__(aggr=aggr, **kwargs)
+        self.normalize = normalize
+
+    def forward(self, x: Tensor, edge_index: Adj, size: Size = None) -> Tensor:
+        """"""
+        norm = ones(x.size(0))
```
rusty1s, line 38: We may want to use `torch_geoemtric.nn.conv.gcn_conv.gcn_norm(...)` for this :)

torch_geometric/nn/conv/lg_conv.py
```diff
@@ -0,0 +1,49 @@
...
+        **kwargs (optional): Additional arguments of
+            :class:`torch_geometric.nn.conv.MessagePassing`.
+    """
+    propagate_type = {'x': Tensor, 'norm': Tensor}
+
+    def __init__(self, aggr: str = 'add', normalize: bool = True, **kwargs):
+        super().__init__(aggr=aggr, **kwargs)
+        self.normalize = normalize
+
+    def forward(self, x: Tensor, edge_index: Adj, size: Size = None) -> Tensor:
```
rusty1s, line 36:
```suggestion
    def forward(self, x: Tensor, edge_index: Adj) -> Tensor:
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        """Computes rankings or link probabilities for pairs of nodes.
+
+        Args:
+            edge_index (Adj): Edge tensor specifying the connectivity of the
+                graph.
+            edge_label_index (Adj, optional): Edge tensor specifying the start
+                and end nodes for which to compute rankings or proabilities.
+                If :obj:`edge_label_index` is set to :obj:`None`, all edges in
+                :obj:`edge_index` will be used (default: :obj:`None`).
+            weights (Tensor, optional): Pre-computed embeddings or node
```
rusty1s, line 118: I think this is pretty unusual and I prefer to remove this functionality. If the user wants to modify the initial embeddings, the user should do it directly by accessing `model.embedding`.

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+
+        return self._propagate_embeddings(edge_index)[indices]
+
+    def _propagate_embeddings(self, edge_index: Adj) -> Tensor:
+        x = self.embeddings.weight
+        out = x * self.alpha[0]
+
+        for i in range(self.num_layers):
+            x = self.convs[i](x, edge_index)
+            out = torch.add(out, x * self.alpha[i + 1])
```
rusty1s, line 251:
```suggestion
            out = out + x * self.alpha[i + 1])
```

torch_geometric/nn/models/lightgcn.py
rusty1s: Prefer
```suggestion
        out_src = out[edge_label_index[0]]
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        if weights is not None:
+            assert (weights.shape == self.embeddings.shape), \
+                'Pre-computed embeddings weights must match shape of \
+                embedding layer.'
+            self.embeddings.weight.data.copy_(weights)
+
+        out = self._propagate_embeddings(edge_index)
+
+        nodes_fst = torch.index_select(out, 0, edge_label_index[0, :].long())
+        nodes_sec = torch.index_select(out, 0, edge_label_index[1, :].long())
```
rusty1s, line 134:
```suggestion
        out_dst = out[edge_label_index[1]]
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+                :obj:`edge_index` will be used (default: :obj:`None`).
+            prob (bool): Whether probabilities should be returned.
+        """
+        assert (self.objective == LightGCN.objectives[1]), 'Prediction ' \
+            'only works for "link_prediction" objective.'
+
+        if edge_label_index is None:
+            edge_label_index = edge_index
+
+        predictions = self.forward(edge_index, edge_label_index)
```
rusty1s, line 163:
```suggestion
        predictions = self(edge_index, edge_label_index)
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+    def recommend(self, edge_index: Adj, source_indices: LongTensor,
+                  target_indices: LongTensor, topK: int = 1) -> Tensor:
+        """Get :obj:`topK` recommendations for nodes in :obj:`source_indices`.
+
+        This prediction head works only for the "ranking" objective.
+
+        Args:
+            edge_index (Adj): Edge tensor specifying the connectivity of the
+                graph.
+            source_indices (LongTensor): Nodes for which recommendations
```
rusty1s, line 178: Can we make this optional? For example, finding the top-k entries for only a set of source or target nodes (not both).

torch_geometric/nn/models/lightgcn.py
```diff
@@ -0,0 +1,315 @@
...
+        target_embeddings = prop_embeddings[target_indices]
+
+        rankings = torch.mm(source_embeddings, target_embeddings.t())
+        top_indices = torch.topk(rankings, topK, dim=-1).indices
+        index_matrix = target_indices.reshape((1, -1))
+        index_matrix = index_matrix.repeat(source_indices.size(0), 1)
+
+        return index_matrix.gather(-1, top_indices)
+
+    def loss(self, preds: Tensor = None, edge_label: Tensor = None,
```
rusty1s, line 199: Let's outsource that into `self.recommendation_loss` and `self.link_prediction_loss`.

rusty1s (commented):
This already looks great. I made a few comments, but most of them are minor ones :)

torch_geometric/nn/models/lightgcn.py
LukasHaas: This would be nice, but I think having the "objective" helps to enforce the right model usage + help the developer decide which kind of model to use depending on the task (i.e. choosing "recommendation_loss" vs. "link_prediction_loss"). 

Eventually, it would be cool if people could implement their own "prediction head" classes to stack on top of a model depending on the objective (node or edge classification / regression, link prediction, ranking etc.) similar to how models make use of convolutional layers in /nn/conv it would be nice to have a /nn/heads directory which implemented common model prediction heads. This would essentially be an abstraction of the what is on slide 24 here: [CS224W Stanford](http://snap.stanford.edu/class/cs224w-2020/slides/08-GNN-application.pdf). Would this work for you?

rusty1s: Yes, I absolutely love this idea, in which we provide prediction heads for link-level or graph-level tasks.

rusty1s (approved):
Thanks for all the changes, I also made a pass :)

Specifically, I removed the `objective` argument as I think that it isn't really needed (e.g., the same model can be used for link prediction and recommendation), and the logic for evaluation and training is already well-separated via `recommend`/`recommendation_loss` and `predict_link/link_pred_loss`.

I also do not apply a `sigmoid` for `link_pred_loss` and use `torch.nn.BCEWithLogitsLoss` instead (which should result in more stable training). The `sigmoid` has been moved to `predict_link` instead.

Anyway, this was really good. Thank you all for your contributions!

PR #3686, opened 2021-12-12
fix attentive_fp

Phimos:
The implementation in PyG is a little bit different with the original repo
https://github.com/OpenDrugAI/AttentiveFP/blob/master/code/AttentiveFP/AttentiveLayers.py

The alignment step in attention layer should use `concat` not `add`. And an extra linear layer should be used to extract feature from neightbor nodes & edges. 

The new model can achieve better performace in my test.

### DIFF
--- a/torch_geometric/nn/models/attentive_fp.py
+++ b/torch_geometric/nn/models/attentive_fp.py
@@ -12,8 +12,8 @@ from torch_geometric.utils import softmax
 
 
 class GATConv(MessagePassing):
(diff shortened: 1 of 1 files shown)


Phimos:
sorry, it seems that there is no differences.
I'm confused why my version works better than PyG version.

Phimos:
maybe `dropout` is the reason, I removed `dropout` from the timestamp loop

Phimos:
> maybe `dropout` is the reason, I removed `dropout` from the timestamp loop

And the original repo did not use `dropout` there

PR #3690, opened 2021-12-12
Fix FastRGCNConv and add a new test

Jokeren:
Change the index argument to edge_index_j in FastRGCNConv's message method


rusty1s (approved):
Thank you for the fast fix :)

Issue #3573, opened 2021-11-26
Adding input and output shapes to docs


Gaurangkarwande:
Can you add to the documentation the input shape to a layer and the corresponding output shape after forward() has been executed. This is similar to the Pytorch documentation.

For e.g. - the documentation for 2D Convolution layer, nn.torch.Conv2d( ), states that:
input  = (N, Cin, H, W)
output = (N, Cout, Hout, Wout)

## Motivation

I am fairly new to GNNs and pytorch-geometric. I think it will be a lot easier and smooth for newbies/noobies like me to get a sense of the dimensional transformation within layer.

## Additional context

nn.torch.Conv2d( ) - 

![image](https://user-images.githubusercontent.com/49279636/143517046-8ae3e10c-71e5-4bde-abdb-c2de265be92f.png)

rusty1s:
This is a great point, but it will involve a lot of work. Let me know if you want to help me with that. 

In general, The input of a `GNN` layer will be `[num_nodes, in_channels]` and will output a tensor of shape `[num_nodes, out_channels]`.

Gaurangkarwande:
Yeah, I sure can contribute. I am new to GNNs but I can try adding on the documentation for the models I know of are have used.

rusty1s:
Super. Let's start with a single layer, e.g., `GraphConv` and go from there. Let me know if you need any more guidance.

saiden89:
Do you think  `edge_index` and optional tensors like `edge_weight` should also be acknowledged as inputs?

rusty1s:
Yeah, that would be something like:
* `edge_index: (2, num_edges)`
* `edge_weight: (num_edges)`

saiden89:
Here is a proposal for `GraphConv`:

![image](https://user-images.githubusercontent.com/45285915/145580021-ddf1eb44-2962-4825-979f-974c10db5660.png)

I've tried to follow as closely as possible the conventions already established in PyTorch, as well as mathematical notations present in the literature. Along these guidelines, the `x` and `edge_index` are proxied by the calligraphic typeface V and E respectively, while enclosure with vertical bars (|) refers to the number of elements.
It would also be possible to ditch the mathematical notation altogether in favor of more agnostic solutions, but considering that this framework is specifically aimed towards GNN, I believe that some mathematic rigor can be justified.
Of course, I would love to hear the community's opinion 😄.

rusty1s:
I think this is great :) Please feel free to contribute!

PR #3697, opened 2021-12-14
Updated docstring for shape info

saiden89:
As proposed in #3573, here is my take on the first chunk of convolutional layers in PyG. More will follow soon.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -125,7 +125,7 @@ h2,
   text-transform: uppercase;
   background: inherit;
   border: none;
-  padding: 6px 0;
+  margin-bottom: 6px !important;
 }
 
 .rst-content dl:not(.docutils) .property:first-child .pre {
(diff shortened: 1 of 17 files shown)


saiden89:
Thanks for the corrections. Indeed, I was also wondering how to solve some inconsistencies in notation. So I guess now all features are F instead of H. How about the attention heads? Some papers refer to them as K (like GAT) some others to H. Do you think it would be better to follow the original article of to use a universal notation in PyG?

rusty1s (approved):
This is great, thank you :)
I further changed the CSS layout and made some style changes.

rusty1s:
I think it's better to have a universal notation, and I like the following:
* `|V|`: number of nodes
* `|E|`: number of edges
* `F`: node feature size
* `D`: edge feature size
* `H`: number of heads
* `L`: number of layers

saiden89:
I have another batch of the updated docstring. Is it possible to reopen this pull request or do I make a new one?

Issue #3713, opened 2021-12-16
Remove size parameter from forward methods


saiden89:
## 🛠  Proposed Refactor

Modify all `forward` methods not to reference `size` in their signature and arguments, and modify the tests to pass accordingly.

### Motivation

Taking hint from e42ac59f1218047038df8241672d944dd06f082d and 6c7965291b3cd24db48ef57105ae5742a3932dec it seems that the `size` parameter of the `forward` function is being deprecated, as calling without passing any value still results in the correct propagation of tensors.

### Pitch

To the best of my knowledge, there is no need to let the user explicitly set the `size`, as already many networks can safely infer it based on input values.

### Additional context
I'm unsure whether this refactoring can apply to *all* layers. Please let me know if there are specific use cases where it is best to leave `size` in the signature.

rusty1s:
Yes, this is a great point. In general, we only need support for `size` in case we operate on a bipartite graph in which one of the node features matrices (source or destination) can be `None`. For example, this is possible in `GATConv`, where the input can be `OptPairTensor = Tuple[Tensor, Optional[Tensor]]`.

However, the commits indeed suggest that this is not consistent across all PyG operators, so I'm happy to refactor this :)

saiden89:
Cool! Reading from the documentation, currently it seems that the following convolutional layers still use the `size`:
- `SAGEConv`
- `GraphConv`
- `GATConv`
- `GATv2Conv`
- `GINConv`
- `GINEConv`
- `MFConf`
- `GMMConv`
- `SplineConv`
- `NNConv`
- `GENConv`
- `GeneralConv`
- `HEATConv`

As I only recently took an interest towards PyG's technical internals, it would be better to have a list of layers that are to be refactored, leaving the rest as they are.

rusty1s:
Thank you! I just checked and it looks like only `GATv2Conv` and `HEATConv` do not support `OptPairTensor` while the have support for `size` in the `forward` call. Let me know if you have any interest in removing this for these two :)

saiden89:
Of course, that's why I opened the issue :).

PR #3744, opened 2021-12-21
Remove size parameter for GATv2 and HEAT

saiden89:
As proposed in #3713, this PR removes the size argument from the forward methods of the two classes and edits the tests to pass accordingly.

p.s. PR #3743 was forked from the wrong branch. Sorry for the inconvenience :(


rusty1s (approved):
Thank you!

PR #3776, opened 2021-12-30
adding ast parsing for function use

Padarn:
Merge request adds a utility function to determine if a function is internally called. Idea would be to use it as

```
self.inspector = Inspector(self)
uses_updater = self.inspector.calls_internal('forward', 'edge_updater')
```
To give a more robust way of determining if `edge_updater` is used inside of the `forward` function.

This came up while trying to figure out if it needed to be converted for `jittable`.


Padarn:
Not sure why the checks are failing on this one - looks like its due to a dataset used in a test not being available.

rusty1s:
Re-started, let's see how it goes.

rusty1s:
The remaining tests should run through if you rebase from master.

rusty1s:
@Padarn Is this PR still needed after #3765 is merged?

Padarn:
Not needed, found a different way to achieve this. Will close this PR.

PR #3796, opened 2022-01-03
Fix Wiki-CS to be undirected by default

pmernyei:
I discovered a mistake I made when assembling the [Wiki-CS dataset](https://github.com/pmernyei/wiki-cs-dataset): I intended to make it an undirected graph and [presented it as such](https://arxiv.org/pdf/2007.02901.pdf), but when making the dataloader for PyG, I forgot the process it accordingly. See also [the relevant issue on my repo](https://github.com/pmernyei/wiki-cs-dataset/issues/3).
 
This PR is updates the dataset to actually be undirected by default, while leaving an optional flag for loading it as a directed dataset for anyone who prefers to use it that way.

I also thought it might be worth adding a warning when loading the dataset for reproducability, since of course this makes new results non-comparble with existing reported results on the default settings. I'm not sure if/how that would be appropriate, though (do we just add a print statement, or is there some more sophisticated logging management in PyG?), so I'll leave that to the maintainers.

Thank you!

### DIFF
--- a/torch_geometric/datasets/fake.py
+++ b/torch_geometric/datasets/fake.py
@@ -31,8 +31,8 @@ class FakeDataset(InMemoryDataset):
             If set to :obj:`"auto"`, will return graph-level labels if
             :obj:`num_graphs > 1`, and node-level labels other-wise.
             (default: :obj:`"auto"`)
-        is_undirected (bool): Whether the graphs to generate are undirected.
-            (default: :obj:`True`)
+        is_undirected (bool, optional): Whether the graphs to generate are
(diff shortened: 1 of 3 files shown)


pmernyei:
Is my impression correct that the failed tests seem very much unrelated, or should I look into them more?

rusty1s:
Yeah, current CI is broken due to a `geodesic` and `numpy` version mismatch. I try to fix it in https://github.com/pyg-team/pytorch_geometric/pull/3792.

Nonetheless, thanks for your PR. I like the option to restore the old behavior. I also added a warning.

PR #3832, opened 2022-01-11
Properly describe ZINC's target

klicperajo:
I've done a bit of digging to figure this out for my ["Directional Message Passing on Molecular Graphs via Synthetic Coordinates"](https://arxiv.org/abs/2111.04718) paper, including some help by Chaitanya Joshi. Should be easier for others if we include this info here as well.


rusty1s (approved):
Thank you!

PR #3867, opened 2022-01-16
Let `TemporalData` inherit from `BaseData` and add docs

otaviocx:
## What

- Refactor the `torch_geometric.data.TemporalData` to get it inheriting from `torch_geometric.data.BaseData`
- Add code documentation to `torch_geometric.data.TemporalData`

## Why

- This PR covers the first 2 items of the (#3230).
- With the `TemporalData` class inheriting from `BaseData`, we turn possible to get it taking advantage of all the features already implemented for `Data` and `HeteroData` classes (such as Batching, DataLoader and Sampling for instance).

**NOTE:** This PR is **not** covering the following item of the Roadmap, which will be covered by another PR.
- Make `TemporalData` work with `torch_geometric.data.DataLoader`

## How

The `TemporalData` class has been refactored to add the missing methods. Since the `BaseData` class has some methods that deals with `NodeStorage` and `EdgeStorage`, the `TemporalData` class has also been refactored to use an internal `_store` field, in a similar way we currently have for `Data` and `HeteroData`

## Testing

The only model that is currently using the `TemporalData` class is `examples/tgn.py`. Thus, in order to test this refactoring, please run this example and it should be working as before.


torch_geometric/data/temporal.py
```diff
@@ -1,24 +1,85 @@
...
 
+from torch_geometric.data.data import BaseData
+from torch_geometric.data.storage import (BaseStorage, NodeStorage,
+                                          EdgeStorage, GlobalStorage)
+from torch_sparse import SparseTensor
+
+
+class TemporalData(BaseData):
+    r"""A data object composed by a stream of events describing a temporal
+    graph. The temporalData object can hold a list of events (that can be
```
rusty1s, line 16:
```suggestion
    graph. The :class:`~torch_geometric.data.TemporalData` object can hold a list of events (that can be
```

torch_geometric/data/temporal.py
```diff
@@ -1,24 +1,85 @@
...
+    understood as temporal edges in a graph) with structured messages.
+    An event is composed by a source node, a destination node, a timestamp
+    and a message. Any Continuous-time dynamic graphs (CTDG) can be
```
rusty1s, line 19:
```suggestion
    and a message. Any Continuous-Time Dynamic Graph (CTDG) can be
```

torch_geometric/data/temporal.py
```diff
@@ -1,24 +1,85 @@
...
+    Args:
+        src (Tensor, optional): A list of source nodes for the events with
+            shape :obj:`[num_events]`. (default: :obj:`None`)
+        dst (Tensor, optional): A list of destination nodes for the events
+            with shape :obj:`[num_events]`. (default: :obj:`None`)
+        t (Tensor, optional): The timestamps for each event with shape
+            :obj:`[num_events]`. (default: :obj:`None`)
+        msg (Tensor, optional): Messages feature matrix with shape
+            :obj:`[num_events, num_msg_features]`. (default: :obj:`None`)
+        y (Tensor, optional): event-level ground-truth labels with
```
rusty1s, line 60: Let's remove `y` here. If the user wants to add ground-truth labels, the user can pass them in via `kwargs` nonetheless.

rusty1s (approved):
This is great. Love it. Left a few comments.

torch_geometric/data/temporal.py
otaviocx: Hi @rusty1s, I've removed `y` and updated the docs with an example adding it. Just checked and the `example/tgn.py` script is still working.

rusty1s:
Thanks a lot. I also made some minor changes (mostly fixing some parts of documentation). Next, I think we should add some basic tests for `TemporalData` as well.

otaviocx:
> Thanks a lot. I also made some minor changes (mostly fixing some parts of documentation). Next, I think we should add some basic tests for `TemporalData` as well.

Thanks, @rusty1s! I will work on tests this week (and also Data Loader).

Issue #3884, opened 2022-01-18
Add torch.jit.script support to GravNetConv layer


RobMcH:
### 🛠 Proposed Refactor

Currently, trying to run torch.jit.script on a GravNetConv layer results in a `RuntimError`. Below a minimal example is given to reproduce this behaviour.

```
import torch
import torch_geometric

# Create layer with some dummy parameters.
layer = torch_geometric.nn.GravNetConv(10, 10, 1, 1, 1).jittable()
# Run torch.jit.script on the layer. Throws RuntimeError.
jitted = torch.jit.script(layer)
```

### Suggest a potential alternative/fix

This is caused by assigning `batch = None` as the default value in the forward method. Without a default value, torch.jit.script executes successfully. Further, the default value is not necessary as in the layer a variable `b = PairOptTensor = (None, None)` is created and then modified depending on whether `batch` is a Tensor or PairTensor (and at no time it is checked whether `batch is None`).

RobMcH:
See [the corresponding MR](https://github.com/pyg-team/pytorch_geometric/pull/3885) for the fix.

PR #3885, opened 2022-01-18
Make gravnet_conv.py compatible with torch.jit.script

RobMcH:
Removes unnecessary default value in order to make the layer compatible with torch.jit.script. See [the corresponding issue](https://github.com/pyg-team/pytorch_geometric/issues/3884).


rusty1s:
Super. I think I also fixed it without making `batch` non-optional by disallowing specific type combinations of `x` and `batch` (see the `#type: ...` comments).

PR #3890, opened 2022-01-19
`hydra.ConfigStore`

rusty1s:
`hydra` setup for instantiating any PyG dataset from `torch_geometric.datasets`. Any (composable) PyG transformation can be applied as well:
```yaml
defaults:
  - dataset: KarateClub
  - transform@dataset.transform:
      - NormalizeFeatures
      - AddSelfLoops

dataset:  # override default values
  transform:
    AddSelfLoops:
      fill_value: 2.0
```
* Implements a basic `hydra.config_store.ConfigStore` with groups `["transform", "dataset"]` by automatically parsing transformations/datasets into `@dataclass` schemas
* Makes use of `defaults` for specifying default configurations
* Passes configuration groups to certain attributes via `@`, *e.g.*, `transform@dataset.transform`
* Makes use of nesting levels to allow for applying multiple transformations, see [here](https://hydra.cc/docs/patterns/select_multiple_configs_from_config_group)


yaoyaowd:
What a familiar name. What's the plan this time?

rusty1s:
This is still a draft. I will update the description with more information once I have a more concrete idea of what I'm actually doing. Happy to chat more afterwards.

.gitignore
```diff
@@ -7,6 +7,7 @@ runs/
 .cache/
 .eggs/
 lightning_logs/
+outputs/
```
yaoyaowd, line 10: why is this change necessary?

test/graphgym/my_config.yaml
```diff
@@ -0,0 +1,5 @@
+defaults:
+  - dataset: KarateClub
+  - transform@dataset.transform:
+      - NormalizeFeatures
```
yaoyaowd, line 4: This is super cool. I think I'm missing a lot of context here. How does these names transfer to a function?

torch_geometric/transforms/compose.py
```diff
@@ -8,7 +8,7 @@ class Compose(BaseTransform):
     """Composes several transforms together.
 
     Args:
-        transforms (List[Callable]): List of transforms to compose.
+        transforms ([callable]): List of transforms to compose.
```
yaoyaowd, line 11: I feel `List[Callable]` is better than `[callable]` since `[]` may sometime mean optional.

torch_geometric/transforms/radius_graph.py
```diff
@@ -17,8 +17,8 @@ class RadiusGraph(BaseTransform):
             with message passing (:obj:`"source_to_target"` or
             :obj:`"target_to_source"`). (default: :obj:`"source_to_target"`)
     """
-    def __init__(self, r, loop=False, max_num_neighbors=32,
-                 flow='source_to_target'):
+    def __init__(self, r: float, loop: bool = False,
```
yaoyaowd, line 20: another coding style would be each argument take a new line so that in the future if we add a new argument, only the new argument line changes.

torch_geometric/transforms/to_device.py
```diff
@@ -12,15 +12,19 @@ class ToDevice(BaseTransform):
...
-            conversion for the given attributes. (default: :obj:`[]`)
+            conversion for the given attributes. (default: :obj:`None`)
         non_blocking (bool, optional): If set to :obj:`True` and tensor
             values are in pinned memory, the copy will be asynchronous with
             respect to the host. (default: :obj:`False`)
     """
-    def __init__(self, device: Union[int, str], attrs: List[str] = [],
-                 non_blocking: bool = False):
+    def __init__(
+        self,
```
yaoyaowd, line 21: just like this.

torch_geometric/graphgym/config_store.py
```diff
@@ -0,0 +1,95 @@
...
+from omegaconf import MISSING
+
+import torch_geometric.datasets as datasets
+import torch_geometric.transforms as transforms
+
+EXCLUDE_ARG_NAMES = {'self', 'args', 'kwargs', 'pre_filter'}
+
+
+def to_dataclass(cls: Any, base: Optional[Any] = None) -> Any:
+    r"""Converts a given class :obj:`cls` to a :obj:`dataclass` schema."""
```
yaoyaowd, line 15: would be great to add an example here.

yaoyaowd (approved):
This is fantastic.

The only drawback I can think of is the limitation of converting class to dataclass. How would we have a more general interface so that if a class cannot be converted to a dataclass directly, we can still initialize the class based on hydra parameter?

One idea I can think of is to directly pass json argument in that case and let the class pass it directly. Any thoughts?

.gitignore
rusty1s: `hydra` adds this automatically as part of its experiment management tracker. Will need to look into how to disable it.

test/graphgym/my_config.yaml
rusty1s: The names are derived from `torch_geometric.transforms`, see [here](https://pytorch-geometric.readthedocs.io/en/latest/modules/transforms.html). That is `torch_geometric.transforms.NormalizeFeatures` will get mapped to the following `@dataclass`:
```python
@dataclass
class NormalizeFeatures:
    _target_: "torch_geometric.transforms.NormalizeFeatures"
    attrs: List[str] = ["x"]
```
This pattern is also used [torch-hydra](https://github.com/pytorch/hydra-torch), a project from the PyTorch team, see, *e.g.*, [here](https://github.com/pytorch/hydra-torch/blob/main/hydra-configs-torch/hydra_configs/torch/optim/adam.py).

torch_geometric/transforms/compose.py
rusty1s: Resolved.

torch_geometric/transforms/radius_graph.py
rusty1s: Resolved.

torch_geometric/graphgym/config_store.py
rusty1s: Added an example.

PR #3897, opened 2022-01-20
`HeteroData`: `data.*_dict` fix in case attribute does not exist

rusty1s:


RexYing (commented):
thx for helping to fix!

Issue #3873, opened 2022-01-17
to_networkx to_undirected option inconsistent with documentation


RemyLau:
### 🐛 Describe the bug

According to the documentation for [``to_networkx``](https://pytorch-geometric.readthedocs.io/en/latest/modules/utils.html#torch_geometric.utils.to_networkx), the ``to_undirected`` option preserves the upper triangle of the adjacency matrix and turn it into an undirected graph, something like the following

![eqn](https://quicklatex.com/cache3/06/ql_77affd715d97f56e1b528dfa14eb9606_l3.png)

But actually, the implementation does the opposite, which preserves the lower triangle and ignores the upper triangle.
The easiest fix would be to rewrite line#95 to ``if to_undirected and v < u``.

https://github.com/pyg-team/pytorch_geometric/blob/ffb70bcf6147b8d5c5a00650e1d1f514771d8898/torch_geometric/utils/convert.py#L93-L104

Despite this fix, the ``to_undirected`` option is a bit misleading in then sense that it is tempting to think it does what [``networkx.DiGraph.to_undirected``](https://networkx.org/documentation/stable/reference/classes/generated/networkx.DiGraph.to_undirected.html#networkx-digraph-to-undirected) does, i.e. if (u, v) in E, then add (v, u) in E if not already existed. My current work-around is the following:
```python
G = torch_geometric.utils.to_networkx(data, to_undirected=False, remove_self_loops=True).to_undirected()
```
Would it make sense to follow the ``to_undirected`` as networkx instead of only keeping track of either out (or in) edges?

### Environment

* PyG version: 2.0.2
* PyTorch version: 1.10
* OS: CentOS Linux
* Python version: 3.9.7
* CUDA/cuDNN version:
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`):

rusty1s:
Thanks for the issue. I might be wrong, but IMO `v > u` (which `u` being the row/source and `v` being the col/destination node maintain the upper triangle. That is, all column indices are greater than all corresponding row indices.

I agree that `to_undirected` naming is a bit confusing. It should probably be renamed to `undirected`. As PyG does not discriminate between undirected and directed graphs, this option is useful to tell `networkx` whether it should return a `nx.DiGraph` or a `nx.Graph`.

RemyLau:
[...]

PR #3901, opened 2022-01-21
Fix to_undirected option in to_networkx, fix test

RemyLau:
This PR aims to fix a minor bug for ``to_networkx`` function with ``to_undirected=True`` (https://github.com/pyg-team/pytorch_geometric/issues/3873).

The corresponding unit test is updated to match the expected behavior of preserving the upper triangle rather than the lower triangle of the adjacency matrix.


rusty1s (approved):

rusty1s:
Thank you!

PR #3930, opened 2022-01-24
Refactor in-degree computation in pna example

### LINKED ISSUE #3925: Computing the size of the in-degree histogram tensor dynamically in the PNA example
### 🛠 Proposed Refactor

The `deg` tensor in the PNA example is initialized with the size `5`. 
https://github.com/pyg-team/pytorch_geometric/blob/50b7bfc4a59b5b6f7ec547ff862985f3b2e22798/examples/pna.py#L23

This value will obviously be different for different datasets. One can iterate over the training data and compute the maximum degree any node has. Then, the histogram tensor can be initialized with that value. Something like this:
```python
# compute the maximum in-degree in the training data
max_degree = 0
for data in train_dataset:
    d = degree(data.edge_index[1], num_nodes=data.num_nodes, dtype=torch.long)
    if d.max().item() > max_degree:
        max_degree = d.max().item()

# create the in-degree histogram tensor
deg = torch.zeros(max_degree + 1, dtype=torch.long)
for data in train_dataset:
    d = degree(data.edge_index[1], num_nodes=data.num_nodes, dtype=torch.long)
    deg += torch.bincount(d, minlength=deg.numel())
``` 

The results of the `degree` function can also be cached to avoid iterating two times. Further, for custom datasets where the split is made with indices or boolean masks e.g `Data(x=[10000, 100], edge_index=[2, 200000], edge_attr=[200000, 20], y=[10000], train_mask=[10000], val_mask=[10000], test_mask=[10000])`, the `subgraph` utility can be used.
```python
tr_subgraph = data.subgraph(data.train_mask)
# compute the in-degree of all the training nodes
d = degree(index=tr_subgraph.edge_index[1], num_nodes=tr_subgraph.num_nodes, dtype=torch.long)

# get the maximum in-degree, this will be the size of the histogram tensor
max_degree = d.max().item()

# create the in-degree histogram tensor
deg = torch.zeros(max_degree + 1, dtype=torch.long)
deg += torch.bincount(d, minlength=deg.numel())
```

hash-ir:


examples/pna.py
```diff
@@ -19,8 +19,15 @@
 val_loader = DataLoader(val_dataset, batch_size=128)
 test_loader = DataLoader(test_dataset, batch_size=128)
 
-# Compute in-degree histogram over training data.
-deg = torch.zeros(5, dtype=torch.long)
+# Compute the maximum in-degree in the training data.
+max_degree = -1
+for data in train_dataset:
+    d = degree(data.edge_index[1], num_nodes=data.num_nodes, dtype=torch.long)
+    if d.max().item() > max_degree:
```
rusty1s, line 26: This is a little bit cleaner :)
```suggestion
max_degree = max(max_degree, int(d.max()))
```

rusty1s (approved):

PR #3948, opened 2022-01-25
New undirected options for to_networkx

RemyLau:
This PR introduces extra options for ``to_undirected`` with ``"upper"`` and ``"lower"``, as discussed in #3873


rusty1s (approved):
This is perfect. Thank you!

PR #3949, opened 2022-01-25
Implement LargestConnectedComponents Transform

abojchevski:
Implements a Transform that selects the subgraph that corresponds to the K largest connected components in the graph.


rusty1s (approved):
Thank you very much :)

PR #3955, opened 2022-01-26
Add MetrLA Dataset

radandreicristian:
With this pull request, the [MetrLA](https://paperswithcode.com/sota/traffic-prediction-on-metr-la) dataset and helper classes were added, including the following files:

- Dataset class
- I/o methods
- Unit tests
- Documentation


radandreicristian:
Well, this is awkward. The docs/unit tests are all working locally (Ubuntu 21.04, Python 3.9, Torch 1.10.1+CU11.3). Any input on how to deal with this would be greatly appreciated, as this is only my second contribution to open-source:)

rusty1s:
I think I fixed this in master, see https://github.com/pyg-team/pytorch_geometric/commit/e04fe09c6937e37d4dd4bb9923d1ed5dabf4bc9c. You might need to rebase your fork from master.

rusty1s:
Replacing all occurrences of `list[...]` with `List[...]` should resolve the remaining errors.

radandreicristian:
> Replacing all occurrences of `list[...]` with `List[...]` should resolve the remaining errors.

Thanks for the help, working on it!

radandreicristian:
There's another error requiring "tables" (for pandas reading from HDF files). I'll add it to the req's on my branch and try again.

rusty1s:
Any chance to get rid of this additional dependency? Otherwise, we may want to add it to the `full_requires` dependency list.

radandreicristian:
Yes, perhaps I can convert the original .hd5 file to .csv.

setup.py
```diff
@@ -4,17 +4,9 @@
...
-    'pandas',
-    'jinja2',
-    'requests',
-    'pyparsing',
-    'hydra-core',
-    'scikit-learn',
-    'googledrivedownloader',
+    'tqdm', 'yacs', 'numpy', 'scipy', 'pandas', 'jinja2', 'requests',
+    'pyparsing', 'hydra-core', 'scikit-learn', 'googledrivedownloader',
+    'tables'
```
rusty1s, line 9:
```suggestion
    'tables',
```

test/datasets/test_metrla.py
```diff
@@ -0,0 +1,38 @@
+import shutil
+
+from torch_geometric.datasets import MetrLa
+
+
+def test_metrla() -> None:
+    root = '/tmp/metr_la'
```
rusty1s, line 7:
```suggestion
    root = osp.join('/tmp', 'metr_la')
```
As the tests suggests, this might need to be fixed on Windows, *e.g.*, via `osp.join('/tmp', 'metr_la')`.

test/datasets/test_metrla.py
```diff
@@ -0,0 +1,38 @@
...
+    n_previous_steps = 6
+    n_future_steps = 6
+
+    print("This may take a couple of minutes...")
```
rusty1s, line 11: How long does it take to download the dataset? We generally do not test downloading and processing datasets as part of our CI. I'm fine with dropping this test completely.

torch_geometric/datasets/__init__.py
```diff
@@ -157,6 +158,7 @@
     'RandomPartitionGraphDataset',
     'LINKXDataset',
     'EllipticBitcoinDataset',
+    'MetrLa'
```
rusty1s, line 161:
```suggestion
    'MetrLa',
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+
+from torch_geometric.data import Data, Dataset
+from torch_geometric.io.metrla import MetrLaIo
+from torch_geometric.utils.sparse import dense_to_sparse
+
+
+class MetrLa(Dataset):
+    r"""The Los Angeles Metropolitan (MetrLA) highway traffic dataset
+    introducted in `"Big Data and its Technical Challenges"
+    <http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.681.7248&rep=rep1&type=pdf>`_
```
rusty1s, line 15: We may need to split the URL into 2 lines:
```
    <http://citeseerx.ist.psu.edu/viewdoc/download?
    doi=10.1.1.681.7248&rep=rep1&type=pdf>`_
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+from google_drive_downloader import GoogleDriveDownloader as gdd
+
+from torch_geometric.data import Data, Dataset
+from torch_geometric.io.metrla import MetrLaIo
+from torch_geometric.utils.sparse import dense_to_sparse
+
+
+class MetrLa(Dataset):
+    r"""The Los Angeles Metropolitan (MetrLA) highway traffic dataset
+    introducted in `"Big Data and its Technical Challenges"
```
rusty1s, line 14:
```suggestion
    introducted in the `"Big Data and its Technical Challenges"
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    <http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.681.7248&rep=rep1&type=pdf>`_
+    paper.
+
+    Nodes represent sensors in a sensor network, each being able to
+    detect cars passing by it. The data points represent the traffic
+    volumes (how many cars have passed by a sensor), aggregated in 5 minutes
+    intervals. The traffic prediciton task is a time series regression,
+    which aims to predict the traffic volumen values in each node for the
+    next *n_next* time steps, given the volumnes in the *n_prev*  previous
```
rusty1s, line 23:
```suggestion
    next :ob:`n_next` time steps, given the volumes in the :obj:`n_prev`  previous
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    time steps. Both n_next and n_prev are pre-set values. Given the
```
rusty1s, line 24:
```suggestion
    time steps. Both :obj:`n_next` and :obj:`n_prev` are pre-set values. Given the
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    time-series nature of the data, each node has an additional, temporal
+    dimension.
+
+    The initial adjacency matrix is constructed by applying a thresholded
+    Gaussian kernel over the real-world distances between the sensors:
+
+    :math:`W_{ij} = exp(-\frac{dist(v_{i}, v_{j})^2}{\sigma^{2}})`,
```
rusty1s, line 31: Let's put that formula in equation mode:
```
.. math::
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    if :math:`dist(v_{i}, v_{j}) >= K`, otherwise :math:`0`, where K is a
+    pre-fixed threshold, and dist(, ) is the real-world road distance
+    between two sensors, and :math:`\sigma^{2}` is the variance of the
+    distances between sensors. A zero-value in the adjacency matrix does
+    not indicate that thers is no road connection between two sensors (
```
rusty1s, line 36:
```suggestion
    not indicate that there is no road connection between two sensors (
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+        """This is a circular dependency. You need this to calculate processed
+        file names, but you need it before the call call to the "process"
+        method in the superclass. The only in-between place you could
+        effectively calculate this is before the process, so either in
+        __init__ (where you don't have access to the downloaded file
+        containing the readings, or in download, but download gets skipped
+        after the first time, so that's not reliable. Open to any
+        workarounds, but the dataset has a fixed size, so
+        this is not a deal breaker imo."""
+        self.n_readings = 34272
```
rusty1s, line 82: It's probably sufficient to just test that the first few processed files exists. Otherwise, checking the existence of 34272 can quickly become the bottleneck.

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    def raw_file_names(self) -> str:
+        r"""The name of the files in the :obj:`self.raw_dir`
+        folder that must be present in order to skip downloading."""
+        return list(self.gdrive_ids.keys())
+
+    @property
+    def processed_file_names(self) -> str:
+        r"""The name of the files in the :obj:`self.processed_dir`
+        folder that must be present in order to skip processing."""
+        return [f'data_{i}.pt' for i in range(self.io.dataset_len)
```
rusty1s, line 104:
```suggestion
        return [f'data_{i}.pt' for i in range(5)
```

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+        edge_attributes_path = Path(
+            self.processed_dir) / self.edge_attributes_file_name
+
+        torch.save(obj=edge_index, f=edge_index_path)
+        torch.save(obj=edge_attributes, f=edge_attributes_path)
+
+        for idx in range(self.io.dataset_len):
+            # Select the "slice" among the first dimension
+            # x_index (n_previous_steps, n_nodes, n_features)
+            x_index = torch.tensor(data=x[idx, ...], dtype=torch.float32)
```
rusty1s, line 146:
```suggestion
            x_index = x[idx]
```
Isn't that sufficient?

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    def get(self, idx: int) -> Data:
+        r"""Gets the data object at index :obj:`idx`."""
+        file_name = self.processed_file_names[idx]
+        edge_index = torch.load(f=Path(self.processed_dir) /
+                                self.edge_index_file_name)
+        edge_attributes = torch.load(f=Path(self.processed_dir) /
+                                     self.edge_attributes_file_name)
+
+        data = torch.load(f=Path(self.processed_dir) / file_name)
+        setattr(data, 'edge_index', edge_index)
```
rusty1s, line 172:
```suggestion
        data.edge_index = edge_index
```

torch_geometric/datasets/metrla.py
radandreicristian: That's just to convert from numpy to torch tensor, since the Data constructor specifies x to be Optional[Tensor].

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @property
+    def processed_file_names(self) -> str:
+        r"""The name of the files in the :obj:`self.processed_dir`
+        folder that must be present in order to skip processing."""
+        return [f'data_{i}.pt' for i in range(self.io.dataset_len)
+                ] + ["edge_index.pt", "edge_attributes.pt"]
+
+    @property
+    def edge_index_file_name(self) -> str:
```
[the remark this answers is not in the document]
radandreicristian: They're not necessary, but since their values are used more than 2 times throughout the code I figured it would be more elegant this way. Would definitely make sense to have them private, thanks for pointing out!

torch_geometric/datasets/metrla.py
```diff
@@ -0,0 +1,177 @@
...
+    between two sensors, and :math:`\sigma^{2}` is the variance of the
+    distances between sensors. A zero-value in the adjacency matrix does
+    not indicate that thers is no road connection between two sensors (
+    because there almost always is, since road networks are connected
+    graphs. Instead, it means that the distance between them is larger than
+    the threshold value.
+
+    Args:
+        root (string): Root directory where the dataset should besaved.
+        n_previous_steps (int): The number of previous time steps to consider
```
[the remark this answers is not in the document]
radandreicristian: Right, I'll just update them in the docstring instead of changing the code.

torch_geometric/datasets/metrla.py
radandreicristian: Right, but then the problem is in the "get" function. But I guess I can solve that with an f-string (assuming all the files are named with the same template). Is that an appropriate way to go?

test/datasets/test_metrla.py
radandreicristian: Downloading takes like 10 sec, processing takes a good 2-3 minutes. I can try to limit it to generate a single example and do the last 5 asserts on it, otherwise will just drop it.

torch_geometric/datasets/metrla.py
radandreicristian: Checking the existence of ~34K file takes 1.2 seconds in my case (on SSD; HDD storage might be slower indeed), so it's probably not that much of a bottleneck for something that happens every time the code is run (if I remove the test case).

Anyway, your suggestion is valid, and we should optimize wherever possible. I modified it to check only the first 5 (needed a bit more changes in the code than replacing the value with 5, i.e. in the "len" method), and it's coming with the next push. Thanks!

torch_geometric/datasets/metrla.py
radandreicristian: I wrapped the getter of the name in a private method (as it was used more than twice, to avoid further problems with renaming, and to avoid possible typos with hard-coded values throughout the code).

radandreicristian:
Alright, integrated all the feedback and pushed. 

Thanks for the guidance, @rusty1s!

PR #3977, opened 2022-01-30
Feat homophily measure

hash-ir:
The code uses the edge homophily measure from PyTorch Geometric and is referenced from the paper [repository](https://github.com/CUAI/Non-Homophily-Large-Scale/blob/eb531f33b2c8f3c14d64287938b761df1838e6d3/homophily.py#L90). Currently, the measure does not incorporate the `batch` vector and returns the total ratio. I also updated the function docstring. Let me know if something is off or needs some additional information! Happy to contribute :)

### DIFF
--- a/torch_geometric/utils/homophily.py
+++ b/torch_geometric/utils/homophily.py
@@ -52,20 +52,19 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
       where :math:`C_k` is the number of samples of class :math:`k`,
       :math:`d_u` is the number of neighbors of node :math:`u`,
       and :math:`d^{(k_u)}_u` is the number of neighbors of u that have
-      the same class label.
-
-      That measure is called the *class insensitive edge homophily ratio*.
+      the same class label. That measure is called the *class
+      insensitive edge homophily ratio*.
 
     Args:
         edge_index (Tensor or SparseTensor): The graph connectivity.
         y (Tensor): The labels.
         batch (LongTensor, optional): Batch vector
-            :math:`\mathbf{b} \in {\{ 0, \ldots,B-1\}}^N`, which assigns
-            each node to a specific example. (default: :obj:`None`)
+        :math:`\mathbf{b} \in {\{ 0, \ldots,B-1\}}^N`, which assigns
+        each node to a specific example. (default: :obj:`None`)
         method (str, optional): The method used to calculate the homophily,
-            either :obj:`"edge"` (first formula), :obj:`"node"`
-            (second formula) or :obj:`"edge-insensitive"`
-             (third formula). (default: :obj:`"edge"`)
+        either :obj:`"edge"` (first formula), :obj:`"node"`
+        (second formula) or :obj:`"edge-insensitive"`
+        (third formula). (default: :obj:`"edge"`)
     """
     assert method in ['edge', 'node', 'edge-insensitive']
     y = y.squeeze(-1) if y.dim() > 1 else y
--- a/torch_geometric/utils/homophily.py
+++ b/torch_geometric/utils/homophily.py
@@ -6,6 +6,7 @@ from torch_scatter import scatter_mean
 from torch_sparse import SparseTensor
 
 from torch_geometric.typing import Adj, OptTensor
+from torch_geometric.utils import degree
 
 
 def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
@@ -21,8 +22,8 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
       that have the same class label:
 
       .. math::
-        \text{homophily} = \frac{| \{ (v,w) : (v,w) \in \mathcal{E} \wedge
-        y_v = y_w \} | } {|\mathcal{E}|}
+        \frac{| \{ (v,w) : (v,w) \in \mathcal{E} \wedge y_v = y_w \} | }
+        {|\mathcal{E}|}
 
       That measure is called the *edge homophily ratio*.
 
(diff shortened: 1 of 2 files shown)


torch_geometric/utils/homophily.py
```diff
@@ -37,17 +37,30 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
 
       That measure is called the *node homophily ratio*.
 
+    - In the "Large-scale learning on non-homophilous graphs: \
```
rusty1s, line 40: Can we make this a link that points to the paper (similar to the other bullet points)?

torch_geometric/utils/homophily.py
```diff
@@ -37,17 +37,30 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
         edge_index (Tensor or SparseTensor): The graph connectivity.
         y (Tensor): The labels.
         batch (LongTensor, optional): Batch vector
             :math:`\mathbf{b} \in {\{ 0, \ldots,B-1\}}^N`, which assigns
             each node to a specific example. (default: :obj:`None`)
         method (str, optional): The method used to calculate the homophily,
-            either :obj:`"edge"` (first formula) or :obj:`"node"`
-            (second formula). (default: :obj:`"edge"`)
+            either :obj:`"edge"` (first formula), :obj:`"node"`
+            (second formula) or `"edge-insensitive"`. (default: :obj:`"edge"`)
```
rusty1s, line 61:
```suggestion
            (second formula) or :obj:`"edge-insensitive"` (third formula). (default: :obj:`"edge"`)
```

torch_geometric/utils/homophily.py
```diff
@@ -37,17 +37,30 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
+      homophily metric better captures the presence or absence of homophily:
+
+      .. math::
+        \text{homophily} = \frac{1}{C-1}\sum_{k=0}^{C-1}\begin{bmatrix}
+        {h_k - \frac{\lvert C_k \rvert}{n}}\end{bmatrix}_+,
+        
+      .. math::
+        h_k = \frac{\sum_{u \in C_k}d_u^{(k_u)}}{\sum_{u \in C_k}d_u}
+
+      That measure is called the *class insensitive edge homophily ratio*
```
rusty1s, line 51:
```suggestion
      That measure is called the *class insensitive edge homophily ratio*.
```

torch_geometric/utils/homophily.py
```diff
@@ -37,17 +37,30 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
+    - In the "Large-scale learning on non-homophilous graphs: \
+      New benchmarks and strong simple methods" paper, the class insensitive 
+      homophily metric better captures the presence or absence of homophily:
+
+      .. math::
+        \text{homophily} = \frac{1}{C-1}\sum_{k=0}^{C-1}\begin{bmatrix}
+        {h_k - \frac{\lvert C_k \rvert}{n}}\end{bmatrix}_+,
+        
+      .. math::
+        h_k = \frac{\sum_{u \in C_k}d_u^{(k_u)}}{\sum_{u \in C_k}d_u}
```
rusty1s, line 49: Can we add some more information to this formula? For example about `C_k` or `d_u`

torch_geometric/utils/homophily.py
```diff
@@ -63,11 +76,27 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
         out[y[row] == y[col]] = 1.
         out = scatter_mean(out, col, 0, dim_size=y.size(0))
         if batch is None:
             return float(out.mean())
         else:
             return scatter_mean(out, batch, dim=0)
+
+    else:
+        c = y.squeeze().max() + 1
+        nonzero_labels = y[y >= 0]
```
rusty1s, line 90: Why do we need this exactly?

hash-ir: It can be removed safely given that there are no negative labels

torch_geometric/utils/homophily.py
rusty1s: We also build doc as part of our CI, see https://pytorch-geometric--3977.org.readthedocs.build/en/3977/

torch_geometric/utils/homophily.py
```diff
@@ -63,11 +85,26 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
+    elif method == 'node':
         out = torch.zeros(row.size(0), device=row.device)
         out[y[row] == y[col]] = 1.
         out = scatter_mean(out, col, 0, dim_size=y.size(0))
         if batch is None:
             return float(out.mean())
         else:
             return scatter_mean(out, batch, dim=0)
+
+    else:
```
wsad1, line 97: The other homophily methods support a `batch` argument. I haven't thought how, but it would be good to support that for this too.

torch_geometric/utils/homophily.py
```diff
@@ -63,11 +85,26 @@ def homophily(edge_index: Adj, y: Tensor, batch: OptTensor = None,
...
+        c = y.squeeze().max() + 1
+        counts = y.unique(return_counts=True)[1]
+        proportions = counts.float() / y.shape[0]
+        h = homophily(edge_index, y, batch=y, method='edge')
+
+        out = 0
+        for k in range(c):
+            class_add = torch.clamp(h[k] - proportions[k], min=0)
+            if not torch.isnan(class_add):
+                out += class_add
```
wsad1, line 107: Can we replace this with something like `torch.clamp(h-proportions).sum()`, I guess you do need to check for `nan` when nodes are isolated, could we could use `torch.nan_to_num` for that.

wsad1 (changes requested):
@hash-ir Nice PR, I liked how you are reusing the `edge homophily` method. I have a few thoughts.

torch_geometric/utils/homophily.py
hash-ir: Okay, that makes sense. I will add it on the weekend

torch_geometric/utils/homophily.py
hash-ir: The refactoring can be done without any issues. For isolated nodes, we can also call the `remove_isolated_nodes` helper before computing the homophily measure. What do you think?

wsad1: Ah now I don't think we need either. `h = homophily(edge_index, y, batch=y, method='edge')` this returns `return scatter_mean(out, batch[col], dim=0)` and as `batch = y` isolated nodes are either going to have h as 0 or h is not going to be as long as we expect( when the isolated nodes class is `max(y)`). For the first case we don't need the check for nan and for the second we just need to append some zeros to `h`
What do you think.

torch_geometric/utils/homophily.py
rusty1s: Added `batch` support.

torch_geometric/utils/homophily.py
rusty1s: Exactly. One can omit appending zeros to `h` by making use of `bincount` rather than `unique(return_counts=True)`.

rusty1s:
Thanks @hash-ir and thanks @wsad1 for the review. I also made a pass and addressed your comments. Overall, I added `batch` argument support, added tests, and got rid of the for-loop.

rusty1s (approved):

hash-ir:
Thanks @wsad1 for the discussions and @rusty1s for incorporating the changes. Looking forward to more contributions in the future!

PR #3978, opened 2022-01-30
Implemented Var misuse dataset

### LINKED ISSUE #3693: VarMisuse dataset
### Motivation

paper "[LEARNING TO REPRESENT PROGRAMS WITH GRAPHS](https://arxiv.org/pdf/1711.00740.pdf)" which encode computer programs as graphs, with rich semantic information, however, most code implementation on this dataset VarMisuse is based on TensorFlow, like [tf-gnn-samples](https://github.com/microsoft/tf-gnn-samples)

### Pitch

include VarMisuse dataset in PyG framework as a class under torch_geometric.datasets

### Additional context

raw dataset provided by author:
[VarMisuse](https://www.microsoft.com/en-us/download/details.aspx?id=56844)

more papers/models on this dataset:
[GNN-FiLM: Graph Neural Networks with Feature-wise Linear Modulation](http://proceedings.mlr.press/v119/brockschmidt20a/brockschmidt20a.pdf)
[ON THE BOTTLENECK OF GRAPH NEURAL NETWORKS AND ITS PRACTICAL IMPLICATIONS](https://arxiv.org/pdf/2006.05205.pdf)

amitamb:


torch_geometric/datasets/var_misuse.py
```diff
@@ -0,0 +1,165 @@
...
+from typing import Callable, List, Optional
+
+import torch
+
+from torch_geometric.data import Data, Dataset, download_url, extract_gz
+
+
+class VarMisuse(Dataset):
+    r"""The VarMisuse dataset from the `"Learning to Represent Programs
+    with Graphs." <https://arxiv.org/abs/1711.00740>`_ paper, consisting of
```
rusty1s, line 12:
```suggestion
    with Graphs" <https://arxiv.org/abs/1711.00740>`_ paper, consisting of
```

torch_geometric/datasets/var_misuse.py
```diff
@@ -0,0 +1,165 @@
...
+                data_list = [self.pre_transform(d) for d in data_list]
+            torch.save(data_list, processed_path)
+
+    def len(self):
+        return len(self.processed_file_names) * 100
+
+    def get(self, idx):
+        file_idx, data_idx = divmod(idx, 100)
+        window, chunk = divmod(file_idx, 10)
+        data_list = torch.load(
```
rusty1s, line 162: How big is the underlying dataset on disk? Can't we use `InMemoryDataset` here?

While I like the current approach of having multiple graphs grouped together in one file, it looks like we are not really making use of this here. Currently, we are loading multiple graphs from disk, use one of them, and throw the remaining ones away.

Sadly, I currently have no good idea how to resolve this though. Any thoughts?

rusty1s:
Thanks for the PR and sorry for my super late review :) Left some comments.

torch_geometric/datasets/var_misuse.py
amitamb: @rusty1s The total sample numbers are following:

Train: 1250 * 100 = 125000
Valid: 214 * 100 = 21400
Test: 586 * 100 = 58600

Total space taken by gzipped files is around ~3.5GB for train, ~500GB for valid and ~1.5 GB for test.

The way I am processing at present,each processed pt file takes around 5 to 10MB for each chunk i.e.100 samples. So storing all train samples in memory would take around 7.5MB *1250 = ~9.375GB+

I thought that is too big for storing in memory. Most of the other in memory datasets are much smaller.

> Currently, we are loading multiple graphs from disk, use one of them, and throw the remaining ones away.

May be we can keep the pointer to the last opened chunk file open and reuse it somehow but that would only help for sequential access. I think samples would be loaded after shuffling so that would not be of much help. 100 samples or 5MB is not too big I thought to access individual sample and with SSD it likely will not be a bottleneck.

torch_geometric/datasets/var_misuse.py
amitamb: Will change that.

amitamb:
No problem. I have example implementation in works and it is working for a single sample/batch but weights are going to 0 for full batch. Would you mind take a look at it?

torch_geometric/datasets/var_misuse.py
rusty1s: Yes, shuffling is indeed a problem for this approach. I think 9GB would be still a reasonable amount for holding data in memory, but we can certainly think of adding a dataset interface in-between, e.g., `InMemoyChunkDataset` that effectively can re-use already loaded chunks.

torch_geometric/datasets/var_misuse.py
amitamb: Changed.

tsadigovAgmail:
Is there any work being done here?

Issue #3858, opened 2022-01-14
`torch_geometric.explain` + built-in `Captum` support


rusty1s:
## Roadmap

1. `Explainer`: `ABC` for explainability
2. `GNNExplainer(Explainer)`
3. `PGExplainer(Explainer)`
4. `to_captum` functionality
5. `to_captum` example
6. `evaluate_explainer` functionality. Some approaches to evaluate explainers are mentioned in [paper1](https://proceedings.neurips.cc/paper/2021/hash/99bcfcd754a98ce89cb86f73acc04645-Abstract.html) and [paper2](https://dl.acm.org/doi/abs/10.1145/3447548.3467283?casa_token=I2_R42o5b_EAAAAA%3AxSqT6VoJw2n6Tbck5_CnTU2u0IEbLisldEWrKWY3UFjF76e6MuQp8cbNU0BFsCPRngh1tyEVCa0r4A)
6. Implement/move to own `torch_geometric.explain` subpackage

```python
def to_captum(model: torch.nn.Module, mask_type: str = "edge",
              node_idx: Optional[int] = None) -> torch.nn.Module

...

model(mask, x, edge_index)
```

```python
def forward(self, mask, *args, **kwargs):
    set_masks(model, mask)
    out = self.model(*args, **kawrgs)
    clear_masks(model)
    return out
```

wsad1:
This is a nice idea. I'd be happy to help with this.
#2576 this is an old open WIP PR for PG explainer,  we could modify it to conform to `Explainer`.
Another thing what functions would `Exaplainer` have, I can think of `set_mask`, `clear_mask` and `train_explainer` thats common across PGexplainer and GNNExplainer.

rusty1s:
That's great, thanks for your help. @RBendias will implement the `Explainer` base class, which takes care of settings mask and providing unified visualizations. After the PR has landed, we can try to map #2576 to respect the new interface.

RBendias:
@wsad1 I already looked into your PR for PGExplainer a few weeks ago & also modified it to the current GNNExplainer state but didn't push it to your PR yet. I'm writing my thesis on Explainability for GNN; therefore, your commits helped me a lot already. Thanks!

wsad1:
@RBendias  all the best with you thesis, it sounds great.
Looking forward to your PR.

wsad1:
[...]

PR #3990, opened 2022-02-02
Make models compatible to Captum

RBendias:
As discussed in issue #3858, we can add a `to_captum` method for simple use of  [Captum](https://captum.ai/). The method adjusts the forward function of a model. I have added an example of edge explainability for node classification.

I have tested several Captum methods (IntegratedGradients, Saliency, InputXGradient, GuidedBackprop, Deconvolution, ShapleyValueSampling, Kernelshap, FeatureAblation), all of which work appropriately with the Captum example given here.

@rusty1s I have implemented a `CaptumModel` class, the `to_captum` method seems redundant. Alternatively, I could just overwrite the forward function instead of creating a new model. Do you have a preference and do you see an advantage in not creating a new model?


torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,92 @@
...
+        assert mask_type in ['edge', 'node', 'node_and_edge']
+
+        self.mask_type = mask_type
+        self.model = model
+        self.node_idx = node_idx
+
+    def forward(self, mask, *args):
+        # Set edge mask
+        if self.mask_type == 'edge':
+            set_masks(self.model, mask.squeeze(0), args[1],
```
rusty1s, line 59: Can we add some documentation about `mask`? Why does it have to be of shape `(1, *)`? How does the input need to look like in case `mask_type == "node_and_edge"`?

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,92 @@
...
+        # Node and edge mask
+        else:
+            x = self.model(mask[0], *args)
+
+        # Clear mask
+        if self.mask_type in ['edge', 'node_and_edge']:
+            clear_masks(self.model)
+
+        if self.node_idx is not None:
+            x = x[[self.node_idx]]
```
rusty1s, line 83: Why not
```suggestion
            x = x[self.node_idx]
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,92 @@
...
+        return x
+
+
+def to_captum(model: torch.nn.Module, mask_type: str = "edge",
+              node_idx: Optional[int] = None):
```
rusty1s, line 88:
```suggestion
              node_idx: Optional[int] = None) -> torch.nn.Module:
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,92 @@
...
+        # Clear mask
+        if self.mask_type in ['edge', 'node_and_edge']:
+            clear_masks(self.model)
+
+        if self.node_idx is not None:
+            x = x[[self.node_idx]]
+        return x
+
+
+def to_captum(model: torch.nn.Module, mask_type: str = "edge",
```
rusty1s, line 87: I like the usage of `to_captum` here, as this indicates that this is just a simple model transformation.

test/nn/models/test_explainer.py
```diff
@@ -0,0 +1,70 @@
+import pytest
+import torch
+import torch.nn.functional as F
+
+from torch_geometric.nn import GATConv, GCNConv, to_captum
+
+
+# TODO: Use same as in GNNExplainer Test
+class GCN(torch.nn.Module):
```
rusty1s, line 9: Define via `torch_geometric.nn.GCN` and `torch_geometric.nn.GAT`?

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+
+node_idx = 10
+captum_model = to_captum(model, mask_type="edge", node_idx=node_idx).to(device)
+target = data.y[node_idx].item()
+input_mask = torch.ones(data.num_edges,
+                        dtype=torch.float).requires_grad_(True).to(device)
+
+# Edge explainability with Integrated Gradients for node 10
+ig = IntegratedGradients(captum_model)
+ig_attr = ig.attribute(input_mask.unsqueeze(0), target=target,
```
rusty1s, line 55: Please add some documentation about `squeeze()` and `unsqueeze()` motivation :)

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+# Edge explainability with Saliency for node 10
+sa = Saliency(captum_model)
+sa_attr = sa.attribute(input_mask.unsqueeze(0), target=target,
+                       additional_forward_args=(x, edge_index))
+
+# Scale attributions to [0, 1]
+ig_attr = ig_attr.squeeze(0).abs() / ig_attr.abs().max()
+sa_attr = sa_attr.squeeze(0).abs() / sa_attr.abs().max()
+
+# Visualize attributions with GNNExplainer visualizer
```
rusty1s, line 62: Add a `TODO` here

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+                       additional_forward_args=(x, edge_index),
+                       internal_batch_size=1)
+
+# Edge explainability with Saliency for node 10
+sa = Saliency(captum_model)
+sa_attr = sa.attribute(input_mask.unsqueeze(0), target=target,
+                       additional_forward_args=(x, edge_index))
+
+# Scale attributions to [0, 1]
+ig_attr = ig_attr.squeeze(0).abs() / ig_attr.abs().max()
```
rusty1s, line 60: Is this correct? In case `ig_attr` contains negative values, you probably want to do:
```python
ig_attr = ig_attr.sub_(ig_attr.min()).div_(ig_attr.max())
```

rusty1s (commented):
Thank you :) This looks great. Left a few but minor comments :)

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+from torch_geometric.nn.models.explainer import to_captum
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+transform = T.NormalizeFeatures()
+dataset = Planetoid(path, dataset, transform=transform)
+data = dataset[0]
+
+
+class Net(torch.nn.Module):
```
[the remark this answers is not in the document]
RBendias: I used the exact same model as in `examples/gnn_explainer.py`. Should I change it nevertheless?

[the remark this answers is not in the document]
rusty1s: Yeah, I think it is better to directly use pre-defined models.

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+for epoch in range(1, 201):
+    model.train()
+    optimizer.zero_grad()
+    log_logits = model(x, edge_index)
+    loss = F.nll_loss(log_logits[data.train_mask], data.y[data.train_mask])
+    loss.backward()
+    optimizer.step()
+
+node_idx = 10
+captum_model = to_captum(model, mask_type="edge", node_idx=node_idx).to(device)
```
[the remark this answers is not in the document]
RBendias: No, i removed it.

torch_geometric/nn/models/explainer.py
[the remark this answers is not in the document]
RBendias: I don't like it so much either. Yes, `args[1] `contains `edge_index`. I will add `x` and `edge_index` to the `forward` as you suggested, however, the variable `x` will then not be used if `mask_type=="node"`. There is no option to add `**kwargs` with Captum only `*args`, so I guess we don't need that here or?

torch_geometric/nn/models/explainer.py
RBendias: In Captum "dimension 0 corresponds to the number of examples". I can also use unsqueeze(0) here.

examples/captum_explainability.py
RBendias: What kind of `TODO`? Change to general explainer?

rusty1s: yes

examples/captum_explainability.py
RBendias: I use the absolute value, because of the visualization, i.e. min is 0.0 (There are edges that are not part of the computational graph -> attribution of 0.0).

examples/captum_explainability.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self):
+        super().__init__()
+        self.conv1 = GCNConv(dataset.num_features, 16)
+        self.conv2 = GCNConv(16, dataset.num_classes)
+
+    def forward(self, x, edge_index):
+        x = F.relu(self.conv1(x, edge_index))
+        x = F.dropout(x, training=self.training)
+        x = self.conv2(x, edge_index)
+        return F.log_softmax(x, dim=1)
```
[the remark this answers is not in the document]
rusty1s: Ok, what does that mean in particular? Is the usage of `log_softmax` correct?

examples/captum_explainability.py
rusty1s: I guess then `abs()` is not needed, right?

test/nn/models/test_explainer.py
```diff
@@ -0,0 +1,42 @@
...
+GCN = GCN(3, 16, 2, 7, dropout=0.5)
+GAT = GAT(3, 16, 2, 7, heads=2, concat=False)
+
+mask_type = ['edge', 'node_and_edge', 'node']
+
+
+@pytest.mark.parametrize('mask_type', mask_type)
+@pytest.mark.parametrize('model', [GCN, GAT])
+@pytest.mark.parametrize('node_idx', [None, 1])
+def test_to_captum(model, mask_type, node_idx):
```
rusty1s, line 15: We do not really test `captum` support here. We could add `captum` dependency to `full_requires`, and only perform this test in case `captum` is installed (via `pytest.mark.skipif`).

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+class CaptumModel(torch.nn.Module):
+    r"""Model with forward function that can be easily used for
+    explainability with `Captum.ai <https://captum.ai/>`_.
+
+    Args:
+        model (torch.nn.Module): Model to be explained.
+        mask_type (str): Denotes the type of mask to be created with a Captum
+            explainer. Valid inputs are :obj:`'edge'`, :obj:`'node'`, and
+            :obj:`'node_and_edge'`. The input for the forward function with
+            mask_type :obj:`'edge'` should be an edge_mask tensor of shape
```
rusty1s, line 43:
```suggestion
            mask type :obj:`'edge'` should be an edge_mask tensor of shape
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+            (1, num_edges), :obj:`x` and :obj:`edge_index`. The input for the
+            forward function with mask_type :obj:`'node'` should be a
+            node_input of shape (1, num_nodes, num_features) and
```
rusty1s, line 46:
```suggestion
            node input of shape :obj:`[1, num_nodes, num_features]` and
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+    r"""Model with forward function that can be easily used for
+    explainability with `Captum.ai <https://captum.ai/>`_.
+
+    Args:
+        model (torch.nn.Module): Model to be explained.
+        mask_type (str): Denotes the type of mask to be created with a Captum
+            explainer. Valid inputs are :obj:`'edge'`, :obj:`'node'`, and
+            :obj:`'node_and_edge'`. The input for the forward function with
+            mask_type :obj:`'edge'` should be an edge_mask tensor of shape
+            (1, num_edges), :obj:`x` and :obj:`edge_index`. The input for the
```
rusty1s, line 44:
```suggestion
            :obj:`[1, num_edges]`, :obj:`x` and :obj:`edge_index`. The input for the
```

torch_geometric/nn/models/explainer.py
rusty1s, line 43:
```suggestion
            mask_type :obj:`'edge'` should be an edge mask tensor of shape
```

torch_geometric/nn/models/explainer.py
rusty1s, line 46:
```suggestion
            node_input of shape :obj:`[1, num_nodes, num_features]` and
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+            forward function with mask_type :obj:`'node'` should be a
+            node_input of shape (1, num_nodes, num_features) and
+            :obj:`edge_index`. The input for the forward function with
+            mask_type :obj:`'node_and_edge'` should be a node_input tensor of
```
rusty1s, line 48:
```suggestion
            mask_type :obj:`'node_and_edge'` should be a node input tensor of
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+            shape (1, num_nodes, num_features), an edge_mask tensor of shape
```
rusty1s, line 49:
```suggestion
            shape :obj:`[1, num_nodes, num_features]`, an edge mask tensor of shape
```

torch_geometric/nn/models/explainer.py
```diff
@@ -0,0 +1,113 @@
...
+            (1, num_edges), and :obj:`edge_index`. (default: :obj:`'edge'`)
```
rusty1s, line 50: Add a note that additional arguments are applicable as long as the order of the first arguments is guaranteed?

rusty1s (commented):
Thanks for the update. After adding a Captum test, I think this PR is ready to get merged.

examples/captum_explainability.py
RBendias: `ig_attr` can have negative values, that's why I used `abs()` -> shifting the negative attributions to positive ones.

RBendias: Min is 0.0 after applying `abs()`. Maybe that was misleading. :)

test/nn/models/test_explainer.py
RBendias: Okay, I'll add that! I wasn't sure if you wanna have Captum as a dependency.

RBendias: I added a test for most of the [attribution methods provided in Captum](https://captum.ai/api/attribution.html). FeaturePermutation, GuidedGradCam, LRP, DeepLift, and DeepLiftShap can not be applied. They either require several samples or modify the input by adding more samples. This doesn't work with our edge_mask approach, at the moment.

rusty1s (approved):

PR #4025, opened 2022-02-08
Add temporal sampling support to `NeighborLoader`

RexYing:
not sure how to use the parameterize thing yet.

I was trying 
@pytest.mark.parametrize('directed', [True, False])
@pytest.mark.parametrize('temporal', [True, False])
but didn't work. Will look into it further.


rusty1s:
> I was trying
> @pytest.mark.parametrize('directed', [True, False])
> @pytest.mark.parametrize('temporal', [True, False])
> but didn't work. Will look into it further.

This should work, but keep in mind to add both `directed` and `temporal` to the function signature.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -207,6 +225,7 @@ def __init__(
         directed: bool = True,
         transform: Callable = None,
         neighbor_sampler: Optional[NeighborSampler] = None,
+        node_time: Optional[Dict[NodeType, Tensor]] = None,
```
Padarn, line 228: Can we add a docstring about the new functionality?

RexYing: done.

torch_geometric/data/lightning_datamodule.py
```diff
@@ -201,7 +201,7 @@ class LightningNodeData(LightningDataModule):
             indices of test nodes. If not given, will try to automatically
             infer them from the :obj:`data` object. (default: :obj:`None`)
         loader (str): The scalability technique to use (:obj:`"full"`,
-            :obj:`"neighbor"`). (default: :obj:`"neighbor"`)
+            :obj:`"neighbor"`, :obj:`"temporal"`). (default: :obj:`"neighbor"`)
```
rusty1s, line 204: As discussed: Let's remove this. `time_attr` can be set within `kwargs`.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -25,19 +25,24 @@ def __init__(
...
         self.replace = replace
         self.directed = directed
+        self.node_time = None
 
         if isinstance(data, Data):
             # Convert the graph data into a suitable format for sampling.
             out = to_csc(data, device='cpu', share_memory=share_memory)
             self.colptr, self.row, self.perm = out
             assert isinstance(num_neighbors, (list, tuple))
+            # TODO: time_attr for homogeneous graphs
```
rusty1s, line 41: Is it possible to fail here, by raising a `ValueError`?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -78,18 +83,34 @@ def __call__(self, index: Union[List[int], Tensor]):
...
+                    self.row_dict,
+                    {self.input_type: index},
+                    self.num_neighbors,
+                    self.num_hops,
+                    self.replace,
+                    self.directed,
+                )
+            else:
+                sample_fn = \
+                    torch.ops.torch_sparse.hetero_neighbor_temporal_sample
```
rusty1s, line 101:
```suggestion
                    torch.ops.torch_sparse.hetero_temporal_neighbor_sample
```

torch_geometric/loader/neighbor_loader.py
```diff
@@ -25,19 +25,24 @@ def __init__(
...
         if isinstance(data, Data):
             # Convert the graph data into a suitable format for sampling.
             out = to_csc(data, device='cpu', share_memory=share_memory)
             self.colptr, self.row, self.perm = out
             assert isinstance(num_neighbors, (list, tuple))
+            # TODO: time_attr for homogeneous graphs
 
         elif isinstance(data, HeteroData):
+            if time_attr is not None:
+                self.node_time = data[time_attr]
```
rusty1s, line 45: This should be a dict, right? I think we need to do `self.node_time_dict = data.collect(time_attr)` instead.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -212,6 +233,16 @@ class NeighborLoader(torch.utils.data.DataLoader):
         transform (Callable, optional): A function/transform that takes in
             a sampled mini-batch and returns a transformed version.
             (default: :obj:`None`)
+        time_attr (str, optional): An optional argument
+            to specify the node attribute that denotes the timestamps for
+            nodes in graph.
+            If set, data[time_attr] has to be a dictionary of type
```
rusty1s, line 239: This is slightly off. Each node type sets its own `time_attr`, e.g., `data[node_type].time = ...`.

rusty1s (changes requested):
This looks pretty good! Left some last comments.

PR #4031, opened 2022-02-08
Graph Relational Attention

fork123aniket:
Added Relational Graph Attention Networks, which computes attention coefficients for relational graphs for each relation type for any number of relations and attention heads. Moreover, also provides functionality for edge attributes to handle.

### DIFF
--- a/examples/rgat_conv.py
+++ b/examples/rgat_conv.py
@@ -2,11 +2,11 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 import torch.optim as optim
-from torch_geometric.nn import RGATConv
 from torch.nn import Linear
 from torch.nn.utils import clip_grad_value_
 
 from torch_geometric.datasets import Entities
+from torch_geometric.nn import RGATConv
 
 
 class RGAT(nn.Module):
--- a/examples/rgat_conv.py
+++ b/examples/rgat_conv.py
@@ -1,0 +1,88 @@
+import torch.nn as nn
+import torch
+import torch.nn.functional as F
+from torch.nn.utils import clip_grad_value_
+from torch.nn import Linear
+
+import torch.optim as optim
+from torch_geometric.datasets import Entities
+from rgat_conv import RGATConv
+
+
+class Net(nn.Module):
+    def __init__(self, d=2):
+        super(Net, self).__init__()
+
+        self.nclass = 4
+        self.readout = "add"
+        self.in_channels = 16
+        self.out_channels = 25
+        self.d = d
+
+        self.conv_layers = nn.ModuleList([
+            RGATConv(16, 20, num_relations=90, num_bases=35, mod="additive",
+                     attention_mechanism="within-relation",
+                     attention_mode="multiplicative-self-attention",
+                     heads=2, d=2),
+            RGATConv(80, 25, num_relations=90, num_blocks=2, mod=None,
+                     attention_mechanism="across-relation",
+                     attention_mode="additive-self-attention",
+                     heads=2, d=1, dropout=0.6, edge_dim=16, bias=False),
+                     ])
+
+        self.fc = Linear(16, self.in_channels, bias=False)
+
+        self.linear = Linear(self.d * self.out_channels, self.nclass)
+
+    def forward(self, x, edge_index, edge_type, edge_attr):
+        if x.dim() == 1:
+            x = x.unsqueeze(-1)
+
+        hid_x = self.fc(x)
+
+        for yt, conv in enumerate(self.conv_layers):
+            if yt == 0:
+                hid_x = conv(hid_x, edge_index, edge_type, edge_attr=None)
+            else:
+                hid_x = conv(hid_x, edge_index, edge_type, edge_attr)
+
+        x = self.linear(hid_x)
+        return F.log_softmax(x, dim=-1)
+
+
+if __name__ == '__main__':
+    dataset = Entities(root='/tmp/AIFB', name='AIFB')
+
+    edge_index, edge_type, train_labels, test_labels, train_mask, test_mask, \
+        num_nodes = dataset[0].edge_index, dataset[0].edge_type, \
+        dataset[0].train_y, dataset[0].test_y, dataset[0].train_idx, \
+        dataset[0].test_idx, dataset[0].num_nodes
+
+    x = torch.randn(num_nodes, 16)
+    edge_attr = torch.randn(58086, 2, 3, 7, 16)
+
+    model = Net()
+    optimizer = optim.Adam(model.parameters(), lr=1e-2, weight_decay=1e-3,
+                           msgrad=False)
+
+    model.train()
+    for epoch in range(100):
+        optimizer.zero_grad()
+        output = model(x, edge_index, edge_type, edge_attr, torch.tensor([0]))
+        loss = F.nll_loss(output[train_mask], train_labels)
+        loss.backward()
+        clip_grad_value_(model.parameters(), 2.0)
+        optimizer.step()
+
+        model.eval()
+        train_acc = torch.sum(output[train_mask].argmax(dim=1) == train_labels)
+        train_acc = train_acc / len(train_mask)
+        test_acc = torch.sum(output[test_mask].argmax(dim=1) == test_labels)
+        test_acc = test_acc.item() / len(test_mask)
+
+        print('Epoch {:03d}'.format(epoch+1),
+              'train_loss: {:.4f}'.format(loss),
+              'train_acc: {:.4f}'.format(train_acc),
+              'test_acc: {:.4f}'.format(test_acc))
+
+    print("Optimization Finished!")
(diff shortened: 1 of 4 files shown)


examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, d=2):
+        super(Net, self).__init__()
+
+        self.nclass = 4
+        self.readout = "add"
+        self.in_channels = 16
+        self.out_channels = 25
+        self.d = d
+
+        self.conv_layers = nn.ModuleList([
```
rusty1s, line 22:
```suggestion
        self.convs= nn.ModuleList([
```

examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+                     attention_mechanism="within-relation",
+                     attention_mode="multiplicative-self-attention", heads=2,
+                     d=2),
+            RGATConv(80, 25, num_relations=90, num_blocks=2, mod=None,
+                     attention_mechanism="across-relation",
+                     attention_mode="additive-self-attention", heads=2, d=1,
+                     dropout=0.6, edge_dim=16, bias=False),
+        ])
+
+        self.fc = Linear(16, self.in_channels, bias=False)
```
rusty1s, line 33: The naming is a bit off. Shall we name this `self.lin1` and `self.lin2`?
```suggestion
        self.lin1 = Linear(16, self.in_channels, bias=False)
```

examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+from torch.nn.utils import clip_grad_value_
+
+from torch_geometric.datasets import Entities
+
+
+class Net(nn.Module):
+    def __init__(self, d=2):
+        super(Net, self).__init__()
+
+        self.nclass = 4
```
rusty1s, line 16: Shall we convert these to input arguments instead of attributes?

examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+                hid_x = conv(hid_x, edge_index, edge_type, edge_attr)
+
+        x = self.linear(hid_x)
+        return F.log_softmax(x, dim=-1)
+
+
+if __name__ == '__main__':
+    dataset = Entities(root='/tmp/AIFB', name='AIFB')
+
+    edge_index, edge_type, train_labels, test_labels, train_mask, test_mask, \
```
rusty1s, line 56:
```python
data = dataset[0]
...
```
should make this easier to read.

torch_geometric/nn/conv/rgat_conv.py
```diff
@@ -0,0 +1,558 @@
...
+        zeros(self.b1)
+        torch.full(self.l2.size(), 1 / self.out_channels)
+        zeros(self.b2)
+        if self.lin_edge is not None:
+            glorot(self.lin_edge)
+            glorot(self.e)
+
+    def forward(self, x: Tensor, edge_index: Adj, edge_type: OptTensor = None,
+                edge_attr: OptTensor = None, size: Size = None,
+                return_attention_weights=None):
```
rusty1s, line 345:
```suggestion
                return_attention_weights: bool = False):
```

torch_geometric/nn/conv/rgat_conv.py
```diff
@@ -0,0 +1,558 @@
...
+        r"""
+        Args:
```
rusty1s, line 347: The arguments here do not match with the `forward` signature.

torch_geometric/nn/conv/rgat_conv.py
```diff
@@ -0,0 +1,558 @@
...
+                Should be only :obj:`None` in case :obj:`edge_index` is of type
+                :class:`torch_sparse.tensor.SparseTensor`.
+                (default: :obj:`None`)
+            return_attention_weights (bool, optional): If set to :obj:`True`,
+                will additionally return the tuple
+                :obj:`(edge_index, attention_weights)`, holding the computed
+                attention weights for each edge. (default: :obj:`None`)
+        """
+
+        if self.add_self_loops:
```
rusty1s, line 363: We should remove this argument here. Adding self-loops will throw errors later on since only `edge_attr` is modified but not `edge_type`.

torch_geometric/nn/conv/rgat_conv.py
```diff
@@ -0,0 +1,558 @@
...
+            alpha = alpha  # original
+
+        if self.attention_mode == "additive-self-attention":
+            return alpha.view(-1, self.heads, 1) * outj.view(
+                -1, self.heads, self.out_channels)
+        elif self.attention_mode == "multiplicative-self-attention":
+            return (alpha.view(-1, self.heads, self.d, 1) *
+                    outj.view(-1, self.heads, 1, self.out_channels))
+
+    def update(self, aggr_out: Tensor):
```
rusty1s, line 531:
```suggestion
    def update(self, aggr_out: Tensor) -> Tensor:
```

torch_geometric/nn/conv/rgat_conv.py
```diff
@@ -0,0 +1,558 @@
...
+            else:
+                aggr_out = aggr_out.mean(dim=1)
+                aggr_out = aggr_out.view(-1, self.d * self.out_channels)
+
+            if self.bias is not None:
+                aggr_out = aggr_out + self.bias
+
+            return aggr_out
+
+    def __repr__(self):
```
rusty1s, line 555:
```suggestion
    def __repr__(self) -> str:
```

rusty1s (commented):
This is great. Left a few comments.

Note that we also need to add `RGATConv` to `nn/conv/__init__.py`. Furthermore, it would be great if you can add some tests for `RGATConv` in `/test/nn/conv/test_rgat_conv.py` (similar to other tests in that folder). Let me know if you need any help for this!

examples/rgat_conv.py
fork123aniket: Already included this line in the code, as can be seen in line 50.

examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+    def forward(self, x, edge_index, edge_type, edge_attr):
+        if x.dim() == 1:
+            x = x.unsqueeze(-1)
+
+        hid_x = self.fc(x)
+
+        for yt, conv in enumerate(self.conv_layers):
+            if yt == 0:
+                hid_x = conv(hid_x, edge_index, edge_type, edge_attr=None)
+            else:
```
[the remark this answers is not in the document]
fork123aniket: Have just included the relevant comments related to "edge_attr" usage in the used GNN module, which can be seen on lines 38-41.

torch_geometric/nn/conv/rgat_conv.py
fork123aniket: This line (attention weights will be returned whenever "return_attention_weights" is set to a value, regardless of its actual value (might be `True` or `False`)) is taken from [https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/gat_conv.py](url). Hence, I've to set "return_attention_weights" to None in order to allow the layer not to return any attention weight (by default).

examples/rgat_conv.py
```diff
@@ -0,0 +1,88 @@
...
+
+    edge_index, edge_type, train_labels, test_labels, train_mask, test_mask, \
+        num_nodes = dataset[0].edge_index, dataset[0].edge_type, \
+        dataset[0].train_y, dataset[0].test_y, dataset[0].train_idx, \
+        dataset[0].test_idx, dataset[0].num_nodes
+
+    x = torch.randn(num_nodes, 16)
+    edge_attr = torch.randn(58086, 2, 3, 7, 16)
+
+    model = Net()
```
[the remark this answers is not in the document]
fork123aniket: Although I ran this experiment on Intel 10th Gen core i5 processor, this implementation will also work for GPUs while employing Torch CUDA.

fork123aniket:
@rusty1s  All the issues raised have been addressed. Two issues above are still open with comments made by me for each of them. Please read them and let me know if any further modification somewhere is required. Furthermore, have also added `RGATConv` to `nn/conv/__init__.py.`

examples/rgat_conv.py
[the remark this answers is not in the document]
rusty1s: Ok, so let's add a
```python
```
call here, similar to [other examples](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/gcn.py#L48) in PyG.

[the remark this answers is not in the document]
fork123aniket: Have just added this `device` line in `rgat_conv.py (example)` on line 63.

fork123aniket:
Have made the text of formulas non-italic in the documentation (https://pytorch-geometric--4031.org.readthedocs.build/en/4031/modules/nn.html#torch_geometric.nn.conv.RGATConv). Furthermore, have also added few tests, as can be seen in `test_rgat_conv.py` (committed to this pull request just now).

rusty1s:
Thanks you for your updates and being so responsive! I will have some last review today or tomorrow and will try to quickly merge afterwards.

rusty1s (approved):
Thanks for the updates. LGTM.

PR #4043, opened 2022-02-09
Fix `BasicGNN` with `return_attention_weights=True`

josephenguehard:
Fix error when using `return_attention_weights=True` with a `GAT`.

Sample of error:

```python
import torch
from torch_geometric.nn import GAT

edge_index = torch.tensor([[0, 1, 1, 2], [1, 0, 2, 1]], dtype=torch.long)
x = torch.tensor([[-1], [0], [1]], dtype=torch.float)
data = Data(x=x, edge_index=edge_index)

gat = GAT(in_channels=1, hidden_channels=5, num_layers=2)
gat.forward(data.x, data.edge_index, return_attention_weights=True)
>>> "TypeError: relu_(): argument 'input' (position 1) must be Tensor, not tuple"
```

After the fix:
```python
output, atts = gat.forward(data.x, data.edge_index, return_attention_weights=True)
outputs.shape
>>> "torch.Size([3, 5])"

### CODE REVIEW

File: torch_geometric/nn/models/basic_gnn.py [hunk 1]
```diff
@@ -118,11 +118,16 @@ def reset_parameters(self):
         if hasattr(self, 'lin'):
             self.lin.reset_parameters()
 
-    def forward(self, x: Tensor, edge_index: Adj, *args, **kwargs) -> Tensor:
+    def forward(self, x: Tensor, edge_index: Adj, *args,
```
rusty1s, line 121: I'm wondering if we solely should add this feature to `GAT` rather than `BasicGNN`. Alternatively, we could make `GAT` inherit from `BasicAttentionalGNN` that again inherits from `BasicGNN`. IMO, I'm not that comfortable adding this feature to every GNN.

Alternatively, on could add a post forward hook to all `GATConv` operators and aggregate the results there. This does not require modifications to `forward`.

It might be also a good idea to document this feature inside the doc-string of `forward`.

rusty1s (commented):
Thank you. I think the main idea is good, but we might want to refactor it a bit.

torch_geometric/nn/models/basic_gnn.py
josephenguehard: I understand the issue regarding implementing GAT-specific components in `BasicGNN`. Creating a `BasicAttentionalGNN` seems to be an overkill and not sure how to use hooks here, so what I did is to modify the forward method of `GATConv`, which now only returns `out`. It also has now a getter to return attention weights based on the latest `forward` call. 

Not sure if this is the best way to do it, let me know what you think! Also, we could then get rid of the `return_attention_weights` argument in the `forward` function. I also don't understand the note regarding this argument: https://github.com/pyg-team/pytorch_geometric/blob/d1aa1837b36dceed9349ec083bba807ac3f58f3f/torch_geometric/nn/conv/gat_conv.py#L189

josephenguehard: P.S. `test_gat_conv` fails for now, I'll modify it if you think this is the good way to go

Padarn: I like the idea of just adding this to `GAT` personally, it seems like a very specific use case (actually I think it would be even okay to simple have)
```
def forward(x, **kwargs):
     if "return_attention_weights" in kawargs:
           raise NotImplementedException("...")
    super().forward(x, **kwargs)
```

inside of the GAT class.

rusty1s: I agree. I think the best solution would be to leverage `forward` hooks on each `GATConv` module in `GAT` which collects all attention weights of all layers, and additionally returns them in case `return_attention_weights=True`.

PR #4065, opened 2022-02-11
Minor improvement to FakeDataset

arunppsg:
I have made a small improvement to `FakeDataset` class.

The node features and edge features generated have same distribution across different labels. I made a small fix to make them have different mean across different labels. This will help when the fake dataset is used for training and testing models. Earlier, since distributions of node features and edge features where same across different labels, a model could not learn effectively.


rusty1s (approved):
LGTM. Thank you!

arunppsg:
Some tests are failing at an assertion for `data.edge_weight` and `data.edge_attr` because the change introduced create a variance in both these features (due to dependency on graph labels during random value generation). 

I am not sure how to handle it here.

rusty1s:
How about we only offset node features?

arunppsg:
> How about we only offset node features?

Sounds good. In the last update I made offset to only node features and it passes the test case. Thanks!

rusty1s:
Super :)

PR #4068, opened 2022-02-11
Feature/add mtl example

sebastian-hatala-leanix:
With this PR I propose to add an example for Multi Task Learning (see https://github.com/pyg-team/pytorch_geometric/discussions/3966)

### DIFF
--- a/examples/ppi_mtl.py
+++ b/examples/ppi_mtl.py
@@ -1,0 +1,96 @@
+import os.path as osp
+
+import torch
+import torch.nn.functional as F
+from sklearn.metrics import f1_score
+
+from torch_geometric.datasets import PPI
+from torch_geometric.loader import DataLoader
+from torch_geometric.nn import BatchNorm, GCNConv
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'PPI')
+train_dataset = PPI(path, split='train')
+val_dataset = PPI(path, split='val')
+test_dataset = PPI(path, split='test')
+train_loader = DataLoader(train_dataset, batch_size=1, shuffle=True)
+val_loader = DataLoader(val_dataset, batch_size=2, shuffle=False)
+test_loader = DataLoader(test_dataset, batch_size=2, shuffle=False)
+
+
+class Net(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+        self.body_convs = nn.ModuleList()
+        self.body_norms = nn.ModuleList()
+
+        for layer in range(2):
+            in_channels = train_dataset.num_features if layer == 0 else 256
+            self.body_convs.append(GCNConv(in_channels, 256))
+            self.body_norms.append(BatchNorm(256))
+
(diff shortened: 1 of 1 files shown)


wsad1:
@sebastian-hatala-leanix Is this very different from the example [here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/ppi.py) other than having some more exclusive parameters between tasks?. Please let me know if I am missing something.

sebastian-hatala-leanix:
@wsad1 as you mentioned these two are very similar - in fact I copied the `ppi.py` and only added the body-head architecture for Multi Task Learning.

The result is also not much different. But after all this is meant to be a blueprint to those who want to tackle a problem with multiple tasks.

To give an example when this architecture is actually useful: We had a setting with three tasks of which one was a binary classification and two multi-class classifications. There having exclusive parameters and being able to use a different loss function per task led to the best result (as opposed to this problem where there is only binary tasks) 💪

rusty1s:
@sebastian-hatala-leanix Thanks for the example. Is there any chance we can enhance this example with a dataset for which one needs to use different loss functions? While I like to include an example for multi-task learning, the current solution is obviously less efficient than just computing outputs at once, leading to showcase „bad-practices“.

sebastian-hatala-leanix:
Jup fair point, I'll have another look at the datasets and give it another go 👍

rusty1s:
I‘ll be happy to discuss more. Feel free to leave this PR open.

sebastian-hatala-leanix:
Thanks! So I can think of two areas of problems where MTL with GCN is really efficient & effective:
1. When predicting a set of independent targets given a common input (e. g. generating metadata from pictures like [here](https://towardsdatascience.com/tuning-a-multi-task-fate-grand-order-trained-pytorch-network-152cfda2e086))
2. When trying to generate embeddings so that nodes are semantically comparable.

I am not sure yet, if any of the `PPI`, `Yelp`, `QM7b`, `QM9` datasets - which are the ones with more than one task - are really suited for this. If not I'll look beyond the data in PyG - maybe someone has a recommendation?

I'll reopen this once we found a good match, how does that sound?

sebastian-hatala-leanix:
**Update** I think I found two approaches to this I'd like to pursue:

1. Given a dataset with a single classification task one could add one head that tries to reconstruct the (sub-)graph using an autoencoder. This could work on e.g. the `IMDb` dataset.
2. Another interesting problem for the approach could be Knowledge Graphs like `WordNet18RR` one head could learn to predict the relation and another one to predict the fact.

Sounds a bit like I have a hammer and am looking for a nail here. But I guess that is sort of what is happening here...😅

rusty1s:
I think those are good options. Alternatively, one can have a look at related literature for applying Graph Neural Networks in a multi-task learning scenario, *e.g.*, [this one](https://arxiv.org/pdf/1809.07695.pdf) or [this one](https://grlearning.github.io/papers/132.pdf), where additional tasks could involve learning some graph statistics.

PR #4074, opened 2022-02-14
added kwargs to fakedataset

arunppsg:
Fix #4067. Changes made in this PR allows users to add additional features which can be generated via `FakeDataset`.

### DIFF
--- a/torch_geometric/datasets/fake.py
+++ b/torch_geometric/datasets/fake.py
@@ -36,7 +36,8 @@ class FakeDataset(InMemoryDataset):
             an :obj:`torch_geometric.data.Data` object and returns a
             transformed version. The data object will be transformed before
             every access. (default: :obj:`None`)
-        **kwargs (optional): Additional attributes and their shapes *e.g.* :obj:`{'global_features':5}`.
+        **kwargs (optional): Additional attributes and their shapes
+            *e.g.* :obj:`{'global_features':5}`.
     """
     def __init__(
         self,
@@ -131,7 +132,8 @@ class FakeHeteroDataset(InMemoryDataset):
             an :obj:`torch_geometric.data.HeteroData` object and returns a
             transformed version. The data object will be transformed before
             every access. (default: :obj:`None`)
-        **kwargs (optional): Additional attributes and their shapes *e.g.* :obj:`{'global_features':5}`.
+        **kwargs (optional): Additional attributes and their shapes
+            *e.g.* :obj:`{'global_features':5}`.
     """
     def __init__(
         self,
--- a/torch_geometric/datasets/fake.py
+++ b/torch_geometric/datasets/fake.py
@@ -36,14 +36,22 @@ class FakeDataset(InMemoryDataset):
             an :obj:`torch_geometric.data.Data` object and returns a
             transformed version. The data object will be transformed before
             every access. (default: :obj:`None`)
-        **kwargs (optional): Additional attributes and their shapes.
+        **kwargs (optional): Additional attributes and their shapes *e.g.* :obj:`{'global_features':5}`.
     """
-    def __init__(self, num_graphs: int = 1, avg_num_nodes: int = 1000,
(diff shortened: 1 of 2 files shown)


torch_geometric/datasets/fake.py
```diff
@@ -36,20 +36,14 @@ class FakeDataset(InMemoryDataset):
...
-        is_undirected: bool = True,
-        transform: Optional[Callable] = None,
-        pre_transform: Optional[Callable] = None,
-    ):
+    def __init__(self, num_graphs: int = 1, avg_num_nodes: int = 1000,
+                 avg_degree: int = 10, num_channels: int = 64,
+                 edge_dim: int = 0, num_classes: int = 10, task: str = "auto",
+                 is_undirected: bool = True,
+                 transform: Optional[Callable] = None,
+                 pre_transform: Optional[Callable] = None, **kwargs):
```
rusty1s, line 46:
```suggestion
                 pre_transform: Optional[Callable] = None, **kwargs,):
```

torch_geometric/datasets/fake.py
```diff
@@ -36,20 +36,14 @@ class FakeDataset(InMemoryDataset):
             an :obj:`torch_geometric.data.Data` object and returns a
             transformed version. The data object will be transformed before
             every access. (default: :obj:`None`)
+        **kwargs (optional): Additional attributes and their shapes.
```
rusty1s, line 39: Can we give an example here? For example,
```
**kwargs (optional): Additional attributes and their shapes, *e.g.*, :obj:`global_feat=(1, 128)`
```

torch_geometric/datasets/fake.py
```diff
@@ -64,10 +58,12 @@ def __init__(
...
         self.is_undirected = is_undirected
 
-        data_list = [self.generate_data() for _ in range(max(num_graphs, 1))]
+        data_list = [
+            self.generate_data(**kwargs) for _ in range(max(num_graphs, 1))
+        ]
         self.data, self.slices = self.collate(data_list)
 
-    def generate_data(self) -> Data:
+    def generate_data(self, **kwargs) -> Data:
```
rusty1s, line 66:
```suggestion
    def generate_data(self) -> Data:
```

torch_geometric/datasets/fake.py
```diff
@@ -93,6 +89,9 @@ def generate_data(self) -> Data:
         elif self.edge_dim == 1:
             data.edge_weight = torch.rand(data.num_edges)
 
+        for feature_name, feature_shape in kwargs.items():
```
rusty1s, line 92:
```suggestion
        for feature_name, feature_shape in self.kwargs.items():
```
Let's use `kwargs` as a property (similar to all other variables).

rusty1s (approved):
LGTM. Thank you! Left a few comments.

torch_geometric/datasets/fake.py
```diff
@@ -36,6 +36,7 @@ class FakeDataset(InMemoryDataset):
             an :obj:`torch_geometric.data.Data` object and returns a
             transformed version. The data object will be transformed before
             every access. (default: :obj:`None`)
+        **kwargs (optional): Additional attributes and their shapes *e.g.* :obj:`{'global_features':5}`.
```
rusty1s: This is still a keyword arg, right? So `global_feature=5` should be correct.

arunppsg: Yeah, I agree.

arunppsg:
> Thanks for the updates. One minor thing: Can we add a test for this as well?

Yeah, I was also thinking about the same. Can we overload `test_fake_dataset` or create a new one `test_fake_dataset_kwargs`? We can check for the shape of attributes.

rusty1s:
I think you can just add the test in the already existing test method :)

arunppsg:
Thanks for your suggestions, I added tests to the fakedataset.

rusty1s:
Awesome, thank you.

PR #4083, opened 2022-02-16
`LightGCN`: Initialize embeddings via `xavier_uniform`

nishithshowri006:
I observed that when we use Xavier method that mentioned in Original LightGCN paper the performance increases than the current model.


rusty1s (approved):
Thank you!

PR #4090, opened 2022-02-17
Add general `Explainer` Class

RBendias:
As discussed in #3858, this includes the first draft of a general `Explainer` class. This can be merged into #2576. And then we follow with an update of `PGExplainer`.

### DIFF
--- a/torch_geometric/nn/models/__init__.py
+++ b/torch_geometric/nn/models/__init__.py
@@ -10,7 +10,7 @@ from .graph_unet import GraphUNet
 from .schnet import SchNet
 from .dimenet import DimeNet
 from .gnn_explainer import GNNExplainer
-from .explainer import to_captum
+from .explainer import Explainer, to_captum
 from .metapath2vec import MetaPath2Vec
 from .deepgcn import DeepGCNLayer
 from .tgn import TGNMemory
@@ -42,6 +42,7 @@ __all__ = [
     'SchNet',
     'DimeNet',
     'GNNExplainer',
+    'Explainer',
     'to_captum',
     'MetaPath2Vec',
     'DeepGCNLayer',
--- a/torch_geometric/nn/models/__init__.py
+++ b/torch_geometric/nn/models/__init__.py
@@ -9,8 +9,8 @@ from .re_net import RENet
 from .graph_unet import GraphUNet
 from .schnet import SchNet
 from .dimenet import DimeNet
-from .gnn_explainer import GNNExplainer
 from .explainer import Explainer, to_captum
+from .gnn_explainer import GNNExplainer
 from .metapath2vec import MetaPath2Vec
 from .deepgcn import DeepGCNLayer
 from .tgn import TGNMemory
@@ -41,9 +41,9 @@ __all__ = [
     'GraphUNet',
     'SchNet',
     'DimeNet',
-    'GNNExplainer',
     'Explainer',
     'to_captum',
+    'GNNExplainer',
     'MetaPath2Vec',
     'DeepGCNLayer',
     'TGNMemory',
--- a/torch_geometric/nn/models/explainer.py
+++ b/torch_geometric/nn/models/explainer.py
@@ -153,8 +153,8 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
 class Explainer(torch.nn.Module):
     r"""An abstract class for integrating explainability into Graph Neural
     Networks, *e.g.* :class:`~torch_geometric.nn.GNNExplainer` and
-    :class:`~torch_geometric.nn.PGExplainer`. It also provides a general
-    visualization methods for graph attributions.
+    :class:`~torch_geometric.nn.PGExplainer`.
+    It also provides general visualization methods for graph attributions.
 
     Args:
         model (torch.nn.Module): The GNN module to explain.
@@ -201,8 +201,9 @@ class Explainer(torch.nn.Module):
 
     def subgraph(self, node_idx: int, x: Tensor, edge_index: Tensor, **kwargs):
         r"""Returns the subgraph of the given node.
+
         Args:
-            node_idx (int): The prediction for node idx to explain.
+            node_idx (int): The node to explain.
             x (Tensor): The node feature matrix.
             edge_index (LongTensor): The edge indices.
             **kwargs (optional): Additional arguments passed to the GNN module.
@@ -251,18 +252,17 @@ class Explainer(torch.nn.Module):
             loss = self._loss(log_logits, prediction, node_idx, **kwargs)
         return loss
 
-    def visualize_subgraph(self, node_idx: int, edge_index: Tensor,
+    def visualize_subgraph(self, node_idx: Optional[int], edge_index: Tensor,
                            edge_mask: Tensor, y: Optional[Tensor] = None,
                            threshold: Optional[int] = None,
                            edge_y: Optional[Tensor] = None,
                            node_alpha: Optional[Tensor] = None, seed: int = 10,
                            **kwargs):
-        r"""Visualizes the subgraph given an edge mask
-        :attr:`edge_mask`.
+        r"""Visualizes the subgraph given an edge mask :attr:`edge_mask`.
 
         Args:
             node_idx (int): The node id to explain.
-                Set to :obj:`-1` to explain graph.
+                Set to :obj:`None` to explain a graph.
             edge_index (LongTensor): The edge indices.
             edge_mask (Tensor): The edge mask.
             y (Tensor, optional): The ground-truth node-prediction labels used
@@ -287,7 +287,7 @@ class Explainer(torch.nn.Module):
 
         assert edge_mask.size(0) == edge_index.size(1)
 
-        if node_idx == -1:
+        if node_idx is None or node_idx < 0:
             hard_edge_mask = torch.BoolTensor([True] * edge_index.size(1),
                                               device=edge_mask.device)
             subset = torch.arange(edge_index.max().item() + 1,
--- a/torch_geometric/nn/models/explainer.py
+++ b/torch_geometric/nn/models/explainer.py
@@ -1,4 +1,3 @@
-from abc import abstractmethod
 from inspect import signature
 from math import sqrt
 from typing import Optional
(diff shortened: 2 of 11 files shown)


torch_geometric/nn/models/__init__.py
```diff
@@ -42,6 +42,7 @@
     'SchNet',
     'DimeNet',
     'GNNExplainer',
```
rusty1s, line 44: Let's swap the order of `GNNExplainer` and `Explainer`.

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+            self._num_hops = num_hops
+        else:
+            # Could this be a general utils function?
+            k = 0
+            for module in self.model.modules():
+                if isinstance(module, MessagePassing):
+                    k += 1
+            self._num_hops = k
+
+    def __flow__(self):
```
rusty1s, line 189:
```suggestion
    @property
    def _flow(self):
```
(just a suggestion)

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
             the index specified. (default: :obj:`None`)
     """
     return CaptumModel(model, mask_type, output_idx)
+
+
+class Explainer(torch.nn.Module):
+    r"""Base Explainer class for Graph Neural Networks, e.g.
+    PGExplainer and GNNExplainer."""
```
rusty1s, line 156:
```suggestion
    :class:`~torch_geometric.nn.GNNExplainer and :class:`~torch_geometric.nn.PGExplainer`."""
```

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
             the index specified. (default: :obj:`None`)
     """
     return CaptumModel(model, mask_type, output_idx)
+
+
+class Explainer(torch.nn.Module):
+    r"""Base Explainer class for Graph Neural Networks, e.g.
```
rusty1s, line 155:
```suggestion
    r"""Base Explainer class for Graph Neural Networks, *e.g.*,
```

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+    PGExplainer and GNNExplainer."""
+
+    # TODO: Change docstring
+    # TODO: Rearange args
```
rusty1s, line 159: why? :)

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+            self._num_hops = k
+
+    def __flow__(self):
+        for module in self.model.modules():
+            if isinstance(module, MessagePassing):
+                return module.flow
+        return 'source_to_target'
+
+    # TODO: Find out, whether to change variable name node_idx here as well?
+    def __subgraph__(self, node_idx, x, edge_index, **kwargs):
```
rusty1s, line 196:
```suggestion
    def subgraph(self, node_idx, x, edge_index, **kwargs):
```
Let's expose this function.

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+                    k += 1
+            self._num_hops = k
+
+    def __flow__(self):
+        for module in self.model.modules():
+            if isinstance(module, MessagePassing):
+                return module.flow
+        return 'source_to_target'
+
+    # TODO: Find out, whether to change variable name node_idx here as well?
```
rusty1s, line 195: I don't think so. Currently, we can only collect the subgraph around a node, not around a link.

torch_geometric/nn/models/gnn_explainer.py
```diff
@@ -95,65 +85,33 @@ def __set_masks__(self, x, edge_index, init="normal"):
             self.node_feat_mask = torch.nn.Parameter(torch.randn(1, F) * std)
 
         std = torch.nn.init.calculate_gain('relu') * sqrt(2.0 / (2 * N))
+
+        # Do we need to initialize the edge mask for allow_edge_mask == False?
```
rusty1s, line 89: Yes, we shouldn't initialize it in this case.

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,213 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+                kwargs_new[key] = value[edge_mask]
+            kwargs_new[key] = value  # TODO: this is not in PGExplainer
+        return x, edge_index, mapping, edge_mask, subset, kwargs_new
+
+    def __to_log_prob__(self, x):
+        x = x.log_softmax(dim=-1) if self.return_type == 'raw' else x
+        x = x.log() if self.return_type == 'prob' else x
+        return x
+
+    def get_initial_prediction(self, x, edge_index, batch=None, **kwargs):
```
rusty1s, line 225: Use a decorator here: `@torch.no_grad()`.

rusty1s (commented):
LGTM, thanks. It would be great if we can re-place all `__*__` functions with `_*` names, *e.g.*, `__loss__` to `_loss`. This is bad-practice introduced by me :)

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +149,211 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+        else:
+            out = self.model(x, edge_index, **kwargs)
+        if self.return_type == 'regression':
+            prediction = out
+        else:
+            log_logits = self._to_log_prob(out)
+            prediction = log_logits.argmax(dim=-1)
+        return prediction
+
+    def get_loss(self, out, prediction, node_idx=None, **kwargs):
```
RBendias, line 237: Is `node_idx` better `-1` or `None` in graph classification cases?

rusty1s: I prefer `None`.

torch_geometric/nn/models/gnn_explainer.py
```diff
@@ -95,71 +85,34 @@ def __set_masks__(self, x, edge_index, init="normal"):
...
-        set_masks(self.model, self.edge_mask, edge_index, apply_sigmoid=True)
+        if self.allow_edge_mask:
+            self.edge_mask = torch.nn.Parameter(torch.randn(E) * std)
 
-    def __clear_masks__(self):
+    def _clear_masks(self):
         module = clear_masks(self.model)
         self.node_feat_masks = None
         self.edge_mask = None
         module.loop_mask = None
```
RBendias, line 96: I could not find `module.loop_mask` anywhere else in the code. So can we remove this line?

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +148,226 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+            :obj:`model`. Valid inputs are :obj:`"log_prob"` (the model
+            returns the logarithm of probabilities), :obj:`"prob"` (the
+            model returns probabilities), :obj:`"raw"` (the model returns raw
+            scores) and :obj:`"regression"` (the model returns scalars).
+            (default: :obj:`"log_prob"`)
+        log (bool, optional): If set to :obj:`False`, will not log any learning
+            progress. (default: :obj:`True`)
+    """
+
+    # TODO: Change docstring
```
rusty1s, line 181: remove TODO

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +148,226 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+
+        self.model = model
+        self.lr = lr
+        self.epochs = epochs
+        self.num_hops = num_hops
+        self.return_type = return_type
+        self.log = log
+
+    @property
+    def num_hops(self):
```
rusty1s, line 196: IMO, `num_hops()` should either return `self._num_hops` or compute the number of hops based on the number of message passing layers, right? And the setter simply sets `self._num_hops`.

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +148,226 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+        if num_hops is not None:
+            self._num_hops = num_hops
+        else:
+            k = 0
+            for module in self.model.modules():
+                if isinstance(module, MessagePassing):
+                    k += 1
+            self._num_hops = k
+
+    def _flow(self):
```
rusty1s, line 210:
```suggestion
    def _flow(self) -> str:
```

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +148,226 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+
+    def get_loss(self, out, prediction, node_idx=None, **kwargs):
+        if self.return_type == 'regression':
+            loss = self._loss(out, prediction, node_idx, **kwargs)
+        else:
+            log_logits = self._to_log_prob(out)
+            loss = self._loss(log_logits, prediction, node_idx, **kwargs)
+        return loss
+
+    def visualize_subgraph(self, node_idx, edge_index, edge_mask, y=None,
```
rusty1s, line 266: Add type hints

torch_geometric/nn/models/gnn_explainer.py
rusty1s: Yes, I guess it refers to `MessagePassing._loop_mask`.

torch_geometric/nn/models/explainer.py
```diff
@@ -144,3 +148,226 @@ def to_captum(model: torch.nn.Module, mask_type: str = "edge",
...
+        assert return_type in ['log_prob', 'prob', 'raw', 'regression']
+
+        self.model = model
+        self.lr = lr
+        self.epochs = epochs
+        self.num_hops = num_hops
+        self.return_type = return_type
+        self.log = log
+
+    @property
```
wsad1, line 195: Please don't let this comment stop you from merging, it's just a nitpick.

Do we need to have `num_hops` as a property of the class with a `setter`. At the moment the `setter` does prevent the user from setting `_num_hops` to `None` other than that it doesn't seem to serve much of a purpose.
Instead couldn't we just do something like `self.num_hops = num_hops if num_hops is not None else get_hops(module)` where `get_hops` is some utility function like @RBendias  suggested. Please let me know if I am missing something.

RBendias: Thanks for your comment! :D  I see what you mean.
@rusty1s Should we add `get_hops` to utils or leave it as a method in `Explainer`?

RBendias: I added `get_num_hops` to utils/subgraph, if I should put it somewhere else or in case this was not the solution you had in mind, let me know.

Issue #3996, opened 2022-02-03
Eigenvalues in ChebConv


daniel-unyi-42:
### 🚀 The feature, motivation and pitch

In the forward function of ChebConv, there's this error message:

```
if self.normalization != 'sym' and lambda_max is None:
            raise ValueError('You need to pass `lambda_max` to `forward() in`'
                             'case the normalization is non-symmetric.')
```

Based on the Gershgorin circle theorem, 2.0 is the upper bound of eigenvalues for **both** normalizations ('sym' and 'rw').
Furthermore, 2.0 * edge_weight.max() is the upper bound of eigenvalues for the non-normalized Laplacian.

I'd suggest to remove this error message for 'rw' normalization and consider to remove it for non-normalization.
It makes the usage of ChebConv way easier, especially for the regular user.

rusty1s:
Sounds good. Let me know if you want to contribute this feature.

daniel-unyi-42:
I'd be glad to contribute.

PR #4106, opened 2022-02-21
Automatic eigenvalue approximation in ChebConv

daniel-unyi-42:
I modified ChebConv according to the issue [Eigenvalues in ChebConv](https://github.com/pyg-team/pytorch_geometric/issues/3996) and updated the documentation as well.


torch_geometric/nn/conv/cheb_conv.py
```diff
@@ -113,6 +111,13 @@ def __norm__(self, edge_index, num_nodes: Optional[int],
                                                 normalization, dtype,
                                                 num_nodes)
 
+        if lambda_max is None:
+            lambda_max = 2.0 * edge_weight
```
rusty1s, line 115: Shouldn't this be?
```python
lambda_max = torch.tensor(2.0, dtype=dtype, device=edge_index.device)
```

torch_geometric/nn/conv/cheb_conv.py
```diff
@@ -113,6 +111,13 @@ def __norm__(self, edge_index, num_nodes: Optional[int],
...
+        if not isinstance(lambda_max, torch.Tensor):
```
rusty1s, line 116:
```suggestion
        elif not isinstance(lambda_max, torch.Tensor):
```

torch_geometric/nn/conv/cheb_conv.py
```diff
@@ -113,6 +111,13 @@ def __norm__(self, edge_index, num_nodes: Optional[int],
...
+            lambda_max = torch.tensor(lambda_max, dtype=edge_weight.dtype,
```
rusty1s, line 117:
```suggestion
            lambda_max = torch.tensor(lambda_max, dtype=dtype,
```

torch_geometric/nn/conv/cheb_conv.py
```diff
@@ -113,6 +111,13 @@ def __norm__(self, edge_index, num_nodes: Optional[int],
...
+                                      device=edge_weight.device)
```
rusty1s, line 118:
```suggestion
                                      device=edge_index.device)
```

torch_geometric/nn/conv/cheb_conv.py
daniel-unyi-42: ### Sorry, I made a mistake. It should be:
`lambda_max = 2.0 * edge_weight.max()`
If normalization is 'sym' or 'rw' then `edge_weight.max()` equals 1.

torch_geometric/nn/conv/cheb_conv.py
daniel-unyi-42: I preserved this if statement because the original code included it.
But since `lambda_max` is an OptTensor, it is either `None` or an instance of `torch.Tensor`.
So I think it could be deleted.

rusty1s (approved):
Thank you!

PR #4114, opened 2022-02-22
Spectral Modularity Pool layer

fork123aniket:
Implementation of spectral modularity pooling that not only coarsen the graphs but can also be used to cluster the nodes of any graph.

### DIFF
--- a/torch_geometric/nn/dense/dense_dmon_pool.py
+++ b/torch_geometric/nn/dense/dense_dmon_pool.py
@@ -1,0 +1,148 @@
+import torch
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.dense.linear import Linear
+from torch.nn import SELU, Dropout, Sequential
+
+EPS = 1e-15
+
+
+class DMonPooling(MessagePassing):
+    r"""Spectral modularity pooling operator from the `"Graph Clustering with
+    Graph Neural Networks" <https://arxiv.org/abs/2006.16904>`_ paper
+
+    .. math::
+        \mathbf{X}^{\prime} &= {\mathrm{softmax}(\mathbf{S})}^{\top} \cdot
+        \mathbf{X}
+
+        \mathbf{A}^{\prime} &= {\mathrm{softmax}(\mathbf{S})}^{\top} \cdot
+        \mathbf{A} \cdot \mathrm{softmax}(\mathbf{S})
+
+    based on dense learned assignments :math:`\mathbf{S} \in \mathbb{R}^{B
+    \times N \times C}`.
+    Returns learned cluster assignment matrix, pooled node feature matrix,
+    coarsened symmetrically normalized adjacency matrix, and two auxiliary
+    objectives: (1) The spectral loss
+
+    .. math::
+        \mathcal{L}_s = - \frac{1}{2m}
+        \cdot{\mathrm{Tr}(\mathbf{S}^{\top} \mathbf{B} \mathbf{S})}
+
+    where :math:`\mathbf{B}` is the modularity matrix, (2) the orthogonality
+    loss
+
+    .. math::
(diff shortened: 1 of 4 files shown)


torch_geometric/nn/dense/dense_dmon_pool.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Args:
+        dense_size (list): A 2D list whose each 1D list contains size for each
+            Linear layer being used.
+        k (int, optional): Number of clusters.
+            (default: :obj:`16`)
+        p (float, optional): Probability value for dropout layer.
+            (default: :obj:`0.2`)
+    """
+    def __init__(self, dense_size, k=16, p=0.2):
```
rusty1s, line 53: Type hints :)

torch_geometric/nn/dense/dense_dmon_pool.py
```diff
@@ -0,0 +1,150 @@
...
+        super().__init__()
+        self.dense_size = dense_size
+        self.k = k
+        self.dropout_rate = p
+
+        mlp = []
```
rusty1s, line 59: Any chance we can use `torch_geometric.nn.MLP` for this?

torch_geometric/nn/dense/dense_dmon_pool.py
```diff
@@ -0,0 +1,150 @@
...
+        {{\|\mathbf{S}^{\top} \mathbf{S}\|}_F} -\frac{\mathbf{I}_C}{\sqrt{C}}
+        \right\|}_F
+
+    where :math:`C` is the number of clusters, and (3) the cluster loss
+
+    .. math::
+        \mathcal{L}_c = \frac{\sqrt{C}}{n}
+        {\left\|\sum_i\mathbf{C_i}^{\top}\right\|}_F - 1.
+
+    Args:
```
rusty1s, line 45: Can we link to the example (similar to how we have done in `RGATConv`)?

torch_geometric/nn/dense/dense_dmon_pool.py
fork123aniket: Unfortunately, it is required to apply **_Dropout_** to only **_Dense_** layer being used to compute soft cluster assignment matrix which `torch_geometric.nn.MLP` currently skips. Therefore, this method seems of no use to us. 

In essence, if we wanna use only one Dense layer, then, in the `forward` function of `torch_geometric.nn.MLP`, **_Dropout_** is gonna skip through as the length of `channel_list` is 2, which in turn, makes only one Dense layer. So, don't think we should use this method. Please comment on this if you've some suggestions on whether to use this method.......

torch_geometric/nn/dense/dense_dmon_pool.py
fork123aniket: Have just added a link to the example, as can be seen on lines **_45-50_**.

torch_geometric/nn/dense/dense_dmon_pool.py
fork123aniket: Can you please help me with this just by giving a quick example of how to type hints for documentation as I don't see an example of it anywhere??

rusty1s: I just mean
```python
def __init__(self, dense_size: int, k: int = 16, p: float = 0.2):
```

torch_geometric/nn/dense/dense_dmon_pool.py
rusty1s: Can't you use `MLP` and apply a final `Dropout` layer afterwards?

fork123aniket:
Unfortunately, my forked repo has been deleted accidently. To address the issues listed in this PR, raising a new PR.

PR #4133, opened 2022-02-24
Update `setup.py`: Make more requirements optional

adelizer:
Moving the following packages to `full_install_requires`:
  - `yacs`
  - `hydra-core`
  - `class-resolver>=0.3.2`
  - `googledrivedownloader`

Keeping only essential packages in `install_requires` will enable `torch-geometric` to be installed in secure environments with strict package management constraints.


adelizer:
As the community for `torch-geometric` grows, this PR will make it more accessible to be installed and used within highly secure environments. This is the case for example at Adyen https://www.adyen.com/ where we would like to start using `torch-geometric` but couldn't install some of the dependencies due to security concerns.

rusty1s:
Thanks for the PR. I'm happy to move some of these dependencies to optional one. We would need to add a new CI test to ensure that PyG can still successfully run even when dropping these dependencies. Furthermore, I wonder what are the security concerns of these packages? These are all pretty popular packages.

adelizer:
Sure, I know of this tool https://snyk.io/advisor/python being used to check the health of different packages and in the case of `googledrivedownloader` it is definitely [low](https://snyk.io/advisor/python/googledrivedownloader) as well as for `yacs`. I can modify the PR to only move these 2 if you suggest so. 
Furthermore, I can also help with the new CI test if needed.

rusty1s:
@adelizer I added a GitHub action to test a minimal installation (currently checks if imports work as expected). As the test  fails, we need to make the dependencies truly optional before we can merge this PR.

rusty1s:
All issues are now resolved. Thank you for the PR!

rusty1s (approved):

adelizer:
@rusty1s Thanks a lot!! Any proposed date for the next release with this change included?

rusty1s:
Next release is scheduled around mid March. Let me know if this is early enough for you.

adelizer:
If there is an option for an earlier one that would be greatly helpful!

rusty1s:
Do you need to install from PyPi or does installing via GitHub work as well?
```
pip install git+https://github.com/pyg-team/pytorch_geometric.git
```
AFAIK, PyTorch 1.11 is coming soon, and we would like to wait for it before releasing any new version.

adelizer:
Great, thanks for explaining! I will wait for mid March then since it is not possible to install directly from GitHub. As in our use case all packages are mirrored on an internal PyPI

rusty1s:
Hey @adelizer, can you reach out to me on `matthias.fey@tu-dortmund.de`? I would like to get in touch with you about the use-cases of PyG @ Adyen.

rusty1s:
Just released **PyG 2.0.4**, see [here](https://github.com/pyg-team/pytorch_geometric/releases/tag/2.0.4) :)

PR #4166, opened 2022-02-27
Spectral Modularity Pool layer

fork123aniket:
Implementation of spectral modularity pooling that not only coarsen the graphs but can also be used to cluster the nodes of any graph.

### DIFF
--- a/.github/workflows/building-pyg-conda.yml
+++ b/.github/workflows/building-pyg-conda.yml
@@ -10,21 +10,29 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04, macos-10.15, windows-latest]
-        python-version: [3.7, 3.8, 3.9]
-        torch-version: [1.9.0, 1.10.0]
-        cuda-version: ['cpu', 'cu102', 'cu111', 'cu113']
+        os: [ubuntu-18.04, macos-10.15, windows-2019]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        torch-version: [1.10.0, 1.11.0]
+        cuda-version: ['cpu', 'cu102', 'cu113', 'cu115']
         exclude:
-          - torch-version: 1.9.0
-            cuda-version: 'cu113'
           - torch-version: 1.10.0
-            cuda-version: 'cu111'
+            cuda-version: 'cu115'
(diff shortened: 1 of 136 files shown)


fork123aniket:
All the issues listed in PR#4114 have been addressed. Have added **_MLP_**, which can be seen in lines 69, 71, and 73 of **_dense_dmon_pool.py_**. Moreover, have also added hints to the function definition on line 61.

rusty1s:
Any idea why the tests are failing?

fork123aniket:
Yes. `torch_geometric.nn.MLP` is the cause of why the tests have failed. To mitigate this, I just superseded the MLPs with `Linear` layers as what I did in the last commit. Moreover, why `torch_geometric.nn.MLP` not the right choice over `Linear` is because it doesn't allow me to set `act` variable to `torch.nn.SELU` which is what the original solution uses. When I tried to access this `act` variable of `torch_geometric.nn.MLP`, it threw `ArgumentError`.

rusty1s:
Mh weird. You should be able to set the act of `MLP` to SELU, e.g., via `MLP(…, act=SELU())`.

fork123aniket:
image

fork123aniket:
Yes, I did. However, I got the error (shown in the screenshot attached below). Moreover, even if we get it somehow to work, 11 of 18 total tests will again get failed. Please let me know if we can still put `Linear` to some use instead of `torch_geometric.nn.MLP`.

image

rusty1s:
@fork123aniket I modified the code to make use of `MLP`. For this, we need to delay the import to avoid circular import. I think this makes the code more clear. Note that the `act` argument of `MLP` can only be modfied in case you have PyG master installed. This feature just got recently added.

fork123aniket:
Hmm.... Yes, you're right!! The code seems more evident now. So, what are we doing next with this PR? Are all the files ready to get merged or do we still need any further modifications somewhere in the given four committed files?? Please let me know your thoughts.......

rusty1s:
It looks good from my side. I will make another pass tomorrow. Thank you! Feel free to send your other PR in a separate branch.

fork123aniket:
@rusty1s Could you please help closing this PR just by merging the committed files??

rusty1s (approved):
Indeed. Thanks for your contribution!

PR #4187, opened 2022-03-03
Mesh Laplacian computation

daniel-unyi-42:
Mesh Laplacian computation in pure PyTorch.

I'm planning to add another test case (some simple shape) and the formula in the documentation.


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,25 @@
+{
```
rusty1s, line 1: Can we delete this file?

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
+import torch
+from torch_scatter import scatter_add
+
+from torch_geometric.utils import add_self_loops
+
+
+def get_mesh_laplacian(pos, face):
+    """ Computes the mesh Laplacian of the mesh given by
```
rusty1s, line 8:
```suggestion
    r"""Computes the mesh Laplacian of the mesh given by

```

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
+import torch
+from torch_scatter import scatter_add
+
+from torch_geometric.utils import add_self_loops
+
+
+def get_mesh_laplacian(pos, face):
```
rusty1s, line 7:
```suggestion
def get_mesh_laplacian(pos: Tensor, face: Tensor) -> Tuple[Tensor, Tensor]:

```

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
+import torch
+from torch_scatter import scatter_add
+
+from torch_geometric.utils import add_self_loops
+
+
+def get_mesh_laplacian(pos, face):
+    """ Computes the mesh Laplacian of the mesh given by
+    :obj:`pos` and :obj:`face`. It is computed as
+    :math:` \mathbf{L}_{ij} = \begin{cases}
```
rusty1s, line 10: Use math environment rather than in-place formula
```suggestion
    .. math::
        \mathbf{L}_{ij} = \begin{cases} 
```

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
...
+        right_vec = right_pos - central_pos
+        dot = torch.einsum('ij, ij -> i', left_vec, right_vec)
+        cross = torch.norm(torch.cross(left_vec, right_vec, dim=1), dim=1)
+        cot = dot / cross  # cos / sin
+        nonlocal cot_weight, area_weight, edge_index
+        cot_weight = torch.cat([cot_weight, cot / 2.0, cot / 2.0])
+        # one-third of a triangle's area is cross / 6.0
+        # since each edge is accounted twice, we compute cross / 12.0 instead
+        area_weight = torch.cat([area_weight, cross / 12.0, cross / 12.0])
+        edge_index = torch.cat([
```
rusty1s, line 47: I think it would be great to avoid intermediate `torch.cat` calls. IMO, it is better to append them into one single list `edge_index` and concatenate after the final `add_edge` call. WDYT?

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
...
+            torch.stack([right, left], dim=1)
+        ])
+
+    # add all 3 edges of the triangles
+    add_edge(face[2], face[0], face[1])
+    add_edge(face[0], face[1], face[2])
+    add_edge(face[1], face[2], face[0])
+
+    # eliminate duplicate matrix entries by adding them together
+    index_linearizer = torch.Tensor([num_nodes, 1]).to(device)
```
rusty1s, line 59:
```suggestion
    index = edge_index[0] * num_nodes + edge_index[1]
```
I think this is cleaner and avoids the type conversion.

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -0,0 +1,73 @@
...
+    lin_index = torch.matmul(edge_index.float(), index_linearizer).long()
+    y, idx = lin_index.unique(return_inverse=True)
+    edge_index = torch.stack([y // num_nodes, y % num_nodes])
+    cot_weight = scatter_add(cot_weight, idx, dim=0)
+    area_weight = scatter_add(area_weight, idx, dim=0)
```
rusty1s, line 64: It looks like all you are doing here is eliminating duplicate entries. It's probably better to make use of `torch_geometric.utils.coalesce` for this.

rusty1s (commented):
Thanks you for your contributions. Left some comments.

.pytest_cache/v/cache/lastfailed
daniel-unyi-42: Yes, I'm going to delete it.

torch_geometric/utils/get_mesh_laplacian.py
daniel-unyi-42: Great, thank you!

torch_geometric/utils/get_mesh_laplacian.py
daniel-unyi-42: You are right, the function add_edge could return cot & cross, and concatenate all in one single call. I'll reformulate the code accordingly.

torch_geometric/utils/get_mesh_laplacian.py
daniel-unyi-42: I've never used this function, based on the documentation, it has the same effect. Thank you for the suggestion.

.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
rusty1s, line 1: Not sure what's happening here. Please feel free to add `.pytest_cache/` to `.gitignore`.

daniel-unyi-42:
Finished everything.
Formula is rendering correctly in the documentation.
I think it's ready to merge, if you don't have any more suggestions.

rusty1s (approved):
Perfect. Thank you :)

xieweiyi:
The summation row-wise for the computed laplacian is not zero.

Issue #890, opened 2020-01-05
Questions about labels in PPI datasets


tiancanzhilei:
## ❓ Questions & Help
I want to know if the labels of the PPI data are multi-label. If so, why is dataset.num_classes = 121?

rusty1s:
Yes, it is a multilabel dataset where each node belongs to multiple classes within 121 classes.

RemyLau:
Does ``graphgym`` have loss function for dealing with multilabel datasets like PPI?
Looking at ``loss.py``, it seems like currently multilabel classification is not supported.
Will this be updated soon?

https://github.com/pyg-team/pytorch_geometric/blob/d8b8e017c9cd9667b9c3b84a58e65130d54cf50e/torch_geometric/graphgym/loss.py#L37-L45

RemyLau:
Thanks for the quick response @rusty1s . I've tried to test it on the [``GemsecDeezer``](https://pytorch-geometric.readthedocs.io/en/latest/modules/datasets.html#torch_geometric.datasets.GemsecDeezer) dataset from Pytorch Geometric, but here's the error I got:

```txt
Traceback (most recent call last):
  File "/mnt/ufs18/rs-027/compbio/krishnanlab/tmp/remy_tmp_test/GTaxoGym/main.py", line 66, in <module>
    train(loggers, loaders, model, optimizer, scheduler)
  File "/mnt/home/liurenmi/software/anaconda3/envs/gtaxogym/lib/python3.9/site-packages/torch_geometric/graphgym/train.py", line 71, in train
    train_epoch(loggers[0], loaders[0], model, optimizer, scheduler)
  File "/mnt/home/liurenmi/software/anaconda3/envs/gtaxogym/lib/python3.9/site-packages/torch_geometric/graphgym/train.py", line 20, in train_epoch
    loss, pred_score = compute_loss(pred, true)
  File "/mnt/home/liurenmi/software/anaconda3/envs/gtaxogym/lib/python3.9/site-packages/torch_geometric/graphgym/loss.py", line 45, in compute_loss
    return bce_loss(pred, true), torch.sigmoid(pred)
  File "/mnt/home/liurenmi/software/anaconda3/envs/gtaxogym/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1051, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #4206, opened 2022-03-06
Fix graphgym multilabel loss/logger

RemyLau:
This PR proposes a minor modification to the loss function regarding the handling of the binary and multilabel classification settings.
1. The flattening seems unnecessary since the [``BCEWithLogitsLoss``](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html) should work just fine so long as the shapes between ``y_true`` and ``y_pred`` are consistent.
2. In Line36, I added an extra condition of ``true.ndim == 1`` so that the multiclass setting is correctly distinguished from the multilabel setting.

This is should also fix the issue I posted as a comment in another issue https://github.com/pyg-team/pytorch_geometric/issues/890#issuecomment-945037351

The error I encountered previously in the multilabel setting is due to mismatching dimensions between ``y_pred`` and ``y_true`` when calling ``logger.update_stats`` in ``train_epoch`` (see below). In particular, ``pred_score`` is flattened by ``compute_loss``, but ``true`` is not, causing the mismatching dimension error as required by [``logger.update_stats``](https://github.com/pyg-team/pytorch_geometric/blob/4557254c849eda62ce1860a56370eb4a54aa76dd/torch_geometric/graphgym/logger.py#L144).

https://github.com/pyg-team/pytorch_geometric/blob/4557254c849eda62ce1860a56370eb4a54aa76dd/torch_geometric/graphgym/train.py#L21-L26

### DIFF
--- a/torch_geometric/graphgym/loss.py
+++ b/torch_geometric/graphgym/loss.py
@@ -22,11 +22,8 @@ def compute_loss(pred, true):
 
     # default manipulation for pred and true
     # can be skipped if special loss computation is needed
-    # if multi task binary classification, treat as flatten binary
     pred = pred.squeeze(-1) if pred.ndim > 1 else pred
     true = true.squeeze(-1) if true.ndim > 1 else true
-    if true.ndim > 1 and cfg.dataset.task_type == 'classification':
-        pred, true = torch.flatten(pred), torch.flatten(true)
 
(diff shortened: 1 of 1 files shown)


rusty1s (approved):
Looks perfect. Thanks for the fix!

PR #4220, opened 2022-03-08
`pathlib.Path` instead of `os.path`

rietesh:
This PR is a continuation of the earlier PR https://github.com/pyg-team/pytorch_geometric/pull/3207; I had accidentally deleted the fork.

This PR fixes the issues discussed in earlier PR regarding changing `../..` to `parent.parent` and other minor fixes regarding imports and merge conflicts. I haven't changed os.path in test scripts to Pathlib.Path if that is required, I can push another commit.


benchmark/citation/datasets.py
```diff
@@ -1,11 +1,11 @@
-import os.path as osp
+from pathlib import Path
 
 import torch_geometric.transforms as T
 from torch_geometric.datasets import Planetoid
 
 
 def get_planetoid_dataset(name, normalize_features=False, transform=None):
-    path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', name)
+    path = Path.joinpath(Path(__file__).resolve().parent.parent, 'data', name)
```
rusty1s, line 8: I think this would be a little bit cleaner.
```suggestion
    path = Path(__file__).resolve().parent.parent.joinpath('data', name)
```

benchmark/runtime/main.py
```diff
@@ -10,11 +10,11 @@
 from torch_geometric.datasets import Entities, Planetoid
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-root = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data')
-Cora = Planetoid(osp.join(root, 'Cora'), 'Cora')
-CiteSeer = Planetoid(osp.join(root, 'CiteSeer'), 'CiteSeer')
-PubMed = Planetoid(osp.join(root, 'PubMed'), 'PubMed')
-MUTAG = Entities(osp.join(root, 'EntitiesMUTAG'), 'MUTAG')
+root = Path.joinpath(Path(__file__).resolve().parent.parent, 'data')
+Cora = Planetoid(str(root.joinpath('Cora')), 'Cora')
```
rusty1s, line 14:
```suggestion
Cora = Planetoid(str(root.joinpath('Cora')), 'Cora')
```
Why do we need to convert to `str` here? Is this needed to let the `dataset` work as expected? Shouldn't we work on supporting this on the `dataset` side first?

rusty1s:
Thanks for the update. I'm a bit unsure on how to proceed with this PR. In my opinion, it would be great to first lay the foundations of `Path` support inside the library (e.g. inside `Dataset` and `InMemoryDataset`). For example, the PR suggests that we need to convert `Path` to `str` first:
```python
str(root.join('PubMed'))
```
Is that really necessary?

benchmark/runtime/main.py
rietesh: There is no need to convert to `str` Dataset supports both `str` and `Path`. I have pushed a commit to fix this issue.

test/graphgym/example_node/0/test/stats.json
```diff
@@ -0,0 +1,4 @@
+{"epoch": 0, "loss": 1.8685, "lr": 0, "params": 23880, "time_iter": 0.0194, "accuracy": 0.361}
```
rusty1s, line 1: Thank you for confirming. It looks like you accidentally pushed these logs here.

rietesh: I removed these logs

benchmark/citation/datasets.py
rietesh: In the previous PR a reviewer(cthoyt) suggested me to use second parent

rusty1s: Yes, we need two parents here (one for getting the `dirname` and one for going one step back). Just wanted to point out that I think the beauty of `Path` is that we can combine paths via regular slashes. In the end, `Path.joinpath()` syntax is not really needed since we can directly do this, e.g., via `root_path / '..' / 'data'`.

Padarn:
Agree with @rusty1s - maybe we could start by splitting with PR into a smaller one just for `Dataset` and then go from there?

PR #4278, opened 2022-03-16
Edge Masking to generate Explanations

fork123aniket:
This PR contains implementation of how to compute layer-wise weights for each edge in order to produce explanations for both node-level and graph-level tasks. Furthermore, this implementation is different from authors' original implementation and is fast and more memory efficient than theirs.

### DIFF
--- a/torch_geometric/nn/conv/updated_gcn.py
+++ b/torch_geometric/nn/conv/updated_gcn.py
@@ -1,0 +1,254 @@
+from typing import Optional, Tuple
+
+import torch
+from torch import Tensor
+from torch.nn import Parameter, ReLU, Sequential, LayerNorm
+from torch_scatter import scatter_add
+from torch_sparse import SparseTensor, fill_diag, matmul, mul
+from torch_sparse import sum as sparsesum
+
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.dense.linear import Linear
+from torch_geometric.nn.inits import zeros, uniform
+from torch_geometric.typing import Adj, OptTensor, PairTensor
+from torch_geometric.utils import add_remaining_self_loops
+from torch_geometric.utils.num_nodes import maybe_num_nodes
+
+
+@torch.jit._overload
+def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
+             add_self_loops=True, dtype=None):
+    # type: (Tensor, OptTensor, Optional[int], bool, bool, Optional[int]) -> PairTensor  # noqa
+    pass
+
+
+@torch.jit._overload
+def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
+             add_self_loops=True, dtype=None):
+    # type: (SparseTensor, OptTensor, Optional[int], bool, bool, Optional[int]) -> SparseTensor  # noqa
+    pass
+
+
+def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
+             add_self_loops=True, dtype=None):
+
+    fill_value = 2. if improved else 1.
+
+    if isinstance(edge_index, SparseTensor):
+        adj_t = edge_index
+        if not adj_t.has_value():
+            adj_t = adj_t.fill_value(1., dtype=dtype)
+        if add_self_loops:
+            adj_t = fill_diag(adj_t, fill_value)
+        deg = sparsesum(adj_t, dim=1)
+        deg_inv_sqrt = deg.pow_(-0.5)
+        deg_inv_sqrt.masked_fill_(deg_inv_sqrt == float('inf'), 0.)
+        adj_t = mul(adj_t, deg_inv_sqrt.view(-1, 1))
+        adj_t = mul(adj_t, deg_inv_sqrt.view(1, -1))
+        return adj_t
+
+    else:
+        num_nodes = maybe_num_nodes(edge_index, num_nodes)
+
+        if edge_weight is None:
+            edge_weight = torch.ones((edge_index.size(1), ), dtype=dtype,
+                                     device=edge_index.device)
+
+        if add_self_loops:
+            edge_index, tmp_edge_weight = add_remaining_self_loops(
+                edge_index, edge_weight, fill_value, num_nodes)
+            assert tmp_edge_weight is not None
+            edge_weight = tmp_edge_weight
+
+        row, col = edge_index[0], edge_index[1]
+        deg = scatter_add(edge_weight, col, dim=0, dim_size=num_nodes)
+        deg_inv_sqrt = deg.pow_(-0.5)
+        deg_inv_sqrt.masked_fill_(deg_inv_sqrt == float('inf'), 0)
+        return edge_index, deg_inv_sqrt[row] * edge_weight * deg_inv_sqrt[col]
+
+
+class GCNConv(MessagePassing):
+    r"""The graph convolutional operator from the `"Semi-supervised
+    Classification with Graph Convolutional Networks"
+    <https://arxiv.org/abs/1609.02907>`_ paper
+    .. math::
+        \mathbf{X}^{\prime} = \mathbf{\hat{D}}^{-1/2} \mathbf{\hat{A}}
+        \mathbf{\hat{D}}^{-1/2} \mathbf{X} \mathbf{\Theta},
+    where :math:`\mathbf{\hat{A}} = \mathbf{A} + \mathbf{I}` denotes the
+    adjacency matrix with inserted self-loops and
+    :math:`\hat{D}_{ii} = \sum_{j=0} \hat{A}_{ij}` its diagonal degree matrix.
+    The adjacency matrix can include other values than :obj:`1` representing
+    edge weights via the optional :obj:`edge_weight` tensor.
+    Its node-wise formulation is given by:
+    .. math::
+        \mathbf{x}^{\prime}_i = \mathbf{\Theta}^{\top} \sum_{j \in
+        \mathcal{N}(v) \cup \{ i \}} \frac{e_{j,i}}{\sqrt{\hat{d}_j
+        \hat{d}_i}} \mathbf{x}_j
+    with :math:`\hat{d}_i = 1 + \sum_{j \in \mathcal{N}(i)} e_{j,i}`, where
+    :math:`e_{j,i}` denotes the edge weight from source node :obj:`j` to target
+    node :obj:`i` (default: :obj:`1.0`)
+    Args:
+        in_channels (int): Size of each input sample, or :obj:`-1` to derive
+            the size from the first input(s) to the forward method.
+        out_channels (int): Size of each output sample.
+        improved (bool, optional): If set to :obj:`True`, the layer computes
+            :math:`\mathbf{\hat{A}}` as :math:`\mathbf{A} + 2\mathbf{I}`.
+            (default: :obj:`False`)
+        cached (bool, optional): If set to :obj:`True`, the layer will cache
+            the computation of :math:`\mathbf{\hat{D}}^{-1/2} \mathbf{\hat{A}}
+            \mathbf{\hat{D}}^{-1/2}` on first execution, and will use the
+            cached version for further executions.
+            This parameter should only be set to :obj:`True` in transductive
+            learning scenarios. (default: :obj:`False`)
+        add_self_loops (bool, optional): If set to :obj:`False`, will not add
+            self-loops to the input graph. (default: :obj:`True`)
+        normalize (bool, optional): Whether to add self-loops and compute
+            symmetric normalization coefficients on the fly.
+            (default: :obj:`True`)
+        bias (bool, optional): If set to :obj:`False`, the layer will not learn
+            an additive bias. (default: :obj:`True`)
+        explain (bool, optional): If set to :obj:`True`, the layer will be
+            used in explanation of predictions of node-level and graph-level
+            tasks. (default: :obj:`False`)
+        **kwargs (optional): Additional arguments of
+            :class:`torch_geometric.nn.conv.MessagePassing`.
+    Shapes:
+        - **input:**
+          node features :math:`(|\mathcal{V}|, F_{in})`,
+          edge indices :math:`(2, |\mathcal{E}|)`,
+          edge weights :math:`(|\mathcal{E}|)` *(optional)*
+        - **output:** node features :math:`(|\mathcal{V}|, F_{out})`
+    """
+
+    _cached_edge_index: Optional[Tuple[Tensor, Tensor]]
+    _cached_adj_t: Optional[SparseTensor]
+
+    def __init__(self, in_channels: int, out_channels: int,
+                 improved: bool = False, cached: bool = False,
+                 add_self_loops: bool = True, normalize: bool = True,
+                 bias: bool = True, explain: bool = False, **kwargs):
+
+        kwargs.setdefault('aggr', 'add')
+        super().__init__(**kwargs)
+
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.improved = improved
+        self.cached = cached
+        self.add_self_loops = add_self_loops
+        self.normalize = normalize
+        self.explain = explain
+
+        self._cached_edge_index = None
+        self._cached_adj_t = None
+
+        self.lin = Linear(in_channels, out_channels, bias=False,
+                          weight_initializer='glorot')
+
+        if bias:
+            self.bias = Parameter(torch.Tensor(out_channels))
+        else:
+            self.register_parameter('bias', None)
+
+        if self.explain:
+            self.convert = Linear(in_channels, out_channels)
+            self.process_message = Sequential(LayerNorm(out_channels),
+                                              ReLU())
+
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        if self.explain:
+            uniform(self.in_channels, self.convert)
+        else:
+            self.lin.reset_parameters()
+            zeros(self.bias)
+            self._cached_edge_index = None
+            self._cached_adj_t = None
+
+    def forward(self, x: Tensor, edge_index: Adj,
+                edge_weight: OptTensor = None,
+                message_scale=None, message_replacement=None) -> Tensor:
+        """"""
+
+        if self.normalize:
+            if isinstance(edge_index, Tensor):
+                cache = self._cached_edge_index
+                if cache is None:
+                    edge_index, edge_weight = gcn_norm(  # yapf: disable
+                        edge_index, edge_weight, x.size(self.node_dim),
+                        self.improved, self.add_self_loops)
+                    if self.cached:
+                        self._cached_edge_index = (edge_index, edge_weight)
+                else:
+                    edge_index, edge_weight = cache[0], cache[1]
+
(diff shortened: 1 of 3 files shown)


fork123aniket:
Please take a look at functions (**_message_**, **_get_latest_source_embeddings_**, **_get_latest_target_embeddings_**, and **_get_latest_messages_**) which must be copied to other layers' implementation in order to make this **_GraphMask_** implementation model-agnostic. The significance of each of these functions can be seen on lines 241-243 (updated_gcn.py) and 306-308 (updated_rgcn.py).

fork123aniket:
@rusty1s Am just following up with you that did you get a chance to look into the code files pinned with this PR? Am still awaiting any review comments so we can make further progress on how to make this GraphMask Explainer implementation model-agnostic. In the previous comment, I've provided details of which functions to put more focus on.

rusty1s:
Hi @fork123aniket. Yes, sorry for the delay. I will try to have a review ready by tomorrow. I sadly only have two hands :(

torch_geometric/nn/conv/updated_gcn.py
```diff
@@ -0,0 +1,253 @@
...
+                        basis_messages = basis_messages + \
+                                         (1 - message_scale).unsqueeze(
+                                             -1) * message_replacement
+                    else:
+                        basis_messages = basis_messages + \
+                                         ((1 - message_scale).unsqueeze(-1) *
+                                          message_replacement.unsqueeze(0))
+
+            self.latest_messages = basis_messages
+            self.latest_source_embeddings = x_j
```
rusty1s, line 233: This is a bit dangerous to do. We should at least `detach()` them from the computation graph, right?

torch_geometric/nn/conv/updated_gcn.py
```diff
@@ -0,0 +1,253 @@
...
+
+        return out
+
+    def message(self, x_i: Tensor, x_j: Tensor, edge_index: Tensor,
+                edge_weight: OptTensor, message_scale,
+                message_replacement) -> Tensor:
+        if self.explain:
+            basis_messages = x_j if edge_weight is None else edge_weight.view(
+                -1, 1) * x_j
+            basis_messages = self.convert(basis_messages)
```
rusty1s, line 216: It looks like it should be possible to out-source this functionality to its own function (between `message` and `aggregate`). We already do this to some extend here: https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/message_passing.py#L330-L341

I think one option that we have to make this GNN-agnostic is to leverage `MessagePassing.explain_message()` with the implementation of the current `MessagePassing` interface as default implementation, but which we can override for specific explainers. As such, the computation flow would be
```python
out = self.message(...)
out = self.explain_message(out)
out = self.aggregate(out, ...)
```
WDYT?

torch_geometric/nn/conv/updated_gcn.py
```diff
@@ -0,0 +1,253 @@
...
+            self.latest_target_embeddings = x_i
+
+            return basis_messages
+
+        return x_j if edge_weight is None else edge_weight.view(-1, 1) * x_j
+
+    # functions (get_latest_source_embeddings, get_latest_target_embeddings,
+    # and get_latest_messages) are necessary to compute edge_weights
+    # through hard_concrete_distribution
+    def get_latest_source_embeddings(self):
```
rusty1s, line 243: I think it is save to drop these functions and just access, *e.g.*, `self.latest_source_embeddings`.

rusty1s:
Sorry for the delay. I left some ideas on how to make this GNN-agnostic :) Looking forward to your reply.

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: These `self.latest_messages`, `self.latest_source_embeddings`, and `self.latest_target_embeddings` are accessed through their respective functions `get_latest_messages()`, `get_latest_source_embeddings()`, and `get_latest_target_embeddings()` and used in Lines **_512-527_** of `GraphMask.py` for computing `edge_weights` to drop irrelevant edges using MLPs (after training of GNN Model). So, is it safe to `detach()` them from the computation graph?? Would love to know your thoughts on this.......

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: I think that would be great if we follow your above mentioned notion of first letting the layer (GAT, GCN, RGCN, etc.) prepare the message then compute the `explain_message` by defining its own function in `MessagePassing` interface. This will not only save us from changing every single layer's implementation but also quite easy to implement. 

1. as we need to take into account relations for heterogeneous graphs, the implementation of computing `edge_weights` is somewhat different:-
For example, after computing messages, GCN explainable layer (updated_gcn.py) implements:

```
basis_messages = self.convert(basis_messages)
basis_messages = self.process_message(basis_messages)
```

and RGCN explainable layer (updated_rgcn.py) implements:

```
basis_messages = self.process_message(basis_messages)
```

So, is it possible if we pass in to this `explain_message()` function a flag which lets it to compute this explainable message for each type of graphs accordingly??

2. Since `self.convert()` and `self.process_message()` are instances of `torch_geometric.nn.dense.linear` and `torch.nn.Sequential` respectively, is it possible to use them in `MessagePassing` interface just by initializing them and set their initial weights somewhere in it, which let us employing these two functions (`self.convert()` and `self.process_message()`) in `explain_message()` I believe?? If yes, please suggest any way to achieve this or do you think any other best way to achieve this??

3. Do we have to register hooks for `self._explain_message_forward_pre_hooks.values()` and `self._explain_message_forward_hooks.values()` as `MessagePassing` interface implements for `message()` function given below:

```
for hook in self._message_forward_pre_hooks.values():
    res = hook(self, (msg_kwargs, ))
    if res is not None:
        msg_kwargs = res[0] if isinstance(res, tuple) else res
for hook in self._message_forward_hooks.values():
    res = hook(self, (msg_kwargs, ), out)
    if res is not None:
        out = res
```

If yes, can you help me with it just by writing here these two hooks statements for `explain_message()`, I'll take care of the rest?? That would be a great help, actually!!! :)

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: As mentioned in the above comments that these three `get_latest_messages()`, `get_latest_source_embeddings()`, and `get_latest_target_embeddings()` functions act as containers for `messages`, `source_embeddings`, and `target_embeddings` respectively for each layer in the trained GNN Model. Since these functions help saving these (message, embeddings for both source and target nodes for each edge) for each layer, is it safe to drop these functions and just directly access `self.latest_messages`, `self.latest_source_embeddings`, and `self.latest_target_embeddings`?? What do you suggest??

fork123aniket:
I've just replied to all the comments above. Please look into each of them and let me know your thoughts. Looking forward to hearing from you very soon....... :)

torch_geometric/nn/conv/updated_gcn.py
rusty1s: Yeah, I agree. `detach()` would destroy the purpose of the `GraphMask` model. I think we have two options here:
* Save the features only in case `_explain=True`, and set them to `None` after optimization is done. Otherwise, the computation graph is never freed.
* A cleaner option is to never save them at all IMO. We should be able to make use of [message passing hooks](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.message_passing.MessagePassing.register_propagate_forward_pre_hook) inside `GraphMask` to obtain these features.

torch_geometric/nn/conv/updated_gcn.py
rusty1s: I don't see where these functions are used to save these features. I think it is safe to drop them and just access the features inside `GraphMask` directly.

torch_geometric/nn/conv/updated_gcn.py
rusty1s: 1. `convert` is simply used to map messages to the desired output space, right? We could make this application optional inside `process_message`:
```python
if message.size(-1) != desired_size:
    message = self.lin(message)
```
where `self.lin = Linear(-1, desired_size)`.
2. I think we should take care of initializing these modules insice `GraphMask`:
```python
for conv in model.all_mp_modules:
    conv.message_lin = ...
    conv.explain_message = ...
```
3. I don't think we need any hooks for now.

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: Agreed!!! I am also considering to drop these functions (`get_latest_messages()`, `get_latest_source_embeddings()`, and `get_latest_target_embeddings()`) and directly access these three instance attributes (`self.latest_messages`, `self.latest_source_embeddings`, and `self.latest_target_embeddings`) inside `GraphMask`.

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: Yes, we should go with the second point by not saving these features (`basis_messages`, `x_j`, and `x_i`) whatsoever. Can you please elaborate as to how we can obtain these features for each layer of the GNN model using message passing hooks?? Sorry, am bit novice to these hooks and don't know how to achieve this functionality.

rusty1s:
```python
def hook(module, inputs):
    print(inputs[-1].keys())

conv.register_message_forward_pre_hook(hook)
```
should work. Let me know about it!

torch_geometric/nn/models/GraphMask.py
```diff
@@ -0,0 +1,1079 @@
...
+    <https://arxiv.org/abs/2010.00577>`_ paper for identifying layer-wise
+    compact subgraph
+    structures and small subsets node features that play a crucial role in a
+    GNN’s node-level and graph-level predictions.
+
+    .. note::
+
+        For an example of using GraphMask-Explainer,
+        see `examples/graphmask_explainer.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
```
Padarn, line 34: Not yet implemented?

torch_geometric/nn/conv/updated_gcn.py
rusty1s: I think the goal here is to make this procedure GNN agnostic. As such, when we want to implement this inside a model's message function, we basically have to do this for every PyG GNN layer. Let me know if I might be misunderstanding something.

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: For GCN, this only prints (`x_j` and `edge_weight`) which is even expected though as `message()` of GCN has only these two parameters in its definition. Am still using the strategy, mentioned above, of saving these three features (`basis_messages`, `x_j`, and `x_i`) to three instance variables (`self.latest_messages`, `self.latest_source_embeddings`, and `self.latest_target_embeddings`). This simply saves all three features to the layer-specific instance variables and then we can access each of them for each layer of the GNN model while generating explanations. WDYT?? Yet, am still supporting and even looking forward to having such approach of accessing these three features for each layer using message passing hooks (and not saving three features at all) as the above approach:-

```
def hook(module, inputs):
    print(inputs[-1].keys())

conv.register_message_forward_pre_hook(hook)
```
appears to be not enough to consider as our solution. Let me know if we still have another way of achieving this or may be there is some possibility of tweaking this above approach to reach our solution of accessing those three features for each layer while not saving them in any instance variables. Would love to know your thoughts on this.......

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: @rusty1s Your above mentioned recommendations literally worked wonders :) Finally, devised the proposed `explain_message()` function. Have also accessed all the parameters for this function using `self.inspector.distribute('explain_message', coll_dict)` inside `MessagePassing` interface. Moreover, inside GraphMask, initialized `self.convert()` and `self.process_message()`, and accessed them inside `MessagePassing` interface while keeping the `self.convert()` optional to take care of mapping messages to the desired output space.

Furthermore, as discussed above, didn't implement any hooks for this `explain_message()` while employing the discussed computation flow inside `MessagePassing` interface:-

```
out = self.message(...)
out = self.explain_message(out, x_i, x_j, message_scale, message_replacement)
out = self.aggregate(out, ...)
```

Have generated explanations for predictions made by three different layers (GAT, GCN, RGCN) without even making any changes to their respective implementations. Used Datasets: Cora (Node Classification - GAT, GCN), Enzymes (Graph Classification - GAT, GCN), and AIFB (Node Classification - RGCN).

rusty1s: Yeah, awesome! Can we add `explain_message` functionality in this PR (it's always a bit dangerous to make modifications to `MessagePassing` - we need to test that well), and then add support for `GraphMask` in a later PR?

torch_geometric/nn/conv/updated_gcn.py
rusty1s: Indeed, let's go with `explain_message` :)

torch_geometric/nn/conv/updated_gcn.py
fork123aniket: Sure. Have just added updated `MessagePassing` file (`updatedMessagePassing.py`) to this PR. One point to note here is that two parameters (`message_scale` and `message_replacement`) must be passed to both `forward()` and `propagate()` of each layer type (GAT, GCN, RGCN, etc.), as can be seen in files (`updated_gat.py`, `updated_gcn.py`, and `updated_rgcn.py`). If we don't do so, the following will surely be a result:- `__collect__()` of `MessagePassing` will fail to save these two parameters in `coll_dict`, which in turn, will fail to pass on these two parameters to `explain_message()` (after running `self.inspector.distribute('explain_message', coll_dict)`). Do let me know if you know some other way of passing these two parameters to `MessagePassing.propagate()` directly without touching the layer's existing implementation or else we must go with this approach of passing these two parameters to both `forward()` and `propagate()` of each layer type.

Furthermore, have taken `GraphMask.py` down because as discussed above, this PR is merely meant for adding `explain_message()` functionality to the `MessagePassing` interface. Besides, please take a deeper look at `self._gmask_explain` flag created inside `MessagePassing` interface, which helps triggering `explain_message` only when we work with GraphMask explainer or else will ensure the normal execution of the interface. This is to not disturb the original implementation of `MessagePassing` interface while not using this explanation technique.

All the changes made to the interface for this explanation technique to work can be seen inside the hash (#) blocks inside `MessagePassing` interface starts with (`# Support for GraphMask#######`) and ends with (`# block ended#######`). This is only to ease your review process for `updatedMessagePassing.py`.

rusty1s: Resolving this conversation. Will continue in the main thread.

rusty1s:
@fork123aniket Thanks for the update :) Can you apply the modifications directly inside `message_passing.py`? This makes it easier to track the changes (and we need to do that anyway before merging).

What do `message_scale` and `message_replace` represent? Do we really need to pass them as arguments to `forward` or can we simply set them as instance attributes during explainability? This way, we could simply access them via `self.message_replace` since `explain_message` rather than adding them as input arguments to every GNN layer (which again obviously does not scale).

fork123aniket:
@rusty1s **_Bingo!!!_** Successfully set those two required parameters (`message_scale` and `message_replacement`) as instance attributes only at the time of generating explanations. Now, we need not to worry about adding them to each layer type's `forward()` function. Have tested the whole proposed explainability technique on datasets:- Cora (Node Classification - GAT, GCN), Enzymes (Graph Classification - GAT, GCN), and AIFB (Node Classification - RGCN) without touching any of the layer types' original implementation which signals the successful achievement towards model-agnostic graph explainability, WDYT??

Furthermore, as mentioned above, since making changes to any GNN layer types' original implementation is not required, hence have taken all the three previously added files (`updated_gcn.py`, `updated_gat.py`, and `updated_rgcn.py`) down from this PR and added all the modifications, **_made for GraphMask Explainer to work_**, inside `MessagePassing` interface of the recently committed `message_passing.py` file. Now, you'll be easily able to figure out the differences between this committed change and the original PyG `message_passing.py` file.

Please review the changes and let me know if any further change is still required somewhere.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -112,14 +109,16 @@ def __init__(self, aggr: Optional[Union[str, List[str]]] = "add",
 
         self.inspector = Inspector(self)
         self.inspector.inspect(self.message)
+        self.inspector.inspect(self.explain_message, pop_first=True)
```
rusty1s, line 112: I think we only want to do this in case `self._explain` is `True`. Otherwise, the arguments of `explain_message` will get collected even if the function is never called. I think we can achieve this by adding a `requires_explain(True)` function to `MessagePassing`, and do inspecting and adding to `__user_args__` solely there. WDYT?

torch_geometric/nn/conv/message_passing.py
```diff
@@ -130,6 +129,9 @@ def __init__(self, aggr: Optional[Union[str, List[str]]] = "add",
 
         # Support for GNNExplainer.
         self._explain = False
+        self.message_scale = None
+        self.message_replacement = None
+        self._gmask_explain = False
```
rusty1s, line 134: Not sure why we need these arguments here. IMO, the `GraphMask` module should set these attributes (and we solely rely on `_explain` to call `self.explain_message` or not).

torch_geometric/nn/conv/message_passing.py
```diff
@@ -328,6 +329,8 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
                                              size, kwargs)
 
                 msg_kwargs = self.inspector.distribute('message', coll_dict)
+                explain_kwargs = self.inspector.distribute(
```
rusty1s, line 332: only distribute kwargs in case `self._explain` is `True`.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -338,6 +341,9 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
                     if res is not None:
                         out = res
 
+                if self._gmask_explain:
```
rusty1s, line 344:
```suggestion
                if self._explain:
```

torch_geometric/nn/conv/message_passing.py
```diff
@@ -434,15 +439,38 @@ def message(self, x_j: Tensor) -> Tensor:
         """
         return x_j
 
+    def explain_message(self, out: Tensor, x_i: Tensor, x_j: Tensor) -> Tensor:
```
rusty1s, line 442: I would expect that `explain_message` has the default explainability logic integrated, see https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/message_passing.py#L343-L354. `GraphMask` would override this functionality.

rusty1s (commented):
Thanks for the updates :)

fork123aniket:
All the above mentioned 5 issues have been addressed in the recently added `message_passing.py`. Please review the changes.

rusty1s (approved):
Thank you! Cleaned up a bit and added necessary tests :)

Issue #4283, opened 2022-03-17
Using TochMetrics for code simplification


Borda:
### 🛠 Proposed Refactor

We have been developing [TochMetrics](https://github.com/PyTorchLightning/metrics) to be a general-purpose metric as well some domain-specific use-cases. In many cases, we have an exact mapping to `scikit-learn` with verification/testing to the reference metric for its correctness. The TM includes functional as well as `nn.Module` versions and for the most/standard metrics, the only dependency is Pytorch (for the domain-specific metrics you need to install related extras). 

### Suggest a potential alternative/fix

WIth using [TM metrics](https://torchmetrics.readthedocs.io/en/stable/) you may rely on the widely tested correctness (testing against good standard `scikit-learn` in multiple OS environments and all PyTorch versions above v1.4) and later you can use `nn.Module` to leverage update and compute.

Overall if you are fine with it we are happy to draft a PR with a suggested change to verify in place the impact.
If you have any questions, happy to follow up with me or @tchaton

What I have quickly checked, al this [`torch_geometric.utils.metric`]( https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/utils/metric.py) can be replaced by TM

rusty1s:
Thanks for this great issue. We are already using `torchmetrics` in all our PL examples, and so I am happy to make the switch to it completely. `torch_geometric.utils.metric` is somewhat old and not well maintained. In fact, as far as I can see, only the `intersection_and_union` metric is actually used inside PyG and `examples/`.

As such, we could:
* remove `torch_geometric.utils.metric` completely
* Update the examples to make use of `torchmetrics`

Let me know what you think.

Borda:
cool, thank you @rusty1s! I have started a draft PR as I may have a few additional questions to deliver the best...
so let's move the additional/implementation discussion to the open #4287

PR #4287, opened 2022-03-17
Deprecate internal metrics in favor of TorchMetrics

Borda:
This PR is resolving #4283

Still some open questions/tasks:
- [ ] shall it be hard removal or with deprecation warning to not breaking user experience (as it is now in this draft)
- [ ] shall be TorchMetrics `install` or `full_install` dependency


Borda:
@rusty1s seems that this PR a slightly lowering the code coverage but it is quite understandable...

Borda:
@rusty1s may I get your opinion on this suggestion/change as the PR description: :rabbit: 
- hard removal or with deprecation warning to not breaking user experience (as it is now in this draft)
- TorchMetrics install or full_install dependency

rusty1s:
@Borda Yes, sorry for the delay :)

* I think `torchmetrics` should be a dependency of the `full_install`.
* I'm fine with a hard removal TBH since I doubt that these are functions that most PyG users are using. If we go for a deprecation warning, is there any chance we can make use of the in-house [`deprecated` warning](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/deprecation.py) rather than relying on the external dependency?

torch_geometric/utils/metric.py
```diff
@@ -138,41 +58,14 @@ def f1_score(pred: Tensor, target: Tensor, num_classes: int) -> Tensor:
...
 
 
+@deprecated(tm_func.jaccard_index, deprecated_in="v2.0", remove_in="v3.0",
+            args_mapping={
+                "pred": "preds",
+                "batch": None,
+                "omitnans": None
+            })
 def mean_iou(pred: Tensor, target: Tensor, num_classes: int,
              batch: Optional[Tensor] = None, omitnans: bool = False) -> Tensor:
```
rusty1s, line 70: It looks like the `batch` argument support will be dropped here (mean `iou` across all examples). This might be a problem.

Borda: ok, then we can do the batch over the IoU... but shall we do it while you suggest dropping metrics completely?

rusty1s: What do you mean with doing the batch over the IoU?

Borda:
> I'm fine with a hard removal TBH since I doubt that these are functions that most PyG users are actively using.

this is something I also do not know that is why I started wth soft removal and can be dropped completely in net release... but happy to follow any your suggestion :)

> * If we go for a deprecation warning, is there any chance we can make use of the in-house [`deprecated` warning](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/deprecation.py) rather than relying on the external dependency?

sure I was using a deprecation package I know, but happy to use any according to your suggestion :)

rusty1s:
Let's go for hard removal :)

Borda:
> Let's go for hard removal :)

Done :+1:

examples/dgcnn_segmentation.py
```diff
@@ -86,11 +86,8 @@ def test(loader):
     for data in loader:
         data = data.to(device)
         pred = model(data).argmax(dim=1)
-
-        i, u = i_and_u(pred, data.y, loader.dataset.num_classes, data.batch)
-        iou = i.cpu().to(torch.float) / u.cpu().to(torch.float)
-        iou[torch.isnan(iou)] = 1
-
+        iou = jaccard_index(pred, data.y,
```
Borda, line 89: @rusty1s so you say We shall implement the batch strategy here, in the example, right?

rusty1s: Let me verify the output in a mini-batch scenario (with a dynamic number of points per example) and come back to you (probably on Monday). This is my only concern. Otherwise the PR looks good to me (thank you very much).

rusty1s: What I mean by this is that I think we need to compute `jacard_index` with `reduction = None`, and then apply the reduction by ourselves, e.g., via `scatter_mean(iou, batch, dim=0).mean()`.

Borda: sure, because otherwise we can use also the `nn.Module` metrics version with `update` and `compute` :rabbit:

Borda: cc: @SkafteNicki

rusty1s: Updated the metric computation in the examples, in which IoU is now computed for each example individually and averaged afterwards. I think this leads to the correct computation now (please confirm).

I think there exists no alternative way to avoid the for-loop at the moment.

rusty1s (approved):

examples/dgcnn_segmentation.py
```diff
@@ -80,24 +80,16 @@ def train():
...
-        iou = i.cpu().to(torch.float) / u.cpu().to(torch.float)
-        iou[torch.isnan(iou)] = 1
+        out = model(data).argmax(dim=1)
 
-        # Find and filter the relevant classes for each category.
-        for iou, category in zip(iou.unbind(), data.category.unbind()):
-            ious[category.item()].append(iou[y_mask[category]])
+        sizes = (data.ptr[1:] - data.ptr[:-1]).tolist()
+        for pred, y in zip(out.split(sizes), data.y.split(sizes)):
+            iou += float(jaccard_index(pred, y))
```
SkafteNicki, line 90: @Borda in v0.8 the num_classes will be required in `jaccard_index`
```suggestion
            iou += float(jaccard_index(pred, y, num_classes=loader.dataset.num_classes))
```

examples/point_transformer_segmentation.py
```diff
@@ -197,24 +197,16 @@ def train():
...
-        iou = i.cpu().to(torch.float) / u.cpu().to(torch.float)
-        iou[torch.isnan(iou)] = 1
+        out = model(data.x, data.pos, data.batch).argmax(dim=1)
 
-        # Find and filter the relevant classes for each category.
-        for iou, category in zip(iou.unbind(), data.category.unbind()):
-            ious[category.item()].append(iou[y_mask[category]])
+        sizes = (data.ptr[1:] - data.ptr[:-1]).tolist()
+        for pred, y in zip(out.split(sizes), data.y.split(sizes)):
+            iou += float(jaccard_index(pred, y))
```
SkafteNicki, line 207:
```suggestion
            iou += float(jaccard_index(pred, y, num_classes=loader.dataset.num_classes))
```

examples/pointnet2_segmentation.py
```diff
@@ -106,24 +106,16 @@ def train():
...
-        iou = i.cpu().to(torch.float) / u.cpu().to(torch.float)
-        iou[torch.isnan(iou)] = 1
+        out = model(data).argmax(dim=1)
 
-        # Find and filter the relevant classes for each category.
-        for iou, category in zip(iou.unbind(), data.category.unbind()):
-            ious[category.item()].append(iou[y_mask[category]])
+        sizes = (data.ptr[1:] - data.ptr[:-1]).tolist()
+        for pred, y in zip(out.split(sizes), data.y.split(sizes)):
+            iou += float(jaccard_index(pred, y))
```
SkafteNicki, line 116:
```suggestion
            iou += float(jaccard_index(pred, y, num_classes=loader.dataset.num_classes))
```

SkafteNicki (approved):

examples/dgcnn_segmentation.py
rusty1s: This leads to:
`You have set 50 number of classes which is different from predicted (32) and target (3) number of classes`

PR #4292, opened 2022-03-18
extend pre-commit

### LINKED ISSUE #4289: move lints to pre-commit
### 🛠 Proposed Refactor

I would suggest moving most of the extra linting from https://github.com/pyg-team/pytorch_geometric/blob/master/.github/workflows/linting.yml
to recommit as you already use the bot and set the check as mandatory :rabbit: 

check that can be safely moved to pre-commit:
- [flake8](https://flake8.pycqa.org/en/latest/user/using-hooks.html)
- [yamllint](https://yamllint.readthedocs.io/en/stable/integration.html)
- sort - it is already there
- [pylint](https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html)
- [pyroma](https://github.com/regebro/pyroma#version-control-integration)

Also, happy to send a PR if you agree with it... :rabbit:

Borda:
Resolves #4289


.github/workflows/linting.yml
```diff
@@ -8,85 +8,8 @@ on:  # yamllint disable-line rule:truthy
 
 jobs:
 
-  flake8:
```
rusty1s, line 11: Any reason to remove them here? As fas as I know, installation of `pre-commit` is not required, right?

.pre-commit-config.yaml
```diff
@@ -2,6 +2,23 @@ ci:
...
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      #- id: check-docstring-first
+      - id: check-json
+      - id: check-yaml
+      - id: check-toml
+      #- id: check-case-conflict
```
rusty1s, line 14: Do these checks (`docstring-first` and `case-conflict`) fail at the moment?

rusty1s (approved):

.pre-commit-config.yaml
Borda: no, they just were not part of my suggestion so was feeling hesitant to include them without your approval :rabbit:

.github/workflows/linting.yml
Borda: they are just duplicates to what the pre-commit does, so you want to jeep both?
note that your pre-cimmot[bot] is called on each PR too

Borda: image

rusty1s: Yes, thanks for clarifying. It is indeed good to remove them.

Borda:
Btw, have noticed that updating YAPF to the latest v0.32 will generate a significant portion of changes

rusty1s:
Yes, please leave it at 0.31 for now :)

.github/workflows/linting.yml
rusty1s: The required checks should be resolved.

Borda:
> Yes, please leave it at 0.31 for now :)

yes I will, just for the record, the pre-commit[bot] has set bu default auto-update so it may happen very soon that it create a PR here with updating all hooks to the latest version including YAPF v0.32

Issue #4288, opened 2022-03-17
extend line length to 120


Borda:
### 🛠 Proposed Refactor

I would update the YAPF an Isort formating to use 120 line length which is nowadays standard... the default 80 is too restrictive and eve GH web UI is fine with 120 chars per line :+1: 

### Suggest a potential alternative/fix

If you are fine with it I am happy to send this config change and then let pre-commit to execute the update :rabbit: 

```
[yapf]
COLUMN_LIMIT=120

[isort]
...
line_length = 120

[flake8]
...
max-line-length = 120
```

Also, happy to send a PR if you agree with it... :rabbit:

rusty1s:
Actually, I prefer the 80 line length limit and am in favor of keeping it:
* It leaves space for putting two versions side by side for comparison (even on a laptop)
* It limits the complexity of a single line by enforcing that we are not composing too many functions in a single line
* I absolutely hate vertical scrolling or breaking lines

Borda:
btw, a bit off-topic, how about Isort formatting when long imports are on separate lines?
https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html
My preference is for option 3 :rabbit:

rusty1s:
I personally like both option 0 and 3. If you feel so strongly about it, feel free to send a PR to see how it looks like in practice :)

PR #4295, opened 2022-03-18
`isort`: `multi_line_output=3` (vertical hanging indent)

Borda:
Resolves https://github.com/pyg-team/pytorch_geometric/issues/4288#issuecomment-1072183187

The nice feature of this formating is that update of imports is easier to spot as long imports are on individual lines so only the particular line is changed... :rabbit:

### DIFF
--- a/benchmark/kernel/asap.py
+++ b/benchmark/kernel/asap.py
@@ -2,8 +2,12 @@ import torch
 import torch.nn.functional as F
 from torch.nn import Linear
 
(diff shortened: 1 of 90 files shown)


rusty1s:
Any chance we can add a trailing comma to the last element of the list? Otherwise, `yapf` will try to re-format :)

Borda:
> Any chance we can add a trailing comma to the last element of the list? Otherwise, `yapf` will try to re-format :)

yes, I have the solution to this :)

rusty1s (approved):
Thank you. I like it :)

PR #4313, opened 2022-03-21
'added_test_for_elliptic_dataset'

shravankumar147:


rusty1s (approved):
Thanks for reporting and providing the test case, really cool :)

The issue is now resolved by fixing the parentheses in:
```python
train_mask = (time_step < 35) & (y != 2)
test_mask = (time_step >= 35) & (y != 2)
```
I also moved the dataset to `data.pyg.org` for faster downloading.

shravankumar147:
Thanks, really appreciate you for thinking about download time, it was taking too long before. I will check and share my thoughts here.

PR #4338, opened 2022-03-25
Fix graphgym config storage directory

JiaxuanYou:
Previously, GraphGym config YAML was stored in the root directory of the experiments, which was not desirable. Now, the config will be correctly stored in each specific experiment directory.


rusty1s (approved):
Perfect :)

PR #4357, opened 2022-03-27
Add `v2` option for `GAT` model

j-adamczyk:
Added argument for GAT model to be able to use `GATv2Conv`.


rusty1s (approved):
Thank you! Renamed the argument to `v2` and added a basic test :)

PR #4375, opened 2022-03-29
Fix unbound local variable error in `NeighborLoader`

mananshah99:


torch_geometric/loader/neighbor_loader.py
```diff
@@ -236,6 +236,7 @@ def __init__(
                                 collate_fn=self.neighbor_sampler, **kwargs)
 
     def transform_fn(self, out: Any) -> Union[Data, HeteroData]:
+        data = None
```
Padarn, line 239: Does it make sense for the `out` to be None? Should we perhaps not fail silently in this case?

If it does make sense then the return type of this function should now be `Optional[Union[Data, HeteroData]]`

rusty1s: Thanks for the catch. Added an additional `assert` statement to ensure this.

rusty1s (approved):

Issue #1147, opened 2020-04-22
How to implement other GraphSAGE Aggregator functions like Max Pool?


sbonner0:
Thanks for all the great work on the package and sorry if this has been asked before.

I was wondering if it possible to use different aggregation functions, like Max-Pool or LSTM aggregation from the original GraphSAGE paper, with the existing SAGEConv layer? Looking at the documentation, it seems that only the mean aggregation is available.

rusty1s:
Max-Pool is indeed possible, but we do not have an implementation for it. Feel free to add it if you like. A fast LSTM aggregation is currently possible, but I am working on it.

sbonner0:
Hi, thanks so much for the reply. I would love to add it if I am able.
Would it be possibe to implemet it how it is in the paper (with a second paramter matrix W_pool) using the current SAGEConv layer? Or would I need to implement it from scratch using  MessagePassing? Would you have any tips on where to start?

rusty1s:
We should maybe first come up with a separate implementation and then decide how to merge the different variants. I suggest to copy most of the `SAGEConv` layer as a starting point. And yes, you need to introduce `W_pool` for the max aggregation.

sbonner0:
Thanks - I'll hopefully get time to look at this over the next week.

cluel01:
Hey guys, is there an update on the two aggregator functions ? Would be keen to know.

sbonner0:
Hey - I am still planning to have a go at this as soon as possible, I was inundated with work since the lock down. 
If you have any idea, or would like to help, I would be most grateful.

jchaykow:
It looks like maybe you just have to add an additional function [here](https://pytorch-scatter.readthedocs.io/en/latest/_modules/torch_scatter/scatter.html#scatter) maybe called `scatter_maxpool` or something?

rusty1s:
`torch-scatter` does already support `max` reduction :)

vijay2411:
Hi,
Is there any update on max Pool aggregator? I would love to know.
Also I think, can we not just change the aggregator value by letting pass an extra argument in SAGEConv constructor of torch_geometric/nn/conv/sage_conv.py ? Is there anything else that needs to be done? The aggregator value there is hard-coded as 'add'.
Thanks

rusty1s:
You can install from source:
```
pip install git+https://github.com/rusty1s/pytorch_geometric.git
```
or change the `aggr` flag after initialization:
```python
[...]

PR #4379, opened 2022-03-30
Support for `LSTM` aggregator in `SAGEConv`

hunarbatra:
This pull request adds the implementation for LSTM, Max Pool and GCN aggregation for SAGEConv which were implemented in the original paper/code [Inductive Representation Learning on Large Graphs](https://arxiv.org/pdf/1706.02216.pdf) and the absence of these aggregator functions implementation has been brought up in issue #1147. Addition of these aggregators to SAGEConv would help others looking to use these with PyTorch Geometric SAGEConv, since DGL supports them too!

Thank you to @rusty1s for guiding me with the LSTM implementation! :)

#### Notes
- I have removed `kwargs.setdefault('aggr', 'mean')` because that was making SAGEConv inherit 'mean' aggr from `MessagePassing`.
- I've added `aggregator_type` argument which is set to `mean` by default


torch_geometric/nn/conv/sage_conv.py
```diff
@@ -23,6 +25,7 @@ class SAGEConv(MessagePassing):
             A tuple corresponds to the sizes of source and target
             dimensionalities.
         out_channels (int): Size of each output sample.
+        aggregator_type (str): ['mean', 'max', 'gcn', 'lstm'], mean by default
```
rusty1s, line 28: Let's go with `aggr` here to align with PyG's naming.

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -45,30 +54,45 @@ class SAGEConv(MessagePassing):
         - **outputs:** node features :math:`(|\mathcal{V}|, F_{out})` or
           :math:`(|\mathcal{V_t}|, F_{out})` if bipartite
     """
-    def __init__(self, in_channels: Union[int, Tuple[int, int]],
-                 out_channels: int, normalize: bool = False,
+    def __init__(self, in_channels: Union[int, Tuple[int,
+                                                     int]], out_channels: int,
+                 aggregator_type: str = 'mean', normalize: bool = False,
                  root_weight: bool = True, bias: bool = True, **kwargs):
```
rusty1s, line 60:
```suggestion
                 root_weight: bool = True, bias: bool = True, **kwargs,):
```

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -45,30 +54,45 @@ class SAGEConv(MessagePassing):
...
-        kwargs.setdefault('aggr', 'mean')
         super().__init__(**kwargs)
```
rusty1s, line 61:
```suggestion
        super().__init__(aggr=aggr, **kwargs)
```

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -45,30 +54,45 @@ class SAGEConv(MessagePassing):
...
         self.root_weight = root_weight
+        self.aggregator_type = aggregator_type
+
+        assert self.aggregator_type in ['mean', 'max', 'lstm', 'gcn', None]
 
         if isinstance(in_channels, int):
             in_channels = (in_channels, in_channels)
 
-        self.lin_l = Linear(in_channels[0], out_channels, bias=bias)
+        if self.aggregator_type == 'gcn':
```
rusty1s, line 74: I'm not sure why we need `gcn` style aggregation here. This is already implemented in `GCNConv`.

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -45,30 +54,45 @@ class SAGEConv(MessagePassing):
...
+            # Convolutional aggregator does not concatenate the root node
+            # i.e it doesn't concatenate the nodes previous layer
+            self.root_weight = False
+
+        if self.aggregator_type == 'lstm':
```
rusty1s, line 79:
```suggestion
        if self.aggr == 'lstm':
```

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -92,7 +119,19 @@ def forward(self, x: Union[Tensor, OptPairTensor], edge_index: Adj,
...
+        if self.aggregator_type == 'mean' or self.aggregator_type == 'gcn':
+            return matmul(adj_t, x[0], reduce='mean')
+
+        elif self.aggregator_type == 'max':
+            return matmul(adj_t, x[0], reduce='max')
+
+        elif self.aggregator_type == 'lstm':
+            x_j = x[0][edge_index_j]
+            x, mask = to_dense_batch(x_j, edge_index_i)
+            return self.lstm(x)
```
rusty1s, line 137: Doesn't `self.lstm` return a tuple?

rusty1s (commented):
Thank you :) Left some comments.

torch_geometric/nn/conv/sage_conv.py
hunarbatra: Yes, correct. I just tested it out and it is returning a tuple and throwing this error: 
`linear(): argument 'input' (position 1) must be Tensor, not tuple`

Earlier when I was testing out, I realised that I did not transform the dataset with ToSparseTensor() and hence it was going to the regular message(), aggregate(), update() flow instead of going to message_and_aggregate().

torch_geometric/nn/conv/sage_conv.py
hunarbatra: I added that because the paper has "GraphSAGE-GCN" as well. But I can definitely remove that since the same behaviour can be achieved by using:
```SAGEConv(in_channels, out_channels, aggr='mean', root_weight=False)```

**Comment by hunarbatra** [comment]
I have converted LSTM's output from tuple to tensor. Kindly review it and let me know if it seems good to you and if something should be changed.

Also, setting `self.aggr = 'lstm'` throws an error since self.aggr initialises the parent constructor aggr value i.e in MessagePassing (where aggr is in AGGRS = {'add', 'sum', 'mean', 'min', 'max', 'mul'}). 

So, currently with my changes LSTM is only available when the edge_index is a SparseTensor and when message_and_aggregate() is implemented.

Please let me know if you have any suggestions for making LSTM work even with Tensors

File: torch_geometric/nn/conv/sage_conv.py [hunk 7]
```diff
@@ -45,10 +54,18 @@ class SAGEConv(MessagePassing):
...
+        in_channels: Union[int, Tuple[int, int]],
+        out_channels: int,
+        aggr: str = 'mean',
+        normalize: bool = False,
+        root_weight: bool = True,
+        bias: bool = True,
+        **kwargs,
+    ):
+
+        kwargs.setdefault("aggr", aggr if aggr != 'lstm' else 'mean')
```
rusty1s, line 68:
```suggestion
        kwargs.setdefault("aggr", aggr if aggr != 'lstm' else None)
```

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -92,7 +123,21 @@ def forward(self, x: Union[Tensor, OptPairTensor], edge_index: Adj,
...
         adj_t = adj_t.set_value(None, layout=None)
-        return matmul(adj_t, x[0], reduce=self.aggr)
+        if self.lstm is not None:
+            x_j = x[0][edge_index_j]
+            x, mask = to_dense_batch(x_j, edge_index_i)
+            _, (rst, _) = self.lstm(x)
+            out = rst.squeeze(0)
+            return out
+
+        elif self.aggr == 'mean':
```
rusty1s, line 139:
```suggestion
        else:
```
should work here as well.

rusty1s (commented):
Thanks for the updates! One thing I think we still need to fix is that this currently only works in case the input is a `SparseTensor`. IMO, it would be great to have more general support for this directly inside `aggregate` rather than inside `message_and_aggregate`. For `aggr='lstm'`, we would need to set `self.fuse = False` such that `message_and_aggregate` is not called. Furthermore, we would need to ensure that `edge_index_i` is sorted, and fail with a `ValueError` otherwise. We can easily check this via `(edge_index_j[:-1] <= edge_index[1:]).all())`. WDYT?

wilcoln:
> Thanks for the updates! One thing I think we still need to fix is that this currently only works in case the input is a `SparseTensor`. IMO, it would be great to have more general support for this directly inside `aggregate` rather than inside `message_and_aggregate`. For `aggr='lstm'`, we would need to set `self.fuse = False` such that `message_and_aggregate` is not called. Furthermore, we would need to ensure that `edge_index_i` is sorted, and fail with a `ValueError` otherwise. We can easily check this via `(edge_index_j[:-1] <= edge_index[1:]).all())`. WDYT?

Hey @rusty1s , @hunarbatra  How about converting `edge_index` into a `SparseTensor` before passing it to propagate as follows:

```
        # Convert edge_index to a sparse tensor,
        # this is required for propagate to call message_and_aggregate
            num_nodes = int(edge_index.max()) + 1
            edge_index = SparseTensor(row=edge_index[0], col=edge_index[1], sparse_sizes=(num_nodes, num_nodes))

        # propagate_type: (x: OptPairTensor)
        # propagate internally calls message_and_aggregate()
        out = self.propagate(edge_index, x=x, size=size)
        out = self.lin_l(out)
```

rusty1s:
@wilcoln Yes, this works as well. The problem with this approach is that we would need to convert to `SparseTensor` in every GNN layer (which is an expensive op that ideally only needs to be done once).

hunarbatra:
@rusty1s  Thank you! Yes, definitely having a more general support in aggregate() would be better and I have been working on it. I feel just checking if edge_index_i is not really working - because it still fails for some common datasets like Cora with dimension issues (when adding root embeddings in update out += self.lin_r(x_r)) even when edge_index_i is “sorted” - after debugging I noticed that the dimension generated with to_dense_batch() [batch_size part of out i.e first dim] is not correct and hence it’s leading to the dimension issues in further steps. Do you have any suggestion for this?

And thanks @wilcoln, that approach definitely works well - but I noticed that when I am converting it to SparseTensor and then applying LSTM - the F1 score is pretty low (for eg: just 14% for supervised Cora)

rusty1s:
@hunarbatra Can you show me an example on how you tried to integrate it into `message` and `aggregate`?

hunarbatra:
```python

def aggregate(self, inputs: Tensor, edge_index_i, edge_index_j) -> Tensor:
        # LSTM
        if (edge_index_i[:-1] <= edge_index_i[1:]).all().tolist() is True:
            x_j = inputs[edge_index_j]
            x, mask = to_dense_batch(x_j, edge_index_i) # batch=edge_index_i should be "ordered"
            _, (hidden, _) = self.lstm(x)
            out = hidden.squeeze(0)
            return out
        else:
            raise ValueError("edge_index_i is not ordered")
```

rusty1s (approved):
Thank you. I pushed the logic to `aggregate` and added a basic test case.

rusty1s:
Do you have a minimal script to reproduce? Please also have a look at https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/conv/test_sage_conv.py#L59-L73.

germayneng:
Thanks for the test reference. i found the issue:
I was setting aggr based on the latest release of 2.0.4 via:

```
self.conv[i].aggr = "lstm"
```

using the param aggr seems to fixed the issue

PR #4396, opened 2022-04-02
Link level `NeighborLoader`

Padarn:
Adds a link neighbourhood loader that is intended to provide utility for sampling for link prediction tasks.

The current API looks like: 
```
loader = LinkNeighborLoader(normal_data, num_neighbors=[2, 1], batch_size=1, shuffle=False)
data, target_edge_index, target_labels = next(iter(loader))
```
The loader iterates giving the sampled g raph and the target edge index and labels for the edges in the graph. The labels are assumed to be "True" for all edges in the graph.

TODO 
- [x] Add hetro graph compatability
- [x] Cleanup hetro/non-hetro type checking
- [x] Examples
- [x] Tests

Addresses https://github.com/pyg-team/pytorch_geometric/issues/4026


torch_geometric/loader/link_neighbour_loader.py
```diff
@@ -0,0 +1,95 @@
...
+
+        batch_size = len(index)
+
+        # take start and end node from each edge then deduplicate
+        row, col = self.data.edge_index
+        node_index = torch.cat([row[index], col[index]], dim=0)
+        node_index = torch.unique(node_index)
+
+        # get sampled graph
+        node, row, col, edge, _ = super().__call__(index)
```
rusty1s, line 35:
```suggestion
        node, row, col, edge, _ = super().__call__(node_index)
```

torch_geometric/loader/link_neighbour_loader.py
```diff
@@ -0,0 +1,95 @@
...
+class LinkNeighborSampler(NeighborSampler):
+    def __init__(
+        self,
+        data: Data,
+        num_neighbors: NumNeighbors,
+        replace: bool = False,
+        directed: bool = True,
+    ):
+        super().__init__(data, num_neighbors, replace, directed)
+        self.data = data
```
rusty1s, line 23: It is pretty important to not save `data` here (as otherwise the data needs to be shared or replicated across `num_workers` during data loading). We could construct `node_index` directly from `self.colptr` and `self.row`.

torch_geometric/loader/link_neighbour_loader.py
```diff
@@ -0,0 +1,95 @@
...
+        row, col = self.data.edge_index
+        node_index = torch.cat([row[index], col[index]], dim=0)
+        node_index = torch.unique(node_index)
+
+        # get sampled graph
+        node, row, col, edge, _ = super().__call__(index)
+        return node, row, col, edge, batch_size, index
+
+
+class LinkNeighborLoader(BaseDataLoader):
```
rusty1s, line 39:
```suggestion
class LinkNeighborLoader(torch.utils.data.DataLoader):
```
`BaseDataLoader` does no longer exist :(

torch_geometric/loader/link_neighbour_loader.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(
+        self,
+        data: Data,
+        num_neighbors: NumNeighbors,
+        input_edge_idx: InputEdges = None,
```
rusty1s, line 44:
```suggestion
        input_edges: InputEdges = None,
```
Aligns better with `NeighborLoader`. WDYT?

torch_geometric/loader/link_neighbour_loader.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def transform_fn(self, out: Any) -> Tuple[Data, torch.Tensor]:
+        node, row, col, edge, batch_size, index = out
+        data = filter_data(self.data, node, row, col, edge,
+                           self.neighbor_sampler.perm)
+
+        data.batch_size = batch_size
+        data = data if self.transform is None else self.transform(data)
+
+        labels = self.get_input_edge_labels()[index]
```
rusty1s, line 80: Shall we write this directly into `data`? (`edge_label` and `edge_label_index`)?

rusty1s, line 80: Any chance we only compute `get_input_edge_labels` and `get_input_edge_idx` once in `__init__`?

rusty1s (commented):
This is a pretty good start. Really cool :)

Left some comments!

torch_geometric/utils/mask.py
```diff
@@ -20,7 +20,7 @@ def index_to_mask(index: Tensor, size: Optional[int] = None) -> Tensor:
 
 
 def mask_to_index(mask: Tensor) -> Tensor:
-    r"""Converts mask to indicies representation 
+    r"""Converts mask to indicies representation
```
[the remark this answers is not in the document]
Padarn: I need to set up spell checking in my IDE 😅

torch_geometric/loader/link_neighbour_loader.py
Padarn: Oh I must need to rebase on master. Thanks for the heads up.

torch_geometric/loader/link_neighbour_loader.py
Padarn: Yeah I was wondering about that. I thought it might be a little bit confusing to have a property that didn't exist on all edges, but its definitely much cleaner.

Padarn:
Thanks, a lot of really good feedback. I'll address it all as I continue to work on the PR.

torch_geometric/loader/link_neighbour_loader.py
Padarn: Good point!

torch_geometric/loader/link_neighbour_loader.py
Padarn: Oh I see the change in https://github.com/pyg-team/pytorch_geometric/pull/4377 - I don't know much about lightning but its on my list to better understand what assumptions it is making about loaders :-)

torch_geometric/loader/neighbor_loader.py
```diff
@@ -13,12 +13,14 @@
...
-from torch_geometric.typing import EdgeType, InputNodes
-
-NumNeighbors = Union[List[int], Dict[EdgeType, List[int]]]
+from torch_geometric.typing import InputNodes, NodeType, NumNeighbors
+from torch_geometric.utils import mask_to_index
 
 
 class NeighborSampler:
+
+    device = "cpu"
```
Padarn, line 22: I'm really not a fan of doing this, but it seemed a bit pointless to make it configurable, and I think its better than hard coding it in more places

torch_geometric/nn/models/deep_graph_infomax.py
```diff
@@ -51,10 +51,12 @@ def discriminate(self, z, summary, sigmoid=True):
 
         Args:
             z (Tensor): The latent space.
+            summary (Tensor): The summary vector.
```
Padarn, line 54: Not totally sure why this one is here, it also seems to be in master.

torch_geometric/loader/link_neighbour_loader.py
Padarn: The reason I changed this was that I need to eventually recover the index of the edges to look up any edge attributes that exist. So if taking as input the actual edges I need to do a search for each edge in the `data.edge_index`.

But... it does seem somehow more usable if you provide edges, so I'm a fan of your suggestion with the caveat that we'll need to do a mapping from edges -> edges_index.

Updating the PR with this, but let me know if you had something else in mind.

Padarn: Actually, I can't see an efficient way to do this... leaving both in the MR.. but I'm hesitant to keep the format without 
it explicitly being an index.

Padarn: Going to resolve this and move it as this file is gone (renamed)

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -0,0 +1,146 @@
...
+            if isinstance(input_edges, EdgeType):
+                raise RuntimeError("`input_edges` cannot be string for non"
+                                   "hetrogenous graphs")
+
+            input_size = input_edges.size()
+
+            if len(input_size) == 1 and input_edges.dtype == torch.bool:
+                return mask_to_index(self.input_edge_idx)
+
+            if len(input_edges.size()) == 2 and input_edges.size()[0] == 2:
```
Padarn, line 128: Restart the conversation on the input specification here: The reason I initially changed the input name to `input_edge_idx` was so that it was clear we wouldn't do this (an inverse mapping from edges to their index). We need the index eventually because we need to recover the edge attributes in the transform function.

I'd be a fan of removing this personally and assuming its an index that is provided to this class - Although it would be nice to allow people to provide edges that don't exist and sample the graph for them, its also confusing what to do for the edge attributes in that case.

WDYT @rusty1s? Open to other approaches.

torch_geometric/loader/link_neighbour_loader.py
rusty1s: There exists currently a bug in PL such that custom loaders need to directly inherit from `torch.utils.data.DataLoader`.

torch_geometric/loader/link_neighbour_loader.py
rusty1s: I think this replicates our `RandomLinkSplit`quite well in which we have `data.edge_index` for message passing` and `data.edge_label_index` for supervision (with potentially different shape).

torch_geometric/loader/link_neighbor_loader.py
rusty1s: I am not really sure I fully understand. Why do we need to recover edge attributes of supervision edges (that potentially do not even exist)?

In case you require an index to this loader, this means that you can only sample links that actually exist (i.e. which are "indexable" by `edge_index[idx]`), which does not seem to be the preferred solution. Ultimately, we want to infer missing links here as well.

torch_geometric/loader/neighbor_loader.py
rusty1s: What's the reason for this change?

torch_geometric/nn/models/deep_graph_infomax.py
rusty1s: Looks like something went wrong during rebase.

torch_geometric/loader/link_neighbor_loader.py
Padarn: Got your point. I was being stupid.

I've updated the PR (it's still messy, but hopefully we're more aligned on the design). The only part I'm uncertain about is on line 85. Will open a conversation there.

torch_geometric/loader/neighbor_loader.py
Padarn: I did this initially because I was storing something on CPU in the LinkNeighbourSampler too, so it seemed better to make sure they all shared the same device.. I'll revert

torch_geometric/loader/link_neighbour_loader.py
Padarn: I've updated this to store `sampled_edges` and `sampled_edge_labels` (and `sampled_index`, maybe we can drop this one), so that it's easy to work with labels of edges that don't exist in the original graph.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -0,0 +1,108 @@
...
+
+        data.batch_size = batch_size
+        data = data if self.transform is None else self.transform(data)
+
+        data.sampled_index = index
+
+        edges = self.input_edges[:, index]
+        data.sampled_edges = edges
+
+        if self.input_edge_labels is not None:
```
Padarn, line 85: I wanted originally to return automatically the label as 'True' for existing edges and 'false' for non-existing ones, but doing this is a bit expensive (need to figure out which on the input edges are in the graph). Not sure its worth it.

Thoughts?

torch_geometric/nn/models/deep_graph_infomax.py
Padarn: Yeah, I'll rebase and squash my PRs once the rest of the conversations are resolved

test/loader/test_neighbor_loader.py
```diff
@@ -54,7 +54,7 @@ def test_homogeneous_neighbor_loader(directed):
         assert is_subset(batch.edge_index, data.edge_index, batch.x, batch.x)
 
         # Test for isolated nodes (there shouldn't exist any):
-        assert data.edge_index.view(-1).unique().numel() == data.num_nodes
+        assert batch.edge_index.view(-1).unique().numel() == batch.num_nodes
```
Padarn, line 57: I assumed this was a mistake, can revert if not (otherwise same check is done for each batch)

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -0,0 +1,307 @@
...
+                dim=0)
+            query_nodes, reverse_query_nodes = torch.unique(
+                query_nodes, return_inverse=True)
+
+            # get sampled graph
+            node, row, col, edge, _ = super().__call__(query_nodes)
+
+            # assume order of nodes in the new graph is same as input
+            assert all(query_nodes == node[:len(query_nodes)])
+            out_graph_nodes = torch.argsort(query_nodes)
```
Padarn, line 70: Any idea if there is a better way to do this? I do see https://discuss.pytorch.org/t/how-to-quickly-inverse-a-permutation-by-using-pytorch/116205 suggests a minor optimization, and there are a few on StackOverflow (I can benchmark some improvements in another MR), but I'm wondering if I'm missing a way to avoid this entirely.

test/loader/test_neighbor_loader.py
rusty1s: Thank you :)

torch_geometric/loader/link_neighbor_loader.py
rusty1s: This looks good to me. I think it would be interesting to benchmark the speed here compared to not doing any `unique` call.

Furthermore, you may want to remove the `assert` statement since this will always be true by design. Using `all()` will also call a really slow Python iterator implementation.

Padarn: Okay fair enough - I added it incase something changes in the implementation of the base class, but our tests should also catch this.

test/loader/test_neighbor_loader.py
Padarn: Actually.. this test fails now 😓  personally I'm not sure if its a problem if this is not true. There could be isolated nodes in the subgraph?

rusty1s (approved):
Thank you!

PR #4397, opened 2022-04-02
Implement Disentangled GCN from ICML2019

bwdeng20:
This layer can take the `edge_index` or `SparseTensor` subgraph induced by the targeted nodes and their neighbor. The original paper assumes all neighbors are sampled, so `torch_geometric.loader.NeighborLoader` with  `num_neighbors=[-1]` can gather data required by `DisenConv` layer.

### DIFF
--- a/torch_geometric/nn/sequential.py
+++ b/torch_geometric/nn/sequential.py
@@ -69,7 +69,7 @@ def Sequential(
 
     Args:
         input_args (str): The input arguments of the model.
-        modules ([(str, Callable) or Callable]): A list of modules (with
+        modules ([(Callable, str) or Callable]): A list of modules (with
             optional function header definitions). Alternatively, an
             :obj:`OrderedDict` of modules (and function header definitions) can
             be passed.


test/nn/conv/test_disen_gcn_conv.py
```diff
@@ -0,0 +1,41 @@
+import torch
+from torch_sparse import SparseTensor
+
+from torch_geometric.nn import DisenConv
+from torch_geometric.testing import is_full_test
+
+
+def test_gcn_conv():
```
rusty1s, line 8:
```suggestion
def test_disen_gcn_conv():
```

torch_geometric/nn/conv/__init__.py
```diff
@@ -49,64 +49,20 @@
...
+    'GravNetConv', 'GatedGraphConv', 'ResGatedGraphConv', 'GATConv',
+    'GATv2Conv', 'TransformerConv', 'AGNNConv', 'TAGConv', 'GINConv',
+    'GINEConv', 'ARMAConv', 'SGConv', 'APPNP', 'MFConv', 'RGCNConv',
+    'FastRGCNConv', 'RGATConv', 'SignedConv', 'DNAConv', 'PointNetConv',
+    'PointConv', 'GMMConv', 'SplineConv', 'NNConv', 'ECConv', 'CGConv',
+    'EdgeConv', 'DynamicEdgeConv', 'XConv', 'PPFConv', 'FeaStConv',
+    'PointTransformerConv', 'HypergraphConv', 'LEConv', 'PNAConv',
+    'ClusterGCNConv', 'GENConv', 'GCN2Conv', 'PANConv', 'WLConv', 'FiLMConv',
+    'SuperGATConv', 'FAConv', 'EGConv', 'PDNConv', 'GeneralConv', 'HGTConv',
+    'HEATConv', 'HeteroConv', 'HANConv', 'LGConv', 'DisenConv'
```
rusty1s, line 65:
```suggestion
    'HEATConv', 'HeteroConv', 'HANConv', 'LGConv', 'DisenConv',
```

torch_geometric/nn/conv/disen_conv.py
```diff
@@ -0,0 +1,155 @@
...
+import torch.nn.functional as F
+from torch import Tensor
+from torch_sparse import SparseTensor, matmul
+
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.dense.linear import Linear
+
+
+class DisenConv(MessagePassing):
+    r"""The graph convolutional operator from the `"Disentangled graph
```
rusty1s, line 13: Can we add more information about the operator here (similar to other GNN layers)?

torch_geometric/nn/conv/disen_conv.py
```diff
@@ -0,0 +1,155 @@
...
+
+    Args:
+      in_channels (int): The number of input channels.
+      out_channels (int): The number of channels of the output feature.
+      K (int): The number of channels in the output.
+      T (int): The number of iterations. Defaults to 5
+      tao (float): The time decay parameter.
+      separate_channels (List[int]): This is a list of integers, which
+                specifies the number of channels for each factor.
+      weight_initializer (Optional[str]): The initializer for
```
rusty1s, line 25: Is support for `weight_initializer` and `bias_initializer` desperately needed here? I think we should drop this to align better with existing GNN layers.

torch_geometric/nn/conv/disen_conv.py
```diff
@@ -0,0 +1,155 @@
...
+            bias=True,
+            weight_initializer=weight_initializer,
+            bias_initializer=bias_initializer,
+        )
+
+    def reset_parameters(self):
+        self.lin_projector.reset_parameters()
+
+    def get_projector(
+        self, separate=False
```
rusty1s, line 86:
```suggestion
        self, separate: bool=False
```

torch_geometric/nn/conv/disen_conv.py
```diff
@@ -0,0 +1,155 @@
...
+        elif isinstance(edge_index, SparseTensor):
+            dst_nodes, src_nodes, _ = edge_index.coo()
+        else:
+            raise TypeError(f"{type(edge_index)} not a valid graph type")
+        bs = batch_size if batch_size is not None \
+            else dst_nodes.unique().numel()
+
+        Z = self.project2aspects(x)
+        # target nodes always placed first
+        C = [chunk[:bs].clone() for chunk in Z]
```
rusty1s, line 131: Not really sure why we need to slice here based on `batch_size`. Can you explain?

bwdeng20: Assume there are `K` Disentangled factors responsible for `d1,d2,...,dK` output channels separately.  We have `batch_size` target nodes and the induced `1-hop` subgraph with totally `N` nodes.

`Z=[Z1(N, d1), Z2(N, d2), ...,  ZK(N, dK)]` is a list of `K` aspects of `N` nodes.  During the routing iteration described in the following figure, we need to reserve the entire `Z` for message passing and initialize the `K` segments of output of target nodes with `Zk[:batch_szie]`s.  `C` is alongside cloned to simultaneously allocate memory for the final disentangled output and initialize them. 

 
屏幕截图 2022-04-06 154146

torch_geometric/nn/conv/disen_conv.py
bwdeng20: I will remove them.

torch_geometric/nn/conv/disen_conv.py
bwdeng20: Yes. Once we reach a consensus on the code.

torch_geometric/nn/conv/disen_conv.py
bwdeng20: ## 1. Input Tuple (without `batch_size`)
If we input a tuple `(x_src, s_dst)`, the code would be like:
```python

def forward_bipartite(self, x: OptPairTensor, edge_index: Union[SparseTensor, Tensor], size: Size = None) -> Tensor:
            src_nodes, dst_nodes = edge_index
        elif isinstance(edge_index, SparseTensor):
            dst_nodes, src_nodes, _ = edge_index.coo()
        else:
            raise TypeError(f"{type(edge_index)} not a valid graph type")

        x_src, x_dst = x
        bipartite_size = (x_src.shape[0] + x_dst.shape[0], x_dst.shape[0]) if size is None else size

        z_dst = self.project2aspects(x_dst)
        z_src = self.project2aspects(x_src)
        z_all = [torch.cat([z_dst[k], z_src[k]]) for k in range(self.K)]  ## 1st marker 

        c_dst = [ck.clone() for ck in z_dst]

        for i in range(self.T):
            p = self.update_p(z_all, c_dst, dst_nodes, src_nodes)
            for k in range(self.K):
                graph = (
                    edge_index.set_value(p[k], layout="coo")
                    if isinstance(edge_index, SparseTensor)
                    else edge_index
                )
                ck = ( ## 2nd marker 
                        self.propagate(graph, x=(z_all[k], z_dst[k]), edge_weight=p[k],  size=bipartite_size)+ z_dst[k]
                )
                c_dst[k] = ck / ck.norm(dim=-1, keepdim=True)

        out = torch.cat([c_dst[k] for k in range(self.K)], dim=1)
        return out
```

This leads to a concatenation operation in the `1st marker` line. To avoid the allocation of new memory caused by `torch.cat`, it would be better to input `(x_all, x_dst)` to bypass `cat`. 

## 2. Input All (with `batch_size`) like the last commit 
If we input `x` of all nodes and the target nodes are first placed, `batch_szie` is required to locate the target nodes.

Actually, I prefer the solution of inputting `(x_all, x_dst)` without `batch_size` now.

bwdeng20:
## 1. Input Tuple (without `batch_size`)
If we input a tuple `(x_src, s_dst)`, the code would be like:
```python

def forward_bipartite(self, x: OptPairTensor, edge_index: Union[SparseTensor, Tensor], size: Size = None) -> Tensor:
            src_nodes, dst_nodes = edge_index
        elif isinstance(edge_index, SparseTensor):
            dst_nodes, src_nodes, _ = edge_index.coo()
        else:
            raise TypeError(f"{type(edge_index)} not a valid graph type")

        x_src, x_dst = x
        bipartite_size = (x_src.shape[0] + x_dst.shape[0], x_dst.shape[0]) if size is None else size

        z_dst = self.project2aspects(x_dst)
        z_src = self.project2aspects(x_src)
        z_all = [torch.cat([z_dst[k], z_src[k]]) for k in range(self.K)]  ## 1st marker 

        c_dst = [ck.clone() for ck in z_dst]

        for i in range(self.T):
            p = self.update_p(z_all, c_dst, dst_nodes, src_nodes)
            for k in range(self.K):
                graph = (
                    edge_index.set_value(p[k], layout="coo")
                    if isinstance(edge_index, SparseTensor)
                    else edge_index
                )
                ck = ( ## 2nd marker 
                        self.propagate(graph, x=(z_all[k], z_dst[k]), edge_weight=p[k],  size=bipartite_size)+ z_dst[k]
                )
                c_dst[k] = ck / ck.norm(dim=-1, keepdim=True)

        out = torch.cat([c_dst[k] for k in range(self.K)], dim=1)
        return out
```

This leads to a concatenation operation in the `1st marker` line. To avoid the allocation of new memory caused by `torch.cat`, it would be better to input `(x_all, x_dst)` to bypass `cat`. 

## 2. Input All (with `batch_size`) like the last commit 
If we input `x` of all nodes and the target nodes are first placed, `batch_szie` is required to locate the target nodes.

Actually, I prefer the solution of inputting `(x_all, x_dst)` without `batch_size` now.

torch_geometric/nn/conv/disen_conv.py
bwdeng20: thx

test/nn/conv/test_disen_gcn_conv.py
bwdeng20: thx

torch_geometric/nn/conv/__init__.py
bwdeng20: I'm Ok with this. BTW, what benefits this comma can brings?

rusty1s: It will reformat the code so that each entry is inside a new row. This is useful to easily see the diffs in code.

torch_geometric/nn/conv/disen_conv.py
rusty1s: Thanks for clarification. I don't think we need to do anything with `batch_size` at all. The input matrix `x` is used for both central node features and features of neighbors such that `c_k` is just given via `Z[k]`, right? We can then calcuate `p_{v,k}` directly inside message (by querying the central node `c_i` and its neighbors `z_j`).

rusty1s:
Please see my comment above. I think neither `batch_size` nor `(x_all, x_dst)` is required.

torch_geometric/nn/conv/disen_conv.py
bwdeng20: Thanks for the comments. 

> The input matrix `x` is used for both central node features and features of neighbors such that `c_k` is just given via `Z[k]`?


Not exactly correct. `c_k` is initialized with `Z[k][dst_nodes]` but going to deviate from this fixed initialization value during the following iteration.  So we have to preserve both `Z[k]` and `c_k`.

I think it might be not that good to calculate `p_{v,k}` during the propagation. The message propagation is conducted for all `K` factors. For the propagation of `k`-th factor, only the inner product of `z_{v,k}` and `c_{k}` is computed, but  `p_{v,k}` is obtained via a softmax operation across all `K` such inner products.  This is why I prefer to compute all   `p_{v,k}` s before propagation.

rusty1s: Sounds good. In this case, we should go for the input tuple approach as illustrated by your earlier. Importantly, we can also support `x` as a `torch.Tensor` via:
```python
if isinstance(x, Tensor):
    x = (x, x)
    
x_src, x_dst = x
```

bwdeng20: OK. I will commit later. 😃

PR #4398, opened 2022-04-02
Add example for GraphRNN

### LINKED ISSUE #983: Generative Modelling on Graphs Request.
I love Pytorch Geometric and it's ease of use over other graph libs. However, one pain point is that there doesn't seem to be easy to use graph generative models. In particular, it would be really to have some canonical Generative models on graphs which already have pytorch code. Personally, I would love if these ones could be added:

1) GRAN: https://github.com/lrjconan/GRAN
2) GraphRNN: https://github.com/JiaxuanYou/graph-generation

I'm happy to help build this together as well :)

### LINKED ISSUE #3311: Example section for graph based generative models
## 🚀 Feature

I was wondering if there is a plan to provide examples for graph based generative models/if you think that the field is not mature enough when compared with node representation learning/link prediction. For example, I see that graph autoencoders with the inner product decoder are provided as an example of the link prediction task. It would be nice to see an example of the same graph autoencoder but with a probabilistic decoder to show capabilities in graph generation. 

Let me know what you think!

orausch:
This PR adds an example implementation of GraphRNN-S, a generative graph model described in 
[GraphRNN: Generating Realistic Graphs with Deep Auto-regressive Models
](https://arxiv.org/abs/1802.08773).

The implementation consists of two transforms (`BFS`, `EncodeGraphRNNFeature`) and a model.
Together, the transforms generate the sequences described in the paper and add these as a feature of the graphs. The model then trains the generative, RNN-based model on these sequences. The model includes a `sample` method that is used to generate graphs.

We're happy to discuss if any parts of the example (e.g. the model or transforms) could be moved into the main project.

The example trains such a model on a dataset consisting of only cycles. It takes around 1 min to run on my GTX 960.
Here is some sample output from the example:
```
Percentage of generated graphs that are cycles at epoch 0: 0.0
Percentage of generated graphs that are cycles at epoch 100: 0.296875
Percentage of generated graphs that are cycles at epoch 200: 0.265625
Percentage of generated graphs that are cycles at epoch 300: 0.28125
Percentage of generated graphs that are cycles at epoch 400: 0.234375
Percentage of generated graphs that are cycles at epoch 500: 0.265625
Percentage of generated graphs that are cycles at epoch 600: 0.953125
Percentage of generated graphs that are cycles at epoch 700: 1.0
Percentage of generated graphs that are cycles at epoch 800: 0.984375
Percentage of generated graphs that are cycles at epoch 900: 1.0
Percentage of generated graphs that are cycles at epoch 1000: 0.984375
```

Closes #3311, closes #983

cc @JiaxuanYou

### DIFF
--- a/examples/graphrnn_s.py
+++ b/examples/graphrnn_s.py
@@ -222,8 +222,7 @@ class GraphRNN_S(nn.Module):
             enforce_sorted=False,
         )
         if hidden is not None:
-            output_sequences, self.hidden = self.rnn(input_sequences,
-                                                     hidden)
+            output_sequences, self.hidden = self.rnn(input_sequences, hidden)
         else:
             output_sequences, self.hidden = self.rnn(input_sequences)
         # Unpack RNN output.
--- a/examples/graphrnn_s.py
+++ b/examples/graphrnn_s.py
@@ -38,7 +38,6 @@ class EncodeGraphRNNFeature(T.BaseTransform):
         :rtype: Tensor
         """
         N = adj.shape[1]
-        adj = adj.reshape(N, N)
         padded_adj = torch.zeros((N + M - 1, N))
         padded_adj[M - 1:, :] = adj
         return padded_adj.as_strided(size=(N - 1, M), stride=(N + 1, N),
@@ -202,7 +201,7 @@ class GraphRNN_S(nn.Module):
                                                  batch_first=True)[0]
         return sequences
 
-    def forward(self, input_sequences, input_length, sampling=False):
+    def forward(self, input_sequences, input_length, hidden=None):
         r"""Forward pass of the GraphRNN-S model.
 
         Args:
@@ -211,6 +210,7 @@ class GraphRNN_S(nn.Module):
                 :obj:`(batch_size, max_num_nodes, adjacency_size)`
             input_length (Tensor): For each graph in the batch, the length of
                 the sequence of adjacency vectors. Shape: :obj:`(batch_size,)`
+            hidden (Tensor, optional): Initial hidden state.
         """
         input_sequences = self.embedding(input_sequences)
 
@@ -221,9 +221,9 @@ class GraphRNN_S(nn.Module):
             batch_first=True,
             enforce_sorted=False,
         )
-        if sampling:
+        if hidden is not None:
             output_sequences, self.hidden = self.rnn(input_sequences,
-                                                     self.hidden)
+                                                     hidden)
         else:
             output_sequences, self.hidden = self.rnn(input_sequences)
         # Unpack RNN output.
@@ -264,17 +264,20 @@ class GraphRNN_S(nn.Module):
                 if node_id == max_num_nodes:
                     break
 
+                # generate the sampling probabilities for the adjacency vectors
                 output_sequence_probs = self.forward(input_sequence,
                                                      torch.ones(batch_size),
-                                                     sampling=True)
+                                                     hidden=self.hidden)
+                # sample the adjacency vectors
                 mask = torch.rand_like(output_sequence_probs)
                 output_sequence = torch.gt(output_sequence_probs, mask)
 
-                # Identify the EOS sequences.
+                # Identify the EOS sequences
+                # (these are the adjacency where no edge was sampled).
                 is_not_eos *= output_sequence.any(dim=-1).squeeze().cpu()
                 seq_lengths += is_not_eos
 
-                sequences[:, node_id - 1] = output_sequence[:, 0]
+                sequences[:, node_id - 1] = output_sequence.squeeze(1)
                 input_sequence = output_sequence.float()
 
         # Clean irrelevant bits and enforce creation of connected graph.
--- a/examples/graphrnn_s.py
+++ b/examples/graphrnn_s.py
@@ -1,16 +1,16 @@
 import argparse
 import random
 
-import networkx as nx
 import matplotlib.pyplot as plt
-
+import networkx as nx
 import torch
 from torch import nn
 from torch.nn import functional as F
 from torch.nn.utils import rnn as rnnutils
 
 import torch_geometric
-from torch_geometric import data, transforms as T, loader
+from torch_geometric import data, loader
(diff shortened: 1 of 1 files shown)


examples/graphrnn_s.py
```diff
@@ -0,0 +1,362 @@
...
+        diagonal of the given square matrix.
+
+        Args:
+            adj (Tensor): The adjacency matrix of shape :obj:`(N, N)`.
+            M (int): The number of bands to extract.
+
+        :rtype: Tensor
+        """
+        N = adj.shape[1]
+        adj = adj.reshape(N, N)
```
Padarn, line 41: I don't think you need this? It'll either do nothing or not work?

examples/graphrnn_s.py
```diff
@@ -0,0 +1,362 @@
...
+            y (Tensor): The encoded sequence of shape :obj:`(N, M)`.
+
+        :rtype: Tensor
+        """
+        bands = torch.flip(y, dims=[1])
+        adj = EncodeGraphRNNFeature.bands_to_matrix(bands)
+        return adj
+
+    def __call__(self, data):
+        adj = torch_geometric.utils.to_dense_adj(data.edge_index)
```
Padarn, line 80: Is it possible to do this without going to the dense adjacency matrix? Perhaps it should be quite easy to extract the bands from a CSR format as sparse tensors?

examples/graphrnn_s.py
```diff
@@ -0,0 +1,362 @@
...
+        self.mask_out_bits_after_length(sequences, seq_lengths + 1)
+        sequences = sequences.tril()
+
+        return sequences[:, :seq_lengths.max()], seq_lengths
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--plot", action="store_true",
+                    help="Whether to plot graphs.")
+args = parser.parse_args()
```
Padarn, line 291: maybe good to have a default or "parse_known_args" so the example can be run easily without arguments?

examples/graphrnn_s.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def forward(self, input_sequences, input_length, sampling=False):
+        r"""Forward pass of the GraphRNN-S model.
+
+        Args:
+            input_sequences (Tensor): For each graph in the batch, the sequence
+                of adjacency vectors (including the first SOS token). Shape:
+                :obj:`(batch_size, max_num_nodes, adjacency_size)`
+            input_length (Tensor): For each graph in the batch, the length of
+                the sequence of adjacency vectors. Shape: :obj:`(batch_size,)`
```
Padarn, line 213: sampling missing from docstring btw (I noticed because I don't understand what it does 😄

Padarn:
I don't know the paper (yet, I'll take a read), but just left a few general comments from reading the codde

examples/graphrnn_s.py
orausch: Yes, there are several options here:
Using a sparse mask (which is then used with `csr.multiply(mask).todia()` or `csr[mask]`)
```python
def construct_sparse_mask(N, M):
    """
    Construct a sparse mask consisting of the M diagonals above the main
    diagonal of a NxN matrix.
    """

    sum_to_M = int((M * (M + 1)) / 2)
    num_entries = N * M - sum_to_M
    data = np.ones((num_entries, ), dtype=np.bool_)

    start_x = np.repeat(np.arange(N - M), M)
    end_x = np.concatenate([np.full((i, ), N - i -1) for i in reversed(range(M))])
    indices_x = np.concatenate([start_x, end_x])

    ending_y = np.concatenate([np.arange(N - i, N) for i in reversed(range(M))])
    start_y = np.tile(np.arange(M) + 1, N - M)
    view_start_y = start_y.reshape(-1, M)
    view_start_y += np.arange(N - M).reshape(-1, 1)
    indices_y = np.concatenate([start_y, ending_y])

    return sp.coo_matrix((data, (indices_x, indices_y)), shape=(N, N))
```

Using diagonals:
```python
def extract_bands_sparse_dia(adj_sparse, M):
    diags = [np.pad(adj_sparse.diagonal(k=i), (i - 1, 0), mode='constant')
             for i in range(1, M + 1)]
    return np.stack(diags).T
```

However, it's a bit unclear what the best option is. For M << N `extract_bands_sparse_dia` is clearly faster, but when the graph is quite full (and `M` is high as a result, this may no longer be the case).

Any thoughts?

examples/graphrnn_s.py
orausch: With `argparse`, `--<...>` arguments are optional, so this example can already be run without arguments.

examples/graphrnn_s.py
orausch: Renamed to `hidden`

examples/graphrnn_s.py
Padarn: Thanks for the detailed ideas! 

If I understand correctly, all three methods will essentially create a dense matrix anyway - in the first method above you create a dense mask, in the second the call `adj_sparse.diagonal` creates dense vectors, so if M is large it'll still be pretty large.

I can't see a big advantage between any of the approaches, except maybe to me the `extract_bands_sparse_dia` is most readable (but I agree it could be slow. In terms of performance, we might have to rely on the sampled graph being small enough to easily fit in memory as a dense graph?

What do you think?

examples/graphrnn_s.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class BFS(T.BaseTransform):
+    r"""Start a breath first search from a random node and reorder the edge
+    list so that the node indices correspond to the breadth-first search order.
+    """
+    def __call__(self, data):
+        x = data.x
+        edge_index = data.edge_index
+        assert (data.is_undirected()
```
Padarn, line 101: just FYI I noticed that `EncodeGraphRNNFeature` works with an undirected graph right now, but I don't know if it makes sense in this case. Maybe its worth having the same assert there if not.

rusty1s:
@JiaxuanYou Might be best to get some further input from the author as well :)

PR #4411, opened 2022-04-04
Optional dependencies

rusty1s:


torch_geometric/nn/conv/message_passing.py
```diff
@@ -630,6 +629,8 @@ def jittable(self, typing: Optional[str] = None):
                 instance with :meth:`forward` types based on :obj:`typing`,
                 *e.g.*: :obj:`"(Tensor, Optional[Tensor]) -> Tensor"`.
         """
+        from jinja2 import Template
```
Padarn, line 632: Do you think it might be worth having a warning if jinja2 is not installed when importing message_passing, otherwise when the import error comes later it can be a bit confusing.

rusty1s: Done.

Padarn (approved):

PR #4416, opened 2022-04-05
Unsupervised `GraphSAGE` example on `PPI`

wilcoln:
This PR adds an example of unsupervised training of an MP-based GNN (in our case GraphSAGE) on the graph-level PPI task. An attempt to reproduce the corresponding experiment from [Inductive Representation Learning on Large Graphs](https://arxiv.org/abs/1706.02216)

We build upon the existing PPI supervised example ([here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/ppi.py)), and the node-level unsupervised GraphSAGE example on the Citation Dataset ([here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/https://github.com/pyg-team/pytorch_geometric/blob/master/examples/graph_sage_unsup.py)).

The Downstream classifier used is the logistic SGDClassifier from the scikit-learn library, the same as in the original paper.

The authors report a test F1 score of `0.486` when using the `mean` aggregator after 10 epochs. This code produces slightly lower results with a Test F1 of around `0.45`.

In collab with @hunarbatra


examples/ppi_unsup.py
```diff
@@ -0,0 +1,164 @@
...
+        return super().sample(batch)
+
+
+train_loader_list = []
+
+for curr_graph in train_data_list:
+    _train_loader = NeighborSampler(curr_graph.edge_index, sizes=[10, 10],
+                                    batch_size=256, shuffle=True,
+                                    num_nodes=curr_graph.num_nodes)
+    train_loader_list.append(_train_loader)
```
rusty1s, line 50: I don't think we want to use a list of samplers here. IMO, it is a better idea to first batch all graphs together, and then apply `NeighborSampler` on the entire set of nodes:
```python
batch = Batch.from_data_list(train_dataset)
loader = NeighborSampler(batch)
```

rusty1s (commented):
Thank your for this example. Since `NeighborSampler` is deprecated in favor of `NeighborLoader`, we might want to wait for https://github.com/pyg-team/pytorch_geometric/pull/4396 to get merged first and then update the example accordingly. Would that work for you?

examples/ppi_unsup.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class GraphSAGE(torch.nn.Module):
```
Padarn, line 53: I'm wondering if this can be written using the `torch_geometric.nn.models.GraphSage` (although would need to also do the conversion to `NeighborLoader` suggested above.

examples/ppi_unsup.py
```diff
@@ -0,0 +1,164 @@
...
+    train_predictions = clf.predict(train_embeddings)
+    train_f1 = f1_score(train_labels, train_predictions, average='micro')
+
+    # Evaluate on validation set
+    val_embeddings, val_labels = get_representations_and_labels(
+        model, val_loader)
+    val_predictions = clf.predict(val_embeddings)
+    val_f1 = f1_score(val_labels, val_predictions, average='micro')
+
+    # Evaluate on validation set
```
Padarn, line 150:
```suggestion
    # Evaluate on testing set
```

rusty1s:
**Update:** The `LinkNeighborLoader` now supports the `neg_sampling_ratio` argument (see https://github.com/pyg-team/pytorch_geometric/pull/4446, thanks to @Padarn). We could update this PR to make use of it instead of relying on the old `NeighborSampler` approach. Let me know what you think :)

examples/ppi_unsup.py
wilcoln: Thanks, the results are 40% better now that I'm batching all train graphs together.

wilcoln:
@Padarn I'm not entirely sure if I'm making good use of `LinkNeighborLoader`'s pos neg sampling here. I've just assumed that first, we have batch nodes, then positive samples, then negative samples. As was the case with its predecessor `NeighborSampler`.

Padarn:
You're right in your assumption - but we also store the labels in the returned data object
```
from torch_geometric.loader import LinkNeighborLoader

data = Data(x=torch.ones(3), edge_index=torch.tensor([[0, 1], [1, 2]]))
sub_data = next(iter(LinkNeighborLoader(
    data,
    num_neighbors=[1],
    neg_sampling_ratio=0.5,
    batch_size=2
)))
sub_data.edge_labels
>>> tensor([1., 1., 0.])
```

examples/ppi_unsup.py
rusty1s: +1 on using `torch_geometric.nn.GraphSAGE` directly: `model = GraphSAGE(in_channels, hidden_channels, num_layers=3, out_channels=hidden_channels)`

examples/ppi_unsup.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def train():
+    model.train()
+
+    total_loss = 0
+    for data in loader:
+        data = data.to(device)
+        optimizer.zero_grad()
+        out = model(data.x, data.edge_index)[:data.edge_label.shape[0]]
```
rusty1s, line 62: There is no need for this separation anymore. You should be able to simply do:
```python
out = model(data.x, data.edge_index)
link_pred = (out[data.edge_label_index[0]] * out[data.edge_label_index[1]]).sum(dim=-1)
loss = F.bce_loss(out, data.edge_label)
```

examples/ppi_unsup.py
```diff
@@ -0,0 +1,119 @@
...
+    test_predictions = clf.predict(test_embeddings)
+    test_f1 = f1_score(test_labels, test_predictions, average='micro')
+
+    return train_f1, val_f1, test_f1
+
+
+for epoch in range(1, 51):
+    loss = train()
+    train_f1, val_f1, test_f1 = test(train_loader, val_loader, test_loader)
+    print('Epoch: {:03d}, Loss: {:.4f}, Train F1: {:.4f}, '
```
rusty1s, line 117: `test()` might be expensive since we run a `clf` classifier on CPU in every epoch. Shall we only run this every 5th epoch?

examples/ppi_unsup.py
wilcoln: Thanks @rusty1s, did you mean `F.bce_loss(out, data.edge_label)`

rusty1s: Yes, sorry :)

rusty1s:
Thanks for the updates. This looks pretty good. I will make a last pass tomorrow and merge afterwards.

Just curious: Do you have any feedback to the new `LinkNeighborLoader`? Anything unexpected or weird going on in your opinion? Also: Did model performance change or did it stay the same?

wilcoln:
> Regarding `LinkNeighborLoader`.
Well, It took me a while to understand how it works since it operates on edges instead of nodes, but in the end, it leads to a more elegant and concise code, so I like it. A minor suggestion would be to rename it to `EdgeNeighborLoader`, for more consistency.

> Regarding the performance.
I'm getting the same as yesterday (~`0.45` in the first epoch). I actually checked the practical correctness of what you proposed ([here](https://github.com/pyg-team/pytorch_geometric/pull/4416#discussion_r855003671)) by comparing it to what I get when separately computing the positive and negative loss for each batch using the old code, and they match.

In short, everything looks good as far as I can tell.

rusty1s (approved):
Thank you!

PR #4418, opened 2022-04-05
Update message_passing.py

xionggziran:
When using a bipartite graph for reverse message passing(set flow = 'target_to_source'), the message aggregation fails or  returns a tensor of the wrong dimension due to incorrect dim_size, although it doesn't really matter and different edge_index can be used to achieve the same effect.
ps:I'm new to pyg, so please forgive me if it's not correct.


rusty1s (approved):
Thanks for catching!

PR #4424, opened 2022-04-06
Updates to dimenet model

arunppsg:
Two minor fixes to dimenet model:
1. When initializing `model = DimeNet(hidden_channels = 5, out_channels=19, num_blocks=1, num_bilinear=1, num_spherical=1, num_radial=1)`, it throws a `RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.`. It occurs in the `reset_parameters` method of `BesselBasisLayer`.  The cause is that we are performing an in-place operation in the leaf variable when resetting parameters.
2. The model requires `num_spherical` to be at least 2. I am not sure why it is, but it fixes. Sorry about it.

### DIFF
--- a/torch_geometric/nn/models/dimenet.py
+++ b/torch_geometric/nn/models/dimenet.py
@@ -59,7 +59,9 @@ class BesselBasisLayer(torch.nn.Module):
         self.reset_parameters()
 
     def reset_parameters(self):
-        torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)
+        with torch.no_grad():
+            torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)
+        self.freq.requires_grad_()
 
     def forward(self, dist):
         dist = dist.unsqueeze(-1) / self.cutoff
@@ -293,6 +295,9 @@ class DimeNet(torch.nn.Module):
                  num_after_skip: int = 2, num_output_layers: int = 3,
                  act: Callable = swish):
         super().__init__()
+
+        if num_spherical < 2:
+            raise ValueError("num_spherical should be greater than 1")
 
         self.cutoff = cutoff
         self.max_num_neighbors = max_num_neighbors


arunppsg:
More details, the full error traceback in first scenraio:
```
RuntimeError                              Traceback (most recent call last)
Input In [4], in <module>
----> 1 model = DimeNet(hidden_channels = 5, out_channels=19, num_blocks=1, num_bilinear=1, num_spherical=1, num_radial=1)

File ~/proj/pyg/torch_geometric/nn/models/dimenet.py:301, in DimeNet.__init__(self, hidden_channels, out_channels, num_blocks, num_bilinear, num_spherical, num_radial, cutoff, max_num_neighbors, envelope_exponent, num_before_skip, num_after_skip, num_output_layers, act)
    298 self.max_num_neighbors = max_num_neighbors
    299 self.num_blocks = num_blocks
--> 301 self.rbf = BesselBasisLayer(num_radial, cutoff, envelope_exponent)
    302 self.sbf = SphericalBasisLayer(num_spherical, num_radial, cutoff,
    303                                envelope_exponent)
    305 self.emb = EmbeddingBlock(num_radial, hidden_channels, act)

File ~/proj/pyg/torch_geometric/nn/models/dimenet.py:59, in BesselBasisLayer.__init__(self, num_radial, cutoff, envelope_exponent)
     55 self.envelope = Envelope(envelope_exponent)
     57 self.freq = torch.nn.Parameter(torch.Tensor(num_radial))
---> 59 self.reset_parameters()

File ~/proj/pyg/torch_geometric/nn/models/dimenet.py:62, in BesselBasisLayer.reset_parameters(self)
     61 def reset_parameters(self):
---> 62     torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)

RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
```

For the second scenario, when num_spherical is less than two, it throws an error in this line:
https://github.com/pyg-team/pytorch_geometric/blob/c30d37158919cee1b3ddceb2b41b289006ed4929/torch_geometric/nn/models/dimenet_utils.py#L79
Here, it requires the value to be at-least two for the assignment statement.

torch_geometric/nn/models/dimenet.py
```diff
@@ -59,7 +59,9 @@ def __init__(self, num_radial, cutoff=5.0, envelope_exponent=5):
         self.reset_parameters()
 
     def reset_parameters(self):
-        torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)
+        with torch.no_grad():
```
rusty1s, line 62: Do we need the `requires_grad_()` call here? IMO, it should be sufficient to just do:
```python
@torch.no_grad()
def reset_parameters(self):
```

rusty1s (approved):

torch_geometric/nn/models/dimenet.py
arunppsg: I am not sure about it, one question I have is that if we don't use `requires_grad_()` how do we track gradient for `self.freq`?

arunppsg: I am not completely sure about it, I tried the following code snippets:
Snippet 1, with torch.no_grad:
```
PI = 3.14
class BBL1(torch.nn.Module):
    def __init__(self, num_radial, cutoff=5.0, envelope_exponent=5):
        super().__init__()
        self.freq = torch.nn.Parameter(torch.Tensor(num_radial))
        self.reset_parameters()

    @torch.no_grad()
    def reset_parameters(self):
        torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)

    def forward(self, dist):
        raise NotImplementError

bbl1 = BBL1(4)
bbl1.freq.requires_grad  # False
```
The output was `False`. Snippet 2, with require_grad
```
class BBL2(torch.nn.Module):
    def __init__(self, num_radial, cutoff=5.0, envelope_exponent=5):
        super().__init__()
        self.freq = torch.nn.Parameter(torch.Tensor(num_radial))
        self.reset_parameters()

    def reset_parameters(self):
        with torch.no_grad():
            torch.arange(1, self.freq.numel() + 1, out=self.freq).mul_(PI)
        self.freq.requires_grad_()

    def forward(self, dist):
        raise NotImplementError

bbl2 = BBL2(4)
bbl2.freq.requires_grad  # True
```
The output was `True.`

Each returns a different output. Am I missing something here?

arunppsg: Thanks!

PR #4432, opened 2022-04-07
`DimeNet++` implementation

### LINKED ISSUE #4427: Support for DimeNet++ model

The DimeNet model is a model which can be used for molecular property prediction. DimeNet++ model is an upgrade to the DimeNet model. DimeNet++ model is faster than DimeNet and more accurate ([ref](https://arxiv.org/abs/2011.14115)). It will be nice to have DimeNet++ model in PyG.

There are two possible ways to add DimeNet++:

1. The key difference between DimeNet and DimeNet++ is the InteractionBlock (DimeNet++ uses an updated version of InteractionBlock - called as InteractionPPBlock). Also, the authors proposed a minor upgrade to OutputBlock in DimeNet++ (OutputPPBlock). The `forward` method are same in both. Thus, we can subclass DimeNet++ to be a child class of DimeNet, and add an option in DimeNet `__init__`  method to use `DimeNet++` architecture rather than DimeNet architecture (something like `plusplus = True`, then initialize plusplus model, thereby using OutputPPBlock and InteractionPPBlock, we also need to add a couple of additional parameters required by DimeNet++).
2. Create an independent implementation of DimeNet++. There is a pytorch implementation [here](https://github.com/Open-Catalyst-Project/ocp/blob/main/ocpmodels/models/dimenet_plus_plus.py) which I adapted for PyG. The adapted version can be found [here](https://github.com/arunppsg/pytorch_geometric/blob/dimenet_pp/torch_geometric/nn/models/dimenet_plus_plus.py).

Thanks for consideration! Let me know if I can provide any other additional details.

arunppsg:
In this PR, I implemented DimeNet++ by subclassing DimeNet. The `InteractionPPBlock` and `OutputPPBlock` was based on this [implementation](https://github.com/Open-Catalyst-Project/ocp/blob/main/ocpmodels/models/dimenet_plus_plus.py).

Also, I depreciated `OutputBlock`, thereby DimeNet model will also be using `OutputPPBlock`. `OutputPPBlock` allows up-projection and down-projection of embeddings, thereby removing information bottlenecks.

Fixes #4427

### DIFF
--- a/torch_geometric/nn/models/dimenet.py
+++ b/torch_geometric/nn/models/dimenet.py
@@ -227,10 +227,10 @@ class InteractionPPBlock(torch.nn.Module):
     def __init__(self, hidden_channels, int_emb_size, basis_emb_size,
                  num_spherical, num_radial, num_before_skip, num_after_skip,
                  act=swish):
-        super(InteractionPPBlock, self).__init__()
+        super().__init__()
         self.act = act
 
-        # Transformation of Bessel and spherical basis representations
+        # Transformation of Bessel and spherical basis representations:
         self.lin_rbf1 = Linear(num_radial, basis_emb_size, bias=False)
         self.lin_rbf2 = Linear(basis_emb_size, hidden_channels, bias=False)
 
@@ -238,14 +238,15 @@ class InteractionPPBlock(torch.nn.Module):
                                bias=False)
         self.lin_sbf2 = Linear(basis_emb_size, int_emb_size, bias=False)
 
-        # Hidden transformation of input message
+        # Hidden transformation of input message:
         self.lin_kj = Linear(hidden_channels, hidden_channels)
         self.lin_ji = Linear(hidden_channels, hidden_channels)
 
-        # Embedding projections for interaction triplets
+        # Embedding projections for interaction triplets:
         self.lin_down = Linear(hidden_channels, int_emb_size, bias=False)
         self.lin_up = Linear(int_emb_size, hidden_channels, bias=False)
-        # Residual layers before and after skip connection
+
+        # Residual layers before and after skip connection:
         self.layers_before_skip = torch.nn.ModuleList([
             ResidualLayer(hidden_channels, act) for _ in range(num_before_skip)
         ])
@@ -278,24 +279,24 @@ class InteractionPPBlock(torch.nn.Module):
             res_layer.reset_parameters()
 
     def forward(self, x, rbf, sbf, idx_kj, idx_ji):
-        # Initial transformation
+        # Initial transformation:
         x_ji = self.act(self.lin_ji(x))
         x_kj = self.act(self.lin_kj(x))
 
-        # Transformation via Bessel basis
+        # Transformation via Bessel basis:
         rbf = self.lin_rbf1(rbf)
         rbf = self.lin_rbf2(rbf)
         x_kj = x_kj * rbf
 
-        # Down project embedding and generating triple-interactions
+        # Down project embedding and generating triple-interactions:
         x_kj = self.act(self.lin_down(x_kj))
 
-        # Transform via 2D spherical basis
+        # Transform via 2D spherical basis:
         sbf = self.lin_sbf1(sbf)
         sbf = self.lin_sbf2(sbf)
         x_kj = x_kj[idx_kj] * sbf
 
-        # Aggregate interactions and up-project embeddings
+        # Aggregate interactions and up-project embeddings:
         x_kj = scatter(x_kj, idx_ji, dim=0, dim_size=x.size(0))
         x_kj = self.act(self.lin_up(x_kj))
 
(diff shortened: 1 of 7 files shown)


rusty1s (commented):
Thanks for the PR. Can we add a basic test for `DimeNetPlusPlus` as well? Also I am interested in whether it is possible to make use of pre-trained weights for this model as well (similar to what we do for `DimeNet`), see [here](https://github.com/gasteigerjo/dimenet/tree/master/pretrained/dimenet_pp).

arunppsg:
Sure, I will look into using pre-trained weights and also add tests.

torch_geometric/nn/models/dimenet.py
```diff
@@ -242,11 +334,23 @@ def reset_parameters(self):
...
 
 
+class OutputBlock(OutputPPBlock):
+    def __init__(self, num_radial, hidden_channels, out_channels, num_layers,
+                 act=swish):
+        super().__init__(num_radial, hidden_channels, hidden_channels,
+                         out_channels, num_layers, act=swish)
+        import warnings
+        warnings.warn(
+            "OutputBlock will be depreciated in PyG 2.1.0. Use OutputPPBlock instead",  # noqa
```
Padarn, line 350: Any reason we don't use `OutputPPBlock` this will be depreciated?

arunppsg: The DimeNet++ paper says that OutputPPBlock has less information bottlenecks as it allows for information up projection and down projection of embedings. The difference between them is minor and I thought that OutputPPBlock will also benefit DimeNet model users. Hence the change.

But I have to test whether OutputPPBlock will be compatible with DimeNet when using pre-trained weights. If it is not compatible, I will revert back.

rusty1s: Thank you. On that note, it might makes sense to add a test for loading pre-trained weights in general. Can try to do this in a separate PR.

arunppsg: @Padarn I have restored OutputBlock as OutputPPBlock was not compatible with DimeNet when using pretrained weights.

@rusty1s , I would like to add tests in a separate PR.

Padarn: Thanks for the update @arunppsg. Once this is merged could you make an issue describing the `OutputPPBlock` problem? 🙇

arunppsg: I don't think it's a problem - earlier DimeNet was using OutputBlock, I thought it can use OutputPPBlock since it is more efficient but realized that it will not be compatible, then reverted back. So, users who want to use OutputPPBlock can directly use DimeNet++

Padarn: Oh I see now, sorry I got mixed up because I was reading it via email. Thanks.

arunppsg:
@rusty1s regarding tests for DimeNetPlusPlus, I have added a test for checking output size and another one test for checking whether the model is training well or not - test_overfit. The downside of test_overfit is that it is sometimes flaky but otherwise it is a good test for checking whether the model is able to train or not.

rusty1s:
Thank you @arunppsg! I am currently on vacation, will have a final look in the next week.

arunppsg:
The failures in testing are due to not having `sympy` in requirements. It is used by DimeNet++ and DimeNet for basis computation.

rusty1s:
@arunppsg Can we add the necessary dependencies to `full_install_requires`, see [here](https://github.com/pyg-team/pytorch_geometric/blob/master/setup.py#L22-L35)?

arunppsg:
@rusty1s added the necessary requirements.

rusty1s:
Thanks @arunppsg. I think I fixed the dependency issue during testing. I also added a `--use_dimenet_plus_plus` option to the QM9 script. However, the results of the pre-trained model are far off, e.g.:
```
Target: 00, MAE: 4.85522 ± 3.42503
Target: 01, MAE: 15.25113 ± 6.21239
Target: 02, MAE: 603.06146 ± 484.24606
Target: 03, MAE: 994.61578 ± 703.35645
Target: 05, MAE: 977.33118 ± 332.41879
Target: 06, MAE: 594.56433 ± 317.66690
Target: 07, MAE: 3095.07080 ± 1889.53760
```
Do you know what's the reason for this?

arunppsg:
> Thanks @arunppsg. I think I fixed the dependency issue during testing. I also added a `--use_dimenet_plus_plus` option to the QM9 script. However, the results of the pre-trained model are far off, e.g.:
> 
> ```
> Target: 00, MAE: 4.85522 ± 3.42503
> Target: 01, MAE: 15.25113 ± 6.21239
> Target: 02, MAE: 603.06146 ± 484.24606
> Target: 03, MAE: 994.61578 ± 703.35645
> Target: 05, MAE: 977.33118 ± 332.41879
> Target: 06, MAE: 594.56433 ± 317.66690
> Target: 07, MAE: 3095.07080 ± 1889.53760
> ```
> 
> Do you know what's the reason for this?

Not sure about it, the results are far off from the dimenet++ paper. Can we create an issue and track it seperately?

torch_geometric/nn/models/dimenet.py
```diff
@@ -454,8 +454,8 @@ def reset_parameters(self):
         for interaction in self.interaction_blocks:
             interaction.reset_parameters()
 
-    @staticmethod
-    def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
+    @classmethod
```
arunppsg, line 459: Just curious to know, how is using classmethod different from using staticmethod here?

rusty1s: I think this makes more sense here since it can naturally handle inheritance (`cls.url` except for `DimeNet.url`).

rusty1s:
This sounds good to me. I will try to look into this as well.

rusty1s (approved):

arunppsg:
Thank you!

PR #4437, opened 2022-04-09
Add support for projecting features before aggregation in `SAGEConv`

wilcoln:
This pull request adds the implementation for Max and Mean Pool aggregations for SAGEConv which were implemented in the original study [Inductive Representation Learning on Large Graphs](https://arxiv.org/pdf/1706.02216.pdf), it's a sequel to PR https://github.com/pyg-team/pytorch_geometric/pull/4379, which in turn is related to issue https://github.com/pyg-team/pytorch_geometric/issues/1147.

The implementation is a direct application of the Eq 3 of the original paper (for both `mean` and `max`):
image

cc @hunarbatra

### DIFF
--- a/test/nn/conv/test_sage_conv.py
+++ b/test/nn/conv/test_sage_conv.py
@@ -6,14 +6,15 @@ from torch_geometric.nn import SAGEConv
 from torch_geometric.testing import is_full_test
 
 
-def test_sage_conv():
+@pytest.mark.parametrize('project', [False, True])
+def test_sage_conv(project):
     x1 = torch.randn(4, 8)
     x2 = torch.randn(2, 16)
     edge_index = torch.tensor([[0, 1, 2, 3], [0, 0, 1, 1]])
     row, col = edge_index
     adj = SparseTensor(row=row, col=col, sparse_sizes=(4, 4))
 
-    conv = SAGEConv(8, 32)
+    conv = SAGEConv(8, 32, project=project)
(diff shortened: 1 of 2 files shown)


rusty1s (commented):
Thank you for the PR. I am not sure I fully understand this. How does that differ from `aggr=mean` and `aggr=max`? IMO, these already implement the linked Equation (we just added LSTM support to make `SAGEConv` finally support all proposed aggregations).

wilcoln:
Thanks for your comment @rusty1s .

As you know, the mean/max aggregator directly applies the mean/max reduction to all neighboring nodes.




While the mean/max pool aggregator first applies an additional linear layer (+ bias) followed by an activation function, and only then, performs the element-wise mean/max reduction.

rusty1s:
Thanks for the clarification. How about we add an option to apply the transformation before aggregation rather than after it. This way, there would be no need to modify the underlying aggregation scheme.

wilcoln:
That would be much cleaner indeed and would work with all existing aggregators, I will update my PR accordingly, by adding a `pool` boolean parameter to the class constructor, defaulting to `False`. How does that sound?

rusty1s:
Sounds good to me, thank you very much! I am wondering if `pool` is the most descriptive name for this. How about we name it `pre_transform` or something similar?

rusty1s (approved):
Thank you!

PR #4439, opened 2022-04-09
Fix `LinkNeighborLoader` in case `src_node_type = dst_node_type`

Padarn:
This PR fixes a corner case in the `LinkNeighborLoader` introduced in https://github.com/pyg-team/pytorch_geometric/pull/4396 when the `HeteroData` node type is a self-loop type.

- [x] Fix by taking union of nodes 
- [x] Test fix


rusty1s (approved):
LGTM :)

PR #4441, opened 2022-04-09
Convert edge types input for `LinkNeighborLoader` to canonical form.

Padarn:
Modifies the handling of the input types for edges so that the user can create the `LinkNeighborLoader` 
with either
```
loader = LinkNeighborLoader(data, num_neighbors=[-1] * 2,
                            edge_label_index=('paper', 'paper'))
```
or
```
loader = LinkNeighborLoader(data, num_neighbors=[-1] * 2,
                            edge_label_index=('paper', 'to', 'paper'))
```

### DIFF
--- a/torch_geometric/loader/link_neighbor_loader.py
+++ b/torch_geometric/loader/link_neighbor_loader.py
@@ -263,11 +263,13 @@ def get_edge_label_index(
 
     if isinstance(edge_label_index[0], str):
         edge_type = edge_label_index
+        edge_type = data._to_canonical(*edge_type)
         return edge_type, data[edge_type].edge_index
 
     assert len(edge_label_index) == 2
 
     edge_type, edge_label_index = edge_label_index
+    edge_type = data._to_canonical(*edge_type)
     if edge_label_index is None:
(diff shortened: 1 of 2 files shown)


torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -263,11 +263,13 @@ def get_edge_label_index(
 
     if isinstance(edge_label_index[0], str):
         edge_type = edge_label_index
+        edge_type = data._to_canonical(*edge_type)
         return edge_type, data[edge_type].edge_index
 
     assert len(edge_label_index) == 2
 
     edge_type, edge_label_index = edge_label_index
+    edge_type = data._to_canonical(*edge_type)
```
Padarn, line 272: not very happy accessing a "private" function - perhaps there is a better alternative

rusty1s: Currently no. I agree with your concerns, but it is not blocking. Merging nonetheless.

rusty1s (approved):
Thank you!

PR #4442, opened 2022-04-09
Add Geom-GCN splits to `Planetoid` dataset

dtortorella:
Add the option to use the 10-fold splits used in the ["Geom-GCN: Geometric Graph Convolutional Networks"](https://openreview.net/forum?id=S1e2agrFvS) paper.


rusty1s (approved):
Amazing. Thank you :)

PR #4446, opened 2022-04-10
Add random negative sampling to `LinkNeighborLoader`

Padarn:
This PR continues the work discussed in https://github.com/pyg-team/pytorch_geometric/issues/4026, adding a simple mechanism to do negative sampling.

The interface looks like the following:

```
loader = LinkNeighborLoader(..., neg_sampling_ratio=neg_sampling_ratio)
```

Then each batch returned will have `int(neg_sampling_ratio)*batch_size` negative samples, and 
`batch_size - int(neg_sampling_ratio)*batch_size` positive samples.

> Note that the sampling is not checked for false negatives

### DIFF
--- a/torch_geometric/loader/link_neighbor_loader.py
+++ b/torch_geometric/loader/link_neighbor_loader.py
@@ -202,8 +202,11 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             (default: :obj:`None`)
         neg_sampling_ratio (float, optional): The ratio of sampled negative
             edges to the number of positive edges. (default: :obj:`0`).
-            If set, the edge labels cannot be provided, as the labels now
-            become :obj:`0` for positive edges and  :obj:`1` for positive.
+            If set, and edge_labels are not provided, all provided edges will
+            be assumed to be positive and given a label of :obj:`1`. The
+            negative edges will have label :obj:`0`.
+            Note that these edges will be added to the batch, increasing the
+            batch size.
             If the data is :class:`~torch_geometric.data.HeteroData` then
             the edge type that is being sampled will be the edge type for
             which negative edges are added.
--- a/torch_geometric/loader/link_neighbor_loader.py
+++ b/torch_geometric/loader/link_neighbor_loader.py
@@ -17,28 +17,36 @@ class LinkNeighborSampler(NeighborSampler):
                  share_memory: bool = False, neg_sampling_ratio: float = 1.0):
         super().__init__(data, num_neighbors, replace, directed, input_type,
                          share_memory)
+
         self.neg_sampling_ratio = neg_sampling_ratio
+        if issubclass(self.data_cls, Data):
+            self.n_src = self.n_dst = len(data.x)
+        else:
+            self.n_src = len(data[self.input_type[0]].x)
+            self.n_dst = len(data[self.input_type[-1]].x)
+
+    @staticmethod
+    def _random_edges(n_src, n_dest, n_edges):
+        row = torch.randint(n_src, (n_edges, ), dtype=torch.long)
+        col = torch.randint(n_dest, (n_edges, ), dtype=torch.long)
+        return torch.stack([row, col], dim=0)
+
+    def _add_negative_edges(self, edge_label_index, edge_label):
+
+        n_query = edge_label_index.size()[1]
+        n_neg = int(n_query * self.neg_sampling_ratio)
+        if n_neg == 0:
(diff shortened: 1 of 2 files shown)


torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -11,7 +11,42 @@
 
 
 class LinkNeighborSampler(NeighborSampler):
+    def __init__(self, data: Union[Data, HeteroData],
```
Padarn, line 14: this could be put in the base class.. but I don't know if negative sampling makes sense for node sampler

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -11,7 +11,42 @@
...
+                 num_neighbors: NumNeighbors, replace: bool = False,
+                 directed: bool = True, input_type: Optional[Any] = None,
+                 share_memory: bool = False, neg_size: int = 0):
```
Padarn, line 17: I would have preferred to not have `neg_size` here, but instead again use `neg_sampling_ratio`, but to make the batch size not change, this would have meant modifying the ratio which I thought was confusing.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -189,13 +243,19 @@ def __init__(
         self.transform = transform
         self.neighbor_sampler = neighbor_sampler
 
+        batch_size = kwargs.get("batch_size", 1)
```
Padarn, line 246: I was thinking of instead just making negative samples make the batch size larger.. but this seemed like the lesser evil to me

torch_geometric/loader/link_neighbor_loader.py
rusty1s: I think we should just get the batch size directly in the sampler. This should be doable via `edge_label_index.size(1)`.

torch_geometric/loader/link_neighbor_loader.py
rusty1s: It looks good how it is. Alternatively, we could add `kwargs` support to `NeighborSampler` that just stores any additionally given arguments:
```python
for key, value in kwargs.items():
    setattr(self, key, value)
```

torch_geometric/loader/link_neighbor_loader.py
Padarn: So my thinking was, if we provide `neg_sampling_ratio` but don't adjust the batch size, then we'd need to either increase the size of the batch or throw away some of the batch: Throwing away seemed like something people might not want to do.

But, happy to adapt to either of these. What do you think?

torch_geometric/loader/link_neighbor_loader.py
Padarn: I slightly prefer what I've got so will leave it here. Makes it a bit more explicit what properties can be expected.

torch_geometric/loader/link_neighbor_loader.py
rusty1s: Ah, I see. I think there is no reason to modify `batch_size` according to `neg_sampling_ratio` since I personally would have expected that `batch_size` only refers to positive examples. I get your point though. We can also modify `batch_size` in `__init__` and pass `neg_sampling_ratio` down the line to `LinkNeighborSampler`.

Padarn: Yeah thats a little cleaner.. although I'd need to probably modify `neg_sampling_ratio` which is a bit confusing.

Example
- `batch_size=5`
- `neg_sampling_ratio=0.5`
- `new_batch_size=3`
- `new_neg_sampling_ratio= 2/3`

the new `new_neg_sampling_ratio` is needed because of the rounding...

But if you think just adding to the batch is natural for users, I like this better, I'll change to that :-)

rusty1s: If we are operating on fixed `neg_size`, then this will also be a problem for the last batch, so there is always some kind of trade-off we make :)

Padarn: done

Padarn: > If we are operating on fixed neg_size, then this will also be a problem for the last batch, so there is always some kind of trade-off we make :)

Yep good point. I think whatever is most predictable to users is probably the best.

rusty1s (approved):
Amazing :)

PR #4463, opened 2022-04-12
Edge Masking to generate Explanations

fork123aniket:
This PR is an extension of discussion #4440. Moreover, this PR contains implementation of how to compute layer-wise weights for each edge in order to produce explanations for both node-level and graph-level tasks. Furthermore, this implementation is different from authors' original implementation and is fast and more memory efficient than theirs.

### DIFF
--- a/torch_geometric/nn/models/GraphMask.py
+++ b/torch_geometric/nn/models/GraphMask.py
@@ -1,0 +1,865 @@
+from typing import Optional
+
+import math
+from math import sqrt
+from inspect import signature
+import numpy as np
+
+import torch
+from torch import Tensor
+from torch.nn import Linear, LayerNorm, ReLU, Parameter, init, Sequential
+import torch.nn.functional as F
+from torch import sigmoid
+from tqdm import tqdm
+from torch_geometric.data import Data
+from originalMessagePassing import MessagePassing
+from torch_geometric.utils import k_hop_subgraph, to_networkx
+from torch_geometric.nn.models.explainer import Explainer
+
+EPS = 1e-15
+
+
+def explain_message(self, out, x_i, x_j):
+    basis_messages = Sequential(LayerNorm(out.size(-1)), ReLU())(out)
+
+    if getattr(self, 'message_scale', None) is not None:
+        basis_messages = basis_messages * self.message_scale.unsqueeze(
+            -1)
+
+        if self.message_replacement is not None:
+            if basis_messages.shape == self.message_replacement.shape:
+                basis_messages = (basis_messages +
+                                  (1 - self.message_scale).unsqueeze(
+                                      -1) * self.message_replacement)
+            else:
+                basis_messages = (basis_messages +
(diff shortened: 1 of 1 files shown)


Padarn:
Should we add some tests? I'm curious why codecov is not showing coverage checks on this?

Also do you want to add an import to `torch_geometric/nn/models/__init__.py`?

torch_geometric/nn/models/GraphMask.py
```diff
@@ -0,0 +1,840 @@
+import math
+from inspect import signature
+from math import sqrt
+from typing import Optional
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+from originalMessagePassing import MessagePassing
```
Padarn, line 9: is this a mistake? I can't find this module

torch_geometric/nn/models/GraphMask.py
```diff
@@ -0,0 +1,840 @@
+import math
```
Padarn, line 1: I think the filename should be `graph_mask.py`

torch_geometric/nn/models/GraphMask.py
```diff
@@ -0,0 +1,840 @@
...
+            penalty = torch.zeros_like(input_element)
+
+        if summarize_penalty:
+            penalty = penalty.mean()
+
+        s = s * (zeta - gamma) + gamma
+
+        clipped_s = s.clamp(min_val, max_val)
+
+        if True:
```
Padarn, line 161: mistake?

torch_geometric/nn/models/GraphMask.py
```diff
@@ -0,0 +1,840 @@
...
+                        pbar.set_description(f'Explain node {node_idx}')
+                output = self.full_biases[i]
+                for j in range(len(gate_input)):
+                    partial = self.gates[i * 4][j](gate_input[j][i])
+                    result = self.gates[(i * 4) + 1][j](partial)
+                    output = output + result
+                relu_output = self.gates[(i * 4) + 2](output / len(gate_input))
+                sampling_weights = self.gates[(i * 4) +
+                                              3](relu_output).squeeze(dim=-1)
+                sampling_weights, penalty = self.__hard_concrete__(
```
Padarn, line 505: penalty is unused here btw (not sure if its intentional)

Padarn:
Reading this I was very confused at first, I got the context from the discussion linked though, thanks!

It makes me wonder however whether we should separate the functionality here, perhaps we could create a 

`ExplainableMessagePassing` class that looking something like 
```
class ExplainableMessagePassing(MessagePassing):
    def explain_message(self, out, x_i, x_j): 
        ...
   def from_message_passing(self, module: MessagePassing):
       ...
```
where `from_message_passing` would do all of he injection work which is currently inside the GraphMask model?

fork123aniket:
Closing this PR. Will upload a new PR containing correct file.

Issue #4013, opened 2022-02-07
AttributeError: 'BaseStorage' object has no attribute 'edge_index' for hetero graph


Yoloniverse:
### 🐛 Describe the bug

I tried to remove isolated nodes in my hetero graph. My hetero graph is as below:

data
Out[117]: 
HeteroData(
  card={ x=[15910, 300] },
  user={ x=[12363, 300] },
  (user, solved, card)={ edge_index=[88552, 2] }
)

When I use 'has_isolated_nodes()', it shows True

data.has_isolated_nodes()
Out[119]: True

Then I tried to remove the isolated nodes, but I got an error. 

`import torch`
`from torch_geometric.data import HeteroData`
`import torch_geometric.transforms as T`
`data = T.RemoveIsolatedNodes()(data)`

```
Traceback (most recent call last):

  File "/home/hojun/anaconda3/envs/geometric/lib/python3.6/site-packages/torch_geometric/data/storage.py", line 48, in __getattr__
    return self[key]

  File "/home/hojun/anaconda3/envs/geometric/lib/python3.6/site-packages/torch_geometric/data/storage.py", line 68, in __getitem__
    return self._mapping[key]

KeyError: 'edge_index'


During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "<ipython-input-120-2f26aa6d93fc>", line 1, in <module>
    data = T.RemoveIsolatedNodes()(data)

  File "/home/hojun/anaconda3/envs/geometric/lib/python3.6/site-packages/torch_geometric/transforms/remove_isolated_nodes.py", line 13, in __call__
    out = remove_isolated_nodes(data.edge_index, data.edge_attr, num_nodes)

  File "/home/hojun/anaconda3/envs/geometric/lib/python3.6/site-packages/torch_geometric/data/hetero_data.py", line 118, in __getattr__
    return getattr(self._global_store, key)

  File "/home/hojun/anaconda3/envs/geometric/lib/python3.6/site-packages/torch_geometric/data/storage.py", line 51, in __getattr__
    f"'{self.__class__.__name__}' object has no attribute '{key}'")

AttributeError: 'BaseStorage' object has no attribute 'edge_index'
```

As it is hetero graph, I do not have edge_index alone. My metadata() shows as below:

data.metadata()
Out[124]: (['card', 'user'], [('user', 'solved', 'card')])

How can I use RemoveIsolatedNodes in hetero graph?

### Environment

[...]

PR #4479, opened 2022-04-15
Add `HeteroData` support to `RemoveIsolatedNodes` 

Padarn:
This PR adds `HeteroData` support to the `RemoveIsolatedNodes` transform, closing https://github.com/pyg-team/pytorch_geometric/issues/4013.

This is achieved by looping through the different node types and removing them one at a time.

### DIFF
--- a/torch_geometric/transforms/remove_isolated_nodes.py
+++ b/torch_geometric/transforms/remove_isolated_nodes.py
@@ -1,7 +1,9 @@
 import re
+from typing import Union
 
 import torch
 
+from torch_geometric.data import Data, HeteroData
 from torch_geometric.data.datapipes import functional_transform
 from torch_geometric.transforms import BaseTransform
 from torch_geometric.utils import remove_isolated_nodes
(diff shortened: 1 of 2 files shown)


torch_geometric/transforms/remove_isolated_nodes.py
```diff
@@ -11,18 +13,91 @@
...
+
+            return data
+
+        elif isinstance(data, HeteroData):
+            return remove_hetero_isolated_nodes(data)
+
+        raise TypeError(f"`RemoveIsolatedNodes` invalid type: {type(data)}")
+
+
+def remove_hetero_isolated_nodes(data: HeteroData):
```
Padarn, line 42: This cannot go in 'torch_geometric/utils/isolated.py` because that class is imported by `torch_geometric/data/hetero_data.py`.

I thought about writing this taking a dictionary of edge_index, in the same way that `remove_isolated_nodes` works, I can adapt it to work like this if wanted, but I'd rather the `remove_isolated_nodes` were just depreciated in favour of one taking a `Data` object.

Any thoughts?

rusty1s (approved):

torch_geometric/transforms/remove_isolated_nodes.py
rusty1s: Yeah, handling this with `data` only sounds like a better approach :)

Issue #4445, opened 2022-04-09
from_networkx seems not compatible with set_node_attributes


minhtriet:
### 🐛 Describe the bug

I cannot create a graph from networkx if using `set_node_attributes` 

```python
import networkx as nx

G = nx.path_graph(3)
attrs = {0: {"attr1": torch.tensor([1,0,0], dtype=torch.int64)},
         1: {"attr1": torch.tensor([1,0,0], dtype=torch.int64)}, 
         2: {"attr1": torch.tensor([0,1,0], dtype=torch.int64)}}
nx.set_node_attributes(G, attrs)

pytorch_graph = torch_geometric.utils.from_networkx(G, group_node_attrs=['attr1'])
```
Stack trace
```
torch_geometric/utils/convert.py in from_networkx(G, group_node_attrs, group_edge_attrs)
    189         for key in group_node_attrs:
    190             x = data[key]
--> 191             x = x.view(-1, 1) if x.dim() <= 1 else x
    192             xs.append(x)
    193             del data[key]

AttributeError: 'list' object has no attribute 'dim'
```

It is not clear for me from the documentation what I have done wrong

### Environment

* PyG version: 2.0.4
* PyTorch version: 1.11.0
* OS: macOS
* Python version: 3.7.9
* CUDA/cuDNN version: na
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`):

Padarn:
I don't mind taking a look at this if you can assign it to me

rusty1s:
Thank you @Padarn

Padarn:
@minhtriet I fixed this in https://github.com/pyg-team/pytorch_geometric/pull/4486 - but in the meantime you can get around it by providing your input as a list, not a tensor.
```
import networkx as nx

G = nx.path_graph(3)
attrs = {0: {"attr1": [1,0,0]},
         1: {"attr1": [1,0,0]}, 
         2: {"attr1": [0,1,0]}}
nx.set_node_attributes(G, attrs)

pytorch_graph = torch_geometric.utils.from_networkx(G, group_node_attrs=['attr1'])
```

Although your tensor types may not be right.

ShantanuJhaveri:
I've found this same issue when using `nx.set_edge_attributes().` 

[...]

PR #4486, opened 2022-04-15
Fix `from_networkx` in case where attributes are tensors

Padarn:
Fixes problem reported in https://github.com/pyg-team/pytorch_geometric/issues/4445, the problem being that 

```torch.tensor([torch.Tensor([1,2,3], torch.Tensor([1,2,3]])```
 
does not work, they need to be stacked instead.


rusty1s (approved):
Thank you!

PR #4506, opened 2022-04-20
Fix `DimeNet` envelope bug with missing clamping

danielzuegner:
# What does this fix?

In the `Envelope` class used in the radial basis function in DimeNet, there's a bug. More precisely, the `forward()` is not clamping `d/c` to 1, as is done in the [original implementation](https://github.com/gasteigerjo/dimenet/blob/master/dimenet/model/layers/envelope.py#L23). This could lead to exploding output of the Envelope when `d/c` is around 2 (values go up to 300). This lead to numerical instability for me.

Let me know in case you have any questions.


rusty1s (approved):
Thanks for fixing :)

PR #4508, opened 2022-04-21
Test labels in `LinkNeighborLoader`

Padarn:
Fixing a small bug in the return type of the labels and and adding a test.

### DIFF
--- a/torch_geometric/loader/link_neighbor_loader.py
+++ b/torch_geometric/loader/link_neighbor_loader.py
@@ -215,8 +215,9 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             and labels :obj:`1` to :obj:`num_classes` represent the labels of
             positive edges. (default: :obj:`0.0`).
             Note that returned labels are :obj:`torch.float` when it is binary
-            classification (to facilitate ease-of-use with :meth:`torch.binary_cross_entropy_loss`)
-            and :obj:`torch.long` when it is a multi-class classification problem.
+            classification (to facilitate ease-of-use with
+            :meth:`torch.binary_cross_entropy_loss`) and :obj:`torch.long`
+            when it is a multi-class classification problem.
         **kwargs (optional): Additional arguments of
             :class:`torch.utils.data.DataLoader`, such as :obj:`batch_size`,
             :obj:`shuffle`, :obj:`drop_last` or :obj:`num_workers`.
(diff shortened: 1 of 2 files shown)


Padarn:
Faling test here looks irrelevant - flaky test?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -31,7 +31,8 @@ def _create_label(self, edge_label_index, edge_label):
             return edge_label_index, edge_label
 
         if edge_label is None:
-            edge_label = torch.ones(num_pos_edges, device=device)
+            edge_label = torch.ones(num_pos_edges,
```
rusty1s, line 34: Mh, I am actually in favor of keeping it that way since `BCELoss` expects floating-point inputs. Does that work for you?

Padarn: I see, in this case do you think the assert below on line 36 should be removed?

Padarn: But yes generally fine by me, just wanted to make sure the behaviour was consistent

rusty1s: I think the assert is a good check, since in case edge labels are given, we expect that this is a multi-class classification. If it is not, edge labels should not be passed to the loader in the first place.

Padarn: right, okay I'm sold on your idea - I'll revert this and update the docstring a bit

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -215,8 +215,8 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             and labels :obj:`1` to :obj:`num_classes` represent the labels of
             positive edges. (default: :obj:`0.0`).
             Note that returned labels are :obj:`torch.float` when it is binary
-            classification (to facilitate easy use with cross entropy loss)
-            and :obj:`torch.long` when is multiclass.
+            classification (to facilitate easy use with :obj:`BCE_loss`)
```
rusty1s, line 218:
```suggestion
            classification (to facilitate ease-of-use with :meth:`torch.binary_cross_entropy_loss`)
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -215,8 +215,8 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
...
+            and :obj:`torch.long` when is a multiclass classification problem.
```
rusty1s, line 219:
```suggestion
            and :obj:`torch.long` when it is a multi-class classification problem.
```

rusty1s (approved):

rusty1s (approved):

PR #4509, opened 2022-04-21
fix minor bug in `LinkNeighborLoader` when data.x does not exist

Padarn:
Fixing a small bug that exists when `data.x` does not exist


Padarn:
I know sall PR are easier to work with.. but this one might be a bit extreme :-)

rusty1s (approved):
No, this is just perfect :)

Issue #4503, opened 2022-04-20
[Roadmap] GraphGym via PyTorch Lightning  🚀


aniketmaurya:
### PyTorch Lightning Integration

[GraphGym](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/graphgym/train.py) training experience can be improved for scalability, mixed precision support, logging and checkpoints with [PyTorch Lightning](https://github.com/PyTorchLightning/pytorch-lightning) integration.

* [x] Write [`LightningModule`](https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_module.html)
* [ ] Refactor `load_ckpt` and `save_ckpt` with PL checkpoint save and load method
* [x] Update [train](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/graphgym/train.py#L56) method with PL `Trainer` and the `LightningModule` implementation
* [ ] Integrate `LightningDataset`, `LightningNodeData` and `LightningLinkData` modules
* [ ] ...

rusty1s:
Closing this issue in favor of https://github.com/pyg-team/pytorch_geometric/issues/5132. Thanks for your contributions @aniketmaurya. Hope you are also on board on future refactors of GraphGym :)

PR #4511, opened 2022-04-21
Implement Lightning module for GraphGym

aniketmaurya:
Partially address #4503 

This PR implements LightningModule for GraphGym as a first step in the roadmap.

---

cc: @rusty1s @JiaxuanYou

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@ install_requires = [
     'requests',
     'pyparsing',
     'scikit-learn',
-    'pytorch-lightning>=1.6',
 ]
 
 full_install_requires = [
@@ -29,6 +28,7 @@ full_install_requires = [
     'scikit-image',
     'pytorch-memlab',
     'torchmetrics>=0.7',
+    'pytorch-lightning>=1.6',
 ]
 
 test_requires = [
--- a/setup.py
+++ b/setup.py
@@ -13,9 +13,14 @@ install_requires = [
     'scikit-learn',
 ]
 
-full_install_requires = [
-    'h5py',
+graphgym_requires = [
     'yacs',
+    'hydra-core',
+    'pytorch-lightning',
+]
+
+full_install_requires = graphgym_requires + [
+    'h5py',
     'numba',
     'pandas',
     'captum',
@@ -23,11 +28,9 @@ full_install_requires = [
     'trimesh',
     'networkx',
     'tabulate',
-    'hydra-core',
     'matplotlib',
     'scikit-image',
     'pytorch-memlab',
-    'pytorch-lightning',
     'torchmetrics>=0.7',
 ]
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -11,6 +11,7 @@
     'requests',
     'pyparsing',
     'scikit-learn',
+    'pytorch-lightning>=1.6',
```
rusty1s, line 14: Let's move this to `full_install_requires`. In addition, is the `1.6` version requirement necessary? PyG+PL works fine for version 1.5.* as well.

rusty1s (approved):
Amazing. Left one blocking comment, otherwise LGTM.

torch_geometric/graphgym/model_builder.py
```diff
@@ -7,6 +8,32 @@
...
+    def __init__(self, dim_in, dim_out, cfg):
+        super().__init__()
+        self.model = network_dict[cfg.model.type](dim_in=dim_in,
+                                                  dim_out=dim_out)
+
+    def forward(self, *args, **kwargs):
+        return self.model(*args, **kwargs)
+
+    @property
+    def encoder(self):
```
rusty1s, line 21:
```suggestion
    def encoder(self) -> torch.nn.Module:
```
(+ for other functions).

setup.py
aniketmaurya: Sure, It should work well with 1.5* as well.

aniketmaurya: > Let's move this to full_install_requires

This will affect the `Minimal Installation` test since it checks for successful import of `torch_geometric.graphgym` module. Now that we are integrating PL for GraphGym shall we remove this from the minimal installation test?

rusty1s: Sounds good. Let's remove the import from the test until we make some further progress :)

aniketmaurya:
hi @rusty1s, thank you for reviewing the PR. I have updated it with your suggestions.

rusty1s:
Also added `graphgym_requires` to `setup.py` (which looked like a good solution to me).

torch_geometric/graphgym/model_builder.py
```diff
@@ -1,12 +1,47 @@
...
 import torch
 
 from torch_geometric.graphgym.config import cfg
 from torch_geometric.graphgym.models.gnn import GNN
 from torch_geometric.graphgym.register import network_dict, register_network
 
+try:
+    from pytorch_lightning import LightningModule
+except ImportError:
+    LightningModule = object
```
Borda, line 12: I think would be better to fall to `nn.Module`

rusty1s: Good point. Fixed in https://github.com/pyg-team/pytorch_geometric/pull/4516.

Issue #4447, opened 2022-04-10
Global Pooling via `EquilibriumAggregation`


rusty1s:
### 🚀 The feature, motivation and pitch

Support the recently proposed "Equilibrium Aggregation" scheme which defines an implicitly given global readout function. The [blog post](https://fabianfuchsml.github.io/equilibriumaggregation/) goes over it in detail.

* [Paper](https://arxiv.org/abs/2202.12795): The code is given in the appendix.

Padarn:
This paper is pretty cool. I'm happy to start on this.

I wonder if its worth also trying to implement it as a possible aggregation function in message passing, rather than just as a global readout. That way the global readout version could be implemented just by passing a different adjacency matrix to the layer?

rusty1s:
I think providing it as a global readout layer is sufficient for now. I don‘t think we need to add it as a general option for aggregation in any message passing layer for now.

Padarn:
Fair enough, worth doing separately

Padarn:
Have opened a PR, and assigned this to myself - still a number of open questions to discuss and a lot of cleaning to do.

PR #4522, opened 2022-04-24
`EquilibriumAggregation` global aggregation layer

Padarn:
Implements a simple version of the paper ["Equilibrium Aggregation: Encoding Sets via Optimization"](https://arxiv.org/abs/2202.12795). Note that exact details of the paper have not been implemented yet, and I plan to leave specific details about the optimizer out of this MR, it should only add a new readout layer than implements the method.

TODO:
- [x] Example of the 'exact' methods converging (median)
- [x] Cleanup of the readout layer
- [x] Support batched datasets (similar to other glob layers)
- [x] Tests

Addresses https://github.com/pyg-team/pytorch_geometric/issues/4447


Padarn:
This is still pretty heavy WIP (and I'm not even sure if it is working 100% yet). 

@rusty1s do you think there is a good example graph classification problem to try and compare accuracy to? The paper uses the `MOLPCBA` benchmark (would I could not see in the repo yet, though we could add it). I currently just modified the `MUTAG` example, there is no performance improvement (or degradation).

I plan also have the median example from the paper as an example (acts as a reasonable test it works).

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def reg(self, y):
+        return torch.nn.Softplus()(
+            self.lamb) * (y + EPS).square().mean().sqrt()
+
+    def combine_input(self, x, y):
+        return torch.cat([x, y.expand(x.size(0), -1)], dim=1)
+
+    def forward(self, x: torch.Tensor):
+        grad_enabled = torch.is_grad_enabled()
```
Padarn, line 60: I don't know if there is cleaner way to do this - I need to undo other 'torch.no_grad' otherwise the inner loop cannot run in test/validation stages that use it

examples/mutag_gin_equib.py
```diff
@@ -0,0 +1,100 @@
...
+
+train_loader = DataLoader(train_dataset, batch_size=1)
+test_loader = DataLoader(test_dataset, batch_size=1)
+
+
+class Net(torch.nn.Module):
+    def __init__(self, in_channels, dim, out_channels):
+        super().__init__()
+
+        self.conv1 = GINConv(
```
rusty1s, line 25: Replace with `torch.nn.models.GIN`.

examples/mutag_gin_equib.py
```diff
@@ -0,0 +1,100 @@
...
+        self.readout = EquilibriumAggregation(dim, dim, [256, 256],
+                                              grad_iter=10, alpha=0.05)
+
+    def forward(self, x, edge_index, batch):
+        x = self.conv1(x, edge_index)
+        x = self.conv2(x, edge_index)
+        x = self.conv3(x, edge_index)
+        x = self.conv4(x, edge_index)
+        x = self.conv5(x, edge_index)
+        x = self.readout(x).unsqueeze(0)
```
rusty1s, line 56:
```suggestion
        x = self.readout(x)
```
Why do we need the `unsqueeze` here?

torch_geometric/nn/glob/__init__.py
```diff
@@ -3,15 +3,12 @@
...
-    'global_add_pool',
-    'global_mean_pool',
-    'global_max_pool',
-    'global_sort_pool',
-    'GlobalAttention',
-    'Set2Set',
-    'GraphMultisetTransformer',
+    'global_add_pool', 'global_mean_pool', 'global_max_pool',
+    'global_sort_pool', 'GlobalAttention', 'Set2Set',
+    'GraphMultisetTransformer', 'EquilibriumAggregation'
```
rusty1s, line 11:
```suggestion
    'GraphMultisetTransformer', 'EquilibriumAggregation',
```

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+
+        return h
+
+
+class EquilibriumAggregation(torch.nn.Module):
+    """
+    Args:
+        potential (torch.nn.Module): trainable potenial function
+    """
+    def __init__(self, input_dim: int, output_dim: int, layers: List[int],
```
rusty1s, line 39:
```suggestion
    def __init__(self, in_channels: int, out_channels: int, num_layers: List[int],
```
Keep naming consistent :)

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+                 grad_iter: int = 5, alpha: float = 0.1):
+        super().__init__()
+
+        self.potential = ResNetPotential(input_dim + output_dim, layers)
+        self.lamb = torch.nn.Parameter(torch.Tensor([1]), requires_grad=True)
```
rusty1s, line 44: Looks like we need to fill this tensor in `reset_parameters`.

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+        self.grad_iter = grad_iter
+        self.alpha = alpha
+        self.output_dim = output_dim
+
+    def init_output(self):
+        return torch.zeros(self.output_dim, requires_grad=True)
+
+    def reg(self, y):
+        return torch.nn.Softplus()(
```
rusty1s, line 53: Init `Softplus` in the module constructor?

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+            self.lamb) * (y + EPS).square().mean().sqrt()
```
rusty1s, line 54: Why do we need the `EPS` here?

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def combine_input(self, x, y):
+        return torch.cat([x, y.expand(x.size(0), -1)], dim=1)
+
+    def forward(self, x: torch.Tensor):
```
rusty1s, line 59:
```suggestion
    def forward(self, x: torch.Tensor) -> torch.Tensor:
```

torch_geometric/nn/glob/equilibrium_aggregation.py
rusty1s: I think the preferred approach is:
```python
with torch.enable_grad():
    pass
```

examples/mutag_gin_equib.py
Padarn: Yeah sorry this is just lifted from the previous example for now

examples/mutag_gin_equib.py
Padarn: right now the output shape of self.readout(x) is `(input_size)` but needs to be `(batch_size, input_size)` - the disparity comes because the model here is using batch but I've currently ignored that by setting batch to 1 because I wanted to simplify it as a I figured it out

torch_geometric/nn/glob/equilibrium_aggregation.py
Padarn: because I didn't realize I should be using `.norm` 😓 https://github.com/pytorch/pytorch/issues/6394

Padarn:
Hey @rusty1s this is mostly ready for review (though the example I use is of the median aggregation).

I'm currently trying to play around to get this to converge well, but I haven't had much luck. I'm not sure if its because of parameters I've used or details missing in the paper (like the alpha used in the inner optimization). But perhaps convergence is just slow - they run for 10 million steps which I'm trying to do now but will take quite some time on my resources :-)

Padarn:
after ~1 million iterations no signs of convergence, guessing I've got something wrong

torch_geometric/nn/glob/equilibrium_aggregation.py
```diff
@@ -0,0 +1,183 @@
...
+        batch: Optional[torch.Tensor],
+        func: Callable[[torch.Tensor, torch.Tensor, Optional[torch.Tensor]],
+                       torch.Tensor],
+        iterations: int = 5,
+    ) -> Tuple[torch.Tensor, float]:
+
+        momentum = torch.zeros_like(y)
+        for _ in range(iterations):
+            val = func(x, y, batch)
+            grad = torch.autograd.grad(val, y, create_graph=True,
```
Padarn, line 96: The authors of the paper suggest adding an axillary loss on the `grad` here (`grad.square().sum()/iterations`) to be precise.

Padarn:
I've moved this to the aggregation module but there are a few things missing which I'll need opinions on. Leaving comments in the code.

torch_geometric/nn/aggr/equilibrium.py
```diff
@@ -0,0 +1,181 @@
...
+        return torch.zeros(index_size, self.output_dim,
+                           requires_grad=True).float()
+
+    def reg(self, y: Tensor) -> float:
+        return self.lamb * y.square().mean(dim=-2).sum(dim=0)
+
+    def energy(self, x: Tensor, y: Tensor, index: Optional[Tensor]):
+        return self.potential(x, y, index) + self.reg(y)
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
```
Padarn, line 164: Just a note this was all written before the `Aggregation` module, so its not using `reduce` yet - need to rethink some of the implementation to make use of it.

torch_geometric/nn/aggr/equilibrium.py
```diff
@@ -0,0 +1,181 @@
...
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+
+        if ptr is not None:
+            raise ValueError(f"{self.__class__} doesn't support `ptr`")
+
+        if dim_size is not None:
```
Padarn, line 171: I could technically support this, but I don't know what you'd expect the behavior to be: Unlike sum/mean etc, we can't just assume the input is zero (or if we do, the output would just be random).

Thoughts?

torch_geometric/nn/aggr/equilibrium.py
rusty1s: No need to make use of `reduce`. The `LSTMAggregation` module doesn't make use of it as well.

torch_geometric/nn/aggr/equilibrium.py
rusty1s: Please have a look at https://pytorch-scatter.readthedocs.io/en/latest/functions/scatter.html. The `dim_size` argument ideally is identical to `index_size = index.max() + 1` in your case. If passed, there is no need to compute it in the first place.

Padarn: I didn't fully get that sorry. If `dim_size` is passed 
- if `dim_size < index_size` I'd expect error,
- if  `dim_size > index_size` I'm not sure what to expect - I guess zero in for those entries?

rusty1s: Yes, that is correct.

examples/equilibrium_median.py
```diff
@@ -0,0 +1,37 @@
...
+gamma = torch.distributions.gamma.Gamma(0.2, 0.5)
+uniform = torch.distributions.uniform.Uniform(0, 1)
+total_loss = 0
+n_loss = 0
+for i in range(steps):
+    optimizer.zero_grad()
+    dist = np.random.choice([norm, gamma, uniform])
+    x = dist.sample((input_size, 1))
+    y = model(x)
+    loss = (y - x.median()).norm(2)
```
lightaime, line 31: It seems the loss is not normalized by the input size.
```suggestion
    loss = (y - x.median()).norm(2) / input_size
```

torch_geometric/nn/aggr/equilibrium.py
```diff
@@ -0,0 +1,181 @@
...
+    def forward(
+        self,
+        x: Tensor,
+        y: Tensor,
+        index: Optional[Tensor],
+        func: Callable[[Tensor, Tensor, Optional[Tensor]], Tensor],
+        iterations: int = 5,
+    ) -> Tuple[Tensor, float]:
+
+        momentum = torch.zeros_like(y)
```
lightaime, line 93:
```suggestion
        momentum_buffer = torch.zeros_like(y)
```

torch_geometric/nn/aggr/equilibrium.py
```diff
@@ -0,0 +1,181 @@
...
+        for _ in range(iterations):
+            val = func(x, y, index)
+            grad = torch.autograd.grad(val, y, create_graph=True,
+                                       retain_graph=True)[0]
+            momentum = self.momentum * momentum - self.learning_rate * grad
```
lightaime, line 98:
```suggestion
            momentum_buffer = self.momentum * momentum_buffer - self.learning_rate * grad
```

lightaime (commented):
Great work! Added some minor comments.

Padarn:
thanks @lightaime - I've got a busy week, I'll address your comments ASAP :-)

Padarn:
Thanks for the reviews - I've updated based on most comments. I still have an uncertainty here: https://github.com/pyg-team/pytorch_geometric/pull/4522#discussion_r872945963 but it maybe for another PR

rusty1s:
Thanks @Padarn. Can you also resolve the merge conflicts? I will take a look soon, and address the unresolved comment as well.

Padarn:
Hey @rusty1s, any thoughts on this one? Happy to keep working on it or break it up, but might be good to not leave it here unfinished.

rusty1s:
Yes, no that #4779 is merged, let's try to integrate this next :)

Padarn:
Cool. I'll make another pass to see if there is anything I can clean up based on https://github.com/pyg-team/pytorch_geometric/pull/4779

lightaime (approved):
Thanks, LGTM!

Padarn:
Thanks @lightaime - I'm going to merge this one, but will try and build some more example use cases before we release the new version so make sure its good.

PR #4531, opened 2022-04-25
Implement `LightningModule` & `LoggerCallback` in GraphGym

aniketmaurya:
Partially address 2nd task from the roadmap #4503.

Lightning Module implementation for model training/val/test step and implement Logger as PL Callback.


test/graphgym/test_graphgym.py
```diff
@@ -110,3 +110,77 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
     shutil.rmtree(cfg.out_dir)
+
+
+@withPackage('yacs')
+@withPackage('pytorch_lightning')
+@pytest.mark.parametrize('auto_resume', [True, False])
+@pytest.mark.parametrize('skip_train_eval', [True, False])
+@pytest.mark.parametrize('use_trivial_metric', [True, False])
+def test_graphgym_module(auto_resume, skip_train_eval, use_trivial_metric):
+    import pytorch_lightning as pl
```
Borda, line 121: consider moving this out with conditional import as this may do import with each test call (and this has quite large parametrizations)

test/graphgym/test_graphgym.py
```diff
@@ -110,3 +110,77 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
+
+    Args = namedtuple('Args', ['cfg_file', 'opts'])
+    root = osp.join(osp.dirname(osp.realpath(__file__)))
```
Borda, line 124: lets move this also to the top file

test/graphgym/test_graphgym.py
```diff
@@ -110,3 +110,77 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
+    args = Args(osp.join(root, 'example_node.yml'), [])
+
+    load_cfg(cfg, args)
+    cfg.out_dir = osp.join('/', 'tmp', str(random.randrange(sys.maxsize)))
```
Borda, line 128: consider using `tmpdir` fixture https://docs.pytest.org/en/6.2.x/tmpdir.html

torch_geometric/graphgym/logger.py
```diff
@@ -10,6 +12,15 @@
 from torch_geometric.graphgym.utils.device import get_current_gpu_usage
 from torch_geometric.graphgym.utils.io import dict_to_json, dict_to_tb
 
+try:
+    import pytorch_lightning as pl
+    from pytorch_lightning import Callback
+
+except ImportError:
+    pl = object
```
Borda, line 20:
```suggestion
    pl = None
```

Borda (approved):

test/graphgym/test_graphgym.py
```diff
@@ -110,3 +118,54 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
+
+    seed_everything(cfg.seed)
+    auto_select_device()
+    set_run_dir(cfg.out_dir)
+
+    loaders = create_loader()
+    assert len(loaders) == 3
+
+    model = create_model()
+    assert isinstance(model, pl.LightningModule)
```
Padarn, line 143: Sorry won't this cause an exception if `pl` is None?

torch_geometric/graphgym/logger.py
```diff
@@ -10,6 +12,15 @@
...
 from torch_geometric.graphgym.utils.io import dict_to_json, dict_to_tb
 
+try:
+    import pytorch_lightning as pl
+    from pytorch_lightning import Callback
+
+except ImportError:
+    pl = None
+    Callback = object
+    logging.warning("pytorch_lightning is not installed")
```
Padarn, line 22: Maybe worth noting in the warning what the implication of this is?

torch_geometric/graphgym/model_builder.py
```diff
@@ -19,12 +23,35 @@
 class GraphGymModule(LightningModule):
     def __init__(self, dim_in, dim_out, cfg):
```
Padarn, line 24: maybe we add some typing here? I wouldn't know what `cfg` is

torch_geometric/graphgym/logger.py
```diff
@@ -247,3 +258,105 @@ def create_logger():
...
+                          pl_module: "pl.LightningModule") -> None:
+        self.test_logger.write_epoch(trainer.current_epoch)
+
+    def on_epoch_end(self, trainer: "pl.Trainer",
+                     pl_module: "pl.LightningModule") -> None:
+        for logger in self._logger:
+            logger.close()
+
+    @property
+    def train_logger(self):
```
Padarn, line 353: Just wondering if "create_loggers" should return a dict, this seems a tiny bit error prone.

torch_geometric/graphgym/model_builder.py
aniketmaurya: thank you for reviewing the PR! I have updated this :)

torch_geometric/graphgym/logger.py
aniketmaurya: Agree! Since this PR doesn't update the training epoch implementation so I would keep this change for the next PR with the `Trainer` integration. Thoughts @rusty1s @Borda?

test/graphgym/test_graphgym.py
```diff
@@ -110,3 +118,54 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
+    assert isinstance(scheduler[0], torch.optim.lr_scheduler.CosineAnnealingLR)
+
+    cfg.params = params_count(model)
+    assert cfg.params == 23880
+
+    keys = {"loss", "true", "pred", "pred_score", "step_end_time"}
+    # test training step
+    batch = next(iter(loaders[0]))
+    outputs = model.training_step(batch)
+    assert keys.issubset(outputs.keys())
```
rusty1s, line 156: Can we test for equality instead?

test/graphgym/test_graphgym.py
```diff
@@ -110,3 +118,54 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
...
+    assert torch.is_tensor(outputs["loss"])
```
rusty1s, line 157:
```suggestion
    assert isinstance(outputs["loss"], torch.Tensor)
```

torch_geometric/graphgym/logger.py
```diff
@@ -10,6 +13,16 @@
 from torch_geometric.graphgym.utils.device import get_current_gpu_usage
 from torch_geometric.graphgym.utils.io import dict_to_json, dict_to_tb
 
+try:
```
rusty1s, line 16: I am wondering if there is some better way to do this besides having this `try`/`catch`-block in every file. Maybe we do this in a separate file once (e.g., `import.py`) and import from there?

torch_geometric/graphgym/logger.py
```diff
@@ -247,3 +260,105 @@ def create_logger():
...
+        self._test_epoch_start_time = None
+
+    def _create_stats(self, time_start, outputs: Dict[str, Any],
+                      trainer: "pl.Trainer") -> Dict:
+        true: torch.Tensor = outputs["true"]
+        pred: torch.Tensor = outputs["pred"]
+        pred_score: torch.Tensor = outputs["pred_score"]
+        loss: torch.Tensor = outputs["loss"]
+        lr = trainer.lr_scheduler_configs[0].scheduler.get_last_lr()[0]
+        stats = dict(true=true.detach().cpu(), pred=pred.detach().cpu(),
```
rusty1s, line 279: Any chance we can test this logger in this PR already?

torch_geometric/graphgym/model_builder.py
```diff
@@ -17,14 +25,37 @@
...
         return self.model(*args, **kwargs)
 
+    def configure_optimizers(self) -> Tuple[Any, Any]:
+        optimizer = create_optimizer(self.model.parameters(), self.cfg.optim)
+        scheduler = create_scheduler(optimizer, self.cfg.optim)
+        return [optimizer], [scheduler]
+
+    def _shared_step(self, batch, split: str) -> Dict:
+        batch.split = split
+        pred, true = self.forward(batch)
```
rusty1s, line 44:
```suggestion
        pred, true = self(batch)
```

torch_geometric/graphgym/model_builder.py
```diff
@@ -17,14 +25,37 @@
...
     def forward(self, *args, **kwargs):
         return self.model(*args, **kwargs)
 
+    def configure_optimizers(self) -> Tuple[Any, Any]:
+        optimizer = create_optimizer(self.model.parameters(), self.cfg.optim)
+        scheduler = create_scheduler(optimizer, self.cfg.optim)
+        return [optimizer], [scheduler]
+
+    def _shared_step(self, batch, split: str) -> Dict:
+        batch.split = split
```
rusty1s, line 43: Any reason we do this?

torch_geometric/graphgym/model_builder.py
```diff
@@ -17,14 +25,37 @@
...
+        pred, true = self.forward(batch)
+        loss, pred_score = compute_loss(pred, true)
+        step_end_time = time.time()
+        return dict(loss=loss, true=true, pred=pred, pred_score=pred_score,
```
rusty1s, line 47: Not sure I understand the output here. What does `true=true` mean? And why do we only set `stop_end_time`?

aniketmaurya: this is for logging `time_used` in the `logger.update(...)` method. And with the `LoggerCallback` I have created the `*_epoch_start_time` at the start of each [epoch like here](https://github.com/aniketmaurya/pytorch_geometric/blob/implement_module/torch_geometric/graphgym/logger.py#L288)

torch_geometric/graphgym/logger.py
aniketmaurya: yeah, make sense

torch_geometric/graphgym/model_builder.py
aniketmaurya: I referred to the [current training implementation](https://github.com/aniketmaurya/pytorch_geometric/blob/implement_module/torch_geometric/graphgym/train.py#L45) for this.

torch_geometric/graphgym/logger.py
aniketmaurya: The Logger is a PL callback, I can mock training and test it out in this PR.

test/graphgym/test_graphgym.py
```diff
@@ -31,8 +31,16 @@
 )
 from torch_geometric.testing import withPackage
 
+try:
```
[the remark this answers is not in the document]
aniketmaurya: actually, the minimal installation test fail once I remove the try block. Maybe would have to remove the `graphgym` module from minimal tests for this?

torch_geometric/graphgym/model_builder.py
aniketmaurya: regarding `true=true`, I have saved the values to access it later via the Logger callback.

```python
print(dict(true=true))
# {"true": [SOME TENSOR VALUE]}
```

torch_geometric/graphgym/logger.py
aniketmaurya: I have added a training test with logger as callback [here](https://github.com/aniketmaurya/pytorch_geometric/blob/implement_module/test/graphgym/test_graphgym.py#L195)

This would only test if the hooks get any error or not. Anything specific you would like to check here @rusty1s ?

Suggestions @Borda?

aniketmaurya: I have added a test for logger callback [here](https://github.com/aniketmaurya/pytorch_geometric/blob/implement_module/test/graphgym/test_graphgym.py#L195)

This would only test if the callback hooks get any error or not. Anything specific you would like to check here @rusty1s ?

Suggestions @Borda?

test/graphgym/test_graphgym.py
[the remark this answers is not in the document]
rusty1s: Not sure I understand. Why can't we do `import pytorch_lightning` inside the `test_*` function?

torch_geometric/graphgym/model_builder.py
rusty1s: Ah, thanks for clarifying.

rusty1s (approved):
Thank you!

test/graphgym/test_graphgym.py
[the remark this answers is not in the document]
aniketmaurya: I think it is because pytest goes to every test module and since PL import is defined globally, it will import PL without executing `@withPackage('pytorch_lightning')`. This results in import error and it fails the minimal installation test.

rusty1s:
Thank you! Looking forward to the follow-up contributions :)

PR #4534, opened 2022-04-25
`FeatureStore` abstraction definition

mananshah99:
Defines a `FeatureStore` abstraction and associated tests as a first step to allow for independent scale-out for a graph's features and its `edge_index`. A roadmap for remaining features and implementation details will follow shortly. More details:
* Defines `FeatureStore(MutableMapping)` with basic CRUD operations as abstract methods as well as support for advanced indexing
* Defines `TensorAttr` and `AttrView` as classes that aid in accessing, modifying, and viewing elements in a feature store
* Defines `CastMixin` for straightforward `TensorAttr` creation
* Defines a basic (incomplete) feature store implementation `MyFeatureStore` along with associated tests to showcase API and indexing functionality


torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,220 @@
...
+    attr: TensorAttr
+
+    def __init__(self, store, attr):
+        self._store = store
+        self.attr = attr
+
+    def __getattr__(self, tensor_type):
+        r"""Supports attr_view.attr"""
+        self.attr.tensor_type = tensor_type
+        return self
```
yaoyaowd, line 73: not 100% sure where did  we use this.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,220 @@
...
+        self.backend = backend
+
+    # Core (CRUD) #############################################################
+
+    @abstractmethod
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
+        r"""Implemented by :obj:`FeatureStore` subclasses."""
+        pass
+
+    def put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
```
yaoyaowd, line 96: I thought that we agreed on we can put everything into features store, not only tensors.

torch_geometric/utils/mixin.py
```diff
@@ -0,0 +1,14 @@
+class CastMixin:
```
yaoyaowd, line 1: why do we need this?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,220 @@
...
+                implementations of a FeatureStore to store this metadata in a
+                meaningful way that allows for tensor retrieval from a
+                :obj:`TensorAttr` object.
+        Returns:
+            FeatureTensorType, optional: a tensor of the same type as the
+            index, or None if no tensor was found.
+        """
+        def maybe_to_torch(x):
+            return torch.from_numpy(x) if isinstance(
+                attr.index, torch.Tensor) and isinstance(x, np.ndarray) else x
```
yaoyaowd, line 139: why `isinstance(attr.index, torch.Tensor)`

torch_geometric/data/feature_store.py
mananshah99: This is to support advanced indexing, see [these lines](https://github.com/pyg-team/pytorch_geometric/pull/4534/files#diff-f591068f1a527a7fde0baa34426082c764ddba485da2aa2346f0cf3ca29981dcR84-R86).

torch_geometric/utils/mixin.py
mananshah99: This is to support the creation of a `TensorAttr` with a tuple or list (as opposed to `TensorAttr(...)`), the use of which can be seen in `test_feature_store.py`.

torch_geometric/data/feature_store.py
mananshah99: Sorry for the confusion, my understanding was that it is certainly reasonable to expect feature stores to store more general objects (not just tensors) in the future, but for the purpose of starting with a well-defined, clear interface and proceeding from there, the first iteration of the feature store abstraction would explicitly store tensors. Happy to move to `Any` now if you'd prefer it, but I think for the first few iterations this constraint would be helpful to better guide implementations.

yaoyaowd: I see. The current one sounds good to me.

torch_geometric/utils/mixin.py
yaoyaowd: I feel it is a little confusion since in the function we have the type hint indicate what users pass in is a `TensorAttr` but in fact it is `Any` so we need to do the cast. I'd imaging some C++ style like directly cast `{...}` into a class and pass into the function.
I tried to google it and see if there's any best practice but I couldn't find any.

torch_geometric/data/feature_store.py
yaoyaowd: I feel advance indexing is cool and let `FeatureStore` inherit from `MutableMapping` is also cool but learning and understand curve is higher.
The tricky part is this indexing only support `tensor_type` but what if we want to do things like `.get(node_type=..., tensor_type=...)` or `.get(node_type=...)`. I think it is a little hard to follow the code if we just read the api.
Similar thing apply to `FeatureStore` where `store[...]` can return either `tensors` or `AttrView`. It will make it hard to read code and also do type check.

yaoyaowd (commented):
Put two thoughts in the code. I think what we wrote is cool and pythonic, but not easy to follow if I read this code for the first time.
I don't have a strong opinion to change the style now but want to hear people thoughts.

torch_geometric/data/feature_store.py
mananshah99: This is to support the documented return value, which is a tensor of the same type as index or None if no tensor was found; the code has been fixed.

torch_geometric/data/feature_store.py
mananshah99: Thanks for the comments! Let me clarify a couple points here. 
* The current design of the feature store assumes that all get calls require a `TensorAttr.index`.
* A `get_tensor` call will either return a Tensor (if found, based on `TensorAttr`), `None` if the tensor is not found (because the attributes were not fully specified, or because the attributes do not match a stored value), or raise an Exception if `index` is `None`. I think this behavior makes sense and aligns with what a user would expect from such a function.
* An indexing call will return a Tensor if the attributes are fully specified (`index` is not `None`). It will return an `AttrView` if the attributes are not fully specified (that is, if `index` is `None`), which allows for further specification of the tensor type and the index via `__getattr__` and `__getitem__`. If the tensor type is specified in `TensorAttr` at the initial indexing call, one can simply do `store[{'node_type': node_type, 'tensor_type': tensor_type}][index]` and obtain the desired values. Imo, it seems a bit confusing to support `store[{'tensor_type': tensor_type}].node_type[index]` as this pattern seems out-of-order (first specifying the tensor type, then a node type): a user must first specify general attributes, then a node type, then an index, which is what we currently support. However, if this is desired, it would be quite easy to implement it.

Wdyt?

torch_geometric/utils/mixin.py
mananshah99: Yes, you're absolutely right that the type hints for the functions are technically incorrect as we are able to accept `Any` and cast to the desired class. I'm not sure there is a best practice for this that maintains the same level of code cleanliness.

In particular, I think that the improvements to ease-of-use as a result of `CastMixin` (e.g. indexing via `store[{'tensor_type': x, 'node_type': y}][index]` instead of `store[TensorAttr(tensor_type=x, node_type=y)][index]`) are quite useful. Imo, the current type hints (instead of `Any` everywhere) are provided for improved code readability and clarity, and since they are not enforced it is reasonable to type hint `TensorAttr` and also support non-`TensorAttr` inputs that we subsequently cast. We can add better documentation for `CastMixin` to make this more clear as well. Wdyt?

yaoyaowd: I'm fine with your current design, I think somewhere in the code when we call cast we should keep a comment explain why.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+
+    As a result, store[TensorAttr(...)].tensor_type[idx] allows for indexing
+    into the store.
+    """
+    _store: 'FeatureStore'
+    attr: TensorAttr
+
+    def __init__(self, store, attr):
+        self._store = store
+        self.attr = attr
```
yaoyaowd, line 68: also add cast support?

torch_geometric/data/feature_store.py
yaoyaowd: I think a `AttrView` is a way of select features from store.
```
a = AttrView(store, {'node_type': x, 'graph_type': y})
b = a.feat
c = a.pkey
# b and c should be different views.
```
In another word, I would prefer not expose getattr but explicitly specify views like this
```
a = AttrView(store, {'node_type': x, 'graph_type': y})
b = AttrView(store, {'node_type': x, 'graph_type': y, 'tensor_type': 'feat'})
c = AttrView(store, {'node_type': x, 'graph_type': y, 'tensor_type': 'pkey'})
```
More over, for the same node_type and graph_type, how about support tensor_type as a list? So that if we decide to implement feature store on a different host, when we send a view from a source host to a destination host, we only pass index once for multiple tensor types.

For feature store, I'd prefer have explicit functions for views and get_tensors, i.e keep get_tensor as it is but also define a function
```
def view(ta: TensorAttr) -> AttrView:
```
It has two benefit, each function will have explicit return type. And in addition to it, the function name also reflect the function workload. Create an `AttrView` by `view` function should be lighter than create a `Tensor` by `get_tensor` function. We don't mix them together by `__getitem__` function.

test/data/test_feature_store.py
```diff
@@ -0,0 +1,86 @@
...
+
+    store = MyFeatureStore()
+    tensor = torch.Tensor([[0, 0, 0], [1, 1, 1], [2, 2, 2]])
+    index = torch.Tensor([0, 1, 2])
+
+    tensor_type = 'feat'
+    node_type = 'A'
+    graph_type = 'graph'
+
+    attr = TensorAttr(index, tensor_type, node_type, graph_type)
```
rusty1s, line 57: Any reason we do not add `tensor` to the `TensorAttr` here?

test/data/test_feature_store.py
```diff
@@ -0,0 +1,86 @@
...
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
+        self.store[MyFeatureStore.key(attr)] = torch.cat(
+            (attr.index.reshape(-1, 1), tensor), dim=1)
+        return True
+
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        tensor = self.store.get(MyFeatureStore.key(attr), None)
+        if tensor is None:
+            return None
+        if attr.index is not None:
```
rusty1s, line 30: What does that do? What are we testing with this implementation?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+
+import numpy as np
+import torch
+
+from torch_geometric.typing import FeatureTensorType
+from torch_geometric.utils.mixin import CastMixin
+
+
+@dataclass
+class TensorAttr(CastMixin):
```
rusty1s, line 35: I think we need to swap the order here: `node_type, tensor_type, index`. This way, we can access attributes via
```python
store[node_type, tensor_type, index]
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+    # The node indices the rows of the tensor correspond to
+    index: Optional[FeatureTensorType] = None
+
+    # The type of the feature tensor (may be used if there are multiple
+    # different feature tensors for the same node index)
+    tensor_type: Optional[str] = None
+
+    # The type of the nodes that the tensor corresponds to (may be used for
+    # hetereogeneous graphs)
+    node_type: Optional[str] = None
```
rusty1s, line 47:
```suggestion
    type: Optional[str] = None
```
or similar? Not sure if we want to be node-specific here. This abstraction should be useful for storing edge features as well.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # The type of the graph that the nodes correspond to (may be used if a
+    # feature store supports multiple graphs)
+    graph_type: Optional[str] = None
```
rusty1s, line 51: I think this is safe to drop (at least for now). Why does a `FeatureStore` need to know to which graphs it belongs to? Isn't the reference from a graph to a `FeatureStore` enough?

torch_geometric/typing.py
```diff
@@ -1,5 +1,7 @@
 from typing import Dict, List, Optional, Tuple, Union
 
+import numpy
```
rusty1s, line 3:
```suggestion
import numpy as np
```

torch_geometric/typing.py
```diff
@@ -20,6 +22,9 @@
 
 Metadata = Tuple[List[NodeType], List[EdgeType]]
 
+# A representation of a feature tensor
+FeatureTensorType = Union[torch.TensorType, numpy.ndarray]
```
rusty1s, line 26:
```suggestion
TensorType = Union[torch.Tensor, numpy.ndarray]
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+        r"""Supports attr_view.attr[idx]"""
+        self.attr.index = index
+        return self._store.get_tensor(self.attr)
+
+    def __repr__(self) -> str:
+        return f'AttrView(store={self._store}, attr={self.attr})'
+
+
+class FeatureStore(MutableMapping):
+    def __init__(self, backend: Any):
```
rusty1s, line 85: `backend` doesn't seem to be used anywhere. Why not drop`__init__` completely here?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+                to the feature tensor. See the :obj:`TensorAttr` documentation
+                for required and optional attributes. It is the job of
+                implementations of a FeatureStore to store this metadata in a
+                meaningful way that allows for tensor retrieval from a
+                :obj:`TensorAttr` object.
+        Returns:
+            bool: whether insertion was successful.
+        """
+        attr = TensorAttr.cast(attr)
+        assert attr.index is not None
```
rusty1s, line 112: Why is `index` required?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,227 @@
...
+                    tensor, np.ndarray) else tensor
+            if isinstance(attr.index, np.ndarray):
+                return tensor.numpy() if isinstance(tensor,
+                                                    torch.Tensor) else tensor
+            raise ValueError
+
+        attr = TensorAttr.cast(attr)
+        assert attr.index is not None
+
+        return to_type(self._get_tensor(attr))
```
rusty1s, line 151: It looks like we have to handle the `index` and `tensor` dtype conversion in `_get_tensor` as well. Is this intended?

rusty1s (commented):
Thanks for the PR. The nesting of `TensorAttr` looks pretty cool. Please see my comments below (especially about some confusions I have regarding `index`). I am not really sure why we store `index` in the first place. It looks to be because we require `TensorAttr` to be fully specified. On the other hand, it doesn't make much sense to require it in case the user wants to access the full matrix. Originally, I would have thought that it makes sense to include the `tensor` inside `TensorAttr`, and move any `index` logic to `TensorAttr` instead:
```python
class FeatureStore:
    def put_tensor(tensor, name, group):
        self._put_tensor(TensorAttr(tensor, name, group))
```
where `TensorAttr` manages any `index` logic:
```
class TensorAttr:
    def __getitem__(self, idx: IndexType):
        return self.tensor[idx]
    
    def __call__(self):
        return self.tensor
```
This would then result in an API similar to
```python
x_full = store['x', 'paper']()
x_selected = store['x', 'paper'][index]
```
Also agree with @yaoyaowd that it is pretty hard to understand on first read (although `BaseStorage` has the same problem).

rusty1s:
There exists two further alternatives as far as I can tell:
* Simply return the full matrix in case `index` is None. This way, `store[attr_name, index]` is required rather than doing `store[attr_name][index]`.
* Allow both `:` and tensors to model indices, e.g., `store[attr_name, :]`, and `store[attr_name][:]`, which aligns well with any PyTorch/numpy logic.

rusty1s:
Two additional thoughts:
* I think the `FeatureStore` needs to provide the capability to determine which attributes in `TensorAttr` are required. For example, for in-memory stores, we probably want to make `index` optional. For `torch_geometric.data.Data`, we also may want to make `node_type` optional/disallow its usage.
* `FeatureStore` may need to provide the capability to determine the order of attributes. For homogeneous data, we probably want to allow access via `store[attr_name]`, while for heterogeneous data, access via `store[group_name, attr_name]` feels more natural to me (this is likely a P1).

yaoyaowd:
@rusty1s is the `group_name` and `attr_name` in your comment match `node_type`, `tensor_type` in the code?
If yes, agree `group_name` is better and it cover `edge_type` too.

I also feel directly use `store[node_type, tensor_type, index], store[node_type, tensor_type, :], store[node_type, tensor_type]` to refer tensors is cleaner. In this way maybe we don't even need `TensorAttr`.

And if necessary, we define `store.view(...)` to get view instead of tensor.

I think it is a better idea to simplify interface now and iterate fast to get back additional features we need.

mananshah99:
@yaoyaowd @rusty1s thank you both for the great discussion. Let me summarize some thoughts and the resulting design changes.
* **Requiring an index.** It seems clear that we do not want `index` to be required. The initial motivation here was to define a `TensorAttr` object as fully specified when `index` is present, but I understand the reasons to avoid this. While large distributed feature stores may choose to raise an exception if a tensor is queried without an index, as @rusty1s pointed out this should be the job of implementor classes, which should be able to set required attributes.
* **`TensorAttr` and `AttrView`.** There seems to be some debate about `TensorAttr` and `AttrView`. My opinion is the following:
  * `TensorAttr` should define all parameters necessary to uniquely identify a tensor from the feature store. In particular, this includes parameters such as the index, tensor type, node type, etc. (of course, the required nature of these parameters is up to the feature store). It can be thought of as a protobuf or other serializable data format that can be sent to a store over-the-wire to perform put or get calls.
  * `AttrView` defines a _view_ on the feature store (emphasizing view due to its additional `_store` parameter that references the feature store it points to) that allows for indexing. For this reason, I think it is more natural to implement indexing logic in `AttrView` and not `TensorAttr`. As @yaoyaowd pointed out, we can allow users to access an `AttrView` with a `store.view(...)` method that explicitly returns an `AttrView`.
* **Handling advanced indexing.** There are a few thoughts on how we can implement advanced indexing in a cleaner way. I agree with the sentiment here, as the current set-up can be a bit hard to read and understand from a user perspective. My opinion is the following:
  * **Fully-specified forms.** We should be able to support the following fully-specified forms: `store[node_type, tensor_type, index]`, `store[node_type, tensor_type, index_slice]`, and `store[node_type].tensor_type[index]`. Index slices will automatically be converted to the internal index representation we use. If a form is fully-specified, a Tensor/None will be returned or an Exception will be thrown.
  * **Partially-specified forms.** In order to support the last format, which begins with a partially-specified form (`store[node_type]`) and adds further specification via `__getattr__`, we will need to have `store[node_type]` return an `AttrView`. More generally, we should have non-fully specified forms always return an `AttrView`. To explicitly perform a GET call with a non-fully specified form, a user will have to make use of `__call__`.
  * In summary, we will have the following API, where all indexing and `__call__` will be handled by `AttrView`.
  ```python
  # Fully-specified forms, produce a Tensor output
  store[node_type, tensor_type, index] 
  store[node_type, tensor_type, None]
  store[node_type, tensor_type, :]
  store[node_type].tensor_type[:]

  # Partially-specified forms, produce an AttrView object
  store[node_type]
  store[node_type].tensor_type

  # Partially-specified forms, when called, produce a Tensor output
  # from the `TensorAttr` that has been partially specified.
  store[node_type]()
  store[node_type].tensor_type()
  ```
* **Naming. **There were a couple minor points made about attribute naming. I think `group_name` makes sense, but I think `tensor_type` is more specific where `attr_name` feels a bit too general and confusing. @rusty1s if you have any strong opinions here, happy to adjust. 

I think this proposal will address many of the main design decisions we discussed above and result in a streamlined, Pythonic API. Please let me know if you have any comments or thoughts!

rusty1s:
I like it very much. Whenever something is not fully-specified, we return a `View`, and what qualifies as fully-specified depends on the `FeatureStore` implementation (where both `group` and `index` may be optional). Let me know when this PR is ready for another round of reviews :)

torch_geometric/data/feature_store.py
mananshah99: Closing in favor of the revised design.

test/data/test_feature_store.py
mananshah99: `TensorAttr` can be thought of as a protobuf that stores all tensor-related metadata that a feature store can use to store (and subsequently, retrieve). Users can index into the feature store with `TensorAttr`. I think this may be more clear in the revised design as well :)

torch_geometric/data/feature_store.py
mananshah99: I don't think that a reference from a graph to a FeatureStore is enough, say if we have a generic KV featurestore that supports multiple graphs. But we can drop this for now to avoid unnecessary complexity.

test/data/test_feature_store.py
mananshah99: Made this cleaner :)

torch_geometric/typing.py
mananshah99: I changed this from `TensorType` to `FeatureTensorType` to contrast with the future `EdgeTensorType`. Do you prefer a different naming?

torch_geometric/data/feature_store.py
mananshah99: I don't think so, why would we need to do this conversion in `_get_tensor`?

torch_geometric/typing.py
mananshah99: Why? We only use it once (`numpy.ndarray`)

mananshah99:
@rusty1s @yaoyaowd changes have been made to align with the discussion above. I've tried to add clear documentation for all of the builtins to improve code readability, and have also added tests that showcase functionality. @rusty1s, both points you listed above regarding the `FeatureStore`'s interactions with `TensorAttr` (determining required nature of attributes and determining the order of attributes) have been implemented and are tested as well.

Let me know what you think!

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
+r"""
+This class defines the abstraction for a Graph feature store. The goal of a
```
rusty1s, line 2:
```suggestion
This class defines the abstraction for a backend-agnostic feature store. The goal of a
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+feature store is to abstract away all node and edge feature memory management
+so that varying implementations can allow for independent scale-out.
+
+This particular feature store abstraction makes a few key assumptions:
+    * The features we care about storing are graph node and edge features. To
```
rusty1s, line 7:
```suggestion
    * The features we care about storing are node and edge features of a graph. To
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        this end, the attributes that the feature store supports include a
+        group_name (e.g. a heterogeneous node name, a heterogeneous edge type,
+        etc.), an attr_name (which defines the name of the feature tensor,
+        e.g. `feat`, `discrete_feat`, etc.), and an index.
+    * A feature can be uniquely identified from any associated attributes
+        specified in :obj:`TensorAttr`.
```
rusty1s, line 13:
```suggestion
        specified in :class:`TensorAttr`.
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+
+import numpy as np
+import torch
+
+from torch_geometric.typing import FeatureTensorType
+from torch_geometric.utils.mixin import CastMixin
+
+_field_status = Enum("FieldStatus", "UNSET")
+
+IndexType = Union[FeatureTensorType, slice]
```
rusty1s, line 38:
```suggestion
IndexType = Union[Tensor, np.array, slice, int]
```
Looks a bit weird to use `FeatureTensor` as `IndexType` as well. Maybe make this here explicit as well:

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+@dataclass
+class TensorAttr(CastMixin):
+    r"""Defines the attributes of a :obj:`FeatureStore` tensor; in particular,
+    all the parameters necessary to uniquely identify a tensor from the feature
+    store.
+
+    Note that the order of the attributes is important; this is the order in
+    which attributes must be provided for indexing calls. Feature store
+    implementor classes can define a different ordering by overriding
+    TensorAttr.__init__.
```
rusty1s, line 50: I'm not really sure why we need to inherit from `TensorAttr` in order to achieve this. In general, I would assume to be able to plug-in any `dataclass` as a `attr_cls` to the `FeatureStore`. This is not blocking for now.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+
+    # The type of the nodes that the tensor corresponds to (may be used for
+    # hetereogeneous graphs)
+    group_name: Optional[str] = _field_status.UNSET
```
rusty1s, line 55: Why can't we use `None` for this? I would expect that all atrributes we define here are somewhat required.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        for field in self.__dataclass_fields__:
+            val = getattr(attr, field)
+            if val != _field_status.UNSET:
+                setattr(self, field, val)
+
+
+class AttrView(CastMixin):
+    r"""Defines a view of a :obj:`FeatureStore` that is obtained from a
+    specification of attributes on the feature store. The view stores a
+    reference to the backing feature store as well as a :obj:`TensorAttr`
```
rusty1s, line 90:
```suggestion
    reference to the backing feature store as well as a :class:`TensorAttr`
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+    object that represents the view's state.
+
+    Users can create views either using the :obj:`AttrView` constructor,
+    :obj:`FeatureStore.view`, or by incompletely indexing a feature store.
```
rusty1s, line 94: This is super cool. Maybe consider adding some `code-block:: python` example directly in this doc-string.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    _store: 'FeatureStore'
```
rusty1s, line 96: Not sure why we need to define those here.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+    _attr: TensorAttr
+
+    def __init__(self, store: 'FeatureStore', attr: TensorAttr):
+        self._store = store
+        self._attr = attr
+
+    # Python built-ins ########################################################
+
+    def __getattr__(self, key) -> 'AttrView':
```
rusty1s, line 105:
```suggestion
    def __getattr__(self, key: str) -> 'AttrView':
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        r"""Sets the attr_name field of the backing :obj:`TensorAttr` object to
```
rusty1s, line 106: I think we could be more general here so that our `FeatureStore` is independent from `attr_name` or `group_name` naming and scales nicely to any other attributes we may define. One option for this is that `key` refers to the first attribute in `TensorAttr` that is not set to `UNSET`. That means we could do:
```python
store.group_name.attr_name
```
as well as
```python
store['group_name'].attr_name
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        different values of attr_name. In particular, for a feature store that
+        has `feat` as an `attr_name`, the following code indexes into `feat`:
+
+        .. code-block:: python
+
+            store[group_name].feat[:]
+
+        """
+        if key in ['_attr', '_store']:
+            return super(AttrView, self).__getattribute__(key)
```
rusty1s, line 117: Not sure why we need this.
```suggestion
            return getattr(self, key)
```
should work as well.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+This class defines the abstraction for a Graph feature store. The goal of a
+feature store is to abstract away all node and edge feature memory management
+so that varying implementations can allow for independent scale-out.
+
+This particular feature store abstraction makes a few key assumptions:
+    * The features we care about storing are graph node and edge features. To
+        this end, the attributes that the feature store supports include a
+        group_name (e.g. a heterogeneous node name, a heterogeneous edge type,
+        etc.), an attr_name (which defines the name of the feature tensor,
+        e.g. `feat`, `discrete_feat`, etc.), and an index.
```
rusty1s, line 11: Adopt PyG naming
```suggestion
        e.g. `x`, `edge_attr`, etc.), and an index.
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+            store[group_name].feat[:]
+
+        """
+        if key in ['_attr', '_store']:
+            return super(AttrView, self).__getattribute__(key)
+
+        self._attr.attr_name = key
+        if self._attr.is_fully_specified():
+            return self._store.get_tensor(self._attr)
+        return self
```
rusty1s, line 122: Is it a good idea to modify this in-place? I think this may lead to hard-to-find bugs later on. Consider `copy.copy(self)` first.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        the attribute. This allows for :obj:`AttrView` to be indexed by
+        different values of attr_name. In particular, for a feature store that
+        has `feat` as an `attr_name`, the following code indexes into `feat`:
+
+        .. code-block:: python
+
+            store[group_name].feat[:]
+
+        """
+        if key in ['_attr', '_store']:
```
rusty1s, line 116: We do not need this here at all. Python will not call `__getattr__` in case there exists an attribute with that name. You can see that `BaseStorage` does not as well.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        their backing attribute values. In particular, the following operation
+        sets the `index` of an :obj:`AttrView`:
+
+        .. code-block:: python
+
+            view = store.view(TensorAttr(group_name))
+            view.index = torch.Tensor([1, 2, 3])
+
+        """
+        if key in ['_attr', '_store']:
```
rusty1s, line 136: Drop.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+            return super(AttrView, self).__setattr__(key, value)
+
+        TensorAttr.__setattr__(self._attr, key, value)
+
+    def __getitem__(
```
rusty1s, line 141: Call `getattr` here to avoid duplicated code?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        return self._store.get_tensor(self._attr)
+
+    def __eq__(self, __o: object) -> bool:
+        r"""Compares two :obj:`AttrView` objects by checking equality of their
+        :obj:`FeatureStore` references and :obj:`TensorAttr` attributes."""
+        if not isinstance(__o, AttrView):
+            return False
+
+        return id(self._store) == id(__o._store) and self._attr == __o._attr
```
rusty1s, line 182: Two feature stores may be equal even is there `id` is different. Maybe just implement a `__eq__` in `FeatureStore` as well?

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        self._attr_cls = attr_cls
+
+    # Core (CRUD) #############################################################
+
+    @abstractmethod
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
+        r"""Implemented by :obj:`FeatureStore` subclasses."""
+        pass
+
+    def put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
```
rusty1s, line 205:
```suggestion
    def put_tensor(self, tensor: FeatureTensorType, *args, **kwargs) -> bool:
```
This would be the syntax I would prefer. Which leads to
```python
store.put_tensor(tensor, group_name=..., attr_name=...)
```
We can do `self._attr_cls.cast(*args, **kwargs)` as well to cast this.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        return self._put_tensor(tensor, attr)
+
+    @abstractmethod
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Implemented by :obj:`FeatureStore` subclasses."""
+        pass
+
+    def get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Synchronously obtains a :obj:`FeatureTensorType` object from the
+        feature store. Feature store implementors guarantee that the call
```
rusty1s, line 230:
```suggestion
        feature store. Feature store implementors guarantee that ... holds.
```

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        attr = self._attr_cls.cast(attr)
+        return self._put_tensor(tensor, attr)
+
+    @abstractmethod
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Implemented by :obj:`FeatureStore` subclasses."""
+        pass
+
+    def get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
```
rusty1s, line 228:
```suggestion
    def get_tensor(self, *args, **kwargs) -> FeatureTensorType:
```
see above. In addition, I don't understand the `Optional[...]` output.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+                return None
+            if isinstance(attr.index, torch.Tensor):
+                return torch.from_numpy(tensor) if isinstance(
+                    tensor, np.ndarray) else tensor
+            if isinstance(attr.index, np.ndarray):
+                return tensor.numpy() if isinstance(tensor,
+                                                    torch.Tensor) else tensor
+            return tensor
+
+        attr = self._attr_cls.cast(attr)
```
rusty1s, line 255: assert fully-specified?

torch_geometric/data/feature_store.py
rusty1s: Yeah, it might not be possible to do that in `_get_tensor` because we do not know the `dtype` of the tensor in advance. Ultimately, this means that the implementor needs to take care of type conversion as well (to not index a PyTorch tensor with a numpy array and vice versa), so I am not sure I see the point of `to_type` after `_get_tensor` has done its job.

torch_geometric/data/feature_store.py
```diff
@@ -0,0 +1,354 @@
...
+        return AttrView(self, self._attr_cls.cast(attr))
+
+    # Python built-ins ########################################################
+
+    def __setitem__(self, key: TensorAttr, value: FeatureTensorType):
+        r"""Supports store[tensor_attr] = tensor."""
+        key = self._attr_cls.cast(key)
+        self.put_tensor(value, key)
+
+    def __getitem__(self, key: TensorAttr):
```
rusty1s, line 321: `getattr` and `setattr` equivalents?

rusty1s (commented):
This looks way better now. Thanks for updating. Left a bunch of comments nonetheless :P

torch_geometric/data/feature_store.py
mananshah99: Excellent point.

torch_geometric/data/feature_store.py
mananshah99: Good to know, thanks!

torch_geometric/data/feature_store.py
mananshah99: Sounds good, for the base class `__eq__` will just be an `id` comparison but subclasses can override.

torch_geometric/data/feature_store.py
mananshah99: I don't think a `FeatureStore` should implement `getattr` and `setattr`; imo, these should only be implemented for views on the store. I don't think it's particularly clean to have
```
store.group_name -> AttrView(group_name=group_name)
```
as this syntax seems more confusing to me than clarifying.

torch_geometric/data/feature_store.py
mananshah99: The Optional output is to indicate that `get` may return `None`, if the tensor is not found.

rusty1s: Python dictionaries throw an error in that case. Do we want to break with this?

torch_geometric/data/feature_store.py
rusty1s: 👍

torch_geometric/data/feature_store.py
rusty1s: Thought it makes sense to implement for stores without group names (like PyG `data` objects).

torch_geometric/data/feature_store.py
mananshah99: This is a cool suggestion, thanks!

torch_geometric/data/feature_store.py
mananshah99: Fair enough. Since it doesn't really ever make sense for a feature store to store `None`, we can raise the `KeyError` in `get`.

torch_geometric/data/feature_store.py
mananshah99: Elaborated more in the longer comment below.

mananshah99:
@rusty1s as always, thank you for the detailed comments :) Cleaned up the interface and builtins more. A couple general notes I wanted to share:
* **On `UNSET` fields in `TensorAttr`.** I've spent some more time thinking about this, and I think we should proceed as follows.
  * We need some sort of tombstone in `TensorAttr` to indicate which attributes have been set; this is important both when checking for full specification and when implementing `__getattr__` and `__getitem__` in `AttrView` (which choose the attribute they will be setting based on the first UNSET attribute). 
  * As a result, I think that all fields in `TensorAttr` should start as UNSET. If a field is not being used by a `TensorAttr` subclass, it can default that field to `None` and move it to the end of the `__init__` list (there should be no modification of `is_fully_specified`, which just checks to see whether there are any UNSET values). Users may set fields however they like (with `None` or with any other value). It is up to the feature store implementation to handle these values properly.
  * Partially-specified views, when called, will set all UNSET values to None and pass this `TensorAttr` to the `get_tensor` method. This way, feature store implementors will never have to interact with UNSET values.
  *  I think the resulting implementation is very clean and reduces the number of moving parts: all `TensorAttr` attributes start as UNSET, they can be set by the user (or default to `None` if a partially-specified view is called), and subclasses of `TensorAttr` just need to modify `__init__` and the default values of attributes they are not using.
* **On subclassing `TensorAttr`.** I have a slight bias towards enforcing that `attr_class` subclasses `TensorAttr` (instead of being any arbitrary class), due to the resulting simplicity of implementation (for both the user and the implementor). If we find the need to implement different `TensorAttr` classes for each feature store, this is of course supported (so long as the class has `is_fully_specified` and `fully_specify` methods), but I would avoid this for the time being.

torch_geometric/data/feature_store.py
mananshah99: Responded in the longer comment below.

torch_geometric/data/feature_store.py
mananshah99: Yeah, I wasn't sure about the best way to handle this - one way to handle this would be to always type convert to numpy before passing to `_get_tensor`, but this felt like unnecessary overhead. But the only other option would be to push down this logic entirely to the feature store, which also doesn't seem ideal. Any thoughts?

torch_geometric/data/feature_store.py
mananshah99: Isn't the only way for an output to be fully specified through `getattr` to allow for chaining, which necessitates that `getattr` can return an `AttrView`? That feels odd to me.

rusty1s: If `TensorAttr` sets both `group_name` and `index` to `None` by default, then `store.{attr_name}` should give you a tensor. I am okay with leaving this out for now.

rusty1s (approved):
Thanks for all the updates :)

PR #4562, opened 2022-04-29
Improve distance cutoff in `DimeNet`

danielzuegner:
Currently (as per my [recent PR](https://github.com/pyg-team/pytorch_geometric/pull/4506)) the clamping of `(dist / cutoff).clamp_max(1.0)` is done outside the envelope. However, this means that we need to repeat this for each place using the envelope (e.g., the [`SphericalBasisLayer`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/dimenet.py#L109)). The clean way to do it is to have the masking inside the envelope; this is also how it's implemented in [`DimeNet`](https://github.com/gasteigerjo/dimenet/blob/09123a0e16e728d0a0e53e6686b04f859802aa81/dimenet/model/layers/envelope.py#L23).

### DIFF
--- a/torch_geometric/nn/models/dimenet.py
+++ b/torch_geometric/nn/models/dimenet.py
@@ -104,7 +104,7 @@ class SphericalBasisLayer(torch.nn.Module):
                 self.bessel_funcs.append(bessel)
 
     def forward(self, dist, angle, idx_kj):
-        dist = (dist / self.cutoff)
+        dist = dist / self.cutoff
         rbf = torch.stack([f(dist) for f in self.bessel_funcs], dim=1)
         rbf = self.envelope(dist).unsqueeze(-1) * rbf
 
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/models/dimenet.py
```diff
@@ -45,7 +45,8 @@ def forward(self, x):
         x_pow_p0 = x.pow(p - 1)
         x_pow_p1 = x_pow_p0 * x
         x_pow_p2 = x_pow_p1 * x
-        return 1. / x + a * x_pow_p0 + b * x_pow_p1 + c * x_pow_p2
+        return (1. / x + a * x_pow_p0 + b * x_pow_p1 +
+                c * x_pow_p2) * (x < 1.0).float()
```
rusty1s, line 49:
```suggestion
                c * x_pow_p2) * (x < 1.0).to(x.dtype)
```

rusty1s (approved):

rusty1s:
Thanks!

PR #4570, opened 2022-04-30
Add missing LINKX datasets

Padarn:
This PR addresses the request in https://github.com/pyg-team/pytorch_geometric/issues/4569 to add the missing datasets from LINKX.

Note that the loading of these datasets requires handling slightly different storage formats. It may be better that we create separate loaders for the datasets that don't match the common pattern.

Also note that `deezer-europe` dataset is stored with a large number of [categorical features](https://snap.stanford.edu/data/feather-deezer-social.html)  which I am currently turning into a dense tensor (as per https://github.com/CUAI/Non-Homophily-Large-Scale/blob/eb531f33b2c8f3c14d64287938b761df1838e6d3/dataset.py#L163). Possibly we can store this as sparse.

### DIFF
--- a/torch_geometric/datasets/linkx_dataset.py
+++ b/torch_geometric/datasets/linkx_dataset.py
@@ -36,6 +36,14 @@ class LINKXDataset(InMemoryDataset):
         'amherst41': f'{url}/facebook100/Amherst41.mat',
         'cornell5': f'{url}/facebook100/Cornell5.mat',
         'johnshopkins55': f'{url}/facebook100/Johns%20Hopkins55.mat',
+        'chameleon': f'{url}/chameleon.mat',
+        'cornell': f'{url}/cornell.mat',
+        'deezer-europe': f'{url}/deezer-europe.mat',
+        'film': f'{url}/film.mat',
+        'genius': f'{url}/genius.mat',
+        'squirrel': f'{url}/squirrel.mat',
+        'texas': f'{url}/texas.mat',
+        'wisconsin': f'{url}/wisconsin.mat',
     }
 
     splits = {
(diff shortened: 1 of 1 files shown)


torch_geometric/datasets/linkx_dataset.py
```diff
@@ -79,20 +87,36 @@ def process(self):
...
-            _, out = x[:, i].unique(return_inverse=True)
-            xs.append(F.one_hot(out).to(torch.float))
-        x = torch.cat(xs, dim=-1)
+        if 'A' in mat:
+            A = mat['A'].tocsr().tocoo()
+            row = torch.from_numpy(A.row).to(torch.long)
+            col = torch.from_numpy(A.col).to(torch.long)
+            edge_index = torch.stack([row, col], dim=0)
+        elif 'edge_index' in mat:
+            edge_index = torch.tensor(mat['edge_index'])
```
rusty1s, line 96:
```suggestion
            edge_index = torch.from_numpy(mat['edge_index']).to(torch.long)
```

torch_geometric/datasets/linkx_dataset.py
```diff
@@ -79,20 +87,36 @@ def process(self):
...
+            xs = []
+            y = metadata[:, 1] - 1  # gender label, -1 means unlabeled
+            x = torch.cat([metadata[:, :1], metadata[:, 2:]], dim=-1)
+            for i in range(x.size(1)):
+                _, out = x[:, i].unique(return_inverse=True)
+                xs.append(F.one_hot(out).to(torch.float))
+            x = torch.cat(xs, dim=-1)
+        else:
+            if 'node_feat' in mat:
+                x = torch.tensor(mat['node_feat'])
```
rusty1s, line 111:
```suggestion
                x = torch.from_numpy(mat['node_feat']).to(torch.float)
```

torch_geometric/datasets/genius.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Callable, List, Optional
+
+import torch
+
+from torch_geometric.data import Data, InMemoryDataset, download_url
+
+
+class GeniusDataset(InMemoryDataset):
+    r"""The 'genius' datasets from  `"Expertise and dynamics within
+    crowdsourced musical knowledge curation"
```
rusty1s, line 11:
```suggestion
    Crowdsourced Musical Knowledge Curation"
```

torch_geometric/datasets/genius.py
```diff
@@ -0,0 +1,71 @@
...
+        pre_transform (callable, optional): A function/transform that takes in
+            an :obj:`torch_geometric.data.Data` object and returns a
+            transformed version. The data object will be transformed before
+            being saved to disk. (default: :obj:`None`)
+    """
+
+    url = 'https://github.com/CUAI/Non-Homophily-Large-Scale/raw/master/' \
+          'data/genius.mat'
+
+    def __init__(self, root: str, transform: Optional[Callable] = None,
```
rusty1s, line 29: `name` missing?

torch_geometric/datasets/genius.py
```diff
@@ -0,0 +1,71 @@
...
+    def raw_dir(self) -> str:
+        return osp.join(self.root, self.name, 'raw')
+
+    @property
+    def processed_dir(self) -> str:
+        return osp.join(self.root, self.name, 'processed')
+
+    @property
+    def raw_file_names(self) -> List[str]:
+        names = [self.datasets[self.name].split('/')[-1]]
```
rusty1s, line 45: Where is `self.datasets` and `self.splits` defined?

Padarn:
Hey @rusty1s sorry for the earlier WIP - I was still figuring out how I thought it best to add these datasets.

I decided it was best to add them to the existing `LINKXDataset` as in my original PR - the reason being that these datasets are modified from the original datasets, and I thought it would be confusing if we named them individually.

I will add the other datasets in another PR if you're okay with this approach.

rusty1s:
Thanks @Padarn. I agree with you. This approach sounds perfect to me!

Issue #4567, opened 2022-04-29
`download` and `process` methods are not called when `Dataset` is subclassed indirectly


raimis:
### 🐛 Describe the bug

Create a dummy dataset `DS`  by directly subclassing `Dataset`, i.e. `DS` --> `Dataset`:
```python
from torch_geometric.data import Dataset

class DS(Dataset):

    @property
    def raw_file_names(self):
        return 'dummy'

    def download(self):
        print(f'Calling {self.__class__.__name__}.download')
``` 
When creating an instance of `DS`, the method `download` is called as expected:
```python
ds = DS('.')
Calling DS.download
```

Create a dummy dataset `DS2` by indirectly subclassing `Dataset`, i.e.  `DS2` --> `DS1` --> `Dataset`:
```python
class DS1(Dataset):

    def download(self):
        print(f'Calling {self.__class__.__name__}.download')

class DS2(DS1):

    @property
    def raw_file_names(self):
        return 'dummy'
```
When creating an instance of `DS2`, the method `download` **is not called**, as expected:
```python
ds2 = DS2('.')
# No output
```
However, the method can be called directly:
```python
ds2.download()
Calling DS2.download
``` 

Additionally, the same issue exists with `process` method.

### Environment

* PyG version: `master` (https://github.com/pyg-team/pytorch_geometric/commit/484b7922ec68d146280acba409fc7170549e2775)
* PyTorch version: 1.10.0
* OS: CentOS 7.9
* Python version: 3.9.10
* CUDA/cuDNN version: 11.2.2/8.2.1.32
* How you installed PyTorch and PyG (`conda`, `pip`, source): `conda` and `source`
* Any other relevant information (*e.g.*, version of `torch-scatter`): N/A

raimis:
I suspect, the problem is here:
https://github.com/pyg-team/pytorch_geometric/blob/484b7922ec68d146280acba409fc7170549e2775/torch_geometric/data/dataset.py#L83-L87

Padarn:
I suggest we replace this with 
```
if hasattr(hasattr, 'download'): 
     self._download() 
```
Thoughts?

rusty1s:
Thanks for reporting, although this is somewhat intended. In general, we only want to call `download()` in case it is overridden by the base class. Note that
```python
if hasattr(self, 'download'):
```
[...]

PR #4586, opened 2022-05-03
change dataset behaviour when dataset is overridden

Padarn:
PR to address https://github.com/pyg-team/pytorch_geometric/issues/4567

Changes in the behavior of how 'download' and 'process' are resolved by Dataset. Prior to this change, the functions needed to be defined in the child class, after this change they only need to be implemented in some class that is not the base `Dataset`.

### DIFF
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -1,6 +1,6 @@
 import torch
 
-from torch_geometric.data import Data, HeteroData, InMemoryDataset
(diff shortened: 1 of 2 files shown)


Padarn:
Hey @rusty1s what is your intended interaction with `CHANGELOG.md`? Do you want all merges to master to add one, or only larger changes?

rusty1s:
Every PR should at least change the `CHANGELOG.md` for now (either by writing a new line or adding the PR link to a larger feature).

rusty1s (approved):

PR #4604, opened 2022-05-09
`HeteroData.is_undirected()`

rusty1s:


torch_geometric/data/hetero_data.py
```diff
@@ -554,4 +517,53 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
...
+
+        num_edges = store.num_edges
+        edge_slices[edge_type] = (cumsum, cumsum + num_edges)
+        cumsum += num_edges
+
+    edge_index = None
+    if len(edge_indices) == 1:  # Memory-efficient `torch.cat`:
+        edge_index = edge_indices[0]
+    elif len(edge_indices) > 0:
+        edge_index = torch.cat(edge_indices, dim=-1)
```
Padarn, line 567: [naive question] would it be better to cat incrementally in the loop which determines the edge_slices?

Padarn (approved):

Padarn:
Out of scope - but do you think we should somehow cache the result of 'is undirected'? I'm wondering if anyone uses it for example in a forward pass of a model, in which case it'll be a very expensive operation to repeat many time.

rusty1s:
I think that is an interesting idea. One thing that complicates this is that PyG would need to track changes to `Data` and invalidate the cache, or define some method to freeze the current data object and disallow modifications. Let me know what you think!

Padarn:
Yeah that is true, invalidating the cache does seem fairly error-prone. It doe seem possible given the `Data` objects do all of their setting of attributes via the `_store`. Perhaps it would be fairly safe to just always clear the cache when an attribute is changed.

I like the idea of something like `FrozenData`, but outside of this I don't know what other use cases it might serve.

PR #4620, opened 2022-05-11
`NeighborLoader`: Optionally use argument data without `to_csc`

ZenoTan:


torch_geometric/data/lightning_datamodule.py
```diff
@@ -219,6 +219,7 @@ def __init__(
         loader: str = "neighbor",
         batch_size: int = 1,
         num_workers: int = 0,
+        keep_data: bool = False,
```
rusty1s, line 222: No need for that. We can pass that as part of `kwargs`.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         input_type: Optional[Any] = None,
         share_memory: bool = False,
         time_attr: Optional[str] = None,
+        keep_data=False,
```
rusty1s, line 29: I personally prefer `is_sorted` argument, which dictates that `edge_index` is sorted by destination. Then we simply need to run `torch.ops.torch_sparse.ind2ptr` to compute the `colptr`.
 ```suggestion
        is_sorted: bool=False,
```

rusty1s (approved):

PR #4625, opened 2022-05-11
Support return embedding for MLP layer

JiaxuanYou:
The added `return_emb` layer is useful when user would like to extract embeddings along with predictions


rusty1s:
Can we do this as part of `forward`?
```python
def forward(self, x, return_emb=True)
```

JiaxuanYou:
> Can we do this as part of `forward`?
> 
> ```python
> def forward(self, x, return_emb=True)
> ```

Sounds good!

JiaxuanYou:
Updated

rusty1s (approved):

PR #4626, opened 2022-05-12
AQSOL dataset

vijaydwivedi75:
This PR adds the AQSOL dataset from updated [Benchmarking GNNs](http://arxiv.org/abs/2003.00982), based on [AqSolDB](https://www.nature.com/articles/s41597-019-0151-1) which is a standardized database of 9,982 molecular graphs with their aqueous solubility values, collected from 9 different data sources. 
    
The aqueous solubility targets are collected from experimental measurements and standardized to LogS units in AqSolDB. These final values are the property to regress in the AQSOL dataset, similar to ZINC. This version of the dataset filters out few graphs with no bonds/edges and a small number of graphs with missing node feature values.

The resultant total molecular graphs are 9,823. For each molecular graph, the node features are the types of heavy atoms and the edge features are the types of bonds between them, similar as ZINC.
  
Dataset overview:
- Task: Graph Regression
- Size of Dataset: 9,982 molecules.
- Split: Scaffold split (8:1:1) following same code as OGB.
- After cleaning: 7,831 train / 996 val / 996 test
- Number of (unique) atoms: 65 (Dict below)
- Number of (unique) bonds: 5 (Dict below)
- Performance Metric: MAE, same as ZINC

```
Atom Dict: {'Br': 0, 'C': 1, 'N': 2, 'O': 3, 'Cl': 4, 'Zn': 5, 'F': 6, 'P': 7, 'S': 8, 'Na': 9,
'Al': 10, 'Si': 11, 'Mo': 12, 'Ca': 13, 'W': 14, 'Pb': 15, 'B': 16, 'V': 17, 'Co': 18,
'Mg': 19, 'Bi': 20, 'Fe': 21, 'Ba': 22, 'K': 23, 'Ti': 24, 'Sn': 25, 'Cd': 26, 'I': 27,
'Re': 28, 'Sr': 29, 'H': 30, 'Cu': 31, 'Ni': 32, 'Lu': 33, 'Pr': 34, 'Te': 35, 'Ce': 36,
'Nd': 37, 'Gd': 38, 'Zr': 39, 'Mn': 40, 'As': 41, 'Hg': 42, 'Sb': 43, 'Cr': 44, 'Se': 45,
'La': 46, 'Dy': 47, 'Y': 48, 'Pd': 49, 'Ag': 50, 'In': 51, 'Li': 52, 'Rh': 53, 'Nb': 54,
'Hf': 55, 'Cs': 56, 'Ru': 57, 'Au': 58, 'Sm': 59, 'Ta': 60, 'Pt': 61, 'Ir': 62, 'Be': 63, 'Ge': 64}
    
Bond Dict: {'NONE': 0, 'SINGLE': 1, 'DOUBLE': 2, 'AROMATIC': 3, 'TRIPLE': 4}
```

### DIFF
--- a/torch_geometric/datasets/aqsol.py
+++ b/torch_geometric/datasets/aqsol.py
@@ -1,0 +1,133 @@
+import os
+import os.path as osp
+import shutil
+import pickle
+
+import torch
+from tqdm import tqdm
+from torch_geometric.data import (InMemoryDataset, Data, download_url,
+                                  extract_zip)
+
+
+class AQSOL(InMemoryDataset):
+    r"""The AQSOL dataset from `Benchmarking GNNs <http://arxiv.org/abs/2003.00982>`_,
+    based on `AqSolDB <https://www.nature.com/articles/s41597-019-0151-1>`
+    which is a standardized database of 9,982 molecular graphs with
+    their aqueous solubility values, collected from 9 different data sources. 
+    
+    The aqueous solubility targets are collected from experimental measurements and standardized
+    to LogS units in AqSolDB. These final values as the property to regress in the AQSOL dataset
+    which is the resultant collection in Benchmarking GNNs after filtering out few graphs
+    with no bonds/edges and a small number of graphs with missing node feature values.
+    
+    Thus, the total molecular graphs are 9,823. For each molecular graph, the node features are the
+    types of heavy atoms and the edge features are the types of bonds between them, similar as ZINC.
+    
+    Size of Dataset: 9,982 molecules.
+    Split: Scaffold split (8:1:1) following same code as OGB.
+    After cleaning: 7,831 train / 996 val / 996 test
+    Number of (unique) atoms: 65
+    Number of (unique) bonds: 5
+    Performance Metric: MAE, same as ZINC
+
+    Atom Dict: {'Br': 0, 'C': 1, 'N': 2, 'O': 3, 'Cl': 4, 'Zn': 5, 'F': 6, 'P': 7, 'S': 8, 'Na': 9, 'Al': 10,
+    'Si': 11, 'Mo': 12, 'Ca': 13, 'W': 14, 'Pb': 15, 'B': 16, 'V': 17, 'Co': 18, 'Mg': 19, 'Bi': 20, 'Fe': 21,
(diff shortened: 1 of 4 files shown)


torch_geometric/datasets/aqsol.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def download(self):
+        shutil.rmtree(self.raw_dir)
+        path = download_url(self.url, self.root)
+        extract_zip(path, self.root)
+        os.rename(osp.join(self.root, 'asqol_graph_raw'), self.raw_dir)
+        os.unlink(path)
+
+    def process(self):
+        for split in ['train', 'val', 'test']:
```
Padarn, line 100: I guess you don't really need to process the splits you're not going to load - but also maybe its worth just doing it ahead of time.

torch_geometric/datasets/aqsol.py
```diff
@@ -0,0 +1,148 @@
...
+            with open(osp.join(self.raw_dir, f'{split}.pickle'), 'rb') as f:
+                graphs = pickle.load(f)
```
Padarn, line 102: I feel loading pickles is a bit dangerous in general (doesn't work well across python versions) - do you know if the data is available in a non-pickle format?

Padarn:
Other than a slight preference not to load pickled data (which maybe we cannot avoid here), this looks good to me!

vijaydwivedi75:
Thanks @Padarn for your comments. The PR is similar to how [ZINC](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/datasets/zinc.html) dataset is maintained, which also loads from pickle files.

The pickle file in AQSOL (this PR) is a list of graphs objects, each of which is a tuple of the graph info (the tuple information is mentioned in comments as well as below).

```
# Each `graph` is a tuple (x, edge_attr, edge_index, y)
#     Shape of x : [num_nodes, 1]
#     Shape of edge_attr : [num_edges]
#     Shape of edge_index : [2, num_edges]
#     Shape of y : [1]
```

rusty1s (approved):

rusty1s:
Thank you!

PR #4628, opened 2022-05-12
Add `unbatch` functionality

tzuhanchang:
Unbatching data from DataLoader batch to a list. This is useful for GNN edge classifier, in which case, graphs are required to be reconstructed with edge predictions.

### DIFF
--- a/torch_geometric/loader/utils.py
+++ b/torch_geometric/loader/utils.py
@@ -34,6 +34,7 @@ def to_csc(
     data: Union[Data, EdgeStorage],
     device: Optional[torch.device] = None,
     share_memory: bool = False,
+    is_sorted: bool = False,
 ) -> Tuple[Tensor, Tensor, OptTensor]:
     # Convert the graph data into a suitable format for sampling (CSC format).
     # Returns the `colptr` and `row` indices of the graph, as well as an
@@ -47,17 +48,18 @@ def to_csc(
 
     elif hasattr(data, 'edge_index'):
         (row, col) = data.edge_index
-        size = data.size()
-        perm = (col * size[0]).add_(row).argsort()
+        if not is_sorted:
+            size = data.size()
+            perm = (col * size[0]).add_(row).argsort()
+            row = row[perm]
         colptr = torch.ops.torch_sparse.ind2ptr(col[perm], size[1])
-        row = row[perm]
     else:
         raise AttributeError("Data object does not contain attributes "
                              "'adj_t' or 'edge_index'")
 
     colptr = colptr.to(device)
     row = row.to(device)
-    perm = perm if perm is not None else perm.to(device)
+    perm = perm.to(device) if perm is not None else None
 
     if not colptr.is_cuda and share_memory:
         colptr.share_memory_()
@@ -72,6 +74,7 @@ def to_hetero_csc(
     data: HeteroData,
     device: Optional[torch.device] = None,
     share_memory: bool = False,
+    is_sorted: bool = False,
 ) -> Tuple[Dict[str, Tensor], Dict[str, Tensor], Dict[str, OptTensor]]:
     # Convert the heterogeneous graph data into a suitable format for sampling
     # (CSC format).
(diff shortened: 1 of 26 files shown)


torch_geometric/loader/utils.py
```diff
@@ -24,6 +25,22 @@ def index_select(value: Tensor, index: Tensor, dim: int = 0) -> Tensor:
     return torch.index_select(value, 0, index, out=out)
 
 
+def unbatcher(src: torch.tensor, index: torch.tensor, dim: int = -1) -> list:
```
rusty1s, line 28: Can you add some documentation to this function? I am unsure what's the exact purpose and use-case of this is.

torch_geometric/loader/utils.py
```diff
@@ -24,6 +25,22 @@ def index_select(value: Tensor, index: Tensor, dim: int = 0) -> Tensor:
...
+    scatter = pandas.DataFrame(
+        index, columns=['scatter']).reset_index().groupby('scatter').tail(1)
```
rusty1s, line 30: Any reason we need `pandas` for this? This should be doable via `torch-scatter` as well.

torch_geometric/loader/utils.py
```diff
@@ -24,6 +25,22 @@ def index_select(value: Tensor, index: Tensor, dim: int = 0) -> Tensor:
...
+    indices = []
+    index_i = 0  # index starts for a graph
+    index_f = 0  # index ends for a graph
+    for i in range(0, len(scatter)):
```
rusty1s, line 34: Any chance we can remove the (expensive) for-loop via by making use of vectorized operations?

torch_geometric/loader/utils.py
tzuhanchang: After edge classification we ended with data batches, we want to extract edge scores and reconstruct networkx digraphs with this new edge information. This application is very specific for usage in collider physics, where we want to assign two jets to form a parton according to the edge scores.

torch_geometric/loader/utils.py
tzuhanchang: We want to find where a graph end, then use ```index_select``` to select out the edges attributes of a graph. This also works for unbatching node_attrs

torch_geometric/loader/utils.py
tzuhanchang: At this moment I don't have any more efficient way, this loop find all node_attrs or edge_attrs for all graphs in the batch. We then use ```Data``` and ```to_networkx``` to reconstruct all graphs.

rusty1s (changes requested):
I think in order to get this merged, we would need to have a pure PyTorch function without for-loops. IMO, this is achievable by
```python
src.split(degree(index).tolist())
```
In addition, we would need to add this function to `utils` rather than `loader/utils.py`, add doccumentation, and write tests for it. WDYT?

tzuhanchang:
@rusty1s Thank you very much! ```src.split(degree(index).tolist())``` works perfectly, it can replace two for loops completely. I also add it to ```utils```.

rusty1s (commented):
This looks better. Thank you! Can we also add some basic tests in `test/utils/test_unbatch.py` and add this operator to `torch_geometric/utils/__init__.py`?

torch_geometric/utils/unbatch.py
```diff
@@ -0,0 +1,18 @@
...
+    r"""Unbatches batched data from DataLoader.
+
+    Args:
+        src (Tensor): source tensor.
+        index (LongTensor): index tensor.
+        dim (int): dimension along which to split the tensor.
+
+    :rtype: :class:`list`
+    """
+    split_sizes = list(map(int, degree(index).tolist()))
```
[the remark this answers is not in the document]
tzuhanchang: Thanks, this works. Since ```tensor.split``` returns a tuple so I added ```unbatched = list(src.split(split_sizes, dim))``` to convert output to a ```list[Tensor]```

rusty1s (approved):
Thanks for the updates!

PR #4635, opened 2022-05-13
`HeteroData.subgraph()`

### LINKED ISSUE #4001: `HeteroData.subgraph()`

Similar to `Data.subgraph()`, there should exist a `HeteroData.subgraph()` method to compute subgraphs in a heterogeneous graph setting, *e.g.*, for obtaining inductive node splits. Here, `mask`/`index` should be of type `dict`, holding masks/indices for each/a subset of node types:
```python
hetero_data.subgraph({'paper': mask})
```

### Alternatives

_No response_

### Additional context

_No response_

wsad1:
Closes #4001.

This PR adds `subgraph` to `HeteroData`.

### DIFF
--- a/torch_geometric/data/hetero_data.py
+++ b/torch_geometric/data/hetero_data.py
@@ -229,6 +229,14 @@ class HeteroData(BaseData):
         r"""Returns a list of all edge storages of the graph."""
         return list(self._edge_store_dict.values())
 
+    def node_items(self) -> List[Tuple[NodeType, NodeStorage]]:
+        r"""Returns a list of node type and node storage pairs."""
+        return list(self._node_store_dict.items())
+
+    def edge_items(self) -> List[Tuple[EdgeType, EdgeStorage]]:
+        r"""Returns a list of edge type and edge storage pairs."""
+        return list(self._edge_store_dict.items())
+
     def to_dict(self) -> Dict[str, Any]:
         out = self._global_store.to_dict()
         for key, store in chain(self._node_store_dict.items(),
--- a/torch_geometric/data/hetero_data.py
+++ b/torch_geometric/data/hetero_data.py
@@ -12,6 +12,7 @@ from torch_sparse import SparseTensor
 from torch_geometric.data.data import BaseData, Data, size_repr
 from torch_geometric.data.storage import BaseStorage, EdgeStorage, NodeStorage
 from torch_geometric.typing import EdgeType, NodeType, QueryType
+from torch_geometric.utils import bipartite_subgraph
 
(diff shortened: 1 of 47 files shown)


torch_geometric/data/hetero_data.py
```diff
@@ -437,6 +437,79 @@ def rename(self, name: NodeType, new_name: NodeType) -> 'HeteroData':
...
+            >>> HeteroData(
+                paper={ x=[10, 16] },
+                author={ x=[5, 32] },
+                conference={ x=[5, 8] },
+                (paper, cites, paper)={ edge_index=[2, 50] },
+                (author, to, paper)={ edge_index=[2, 30] },
+                (paper, to, conference)={ edge_index=[2, 25] }
+                )
+
+            subset ={'paper': torch.tensor([3,4,5,6]),
```
Padarn, line 463:
```suggestion
            subset = {'paper': torch.tensor([3,4,5,6]),
```

torch_geometric/data/hetero_data.py
```diff
@@ -437,6 +437,79 @@ def rename(self, name: NodeType, new_name: NodeType) -> 'HeteroData':
...
+            print(data.subgraph(subset))
+            >>> HeteroData(
+                paper={ x=[4, 16] },
+                author={ x=[2, 32] },
+                (paper, cites, paper)={ edge_index=[2, 24] },
+                (author, to, paper)={ edge_index=[2, 5] }
+                )
+
+        Args:
+            subset (Dict[str, Tensor or BoolTensor]): A dictonary holding the
```
Padarn, line 475: Do you think its worth allowing also `List[str]` when you just want to filter node types out?

torch_geometric/data/hetero_data.py
```diff
@@ -437,6 +437,79 @@ def rename(self, name: NodeType, new_name: NodeType) -> 'HeteroData':
...
+        """
+
+        data = HeteroData(self._global_store)
+
+        for node_type in self.node_types:
+            if node_type not in subset:
+                continue
+            cur_num_nodes = self[node_type].num_nodes
+            for key, value in self[node_type].items():
+                if isinstance(value,
```
Padarn, line 486: maybe you could use the `is_node_attr` and `is_edge_attr` to simplify this loop quite a bit?

torch_geometric/data/hetero_data.py
rusty1s: Good idea. Let's do this in a separate PR since I am not sure we want to re-use `subgraph` for this due its expensive re-labeling computation.

rusty1s (approved):
Thank you (and welcome back :))

PR #4682, opened 2022-05-19
Use class setter properties if present

adrianomartinelli:
# Objective
Add support of class properties when setting attributes. This pull request is based on this [discussion](https://github.com/pyg-team/pytorch_geometric/discussions/4638) with @rusty1s .

# Problem
If `Data` is subclassed all attributes are set with `__setattr__` method. This prevents users from implementing validation logic for attributes (like ensuring a certain `dtype`).  In the example below, the developer would like to ensure that `my_attr` can only hold `int` values. This is currently not possible with the implemented `__setattr__` method.

```python
  class Graph(Data):
      def __init__(self):
          super().__init__()
          self.my_attr = 1.0
          self.my_attr2 = 2

      @property
      def my_attr(self):
          return self._my_attr

      @my_attr.setter
      def my_attr(self, value):
          self._my_attr = int(value)
```

# Solution
The proposed solution checks if the class implements a setter method for the attribute or defaults to the standard behaviour
```python
    def __setattr__(self, key: str, value: Any):
        propobj = getattr(self.__class__, key, None)
        if propobj is None or propobj.fset is None:
            setattr(self._store, key, value)
        else:
            propobj.fset(self, value)
```
This implementation is tested with `test_data_subclass_setter` in `test/data/test_data.py`. Coverage of `data.py` is 89%

Best,
Adriano

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for graph-level outputs in `to_hetero` ([#4582](https://github.com/pyg-team/pytorch_geometric/pull/4582))
 - Added `CHANGELOG.md` ([#4581](https://github.com/pyg-team/pytorch_geometric/pull/4581))
 ### Changed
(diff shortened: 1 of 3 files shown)


rusty1s (approved):
Amazing. Thank you!

PR #4687, opened 2022-05-20
`torch_geometric.nn.aggr` package with base class

rusty1s:


torch_geometric/nn/aggr/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .aggr import Aggr, MeanAggr
+
+__all__ = classes = [
+    'Aggr',
+    'MeanAggr',
+]
```
cthoyt, line 6: good place for class-resolver?

torch_geometric/transforms/base_transform.py
```diff
@@ -1,7 +1,8 @@
+from abc import ABC
 from typing import Any
 
 
-class BaseTransform:
+class BaseTransform(ABC):
```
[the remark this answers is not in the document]
rusty1s: No real advantages, it just prevents that someone instantiates this class directly.

torch_geometric/nn/aggr/aggr.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC
+from typing import Optional
+
+import torch
+from torch import Tensor
+from torch_scatter import scatter, segment_csr
+
+
+class Aggr(torch.nn.Module, ABC):
```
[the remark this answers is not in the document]
rusty1s: Yeah, I thought about that as well. In the long run, I think it is good to deprecate `glob` and use `aggr` both for global pooling and `MessagePassing` aggregation.

torch_geometric/transforms/base_transform.py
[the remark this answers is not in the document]
cthoyt: I think it's really useful to write when classes are meant to be subclassed both for documentation and contributors

torch_geometric/nn/aggr/aggr.py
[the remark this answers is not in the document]
Padarn: That makes sense to me. I'll move my equilibrium aggregation layer in here too once merged

torch_geometric/nn/aggr/__init__.py
rusty1s: Thanks for your input. We will add that in a follow-up PR :)

torch_geometric/nn/aggr/aggr.py
```diff
@@ -0,0 +1,200 @@
...
+            return scatter(x, index, dim=dim, dim_size=dim_size, reduce='sum')
+
+
+class SoftmaxAggr(Aggr):
+    def __init__(self, t: float = 1.0, learn: bool = False):
+        super().__init__()
+        self.init_t = t
+        self.learn = learn
+        if learn:
+            self.t = Parameter(torch.Tensor([t]), requires_grad=True)
```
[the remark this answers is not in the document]
lightaime: I tried it briefly before. It makes minor differences. But sure, we can add this support. It may be worth trying out.

torch_geometric/nn/aggr/__init__.py
cthoyt: I will be happy to take care of it! I'll get a notification when this one is merged and follow up ASAP

torch_geometric/nn/aggr/base.py
```diff
@@ -0,0 +1,39 @@
...
+from torch import Tensor
+from torch_scatter import scatter, segment_csr
+
+
+class BaseAggr(torch.nn.Module, ABC):
+    r"""An abstract base class for writing aggregations."""
+    @abstractmethod
+    def forward(self, x: Tensor, index: Tensor, dim_size: Optional[int] = None,
+                ptr: Optional[Tensor] = None, dim: int = -2) -> Tensor:
+        pass
```
cthoyt, line 14: would be a great opportunity to write a nice docstring about what each of these are in case someone wants to subclass it themselves

torch_geometric/nn/aggr/base.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def reset_parameters(self):
+        pass
+
+    def reduce(self, x: Tensor, index: Tensor, dim_size: Optional[int] = None,
```
cthoyt, line 19: this appears to not use `self`, so it could be annotated with `@staticmethod`

torch_geometric/nn/aggr/base.py
```diff
@@ -0,0 +1,40 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
+import torch
+from torch import Tensor
+from torch_scatter import scatter, segment_csr
+
+
+class BaseAggr(torch.nn.Module, ABC):
```
cthoyt, line 9: I would also just call this `Aggr` instead of `BaseAggr`.

Even further, making the class name descriptive as `Aggregation` or `Aggregator` would make it way easier to read. Not like we have a limit on screen space :)

torch_geometric/nn/aggr/base.py
rusty1s: done :)

torch_geometric/nn/aggr/__init__.py
cthoyt: See https://github.com/pyg-team/pytorch_geometric/pull/4716

PR #4689, opened 2022-05-20
Lightning Trainer integration in GraphGym

aniketmaurya:
Third PR from [the Roadmap](https://github.com/pyg-team/pytorch_geometric/issues/4503).

* Implements PL Trainer
* Soft import for optional dependencies
* Update tests


graphgym/main.py
```diff
@@ -41,20 +41,21 @@
...
         # Start training
         if cfg.train.mode == 'standard':
-            train(loggers, loaders, model, optimizer, scheduler)
+            train(model, datamodule, logger=True)
         else:
+            loaders = create_loader()
+            loggers = create_logger()
+            optimizer = create_optimizer(model.parameters(), cfg.optim)
+            scheduler = create_scheduler(optimizer, cfg.optim)
             train_dict[cfg.train.mode](loggers, loaders, model, optimizer,
```
aniketmaurya, line 59: do we need this else block? What is the other `train.mode`?

torch_geometric/graphgym/utils/agg_runs.py
```diff
@@ -86,8 +86,8 @@ def agg_runs(dir, metric_best='auto'):
         validation performance. Options: auto, accuracy, auc.
 
     '''
-    results = {'train': None, 'val': None, 'test': None}
-    results_best = {'train': None, 'val': None, 'test': None}
+    results = {'train': None, 'val': None}
```
aniketmaurya, line 89: test loop will not run during the training.

graphgym/main.py
Padarn: seems like its just 'standard' or "not" 'standard'

torch_geometric/graphgym/model_builder.py
```diff
@@ -1,22 +1,15 @@
...
 from torch_geometric.graphgym.loss import compute_loss
 from torch_geometric.graphgym.models.gnn import GNN
 from torch_geometric.graphgym.optim import create_optimizer, create_scheduler
 from torch_geometric.graphgym.register import network_dict, register_network
 
-try:
-    from pytorch_lightning import LightningModule
-except ImportError:
-    LightningModule = torch.nn.Module
-    warnings.warn("Please install 'pytorch_lightning' for using the GraphGym "
```
Padarn, line 17: It seems from below that this won't work with `pytorch_lightning` not installed? In this case does it make more sense to just throw the exception here?

graphgym/main.py
rusty1s: I think this can be safely dropped.

test/graphgym/test_graphgym.py
```diff
@@ -85,11 +81,8 @@ def test_run_single_graphgym(auto_resume, skip_train_eval, use_trivial_metric):
         cfg.metric_best = 'auto'
         cfg.custom_metrics = []
 
-    loaders = create_loader()
-    assert len(loaders) == 3
-
-    loggers = create_logger()
-    assert len(loggers) == 3
+    datamodule = GraphGymDataModule()
+    assert len(datamodule._loaders) == 3
```
rusty1s, line 85:
```suggestion
    assert len(datamodule.loaders) == 3
```
Worth it to expose?

torch_geometric/graphgym/model_builder.py
rusty1s: I thought about this as well. Might be worth reconsidering the decision of gracefully warning the user whenever the user imports `torch_geometric.graphgym`. We could just error out in that case. We can fix in a follow-up PR if we are all aligned.

torch_geometric/graphgym/train.py
```diff
@@ -1,95 +1,47 @@
...
-                                cfg.train.epoch_resume)
-    if start_epoch == cfg.optim.max_epoch:
-        logging.info('Checkpoint found, Task already done')
-    else:
-        logging.info('Start from epoch {}'.format(start_epoch))
+    trainer_config = trainer_config or {}
+    trainer = pl.Trainer(**trainer_config,
+                         enable_checkpointing=cfg.train.enable_ckpt,
+                         callbacks=callbacks, default_root_dir=cfg.out_dir,
+                         max_epochs=cfg.optim.max_epoch)
```
rusty1s, line 44:
```suggestion
                         max_epochs=cfg.optim.max_epoch,)
```

torch_geometric/graphgym/utils/agg_runs.py
rusty1s: Thanks!

rusty1s (approved):
This is amazingly clean. Thank you so much! I love it :)

PR #4731, opened 2022-05-27
`LSTMAggregation`

rusty1s:


torch_geometric/nn/aggr/lstm.py
```diff
@@ -0,0 +1,48 @@
...
+        self.lstm = LSTM(in_channels, out_channels, batch_first=True, **kwargs)
+
+    def reset_parameters(self):
+        self.lstm.reset_parameters()
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+
+        assert index is not None  # TODO
```
Padarn, line 40: sorry I didn't understand the reason for this, 
```
index = torch.zeros(x.size(0))
```
and 
```
index = None
```
will have the same result no?

Padarn: After working with the class not I see that this is because we don't yet support `ptr` - Should add a more informative exception here?

Padarn, line 40:
```suggestion
        assert index is not None, raise ValueError("`ptr` is not yet supported")
```
or similar?

rusty1s: Good point. I added more descriptive error messages.

Padarn (approved):

torch_geometric/nn/aggr/lstm.py
```diff
@@ -0,0 +1,57 @@
...
+        if x.dim() != 2:
+            raise ValueError(f"'{self.__class__.__name__}' requires "
+                             f"two-dimensional inputs (got '{x.dim()}')")
+
+        if dim not in [-2, 0]:
+            raise ValueError(f"'{self.__class__.__name__}' needs to perform "
+                             f"aggregation in first dimension (got '{dim}')")
+
+        x, _ = to_dense_batch(x, index, batch_size=dim_size)
+        return self.lstm(x)[0][:, -1]
```
lightaime, line 53: Sorry, I am not familiar with LSTM. Some comments: I don't know whether using the output of the last time step is a good idea or not. 1. many padded zeros and fed into LSTM for nodes with smaller degrees. 2. the forgetting issue for long sequences.

How about something like this to remove the effects of padded zeros and average over all the time steps (not sure how it performs):
```
x, mask = to_dense_batch(x, index, batch_size=dim_size)
return self.reduce(self.lstm(x)[0][mask], index, ptr, dim_size, dim, reduce='mean')
```

rusty1s: I think this is a valid point. AFAIK, `GraphSAGE` (and any other LSTM-style aggregation procedure, e.g., Jumping Knowledge), just read out the embeddings after the last element of the sequence has been processed, so I opted to go for this solution. It might be very interesting to explore which reduction performs better here.

lightaime: Sure, it is worth trying it out later.

PR #4749, opened 2022-06-01
`MultiAggregation` and `aggregation_resolver`

lightaime:
- Add `MulAggregation`, `MultiAggregation` and `aggregation_resolver`


torch_geometric/nn/conv/message_passing.py
```diff
@@ -99,8 +101,12 @@ def __init__(self, aggr: Optional[Union[str, List[str]]] = "add",
             assert len(set(aggr) | AGGRS) == len(AGGRS)
             self.aggr: Optional[str] = None
             self.aggrs: List[str] = aggr
+        elif isinstance(aggr, Aggregation):
```
rusty1s, line 104: I think most of the `aggrs` logic can be dropped here since we now simply can utilize `MultiAggregation` for that.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -99,8 +101,12 @@ def __init__(self, aggr: Optional[Union[str, List[str]]] = "add",
...
+            self.aggr: Optional[Aggregation] = aggr
```
rusty1s, line 105: Use `aggr_resolver` to map `str` to `Aggregation`?

torch_geometric/nn/conv/message_passing.py
```diff
@@ -480,12 +486,16 @@ def aggregate(self, inputs: Tensor, index: Tensor,
...
-        if ptr is not None:
-            ptr = expand_left(ptr, dim=self.node_dim, dims=inputs.dim())
-            return segment_csr(inputs, ptr, reduce=aggr)
+        if isinstance(self.aggr, Aggregation):
+            return aggr(inputs, index=index, ptr=ptr, dim_size=dim_size,
+                        dim=self.node_dim)
         else:
-            return scatter(inputs, index, dim=self.node_dim, dim_size=dim_size,
-                           reduce=aggr)
+            if ptr is not None:
```
rusty1s, line 493: I don't think we need this conditional here. The goal is to always simply call `return self.aggr(...)`.

rusty1s (commented):
Thank you. This mostly looks good to me, and I left a few comments.

Notably, this PR groups a bunch of separate functionality (`MulAggregation`, `MultiAggregation`, `MessagePassing` integration, ...) into a single PR, which makes it overly complex. Next time, I think we should work on separating these functionalities into separate PRs to ease reviewing.

torch_geometric/nn/conv/message_passing.py
lightaime: Do we want to resolve all the `str` in all the cases to `Aggregation`? It seems the `str` is used when `edge_index` is `SparseTensor` and used in `__repr__`. For some convs that override the `aggregate ` will also raise errors if `self.aggr` is not `str`. Should we keep both `str` and `Aggregation` as attributes and use them accordingly?

torch_geometric/nn/aggr/multi.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Any, List, Optional, Union
+
+import torch
+from torch import Tensor
+
+from . import basic, lstm
+from .base import Aggregation
+
+
+def normalize_string(s: str) -> str:
```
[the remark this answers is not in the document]
lightaime: Do we want to move the `aggregation_resolver` to `nn/resolver.py`? It raises circular import error since I need to import `nn.aggr` to `nn/resolver.py` and import `aggregation_resolver` to `nn.aggr.multi`. Show we make a python class for the resolver so that we don’t need to import the package to `nn/resolver.py` and can instantiate it for different packages?

torch_geometric/nn/conv/message_passing.py
rusty1s: That's a good point. Let's name it `aggr_module` or something like that.

torch_geometric/nn/aggr/multi.py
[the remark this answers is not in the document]
rusty1s: You can import `nn.aggr` inside the `aggregation_resolver()` which should fix this issue.

torch_geometric/nn/conv/message_passing.py
lightaime: Ok. That should work.

lightaime:
Thanks for the helpful comments @rusty1s! The PR is ready for another review.

test/nn/aggr/test_mp_interface.py
```diff
@@ -0,0 +1,122 @@
...
+    assert out2.size() == (4, 16)
+    assert conv2(x, adj.t()).tolist() == out2.tolist()
+    conv2.fuse = False
+    assert conv2(x, adj.t()).tolist() == out2.tolist()
+    conv2.fuse = True
+
+    assert torch.allclose(out1, out2)
+
+
+@pytest.mark.parametrize('Aggregation',
```
Padarn, line 71:
```suggestion
@pytest.mark.parametrize('aggregation',
```

torch_geometric/nn/aggr/basic.py
```diff
@@ -36,6 +37,18 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
         return self.reduce(x, index, ptr, dim_size, dim, reduce='min')
 
 
+class MulAggregation(Aggregation):
+    def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+        if ptr is not None:  # TODO
+            import warnings
+            warnings.warn(f"'{self.__class__.__name__}' with "
```
Padarn, line 46: I'd prefer raising a `NotImplementedError` here

Padarn (commented):
Looks good to me so far. Happy to review again once no longer WIP.

torch_geometric/nn/aggr/basic.py
lightaime: Thanks, @Padarn. Sounds fair. Will change it.

rusty1s:
Thanks @lightaime. In favor of smaller PRs, I removed the `MessagePassing` integration such that this PR solely adds support for `MulAggregation`, `MultiAggregation` and `aggregation_resolver`. Can you please re-send your modifications in `message_passing.py` and `test_message_passing.py` in a separate PR? Here is the last commit: https://github.com/pyg-team/pytorch_geometric/tree/2b1ac91569c4ea4de2b603e883e6e831cc2c0c0e

PR #4750, opened 2022-06-01
Sampling according to `max_sample` within `AddMetaPaths`

RexYing:
This functionality is used in cases where certain metapath edge types result in very dense adjacency.
If `max_sample` is set, in expectation this number of samples will be sampled per node for each edge index.


torch_geometric/transforms/add_metapaths.py
```diff
@@ -100,7 +100,7 @@ def __init__(self, metapaths: List[List[EdgeType]],
         self.keep_same_node_type = keep_same_node_type
         self.drop_unconnected_nodes = drop_unconnected_nodes
 
-    def __call__(self, data: HeteroData) -> HeteroData:
+    def __call__(self, data: HeteroData, max_sample: int = None) -> HeteroData:
```
rusty1s, line 103: Move to an `__init__` argument.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -110,15 +110,25 @@ def __call__(self, data: HeteroData) -> HeteroData:
...
             adj1 = SparseTensor.from_edge_index(
-                edge_index=data[edge_type].edge_index,
-                sparse_sizes=data[edge_type].size())
+                edge_index=edge_index, sparse_sizes=data[edge_type].size())
 
             for i, edge_type in enumerate(metapath[1:]):
+                print('ADD METAPATH LOOP ========== ', i, edge_type)
+                edge_index = data[edge_type].edge_index
+                print(edge_index.size(), '============== full')
+                if max_sample is not None:
```
rusty1s, line 124:
```suggestion
                if self.max_sample is not None:
```

torch_geometric/transforms/add_metapaths.py
```diff
@@ -145,3 +155,20 @@ def __call__(self, data: HeteroData) -> HeteroData:
                     del data[node]
 
         return data
+
+    def _edge_index_sampling(self, edge_index, max_sample, backward=False):
```
rusty1s, line 159:
```suggestion
    def _edge_index_sampling(self, adj, max_sample, backward=False):
```
Let's work on the `SparseTensor` here.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -145,3 +155,20 @@ def __call__(self, data: HeteroData) -> HeteroData:
...
+        if backward:
+            node_idx = edge_index[1]
+        else:
+            node_idx = edge_index[0]
+        _, counts = torch.unique(node_idx, sorted=False, return_counts=True)
+        sample_prob = torch.repeat_interleave(1 / counts * max_sample, counts)
+        draw = torch.rand(sample_prob.size(), dtype=float)
+        mask = sample_prob > draw
+        return edge_index[:, mask]
```
rusty1s, line 168:
```suggestion
        return adj.masked_select_nnz(mask, layout='coo')
```

rusty1s (commented):
Thanks @RexYing. This mostly looks good. I think `edge_index_sampling` should leverage `SparseTensor` to make this a little bit faster and to ensure that we are always operating on sorted inputs.

torch_geometric/transforms/add_metapaths.py
RexYing: do you think there might be use cases where i use the same AddMetaPath, but want different edge_index with different samples? It might not be a frequent use case, but I just thought it might be more flexible than specifying in __init__.

torch_geometric/transforms/add_metapaths.py
RexYing: done

torch_geometric/transforms/add_metapaths.py
```diff
@@ -145,3 +155,20 @@ def __call__(self, data: HeteroData) -> HeteroData:
...
+
+    def edge_index_sampling(self, edge_index, max_sample):
+        edge_index = self._edge_index_sampling(edge_index, max_sample)
+        # sample the reverse direction: restrict the destination degree
+        # to max_sample in expectation
+        return self._edge_index_sampling(edge_index, max_sample, backward=True)
```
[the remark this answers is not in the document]
RexYing: Since it's a bipartite edge_index, I sampled twice. The first pass ensures that source degree is upper-bounded; the second pass ensures that dst degree is upper-bounded.
It will be the same as limiting both the adj row and column sums instead of just rows.

rusty1s (approved):

rusty1s:
I think I fixed the runtime issues. One reason for this was that it is quite expensive to sample in both directions. I don't think this is really needed, and I moved to a solution where we only sample `adj1` after performing the matrix multiplication. Merging this now but happy to extend/improve/make faster in later PRs.

PR #4762, opened 2022-06-03
`nn.aggr.Set2Set`

rusty1s:


torch_geometric/nn/aggr/base.py
```diff
@@ -58,8 +48,43 @@ def __call__(self, x: Tensor, index: Optional[Tensor] = None, *,
             raise ValueError(f"Encountered mismatch between 'dim_size' (got "
                              f"'{dim_size}') and 'ptr' (got '{ptr.size(0)}')")
 
+        if index is None and ptr is None:
```
Padarn, line 51: This is much cleaner than handling index possibly being `None` everywhere 👍

torch_geometric/nn/aggr/base.py
```diff
@@ -58,8 +48,43 @@ def __call__(self, x: Tensor, index: Optional[Tensor] = None, *,
...
+            index = x.new_zeros(x.size(dim), dtype=torch.long)
+
+        if dim_size is None and ptr is not None:
+            dim_size = ptr.numel() - 1
+        elif dim_size is None:
+            dim_size = int(index.max()) + 1 if index.numel() > 0 else 0
+
```
Padarn, line 58:
```suggestion
elif index is not None:
    assert dim_size >= index.numel() + 1
else: 
   assert dim_size >= ptr.numel() -1
```
what do you think?

torch_geometric/nn/aggr/base.py
```diff
@@ -71,8 +96,15 @@ def reduce(self, x: Tensor, index: Optional[Tensor] = None,
...
         return scatter(x, index, dim=dim, dim_size=dim_size, reduce=reduce)
 
-    def __repr__(self) -> str:
-        return f'{self.__class__.__name__}()'
+    def to_dense_batch(self, x: Tensor, index: Optional[Tensor] = None,
+                       ptr: Optional[Tensor] = None,
+                       dim_size: Optional[int] = None,
+                       dim: int = -2) -> Tuple[Tensor, Tensor]:
+
+        self.assert_index_present(index)  # TODO
```
Padarn, line 104:
```suggestion
        self.assert_index_present(index)  # TODO: support non-sorted index
```
Unclear what the TODO was for me exactly (not sure I'm right here, suggest a better message if mine is not accurate)

torch_geometric/nn/aggr/set2set.py
```diff
@@ -0,0 +1,66 @@
...
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        self.lstm.reset_parameters()
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+
+        self.assert_index_present(index)  # TODO
```
Padarn, line 47: Maybe we should comment in the `assert_index_present` to make it clear we want to eventually want `ptr` to be supported everywhere?

torch_geometric/nn/aggr/base.py
rusty1s: done :)

torch_geometric/nn/aggr/base.py
rusty1s: done.

torch_geometric/nn/aggr/set2set.py
rusty1s: done.

Padarn (approved):

PR #4779, opened 2022-06-08
Integration of `nn.aggr` within `MessagPassing`

lightaime:
Support new aggregation modules with `MessagePassing`.


torch_geometric/nn/aggr/basic.py
```diff
@@ -22,7 +22,8 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
         return self.reduce(x, index, ptr, dim_size, dim, reduce='sum')
 
 
-AddAggregation = SumAggregation  # Alias
+AddAggregation = type('AddAggregation', SumAggregation.__bases__,
```
rusty1s, line 25: Mmh, I see the problem. Let us instead add `"add"` support directly in `aggregation_resolver`. This should be cleaner.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -85,22 +91,35 @@ class MessagePassing(torch.nn.Module):
         'size_i', 'size_j', 'ptr', 'index', 'dim_size'
     }
 
-    def __init__(self, aggr: Optional[Union[str, List[str]]] = "add",
-                 flow: str = "source_to_target", node_dim: int = -2,
-                 decomposed_layers: int = 1):
+    def __init__(
+        self,
+        aggr: Optional[Union[str, List[str], Aggregation]] = "add",
```
rusty1s, line 96: We would need some additional `**aggr_kwargs` passed to the `aggregation_resolver`.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -288,7 +307,7 @@ def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
 
         # Run "fused" message and aggregation (if applicable).
         if (isinstance(edge_index, SparseTensor) and self.fuse
-                and not self.explain and len(self.aggrs) == 0):
+                and not self.explain and (self.aggr in SPMM_AGGRS)):
```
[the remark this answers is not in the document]
lightaime: Not all the `aggrs` support `SparseTensor`. If `message_and_aggregate` is specified, some aggrs such as `std` may throw value errors. Do we want to support message passing with these aggrs when using `SparseTensor`?

torch_geometric/nn/conv/message_passing.py
```diff
@@ -497,13 +507,6 @@ def message_and_aggregate(self, adj_t: SparseTensor) -> Tensor:
         """
         raise NotImplementedError
 
-    def combine(self, inputs: List[Tensor]) -> Tensor:
```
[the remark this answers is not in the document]
lightaime: Sounds good. Added it back. But the multi aggrs logic is removed. Not sure if we used `combine` anywhere.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -85,22 +87,25 @@ class MessagePassing(torch.nn.Module):
...
+        decomposed_layers: int = 1,
+    ):
 
         super().__init__()
 
-        if aggr is None or isinstance(aggr, str):
-            assert aggr is None or aggr in AGGRS
-            self.aggr: Optional[str] = aggr
-            self.aggrs: List[str] = []
+        if aggr is None or isinstance(aggr, (str, Aggregation)):
```
lightaime, line 100: TODO: Make SPMM aggrs such as 'sum' and SumAggregation behave the same

torch_geometric/nn/aggr/basic.py
```diff
@@ -22,7 +22,9 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None, *,
         return self.reduce(x, index, ptr, dim_size, dim, reduce='sum')
 
 
-AddAggregation = SumAggregation  # Alias
+class AddAggregation(SumAggregation):
```
Padarn, line 25: Do we need `AddAggregation`?

torch_geometric/nn/aggr/multi.py
```diff
@@ -19,7 +20,18 @@ def __init__(self, aggrs: List[Union[Aggregation, str]]):
             raise ValueError(f"'aggrs' of '{self.__class__.__name__}' should "
                              f"not be empty")
 
-        self.aggrs = [aggregation_resolver(aggr) for aggr in aggrs]
+        if aggrs_kwargs is None:
+            aggrs_kwargs = [{}] * len(aggrs)
+        elif len(aggrs) != len(aggrs_kwargs):
+            raise ValueError(f"'aggrs_kwargs' with invalid length passed to "
```
Padarn, line 26: This error seems like something pretty generic. Do you think there is any value in having a util to generate such an error?

torch_geometric/nn/conv/message_passing.py
Padarn: Curious what the gap is today?

torch_geometric/nn/conv/message_passing.py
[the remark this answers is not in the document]
Padarn: Maybe it makes sense to raise the error within the `aggr` itself? Means one less place to update when and if we update the support.

torch_geometric/nn/conv/message_passing.py
```diff
@@ -466,26 +469,20 @@ def explain_message(self, inputs: Tensor, size_i: int) -> Tensor:
         return inputs * edge_mask.view(size)
 
     def aggregate(self, inputs: Tensor, index: Tensor,
-                  ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
-                  aggr: Optional[str] = None) -> Tensor:
```
Padarn, line 470: any chance we'll be breaking code by doing this? Seems like a class people may already be building ontop of?

torch_geometric/nn/conv/message_passing.py
```diff
@@ -466,26 +469,20 @@ def explain_message(self, inputs: Tensor, size_i: int) -> Tensor:
...
         """
-        aggr = self.aggr if aggr is None else aggr
-        assert aggr is not None
-        if ptr is not None:
-            ptr = expand_left(ptr, dim=self.node_dim, dims=inputs.dim())
-            return segment_csr(inputs, ptr, reduce=aggr)
-        else:
-            return scatter(inputs, index, dim=self.node_dim, dim_size=dim_size,
-                           reduce=aggr)
+        return self.aggr_module(inputs, index=index, ptr=ptr,
```
Padarn, line 484: So much cleaner 💌

torch_geometric/nn/aggr/multi.py
rusty1s: Fixed.

torch_geometric/nn/conv/message_passing.py
rusty1s: `aggr` was just introduced in master and now we are removing it. I think we are safe.

torch_geometric/nn/aggr/basic.py
rusty1s: Removed it.

torch_geometric/nn/conv/message_passing.py
rusty1s: Yes, indeed :)

rusty1s (approved):

rusty1s:
Also fixed jit issues:
* `*` was not allowed as argument in `forward`
* `MultiAggregation` needs to wrap its child aggregation inside `torch.nn.ModuleList`

torch_geometric/nn/conv/message_passing.py
lightaime: The gap is that `self.aggr = str(SumAggregation())` is not the same as 'self.aggr = sum'. Therefore, `self.fuse` is `False` even through `SumAggregation()` should be able to be fused. 
https://github.com/pyg-team/pytorch_geometric/blob/0f7e018aff9f9ca5356bf90b27e3546e26343c37/torch_geometric/nn/conv/message_passing.py#L158

Ways to fix: 
1. Modify the `__repr__` of the fused `Aggregation` functions to return str like `sum`, `mean`, `max`, ...
2. Add their `__repr__` str to `FUSE_AGGRS`. 

Both if them seem to be tricky. Any suggestion?

rusty1s: Yes, it is tricky. I'm not sure we desperately need to fix this right now, but one option would be to maintain a `repr` variable inside each `Aggregation` module for the reverse mapping. One could even automate that potentially (like a reverse `aggregation_resolver`)

rusty1s: Added this to the roadmap.

PR #4805, opened 2022-06-14
Adding Typing to JumpingKnowledge and Torch Script Support

will-leeson:
When attempting to serialize a model with a JumpingKnowledge layer using `torch.jit.script`, I would receive an error due to the lack of typing. In order to type them, I removed the ability to pass a tuple of tensors in, since Torch Script seems to only allow tuples of a set size (I could be wrong, please correct me if possible)

Torch also complained that `self.lstm` and `self.att` didn't exist, so I moved them out of the if statement in `__init__`.  Once again, if there is a way to allow this, please correct me.

### DIFF
--- a/torch_geometric/nn/models/jumping_knowledge.py
+++ b/torch_geometric/nn/models/jumping_knowledge.py
@@ -65,7 +65,7 @@ class JumpingKnowledge(torch.nn.Module):
         r"""Aggregates representations across different layers.
 
         Args:
-            xs (list or tuple): List containing layer-wise representations.
+            xs (list): List containing layer-wise representations.
         """
 
         assert isinstance(xs, list)
--- a/torch_geometric/nn/models/jumping_knowledge.py
+++ b/torch_geometric/nn/models/jumping_knowledge.py
@@ -1,4 +1,7 @@
+from typing import List
+
 import torch
+from torch import Tensor
 from torch.nn import LSTM, Linear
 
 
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/models/jumping_knowledge.py
```diff
@@ -40,13 +43,14 @@ def __init__(self, mode, channels=None, num_layers=None):
         super().__init__()
         self.mode = mode.lower()
         assert self.mode in ['cat', 'max', 'lstm']
+        # Must declare these here in order to use torch.jit.script
+        self.lstm = LSTM(channels, (num_layers * channels) // 2,
```
rusty1s, line 47: Why is that? We are doing something similar in [SAGEConv](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/sage_conv.py#L99) and it is TorchScript compatible. I think the key is to have some additional variable that denotes if this module should be applied.

rusty1s (approved):
Thanks!

torch_geometric/nn/models/jumping_knowledge.py
will-leeson: When it's declared inside of the condtional I get the following error:
`Module 'JumpingKnowledge' has no attribute 'lstm'`
I've changed:
`if mode == 'lstm':`
to
`if self.mode == 'lstm':`
but that doesn't seem to fix it.

will-leeson: Ahh, yes. This makes sense. Thanks!

PR #4807, opened 2022-06-15
Let `Data` and `HeteroData` implement `FeatureStore`

mananshah99:
This PR lets `Data` and `HeteroData` implement the feature store abstraction. In particular, it defines `put_tensor`, `get_tensor`, and `remove_tensor` methods on both classes, and adds basic tests for these functionalities.

### DIFF
--- a/torch_geometric/data/batch.py
+++ b/torch_geometric/data/batch.py
@@ -23,8 +23,12 @@ class DynamicInheritance(type):
             new_cls = base_cls
         else:
             name = f'{base_cls.__name__}{cls.__name__}'
+
+            class MetaResolver(type(cls), type(base_cls)):
+                pass
+
             if name not in globals():
-                globals()[name] = type(name, (cls, base_cls), {})
+                globals()[name] = MetaResolver(name, (cls, base_cls), {})
             new_cls = globals()[name]
 
         params = list(inspect.signature(base_cls.__init__).parameters.items())
--- a/torch_geometric/data/batch.py
+++ b/torch_geometric/data/batch.py
@@ -24,6 +24,10 @@ class DynamicInheritance(type):
         else:
             name = f'{base_cls.__name__}{cls.__name__}'
 
+            # NOTE `MetaResolver` is necessary to resolve metaclass conflict
+            # problems between `DynamicInheritance` and the metaclass of
+            # `base_cls`. In particular, it creates a new common metaclass
+            # from the defined metaclasses.
             class MetaResolver(type(cls), type(base_cls)):
                 pass
 
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -739,11 +739,7 @@ class Data(BaseData, FeatureStore):
             # we set `index`. So, we assume here that indexing by `None` is
             # equivalent to not indexing at all, which is not in line with
             # Python semantics.
-            if attr.index is None:
-                return tensor
-
-            dim = self.__cat_dim__(attr.attr_name, tensor)
-            return torch.index_select(tensor, attr.index, dim=dim)
+            return tensor[attr.index] if attr.index is not None else tensor
         return None
 
     def _remove_tensor(self, attr: TensorAttr) -> bool:
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -605,7 +605,7 @@ class Data(BaseData, FeatureStore):
         data = HeteroData()
 
         for i, key in enumerate(node_type_names):
-            for attr, value in self.items():
+            for attr, value in self._store.items():
                 if attr == 'node_type' or attr == 'edge_type':
                     continue
                 elif isinstance(value, Tensor) and self.is_node_attr(attr):
@@ -616,7 +616,7 @@ class Data(BaseData, FeatureStore):
 
         for i, key in enumerate(edge_type_names):
             src, _, dst = key
-            for attr, value in self.items():
+            for attr, value in self._store.items():
                 if attr == 'node_type' or attr == 'edge_type':
                     continue
                 elif attr == 'edge_index':
@@ -629,7 +629,7 @@ class Data(BaseData, FeatureStore):
 
         # Add global attributes.
         keys = set(data.keys) | {'node_type', 'edge_type', 'num_nodes'}
-        for attr, value in self.items():
+        for attr, value in self._store.items():
             if attr in keys:
                 continue
             if len(data.node_stores) == 1:
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -1,5 +1,6 @@
 import copy
 from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
 from typing import (
     Any,
     Callable,
@@ -17,6 +18,12 @@ import torch
 from torch import Tensor
 from torch_sparse import SparseTensor
 
+from torch_geometric.data.feature_store import (
+    FeatureStore,
+    FeatureTensorType,
+    TensorAttr,
+    _field_status,
+)
 from torch_geometric.data.storage import (
     BaseStorage,
     EdgeStorage,
@@ -300,7 +307,16 @@ class BaseData(object):
 ###############################################################################
 
 
-class Data(BaseData):
+@dataclass
+class DataTensorAttr(TensorAttr):
+    r"""Attribute class for `Data`, which does not require a `group_name`."""
+    def __init__(self, attr_name=_field_status.UNSET,
+                 index=_field_status.UNSET):
+        # Treat group_name as optional, and move it to the end
+        super().__init__(None, attr_name, index)
+
+
+class Data(BaseData, FeatureStore):
     r"""A data object describing a homogeneous graph.
     The data object can hold node-level, link-level and graph-level attributes.
     In general, :class:`~torch_geometric.data.Data` tries to mimic the
@@ -364,6 +380,10 @@ class Data(BaseData):
 
         for key, value in kwargs.items():
             setattr(self, key, value)
+
+        # `Data` does not support group_nae, so we need to adjust `TensorAttr`
+        # accordingly here to avoid requiring `group_name` to be set:
+        FeatureStore.__init__(self, attr_cls=DataTensorAttr)
 
     def __getattr__(self, key: str) -> Any:
         if '_store' not in self.__dict__:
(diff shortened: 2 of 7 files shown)


torch_geometric/data/feature_store.py
```diff
@@ -245,7 +245,7 @@ def __init__(self, attr_cls: Any = TensorAttr):
         attributes by subclassing :class:`TensorAttr` and passing the subclass
         as :obj:`attr_cls`."""
         super().__init__()
-        self._attr_cls = attr_cls
+        self.__dict__['_attr_cls'] = attr_cls
```
yaoyaowd, line 248: why ?

torch_geometric/data/hetero_data.py
```diff
@@ -616,6 +625,48 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
...
+            if val is not None:
+                val[attr.index] = tensor
+        else:
+            # No group name, just store tensor in node storage:
+            setattr(self[attr.group_name], attr.attr_name, tensor)
+        return True
+
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Obtains a feature tensor from node storage."""
+        # Retrieve tensor and index accordingly:
```
yaoyaowd, line 648: In general, should we get `attr.index` first and then `getattr`. The reason I'm asking because it is usually preferred to filter as much data as possible and I think select by index filters more data.

torch_geometric/data/data.py
```diff
@@ -692,6 +712,44 @@ def num_faces(self) -> Optional[int]:
...
+            out[attr.index] = tensor
+        else:
+            # No attr nane, just store tensor:
+            setattr(self, attr.attr_name, tensor)
+        return True
+
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Obtains a feature tensor from node storage."""
+        # Retrieve tensor and index accordingly:
+        tensor = getattr(self, attr.attr_name)
```
yaoyaowd, line 735: ditto.

torch_geometric/data/data.py
```diff
@@ -365,6 +381,10 @@ def __init__(self, x: OptTensor = None, edge_index: OptTensor = None,
         for key, value in kwargs.items():
             setattr(self, key, value)
 
+        # `Data` does not support group_nae, so we need to adjust `TensorAttr`
```
rusty1s, line 384:
```suggestion
        # `Data` does not support group_name, so we need to adjust `TensorAttr`
```

torch_geometric/data/hetero_data.py
```diff
@@ -616,6 +625,48 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
 
         return data
 
+    # :obj:`FeatureStore` interface ###########################################
+
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
+        r"""Stores a feature tensor in node storage."""
+        if not attr.is_set('index'):
+            attr.index = None
```
rusty1s, line 633: Why do we do need to set `attr.index = None` here?

rusty1s (commented):
I think this looks pretty great. Thank you! Let me know if you need further help for the `DynamicInheritance` issue in `Batch`.

On a different note: It looks like `Data` and `FeatureStore` implement, e.g., `__getitem__`. Currently, it looks like we do not really take advantage of the `View` stuff in `FeatureStore` as a result. Might be good to add this as a follow up TODO. WDYT?

torch_geometric/data/data.py
```diff
@@ -692,6 +712,44 @@ def num_faces(self) -> Optional[int]:
...
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Obtains a feature tensor from node storage."""
+        # Retrieve tensor and index accordingly:
+        tensor = getattr(self, attr.attr_name)
+        if tensor is not None:
+            # TODO this behavior is a bit odd, since TensorAttr requires that
+            # we set `index`. So, we assume here that indexing by `None` is
+            # equivalent to not indexing at all, which is not in line with
+            # Python semantics.
+            return tensor[attr.index] if attr.index is not None else tensor
```
rusty1s, line 741: One important thing: I think we need to make use of `cat_dim` here to know which in which index to select. See [here](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/loader/utils.py#L103-L104) for an example:
```python
cat_dim = self.__cat_dim__(attr.attr_name, tensor)
```

torch_geometric/data/data.py
```diff
@@ -365,6 +381,10 @@ def __init__(self, x: OptTensor = None, edge_index: OptTensor = None,
         for key, value in kwargs.items():
             setattr(self, key, value)
 
+        # `Data` does not support group_nae, so we need to adjust `TensorAttr`
+        # accordingly here to avoid requiring `group_name` to be set:
+        FeatureStore.__init__(self, attr_cls=DataTensorAttr)
```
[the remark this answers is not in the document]
mananshah99: Or we could just take advantage of MRO and call `super(Data, self).__init__(attr_cls)`; ended up implementing that way as I think it's cleaner. lmkwyt :)

torch_geometric/data/data.py
```diff
@@ -692,6 +712,44 @@ def num_faces(self) -> Optional[int]:
...
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
+        r"""Stores a feature tensor in node storage."""
+
+        if not attr.is_set('index'):
+            attr.index = None
+
+        out = getattr(self, attr.attr_name, None)
+        if out is not None:
+            # Attr name exists, handle index:
+            out[attr.index] = tensor
```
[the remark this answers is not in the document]
mananshah99: Good catch :)

torch_geometric/data/data.py
mananshah99: I imagined that it would be on the user to provide the correct slice for this; for example,
```
(*([slice(None)] * (cat_dim - 1 if cat_dim > 0 else cat_dim)), attr.index)
```
but I'm happy to implement it this way as well.

torch_geometric/data/hetero_data.py
mananshah99: Not sure I understand. In this implementation, it doesn't matter since all data is in memory, so we first need to get the tensor and then index that tensor. In an out-of-core implementation, all of `TensorAttr` is passed to the implementor, and they can filter however they like.

torch_geometric/data/data.py
rusty1s: I see. That makes sense.

torch_geometric/data/hetero_data.py
```diff
@@ -616,6 +623,50 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
 
         return data
 
+    # :obj:`FeatureStore` interface ###########################################
+
+    def _put_tensor(self, tensor: FeatureTensorType, attr: TensorAttr) -> bool:
```
yaoyaowd, line 628: Do we really want to support this `put_tensor` operation now? It looks like there are a lot of places could go wrong. What if index is out of bound (put new tensors)? What if `out` does not exist and `attr.index` is not continuous? What if `out` exists but `val` does not? I'd rather throw a not implemented error here.

yaoyaowd (approved):
Minor comment, LGTM.

torch_geometric/data/hetero_data.py
```diff
@@ -616,6 +623,50 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
...
+        else:
+            # No node storage found, just store tensor in new one:
+            setattr(self[attr.group_name], attr.attr_name, tensor)
+        return True
+
+    def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
+        r"""Obtains a feature tensor from node storage."""
+        # Retrieve tensor and index accordingly:
+        tensor = getattr(self[attr.group_name], attr.attr_name)
+        if tensor is not None:
```
Padarn, line 648: If it requires we set `index` why do we even need to check?

Padarn: Maybe another way to solve this if its confusing is to have a different value for `UNSET` which indicates its going to index 'all'

torch_geometric/data/batch.py
```diff
@@ -23,8 +23,12 @@ def __call__(cls, *args, **kwargs):
             new_cls = base_cls
         else:
             name = f'{base_cls.__name__}{cls.__name__}'
+
+            class MetaResolver(type(cls), type(base_cls)):
```
rusty1s, line 27: Can we add a short description to why this is necessary?

torch_geometric/data/data.py
```diff
@@ -348,7 +364,10 @@ class Data(BaseData):
     def __init__(self, x: OptTensor = None, edge_index: OptTensor = None,
                  edge_attr: OptTensor = None, y: OptTensor = None,
                  pos: OptTensor = None, **kwargs):
-        super().__init__()
+        # `Data` doesn't support group_name, so we need to adjust `TensorAttr`
+        # accordingly here to avoid requiring `group_name` to be set:
+        super(Data, self).__init__(attr_cls=DataTensorAttr)
```
rusty1s, line 369:
```suggestion
        super().__init__(attr_cls=DataTensorAttr)
```
should work as well?

torch_geometric/data/data.py
```diff
@@ -692,6 +714,45 @@ def num_faces(self) -> Optional[int]:
             return self.face.size(self.__cat_dim__('face', self.face))
         return None
 
+    # FeatureStore interface ###########################################
+
+    def items(self):
+        return self._store.items()
```
rusty1s, line 720: Add type hint + doc-string. Why is this necessary for implementing the `FeatureStore` interface?

torch_geometric/data/hetero_data.py
```diff
@@ -92,6 +97,8 @@ class HeteroData(BaseData):
     DEFAULT_REL = 'to'
 
     def __init__(self, _mapping: Optional[Dict[str, Any]] = None, **kwargs):
+        super(HeteroData, self).__init__()
```
rusty1s, line 100:
```suggestion
        super().__init__()
```

rusty1s (approved):
I think this looks great. Thank you! One thing that would be good to confirm is whether pickling of `data` objects between PyG 2.0.4 and PyG master still works, e.g., running an example in PyG 2.0.4 and re-run with pre-processed dataset in master. Otherwise, we have to list this as a breaking change.

torch_geometric/data/hetero_data.py
mananshah99: I think we should support `put_tensor` since all of the cases you mentioned (and others) are either covered by the function or error out in the same way that PyG would normally error out. In particular,

> What if index is out of bound (put new tensors)?

Not sure what you mean here. If we are putting a new tensor, then no node storage would be found and we would just store the tensor as intended. If you mean an index out of bounds exception for replacing a tensor, PyG would fail here as well (as expected)

> What if out does not exist and attr.index is not continuous? 

This is fine, we support any index that torch supports

> What if out exists but val does not?

Good point! Fixed.

torch_geometric/data/data.py
mananshah99: I spent far too long on this; it's because `Data` is now a `MutableMapping` (inherited from `FeatureStore`), which defines an `items` method, so we now have to define `items` explicitly. I'll add a docstring to this effect; the end result is that the behavior remains the same.

torch_geometric/data/hetero_data.py
mananshah99: Thanks for the comment! Not sure what you mean by needing to check `index`; this is because `TensorAttr` just requires that its attributes are set (they can be set to None). The current way to index all would be `None`, which I think is acceptable; although I'm happy to define a custom value for the `UNSET` enum to indicate `all` indexing in a follow-up PR.

PR #4816, opened 2022-06-17
`GraphStore` definition + `Data` and `HeteroData` integration

mananshah99:
This PR defines a very lightweight `MaterializedGraph` that supports basic put and get operations of edge indices. It additionally lets `Data` and `HeteroData` inherit from and implement this abstraction.


torch_geometric/data/hetero_data.py
```diff
@@ -667,6 +674,24 @@ def __len__(self) -> int:
     def __iter__(self):
         raise NotImplementedError
 
+    # MaterializedGraph interface #############################################
+
+    def _layout_to_attr_name(self, layout: EdgeLayout) -> str:
```
Padarn, line 679: can we make this part of `MaterializedGraph`

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+    CSR = 'CSR'
+    LIL = 'LIL'
+
+
+@dataclass
+class EdgeAttr(CastMixin):
+    r"""Defines the attributes of an :obj:`MaterializedGraph` edge."""
+
+    # The layout of the edge representation
+    layout: Optional[EdgeLayout] = None
```
Padarn, line 25: does it make sense for this to be optional? if not provided is it a valid `EdgeAttr`?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class MaterializedGraph(MutableMapping):
+    def __init__(self, edge_attr_cls: Any = EdgeAttr):
+        r"""Initializes the materialized graph. Implementor classes can
+        customize the ordering and required nature of their :class:`EdgeAttr`
+        edge attributes by subclassing :class:`EdgeAttr` and passing the
+        subclass as :obj:`attr_cls`."""
+        super().__init__()
+        self.__dict__['_edge_attr_cls'] = edge_attr_cls
```
Padarn, line 38: Sorry naive question, why not just 
```
self._edge_attr_cls = edge_attr_cls
```

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Core ####################################################################
+
+    @abstractmethod
+    def _put_edge_index(self, edge_index: EdgeTensorType,
+                        edge_attr: EdgeAttr) -> bool:
+        return None
```
Padarn, line 45:
```suggestion
        raise NotImplementedErrror()
```

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+            attr.
+            **attr(EdgeAttr): the edge attributes.
+        """
+        edge_attr = self._edge_attr_cls.cast(*args, **kwargs)
+        if not edge_attr.layout:
+            raise AttributeError(
+                "An edge layout is required to store an edge index, but one "
+                "was not provided.")
+
+        # NOTE implementations should take care to ensure that `SparseTensor`
```
Padarn, line 62: What does that mean in this case?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Notes:
+            Most sampling methods operate efficiently on sorted edge indices.
+            This class makes no guarantees on the formats of the edge indices
+            that are passed as input. Implementations should account for this
+            (e.g. by converting all edge indices to CSC) before sampling.
+        """
+        # NOTE for the future: this is expected to call a C++ implementation
+        # in pyg-lib.
+        return None
```
Padarn, line 106:
```suggestion
       raise NotImplementedError()
```
Does this make more sense?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Python built-ins ########################################################
+
+    def __setitem__(self, key: EdgeAttr, value: EdgeTensorType):
+        key = self._edge_attr_cls.cast(key)
+        self.put_edge_index(value, key)
+
+    def __getitem__(self, key: EdgeAttr) -> EdgeTensorType:
+        key = self._edge_attr_cls.cast(key)
+        return self.get_edge_index(key)
```
Padarn, line 116: maybe worth adding some tests to cover there - but not neccessary

torch_geometric/data/materialized_graph.py
wsad1: Yeah thats not needed. A child class has to override methods with `@abstractmethod` decorator.

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        edge_attr = self._edge_attr_cls.cast(*args, **kwargs)
+        if not edge_attr.layout:
+            raise AttributeError(
+                "An edge layout is required to store an edge index, but one "
+                "was not provided.")
+
+        # NOTE implementations should take care to ensure that `SparseTensor`
+        # objects are treated properly here.
+        return self._put_edge_index(edge_index,
+                                    self._edge_attr_cls.cast(*args, **kwargs))
```
wsad1, line 65:
```suggestion
                                    edge_attr)
```

torch_geometric/data/materialized_graph.py
wsad1: +1. With this as non-optional you can avoid checking `edge_attr` in `put_edge_index`.

test/data/test_data.py
```diff
@@ -264,3 +266,25 @@ def test_basic_feature_store():
     assert 'x' in data.__dict__['_store']
     data.remove_tensor(attr_name='x', index=None)
     assert 'x' not in data.__dict__['_store']
+
+
+# Materialized Graph ##########################################################
```
rusty1s, line 271: I personally find the naming of `MaterializedGraph` somewhat confusing - it is not clear why we do not just simply name it `Graph` or `GraphStore`.

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+class EdgeLayout(Enum):
+    COO = 'COO'
+    CSC = 'CSC'
+    CSR = 'CSR'
+    LIL = 'LIL'
+
+
+@dataclass
+class EdgeAttr(CastMixin):
```
rusty1s, line 21: This term is now a bit overloaded.

torch_geometric/data/materialized_graph.py
rusty1s, line 45:
```suggestion
        pass
```

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        self.__dict__['_edge_attr_cls'] = edge_attr_cls
+
+    # Core ####################################################################
+
+    @abstractmethod
+    def _put_edge_index(self, edge_index: EdgeTensorType,
+                        edge_attr: EdgeAttr) -> bool:
+        return None
+
+    def put_edge_index(self, edge_index: EdgeTensorType, *args,
```
rusty1s, line 47: Can we support `layout="coo"` as well?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+
+@dataclass
+class EdgeAttr(CastMixin):
+    r"""Defines the attributes of an :obj:`MaterializedGraph` edge."""
+
+    # The layout of the edge representation
+    layout: Optional[EdgeLayout] = None
+
+    # The type of the edge
+    edge_type: Optional[Any] = None
```
rusty1s, line 28: would be great to also add `is_sorted: bool = False`?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        customize the ordering and required nature of their :class:`EdgeAttr`
+        edge attributes by subclassing :class:`EdgeAttr` and passing the
+        subclass as :obj:`attr_cls`."""
+        super().__init__()
+        self.__dict__['_edge_attr_cls'] = edge_attr_cls
+
+    # Core ####################################################################
+
+    @abstractmethod
+    def _put_edge_index(self, edge_index: EdgeTensorType,
```
rusty1s, line 43: I think we should drop `SparseTensor` support here and instead support `Tuple[Tensor, Tensor]`, e.g., `(row, col)`, `(rowptr, col)`, `(colptr, row)`.

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        Args:
+            **attr(EdgeAttr): the edge attributes.
+
+        Returns:
+            EdgeTensorType: an edge_index tensor corresonding to the provided
+            attributes, or None if there is no such tensor.
+        """
+        return self._get_edge_index(self._edge_attr_cls.cast(*args, **kwargs))
+
+    def sample(
```
rusty1s, line 84: I think we should add this in a later step once we have clarity on how to frame the the `pyg-lib` `graph` interface.

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        # in pyg-lib.
+        return None
+
+    # Python built-ins ########################################################
+
+    def __setitem__(self, key: EdgeAttr, value: EdgeTensorType):
+        key = self._edge_attr_cls.cast(key)
+        self.put_edge_index(value, key)
+
+    def __getitem__(self, key: EdgeAttr) -> EdgeTensorType:
```
rusty1s, line 114: How do these methods relate to `BaseData` and `FeatureStore`?

torch_geometric/data/materialized_graph.py
```diff
@@ -0,0 +1,125 @@
...
+        # NOTE implementations should take care to ensure that `SparseTensor`
+        # objects are treated properly here.
+        return self._put_edge_index(edge_index,
+                                    self._edge_attr_cls.cast(*args, **kwargs))
+
+    @abstractmethod
+    def _get_edge_index(self, edge_attr: EdgeAttr) -> EdgeTensorType:
+        return None
+
+    def get_edge_index(self, *args, **kwargs) -> EdgeTensorType:
```
rusty1s, line 71: Shouldn't we expose additional methods such as `coo()`, `csr()` and `csc()`?

torch_geometric/data/hetero_data.py
mananshah99: I don't think we want this, since `_layout_to_attr_name` is specific to PyG (and will likely not generalize to any other backing feature stores).

torch_geometric/data/materialized_graph.py
mananshah99: For the same reason as in `feature_store`; we don't want to call the `__setattr__` method of `Data` here, so we have to set it in the class `__dict__` explicitly.

torch_geometric/data/materialized_graph.py
mananshah99: In what manner? I think it aligns well with `TensorAttr` in `FeatureStore` (they are both attributes of objects we are trying to add to the respective stores).

torch_geometric/data/materialized_graph.py
mananshah99: The reason behind marking this as optional earlier was that, in an earlier design, `layout` was not required for `get_edge_index`. But this has since changed, and you are correct that it no longer makes sense to have `layout` be optional. Updating to reflect this.

torch_geometric/data/materialized_graph.py
mananshah99: Yep, we currently support this via Enum casting. Will make this more clear.

torch_geometric/data/materialized_graph.py
mananshah99: That's fair. I currently included it just as a reference marker to indicate where we'd want to implement a method in the future, but I'm also happy to remove it since we won't have a corresponding implementation today. No strong preference, lmkwyt!

torch_geometric/data/materialized_graph.py
mananshah99: They are overridden by `BaseData` and its implementations. I think that both `FeatureStore` and `MaterializedGraph` should define these dunder methods internally as both a feature store and materialized graph can be used independently of one another, but implementations that combine the two will need to intelligently override them or deal with some nasty bugs.

torch_geometric/data/materialized_graph.py
mananshah99: Yeah, I think this is a good idea. We can have the default implementation read the edge index from the materialized graph, convert in memory, and push the edge index out to the materialized graph, and we can have custom materialized graphs override this. Will add.

torch_geometric/data/materialized_graph.py
mananshah99: This is a great suggestion and is indeed much cleaner, thanks! Will implement.

torch_geometric/data/graph_store.py
```diff
@@ -0,0 +1,126 @@
...
+    r"""Defines the attributes of an :obj:`GraphStore` edge."""
+
+    # The type of the edge
+    edge_type: Optional[Any]
+
+    # The layout of the edge representation
+    layout: EdgeLayout
+
+    # Whether the edge index is sorted
+    is_sorted: bool = False
```
rusty1s, line 30: Can we define what that means? Ideally by destination node. Also mention that it can only be set for `COO` format?

torch_geometric/data/graph_store.py
```diff
@@ -0,0 +1,126 @@
...
+    CSC = 'csc'
+    CSR = 'csr'
+
+
+@dataclass
+class EdgeAttr(CastMixin):
+    r"""Defines the attributes of an :obj:`GraphStore` edge."""
+
+    # The type of the edge
+    edge_type: Optional[Any]
```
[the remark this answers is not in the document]
mananshah99: I'm biased against doing this since we don't need views just yet in `GraphStore`, and every use of `EdgeAttr` requires `edge_type` and `layout` to be set. I think we can avoid this complexity for the time being, and introduce it when necessary.

torch_geometric/data/graph_store.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # The layout of the edge representation
+    layout: EdgeLayout
+
+    # Whether the edge index is sorted
+    is_sorted: bool = False
+
+
+class GraphStore(MutableMapping):
```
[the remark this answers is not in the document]
mananshah99: I think one can conceptualize of a `GraphStore` as a mapping in the same way we can conceptualize of a `FeatureStore` as a mapping, but I agree that these additional methods aren't totally necessary and can be removed. If we find the need to inherit in the future, we can do so, but removing for now.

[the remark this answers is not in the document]
mananshah99: See above comment on views.

torch_geometric/data/graph_store.py
```diff
@@ -0,0 +1,126 @@
...
+                       **kwargs) -> bool:
+        r"""Synchronously adds an edge_index tensor to the graph store.
+
+        Args:
+            tensor(EdgeTensorType): an edge_index in a format specified in
+            attr.
+            **attr(EdgeAttr): the edge attributes.
+        """
+        edge_attr = self._edge_attr_cls.cast(*args, **kwargs)
+        edge_attr.layout = EdgeLayout(edge_attr.layout)
```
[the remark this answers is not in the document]
mananshah99: Yep, only need to cast in the case of `str`, but it also doesn't matter so leaving this here for simplicity.

torch_geometric/data/data.py
```diff
@@ -755,10 +777,75 @@ def _remove_tensor(self, attr: TensorAttr) -> bool:
...
+    r"""Converts an EdgeTensorType tensor tuple to a PyG Adj tensor."""
+    if attr.layout == EdgeLayout.COO:
+        # COO: 2 x n
+        return torch.stack(tensor_tuple)
+    elif attr.layout == EdgeLayout.CSR:
+        # CSR: (rowptr, col)
+        return SparseTensor(rowptr=tensor_tuple[0], col=tensor_tuple[1],
+                            is_sorted=attr.is_sorted)
+    elif attr.layout == EdgeLayout.CSC:
+        # CSC: (colptr, row) this is a transposed adjacency matrix, so rowptr
```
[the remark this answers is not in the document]
mananshah99: I was conflicted about this as well :) I think we should stick with `(row, col)` always (as you said, scipy CSC is `indices`, `indptr`, which aligns with this), and so I'll make the adjustment to `(row, colptr)` for CSC.

torch_geometric/data/materialized_graph.py
mananshah99: Sounds good. Removed for now :)

torch_geometric/data/materialized_graph.py
mananshah99: Ah, right. I think it's probably okay to leave this overloaded term for now (mostly since I can't think of a better one), and if you have any ideas we can do a quick refactor.

Padarn:
Sorry I was really busy and didn't read your updates earlier @mananshah99 - looks awesome and thanks for your responses.

PR #4817, opened 2022-06-17
Let `NeighborLoader` accept `Tuple[FeatureStore, GraphStore]`

mananshah99:
This PR lets `NeighborLoader` respect `Tuple[FeatureStore, GraphStore]` so users can separate their features from their graph and work directly through the `FeatureStore` and `GraphStore` APIs. In particular, it makes the following changes:

* Adds `get_tensor_size` and `get_all_tensor_attrs` to `FeatureStore` and corresponding implementations in `Data` and `HeteroData`
* Adds `get_all_edge_attrs` to `GraphStore` and corresponding implementations in `Data` and `HeteroData`
* Modifies `NeighborSampler` to accept `Tuple[FeatureStore, GraphStore]` and utilizes the APIs of these two classes to store the CSC representation of the graph in `NeighborSampler`'s class state
* Modifies `NeighborLoader` to accept `Tuple[FeatureStore, GraphStore]` and utilizes the APIs of these two classes to produce a `HeteroData` object that can be ingested downstream.

It also leaves scattered TODOs for better support of, e.g., `GraphStore` implementations with no edge type. These are deferred for the time being but will be implemented in follow-up PRs.


torch_geometric/data/feature_store.py
```diff
@@ -370,7 +370,17 @@ def update_tensor(self, tensor: FeatureTensorType, *args,
         self.remove_tensor(attr)
         return self.put_tensor(tensor, attr)
 
-    # :obj:`AttrView` methods #################################################
+    # Additional methods ######################################################
+
+    @abstractmethod
+    def _get_tensor_size(self, attr: TensorAttr) -> List[int]:
```
rusty1s, line 376: Expose `size()` and/or `shape` in `AttrView`?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -19,7 +25,8 @@
 class NeighborSampler:
     def __init__(
         self,
-        data: Union[Data, HeteroData],
+        data: Union[Union[Data, HeteroData], Tuple[FeatureStore,
```
rusty1s, line 28: `data: Union[Data, HeteroData, Tuple[FeatureStore, MaterializedGraph]]`

test/data/test_materialized_graph.py
```diff
@@ -24,6 +32,9 @@ def _put_edge_index(self, edge_index: EdgeTensorType,
     def _get_edge_index(self, edge_attr: EdgeAttr) -> EdgeTensorType:
         return self.store.get(MyMaterializedGraph.key(edge_attr), None)
 
+    def get_all_edge_types(self):
```
rusty1s, line 35: I think it makes sense to define the same for `FeatureStore`.

test/data/test_data.py
```diff
@@ -261,6 +262,14 @@ def test_basic_feature_store():
     out = data.get_tensor(attr_name='x', index=None)
     assert torch.equal(x, out)
 
+    # Get tensor size:
+    assert data.get_tensor_size(attr_name='x') == [20, 20]
```
rusty1s, line 266: nit: Can we return a `Tuple` to be inline with PyTorch?

test/data/test_feature_store.py
```diff
@@ -51,6 +50,17 @@ def _remove_tensor(self, attr: TensorAttr) -> bool:
         del self.store[MyFeatureStore.key(attr)]
         return True
 
+    def _get_tensor_size(self, attr: TensorAttr):
+        attr.index = None
```
rusty1s, line 54: Why do we need to set it to `None`. If `index` is set, I would assume that we would return the `size` of the filtered data.

test/data/test_feature_store.py
```diff
@@ -51,6 +50,17 @@ def _remove_tensor(self, attr: TensorAttr) -> bool:
...
+        return self._get_tensor(attr).size()
+
+    def get_all_tensor_attrs(self):
```
rusty1s, line 57:
```suggestion
    def get_all_tensor_attrs(self) -> List[str]:
```

test/data/test_graph_store.py
```diff
@@ -42,7 +58,7 @@ def assert_equal_tensor_tuple(expected, actual):
     # to confirm that `GraphStore` works as intended.
     coo = adj.coo()[:-1]
     csr = adj.csr()[:-1]
-    csc = adj.csc()[:-1]
+    csc = list(adj.csc()[:-1])[::-1]  # (row, colptr)
```
rusty1s, line 61:
```suggestion
    csc = adj.csc()[:-1:-1]  # (row, colptr)
```

torch_geometric/data/hetero_data.py
```diff
@@ -673,6 +674,20 @@ def _remove_tensor(self, attr: TensorAttr) -> bool:
             return True
         return False
 
+    def _get_tensor_size(self, attr: TensorAttr) -> List[int]:
+        if not attr.is_set('index'):
+            attr.index = None
+        tensor = self._get_tensor(attr)
+        return list(tensor.size()) if tensor is not None else []
```
rusty1s, line 681: Let's error out here in case the `attr` does not exist. I think `_get_tensor` should error out as well.

torch_geometric/data/hetero_data.py
```diff
@@ -673,6 +674,20 @@ def _remove_tensor(self, attr: TensorAttr) -> bool:
             return True
         return False
 
+    def _get_tensor_size(self, attr: TensorAttr) -> List[int]:
+        if not attr.is_set('index'):
```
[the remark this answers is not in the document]
mananshah99: This is a different case: if the index is not set, I think it's reasonable to get the size of the whole tensor (which is what `None` does here).

torch_geometric/loader/neighbor_loader.py
```diff
@@ -74,14 +90,96 @@ def __init__(
...
+            # Obtain CSC representation of graph for in-memory sampling:
+            # TODO this code will be replaced with a `GraphStore.sample` call
+            # when sampling routines are factored out to work with pyg-lib and
+            # GraphStore
+            edge_type_to_layouts: Dict[Any,
+                                       List[EdgeLayout]] = defaultdict(list)
+            for attr in edge_attrs:
+                edge_type_to_layouts[attr.edge_type].append(attr.layout)
+
+            self.colptr_dict, self.row_dict, self.perm_dict = {}, {}, {}
```
rusty1s, line 138: I think this part is too complex. For now, let us just assume that `EdgeLayout.CSC` is present. In a later PR, we can add automatic conversion of formats inside `GraphStore`.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -329,28 +436,58 @@ def __repr__(self) -> str:
...
+    else:  # Tuple[FeatureStore, GraphStore]
+        # NOTE FeatureStore and GraphStore are treated as separate
+        # entities, so we cannot leverage the custom structure in Data and
+        # HeteroData to infer the number of nodes. As a result, here we expect
+        # that the input nodes are either explicitly provided or can be
+        # directly inferred from the feature store.
+        feature_store, _ = data
+
+        # Explicit tensor:
+        if isinstance(input_nodes, Tensor):
```
rusty1s, line 480: Only if type `bool`? I feel like `from_bool_tensor` should do the `torch.bool` check, and otherwise just return the tensor.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -329,28 +436,58 @@ def __repr__(self) -> str:
...
+            return None, from_bool_tensor(input_nodes)
+
+        if isinstance(input_nodes, tuple) and isinstance(
+                input_nodes[0], str) and isinstance(input_nodes[1], Tensor):
```
rusty1s, line 484: Again.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -329,28 +436,58 @@ def __repr__(self) -> str:
...
-    node_type, input_nodes = input_nodes
-    if input_nodes.dtype == torch.bool:
-        input_nodes = input_nodes.nonzero(as_tuple=False).view(-1)
-    return node_type, input_nodes
+        node_type, input_nodes = input_nodes
+        if input_nodes.dtype == torch.bool:
+            input_nodes = from_bool_tensor(input_nodes)
+        return node_type, input_nodes
+
+    else:  # Tuple[FeatureStore, GraphStore]
```
rusty1s, line 471: This needs to support the same cases as `HeteroData` branch.

rusty1s (approved):
Thanks! It looks like I broke some tests though.

torch_geometric/data/feature_store.py
```diff
@@ -298,10 +298,11 @@ def get_tensor(self, *args, **kwargs) -> Optional[FeatureTensorType]:
         Returns:
             FeatureTensorType, optional: a Tensor of the same type as the
             index, or :obj:`None` if no tensor was found.
+
+        Raises:
+            KeyError: if the tensor corresponding to attr was not found.
```
ZenoTan, line 303: Also `ValueError` if not fully specified?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -329,28 +436,58 @@ def __repr__(self) -> str:
...
+
+        if isinstance(input_nodes, tuple) and isinstance(
+                input_nodes[0], str) and isinstance(input_nodes[1], Tensor):
+            return input_nodes[0], from_bool_tensor(input_nodes[1])
+
+        # Implicit from TensorAttr (infer number of nodes from feature tensor):
+        assert isinstance(input_nodes, TensorAttr)
+        assert input_nodes.is_set('attr_name')
+        return getattr(input_nodes, 'group_name', None), range(
+            feature_store.get_tensor_size(input_nodes)[0])
```
ZenoTan, line 491: Are we calling `get_tensor_size` but only using the first dim?

torch_geometric/loader/utils.py
```diff
@@ -176,3 +184,34 @@ def filter_hetero_data(
...
+    nodes in `node` end edges in `edge` for each node and edge type,
+    respectively."""
+
+    # Construct a new `HeteroData` object:
+    data = HeteroData()
+
+    # Filter edge storage:
+    for key in edge_dict:
+        edge_index = torch.stack([row_dict[key], col_dict[key]], dim=0)
+        data[str_to_edge_type(key)].edge_index = edge_index
```
ZenoTan, line 206: Could we move this out of this func because `feature_store` is not involved?

torch_geometric/loader/neighbor_loader.py
mananshah99: Yes, this is a heuristic for the number of nodes

ZenoTan:
A more general question is, if the user wants to replace only the feature store part of an existing `HeteroData` but keep edge index logic in it, are we happy to support this, or we force it to replace both feature store and graph store?

mananshah99:
@ZenoTan `HeteroData` is a `GraphStore`, so this should work as intended.

ZenoTan (approved):

ZenoTan:
Thanks for making the integration with external data storage easier!

torch_geometric/loader/neighbor_loader.py
mananshah99: Throwing an exception for now to clearly indicate that only CSC and CSR are supported, since I think the added flexibility of working with CSR isn't too complicated (and we can potentially extend this to work with COO in the short term as well). But I agree that in the long term, this will be deprecated in favor of `GraphStore.sample`.

torch_geometric/loader/neighbor_loader.py
mananshah99: I think we do:
* `input_nodes` is a `Tensor` or `[str, Tensor]` tuple = supported exactly
* `input_nodes` is a `str` = `input_nodes` is a `TensorAttr` with a `group_name` set (we also require `attr_name` to be set here for num node inferral, but for now I don't see how to get around that)

If there are others, we can include them in a follow-up PR (don't want to overcomplicate this one).

rusty1s: Oh yeah, seemed like I misunderstood since we pass in `TensorAttr` it can only hold `bool` tensors. Correct?

PR #4837, opened 2022-06-21
Add `batch` and `ptr` vectors for a list of tensors and nested dicts

jan-meissner:
### Feature: 

Adds batching for a list of tensors and nested dicts. Motivation comes from higher order GNNs such as Simplicial Neural Networks or CW complexes, where one often needs to batch over an unknown amount of tensors. Also added support for calculating the batch vector for nested dicts.

### Example:
Batching over a dataset with data objects of the form `Data(xs=[tensor1, tensor2, tensor3])` and using `follow_batch = ['xs']` in the loader now yields `DataBatch(xs = [3], xs_batch = [3], xs_ptr = [3])` where `xs_batch` is the list of batches `[tensor1_batch, tensor2_batch, tensor3_batch]`. Analogous for `ptr`. 

### Additional Issue fixed:
Resolved a issue concerning `separate` (and therefore `InMemoryDataset`) when batching over lists of tensors. Essentially `InMemoryDataset` fails to recover the `data_list` correctly. Here is a code snippet to reproduce the issue:

```python
class MyTestDataset2(InMemoryDataset):
    def __init__(self, data_list):
        super().__init__('/tmp/MyTestDataset2')
        self.data, self.slices = self.collate(data_list)


def test_lists_of_tensors_in_memory_dataset():
    def tr(n, m):
        return torch.rand((n, m))

    d1 = Data(xs=[tr(4, 3), tr(11, 4), tr(1, 2)])
    d2 = Data(xs=[tr(5, 3), tr(14, 4), tr(3, 2)])
    d3 = Data(xs=[tr(6, 3), tr(15, 4), tr(2, 2)])
    d4 = Data(xs=[tr(4, 3), tr(16, 4), tr(1, 2)])

    data_list = [d1, d2, d3, d4]

    dataset = MyTestDataset2(data_list)
    assert isinstance(dataset[0].xs, list)
    assert len(dataset) == 4
    assert dataset[0].xs[1].shape == (11, 4)
    assert dataset[0].xs[2].shape == (1, 2)
    assert dataset[1].xs[0].shape == (5, 3)
    assert dataset[2].xs[1].shape == (15, 4)
    assert dataset[3].xs[1].shape == (16, 4)

### CODE REVIEW

**Comment by janmeissnerRWTH** [comment]
CHANGELOG.md is still missing the link to the pull request.

**Comment by jan-meissner** [comment]
Converted to Draft. I actually don't fully understand `separate`, no longer confident in changes made concerning it.

File: torch_geometric/data/in_memory_dataset.py [hunk 1]
```diff
@@ -130,10 +130,13 @@ def copy(self, idx: Optional[IndexType] = None) -> 'InMemoryDataset':
...
-    for key, value in mapping.items():
-        if isinstance(value, Mapping):
+def nested_iter(node: Union[Mapping, Sequence]) -> Iterable:
+    if isinstance(node, Mapping):
+        for key, value in node.items():
             for inner_key, inner_value in nested_iter(value):
                 yield inner_key, inner_value
-        else:
-            yield key, value
+    if isinstance(node, Sequence):
```
rusty1s, line 138:
```suggestion
    elif isinstance(node, Sequence):
```

torch_geometric/data/separate.py
```diff
@@ -90,6 +90,11 @@ def _separate(
           and not isinstance(value[0], str) and len(value[0]) > 0
           and isinstance(value[0][0], (Tensor, SparseTensor))):
         # Recursively separate elements of lists of lists.
+        return [elem[idx] for i, elem in enumerate(value)]
```
rusty1s, line 93:
```suggestion
        return [elem[idx] for elem in value]
```

torch_geometric/data/collate.py
```diff
@@ -199,6 +196,52 @@ def _collate(
         return values, slices, None
 
 
+def _batch(
```
rusty1s, line 199:
```suggestion
def _get_batch_info(...) -> Tuple:
```

torch_geometric/data/collate.py
```diff
@@ -96,12 +96,9 @@ def collate(
...
 
             # Add an additional batch vector for the given attribute:
-            if (attr in follow_batch and isinstance(slices, Tensor)
-                    and slices.dim() == 1):
-                repeats = slices[1:] - slices[:-1]
-                batch = repeat_interleave(repeats.tolist(), device=device)
-                out_store[f'{attr}_batch'] = batch
-                out_store[f'{attr}_ptr'] = cumsum(repeats.to(device))
+            if attr in follow_batch:
+                out_store[f'{attr}_batch'] = _batch(slices, device)
```
rusty1s, line 100: Can we unify the computation here into a single function? For example, `repeats` is calculated twice otherwise.

rusty1s (commented):
Thank you! This looks pretty clean. If possible, I would like to move computation of `batch` and `ptr` into a single function though.

rusty1s:
Please also add the change to the `CHANGELOG.md`.

torch_geometric/data/collate.py
jan-meissner: Done, see `_batch_and_ptr`. I favored readable code instead of unreadable zip magic.

jan-meissner:
Haven't done a full test so far.

jan-meissner:
> 



> @jan-meissner This looks super good to me. Let me know once this is ready to merge :)

@rusty1s I'm done with it so it should be ready for merge. Well, other than the small `CHANGELOG.md` conflict.

Issue #4791, opened 2022-06-11
Sub-batch of a batch


ArchieGertsman:
### 🚀 The feature, motivation and pitch

I would like to be able to index a batch to obtain a sub-batch, rather than a list of data objects. E.g. if I have a `Batch` object `my_batch`, then something like `my_batch.subbatch(np.arange(5))` would return another `Batch` object containing the first 5 graphs in `my_batch`.

### Alternatives

Of course, `Batch.from_data_list(my_batch[:5])` would achieve this functionality, but this seems slow. I would like to directly construct a sub-batch, if possible.

### Additional context

In my particular application, I have a list of `Data` objects, and I need to include an increasing number of them in a `Batch` every time I pass them into my GNN. I could pass `Batch.from_data_list(my_list[:i])` into my GNN for each iteration `i`, but I presume it would be faster if I could just construct a batch of this whole list at the start, `complete_batch`, and then access sub-batches of that complete batch, passing `complete_batch.subbatch(np.arange(i))` at each iteration `i`.

rusty1s:
Interesting. I agree that one could re-use `data.subgraph()` for this, e.g.:
```python
mask = [True, False, True]  # torch.Tensor that contains which graph in the batch to maintain

mask = mask[data.batch]
data = data.subgraph(mask)
```

ArchieGertsman:
There is one issue with the `subgraph` approach: it doesn't update the properties of the batch, e.g. `data.num_graphs` and `data.batch`.

rusty1s:
Yes, that is correct and would need to be handled afterwards:
```python
assoc = torch.empty(num_examples, dtype=torch.long)
assoc[mask] = torch.arange(mask.sum())
data.batch = assoc[data.batch]
```
Happy to accept a PR on this one :)

josiahbjorgaard:
Also interested in this feature

josiahbjorgaard:
Is anyone working on this? If not I can take it. The approach above using data.batch and subgraph seem to not work for HeteroData. Any thoughts on modifications of the approach above to 'subbatch' HeteroData Batches?

rusty1s:
I don't think anyone is working on this right now, so please feel free to give it a shot!

ArchieGertsman:
[...]

PR #4845, opened 2022-06-22
Added `subbatch` function to `Batch`

ArchieGertsman:
Following up on[ issue #4791](https://github.com/pyg-team/pytorch_geometric/issues/4791)


josiahbjorgaard:
I think that this also needs tests added[ to the batch tests](https://github.com/pyg-team/pytorch_geometric/blob/210ec4bde3dfdb9d587c0caddf2cef92bc557430/test/data/test_batch.py), right?

torch_geometric/data/batch.py
```diff
@@ -183,3 +184,65 @@ def num_graphs(self) -> int:
     def __reduce__(self):
         state = self.__dict__.copy()
         return DynamicInheritanceGetter(), self.__class__.__bases__, state
+
+    def subbatch(self, idx):
+        # convert index to mask
+        if isinstance(idx, slice) or \
+           (isinstance(idx, Tensor) and idx.dtype == torch.long) or \
```
Padarn, line 191: I think in most other parts of the [code](https://github.com/pyg-team/pytorch_geometric/blob/210ec4bde3dfdb9d587c0caddf2cef92bc557430/torch_geometric/data/batch.py#L120) I see that these kinds of conditions are usually split (even when matching result). Maybe we do the same here?

torch_geometric/data/batch.py
```diff
@@ -183,3 +184,65 @@ def num_graphs(self) -> int:
...
+           (isinstance(idx, np.ndarray) and idx.dtype == np.int64) or \
+           (isinstance(idx, Sequence) and not isinstance(idx, str)):
+            mask = torch.zeros(self.num_graphs, dtype=torch.bool)
```
Padarn, line 194:
```suggestion
            mask = index_to_mask(index, self.num_graphs)
```
this function is in `torch_geometric.util`

ArchieGertsman: `index_to_mask` only accepts a `Tensor`, while here we must consider indexes of other types. Maybe `index_to_mask` should consider other types as well?

Padarn: Good point. I agree it could accept other types

PR #4847, opened 2022-06-22
Add a `normalize` parameter to `dense_diff_pool`

Hu-Chuxuan:
I did this modification because:

While I was using this function, I found the link_loss being exceptionally small and thus almost not contributing anything, making the graph structure hard to preserve and all nodes resulting in a few clusters. This is because adj.numel() are generally very large. After I recovered the result by multiplying adj.numel() back, results became reasonable.

According to the original paper, the link_loss is not divided by the size of adjacency matrix
image
image

After communicating with people on slack, I think it makes more sense for us to allow users to have an option of not being normalized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- Added a `normalize` parameter to `dense_diff_pool` ([#4847](https://github.com/pyg-team/pytorch_geometric/pull/4847))
 - Added documentation to the `DataLoaderIterator` class ([#4838](https://github.com/pyg-team/pytorch_geometric/pull/4838))
 - Added `GraphStore` support to `Data` and `HeteroData` ([#4816](https://github.com/pyg-team/pytorch_geometric/pull/4816))
 - Added `FeatureStore` support to `Data` and `HeteroData` ([#4807](https://github.com/pyg-team/pytorch_geometric/pull/4807))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `nn.glob.GlobalPooling` module with support for multiple aggregations ([#4582](https://github.com/pyg-team/pytorch_geometric/pull/4582))
 - Added support for graph-level outputs in `to_hetero` ([#4582](https://github.com/pyg-team/pytorch_geometric/pull/4582))
 - Added `CHANGELOG.md` ([#4581](https://github.com/pyg-team/pytorch_geometric/pull/4581))
+- Added `normalize` parameter([#4847](https://github.com/pyg-team/pytorch_geometric/pull/4847))
```
wsad1, line 40:
```suggestion
- Added `normalize` parameter to `dense_diff_pool` ([#4847](https://github.com/pyg-team/pytorch_geometric/pull/4847))
```

torch_geometric/nn/dense/diff_pool.py
```diff
@@ -44,6 +44,8 @@ def dense_diff_pool(x, adj, s, mask=None):
         mask (BoolTensor, optional): Mask matrix
             :math:`\mathbf{M} \in {\{ 0, 1 \}}^{B \times N}` indicating
             the valid nodes for each graph. (default: :obj:`None`)
+        normalize (Bool, optional): Normalization indicator
+            :if True, will divide link_loss by size of graph. (default: True)
```
wsad1, line 48:
```suggestion
        normalize (Bool, optional): if set to :obj:`True` the link_loss is divided
            by size of graph. (default: :obj:`True`)
```

torch_geometric/nn/dense/diff_pool.py
```diff
@@ -66,7 +68,8 @@ def dense_diff_pool(x, adj, s, mask=None):
 
     link_loss = adj - torch.matmul(s, s.transpose(1, 2))
     link_loss = torch.norm(link_loss, p=2)
-    link_loss = link_loss / adj.numel()
+    if normalize is True:
+        link_loss = link_loss / adj.numel()
```
wsad1, line 72:
```suggestion
    link_loss = link_loss / adj.numel() if normalize else link_loss
```

wsad1 (changes requested):
looks good to me.

rusty1s (approved):

PR #4861, opened 2022-06-25
Fix small bug in `NeighbourLoader` when `input_nodes` is `None`

Padarn:
Fix small bug in `NeighbourLoader` when `input_nodes` is `None`: Seems to just be a typo introduced recentky.


Padarn:
@rusty1s I don't know if its meaningful to make a changelog for this - so I've just ignored that error.

rusty1s (approved):
Thanks @Padarn. Feel free to ignore the changelog error.

PR #4862, opened 2022-06-25
Make benchmark arguments more flexible

Padarn:
This is currently WIP - I needed this locally to make it possible to modify these arguments in the benchmarks. If people like the suggested format I will apply this to all benchmarks.

Note because some argument are list and some are lists of lists, you would need to do
```
python loader/neighbor_loader.py --hetero-neighbor_sizes 5 5 --hetero-neighbor_sizes 10 10
```
and 
```
python loader/neighbor_loader.py --batch-sizes 1 10 20
```


rusty1s:
Can you clarify? Wouldn't it be easier to just pass in a list of lists? No strong opinion though :)

Padarn:
> Wouldn't it be easier to just pass in a list of lists? 

Hmm, maybe I misunderstood how it was meant to be used before. Using argparse I couldn't find an obvious was to pass a list of lists, except using the approach given in this MR.

Padarn:
Curious if anyone else has opinions on this? @mszarma you were the github suggested reviewer

mszarma:
Hi @Padarn  and @rusty1s  , using list of list with argparse is not so easy but i believe there is some way to do that, here is some POC for you: https://github.com/Padarn/pytorch_geometric/pull/1
Please let me know if that something that meets your needs.

Padarn:
I think your idea makes sense, let me integrate your change. Thanks!

Padarn:
Thanks for the suggestion @mszarma - have modified this PR

rusty1s (approved):

PR #4871, opened 2022-06-27
Support `FeatureStore` and `GraphStore` in `infer_input_nodes`

yaoyaowd:
Support `Tuple[FeatureStore, GraphStore]` in `LightningNodeData`

### DIFF
--- a/torch_geometric/data/lightning_datamodule.py
+++ b/torch_geometric/data/lightning_datamodule.py
@@ -1,9 +1,11 @@
 import warnings
-from typing import Optional, Union
+from typing import Optional, Tuple, Union
 
 import torch
 
 from torch_geometric.data import Data, Dataset, HeteroData
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
 from torch_geometric.loader.dataloader import DataLoader
 from torch_geometric.loader.neighbor_loader import (
     NeighborLoader,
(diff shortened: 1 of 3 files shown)


torch_geometric/data/lightning_datamodule.py
```diff
@@ -331,7 +333,13 @@ def __repr__(self) -> str:
 ###############################################################################
 
 
-def infer_input_nodes(data: Union[Data, HeteroData], split: str) -> InputNodes:
+def infer_input_nodes(data: Union[Data, HeteroData, Tuple[FeatureStore,
+                                                          GraphStore]],
+                      split: str) -> InputNodes:
+    if isinstance(data, Tuple):
+        # The GraphStore is the HeteroData at the moment.
+        data = data[1]
```
mananshah99, line 341: I am pretty sure this will not be sufficient for general `GraphStore` objects, and will only work for `HeteroData` objects masquerading as `GraphStores. `GraphStore` does not define `__contains__`, so checking whether certain attributes are in `data` is confusing. Furthermore, in the current definition of `GraphStore`, we only store edge indices in the graph store, so masks and indices would logically be placed in the feature store, not the graph store.

In order to fix this properly, I'd suggest obtaining `feature_store, graph_store = data` and checking for the `attr_name`s in `feature_store` with the feature store API calls.

rusty1s:
Closing this for now since it seems to be resolved.

PR #4880, opened 2022-06-28
Remove unnecessary inclusion of self-loops when sampling negative edges

gengala:
Handling self-loops is only useful when negative edges are not provided.


rusty1s (commented):
Thanks for the PR. Meanwhile, `negative_sampling` will not include self-loops in the computation, so we can drop the removal and addition of self-loops here completely. WDYT?

gengala:
You are right :)

rusty1s (approved):

PR #4882, opened 2022-06-28
Fix: `Data` and `HeteroData` feature store `get_all_tensor_attrs`

mananshah99:
Resolves a bug in `Data` and `HeteroData` which results in `get_all_tensor_attrs` not filtering for true tensor attributes.


rusty1s (approved):

rusty1s:
@mananshah99 I just had a second look and this will somehow break PyG compatibility once we make the full switch to `FeatureStore`: Currently, every attribute will just get passed down as a reference to the mini-batch object (without indexing), in case it is neither a node or edge tensor. Let me know what you think!

PR #4883, opened 2022-06-29
`GraphStore`: support `COO` layouts, refactor conversion logic

mananshah99:
This PR enhances `GraphStore`, refactors key logic out of `NeighborLoader`, and improves integration with `Data` and `HeteroData` (in particular, regarding support for `num_nodes`). Precisely, it does the following:
* Implements conversion between any two `EdgeLayout`s in `GraphStore`, and simple interfaces `coo()`, `csr()`, and `csc()` to do so efficiently. This refactor places all conversion logic in `GraphStore`, which is a step towards unifying behind the `GraphStore` interface.
* Adds a `num_nodes` attribute to `EdgeAttr` to allow for proper sparse size definition for all `SparseTensor` integrations, and conversion from `COO` to other layouts in `GraphStore`. Updates `Data` and `HeteroData` implementations to support `num_nodes`.
* Adds tests for the aforementioned functionality, confirming that all conversions behave as intended.


torch_geometric/data/graph_store.py
```diff
@@ -92,15 +116,134 @@ def get_edge_index(self, *args, **kwargs) -> EdgeTensorType:
...
+            # Ignore if requested layout is already present:
+            if to_layout in edge_layouts:
+                from_attr = edge_attrs[edge_layouts.index(to_layout)]
+                row, col = self.get_edge_index(from_attr)
+                perm = None
+
+            # Convert otherwise:
+            else:
+                # Pick the most favorable layout to convert from. We always
+                # prefer CSR or CSC to COO:
```
rusty1s, line 153: There is no real reason to do that IMO. COO -> CSC is actually preferred over CSR -> CSC because you do not have to unwrap the compressed representation.

torch_geometric/data/graph_store.py
```diff
@@ -92,15 +116,134 @@ def get_edge_index(self, *args, **kwargs) -> EdgeTensorType:
...
+                    from_attr = edge_attrs[edge_layouts.index(EdgeLayout.CSR)]
+                else:
+                    from_attr = edge_attrs[edge_layouts.index(EdgeLayout.COO)]
+
+                from_tuple = self.get_edge_index(from_attr)
+
+                # Convert to the new layout:
+                if to_layout == EdgeLayout.COO:
+                    if from_attr.layout == EdgeLayout.CSR:
+                        adj = SparseTensor(
```
rusty1s, line 167: Can we drop `SparseTensor` here? We can just convert indices back to COO, and, use `sort_edge_index`.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -126,62 +121,16 @@ def __init__(
...
-                                  share_memory=share_memory,
-                                  is_sorted=is_sorted)
+            def update_keys(dict_: Dict, update_fn: Callable):
+                original_keys = list(dict_.keys())
+                for key in original_keys:
+                    dict_[update_fn(key)] = dict_.pop(key)
+
+            # Obtain CSC representations for in-memory sampling:
+            self.row_dict, self.colptr_dict, self.perm_dict = graph_store.csc()
+            update_keys(self.row_dict, edge_type_to_str)
```
rusty1s, line 131:
```python
row_dict = {edge_type_to_str(k): v for k, v in row_dict.items()}
```
No real reason for this in-place update I think.

rusty1s (approved):
I think this looks great. Left a few comments.

torch_geometric/data/graph_store.py
```diff
@@ -28,14 +38,22 @@ class EdgeAttr(CastMixin):
...
 
-    def __init__(self, edge_type: Optional[Any], layout: EdgeLayout,
-                 is_sorted: bool = False):
+    # NOTE we define __init__ to force-cast layout
+    def __init__(
+        self,
+        edge_type: Optional[Any],
+        layout: EdgeLayout,
+        is_sorted: bool = False,
+        size: Optional[Tuple[int, int]] = None,
```
ZenoTan, line 51: Would it be better if we use `sparse_sizes` as `SparseTensor` did?

ZenoTan (approved):

torch_geometric/data/graph_store.py
mananshah99: They have the same meaning, but `sparse_sizes` wouldn't make sense for a general tensor.

torch_geometric/data/graph_store.py
```diff
@@ -117,3 +256,106 @@ def __setitem__(self, key: EdgeAttr, value: EdgeTensorType):
...
+    elif layout == EdgeLayout.CSR:
+        return edge_index.csr()[:-1]  # (rowptr, col)
+    else:
+        return edge_index.csr()[-2::-1]  # (row, colptr)
+
+
+###############################################################################
+
+
+def to_csc(
```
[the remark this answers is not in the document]
mananshah99: Note the comment in `utils`, we will deprecate that in time as we consolidate behind `FeatureStore` and `GraphStore`. But for now, the separation is necessary and cleaner.

torch_geometric/data/graph_store.py
mananshah99: Not sure why we would need `sort_edge_index` here?

torch_geometric/data/graph_store.py
```diff
@@ -92,15 +116,134 @@ def get_edge_index(self, *args, **kwargs) -> EdgeTensorType:
...
         if edge_index is None:
             raise KeyError(f"An edge corresponding to '{edge_attr}' was not "
                            f"found")
         return edge_index
 
-    # TODO implement coo(), csc(), csr() methods on GraphStore, which perform
-    # conversions of edge indices between formats. These conversions can also
-    # automatically be performed in `get_edge_index`
+    def _to_layout(self, to_layout: EdgeLayout,
+                   store: bool = False) -> ConversionOutputType:
```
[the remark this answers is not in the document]
mananshah99: Since we don't currently support `remove_edge_index`, this isn't possible in the current PR; we can leave a TODO for it in the future.

torch_geometric/data/graph_store.py
mananshah99: Gotcha, changed - thanks!

PR #4889, opened 2022-06-30
Disjoint sampling

RexYing:
The main goal is to allow disjoint sampling (mostly used for temporal sampling use cases), in order to avoid the same node that are sampled multiple times have its timestamps overriding each other.

This is only a tentative prototype, since I find it somewhat difficult to directly integrate this feature into the c++ sampling routine in torch_sparse. But I'm happy to discuss alternatives and pros / cons and potentially revise it significantly.

I'm also yet to do the collation (mentioned in TODO), but it's not hard in python since we have existing routines. It's more difficult if I were to write this in c++.

One disadvantage of this workaround is that the sampling logic is somewhat exposed to the python side in PyG, which could be confusing. There might also be efficiency concerns, since we now loop it over in Python, rather than in c++.

### DIFF
--- a/torch_geometric/loader/neighbor_loader.py
+++ b/torch_geometric/loader/neighbor_loader.py
@@ -86,6 +86,11 @@ class NeighborSampler:
 
             assert input_type is not None
             self.input_type = input_type
+            if self.node_time_dict is not None:
+                # if temporal sampling is enabled, we force disjoint sampling
+                # to avoid the issue where a node is sampled multiple times
+                # and its associated timestamps are overridden each time.
+                self.disjoint = True
 
         # If we are working with a `Tuple[FeatureStore, GraphStore]` object,
         # obtain edges from GraphStore and convert them to CSC if necessary,
@@ -200,8 +205,15 @@ class NeighborSampler:
 
         elif self.data_cls == 'custom' or issubclass(self.data_cls,
                                                      HeteroData):
-            return self._hetero_sparse_neighbor_sample(
-                {self.input_type: index}) + (index.numel(), )
+            if not self.disjoint:
+                return self._hetero_sparse_neighbor_sample(
+                    {self.input_type: index}) + (index.numel(), )
+            else:
+                result = []
+                for i in index.size(0):
+                    result.append(self._hetero_sparse_neighbor_sample(
+                        {self.input_type: index[i:i+1]}))
+                return tuple(map(list, zip(*result))) + (index.numel(), )
 
 
 class NeighborLoader(torch.utils.data.DataLoader):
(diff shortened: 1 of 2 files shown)


torch_geometric/loader/neighbor_loader.py
```diff
@@ -198,8 +205,16 @@ def __call__(self, index: Union[List[int], Tensor]):
...
         elif self.data_cls == 'custom' or issubclass(self.data_cls,
                                                      HeteroData):
-            return self._hetero_sparse_neighbor_sample(
-                {self.input_type: index}) + (index.numel(), )
+            if not self.disjoint:
+                return self._hetero_sparse_neighbor_sample(
+                    {self.input_type: index}) + (index.numel(), )
+            else:
+                result = []
+                for i in index.size(0):
```
Padarn, line 213: Does this seem like it would be very slow?

torch_geometric/loader/utils.py
```diff
@@ -187,6 +187,31 @@ def filter_hetero_data(
     return out
 
 
+def filter_hetero_data_disjoint(
```
Padarn, line 190: Not used yet?

Padarn:
@RexYing can you help me understand the problem a bit more? Does this have any impact in the case of graphs without a temporal component?

rusty1s:
@Padarn This will have no real effect on graphs without temporal sampling. It will create disjoint/isolated graphs during sampling instead of merging nodes that appear in different subgraphs (which we currently do). This scenario is likely not happening too frequently anyway for very large graphs.

Padarn:
Yep makes sense 👍 This is what I understood from the current implementation.

torch_geometric/loader/neighbor_loader.py
RexYing: > Yep makes sense 👍 This is what I understood from the current implementation.

Thanks for looking into it! I'll make the suggested change you made:)

I'm testing out both approaches of doing it in PyG v.s. c++ implementation in torch sparse (and later pyg-lib) to see which one could be better. Yes it does seem slower when doing it in PyG. However, this approach of separate calls to c++ could potentially benefit from easy parallelization (which is currently not utilized here).
So I'll get back to you with more insights once I can test both options out more thoroughly.

RexYing:
> Yep makes sense 👍 This is what I understood from the current implementation.

Thanks for looking into it! I'll make the suggested change you made:)

I'm testing out both approaches of doing it in PyG v.s. c++ implementation in torch sparse (and later pyg-lib) to see which one could be better. Yes it does seem slower when doing it in PyG. However, this approach of separate calls to c++ could potentially benefit from easy parallelization (which is currently not utilized here).
So I'll get back to you with more insights once I can test both options out more thoroughly.

rusty1s:
Closing this since this functionality has been directly integrated into `torch-sparse`.

PR #4910, opened 2022-07-01
Fixing `to_heterogeneous` to work with GPU

azrael417:
Dear Sir/Madam,

this small PR makes the to_heterogeneous function of torch_geometric.data.Data compatible with non CPU devices.

Previously, if the Data object was model to the GPU, the to_heterogeneous call would fail with:

```
  File "utils/transforms.py", line 102, in <module>
    main()
  File "utils/transforms.py", line 95, in main
    hdata = data.to_heterogeneous()
  File "/opt/pytorch_geometric/torch_geometric/data/data.py", line 660, in to_heterogeneous
    index_map[node_ids[i]] = torch.arange(len(node_ids[i]))
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

Please let me know what you think.

Best regards
Thorsten


rusty1s (approved):
Thank you!

PR #4922, opened 2022-07-06
`GraphStore`: `Data`, `HeteroData` respect `is_sorted`

mananshah99:
Allows `NeighborLoader` `csc()` conversion with `GraphStore` to respect `is_sorted` if `GraphStore` is a `Data` or `HeteroData` object.


ZenoTan (approved):

rusty1s (commented):
I might be missing information here. Why would we want to allow override of `is_sorted`? IMO, one should take care of correctly inserting `COO` representation in the first place.

sdulloor (approved):

mananshah99:
@rusty1s this is now updated to have `Data` and `HeteroData` respect `is_sorted` properly. PTAL

torch_geometric/data/data.py
```diff
@@ -842,6 +842,9 @@ def _put_edge_index(self, edge_index: EdgeTensorType,
         attr_val = edge_tensor_type_to_adj_type(edge_attr, edge_index)
         setattr(self, attr_name, attr_val)
 
+        # Set edge attributes:
+        setattr(self, f'{attr_name}_edge_attr', edge_attr)
```
rusty1s, line 846: How does this work when we add multiple `edge_attr` to a single object? Won't we need to condition that on layout?
```python
self._edge_attrs = {
    EdgeLayout.COO: copy.copy(edge_attr)
}
```
I would also make this a private attribute to not expose it as part of `print(data)`.

torch_geometric/data/data.py
```diff
@@ -869,8 +872,10 @@ def get_all_edge_attrs(self) -> List[EdgeAttr]:
         out = []
         for layout, attr_name in EDGE_LAYOUT_TO_ATTR_NAME.items():
             if attr_name in self:
+                attr_val = self[f'{attr_name}_edge_attr']
```
rusty1s, line 875: We could just return `self._edge_attrs.values()`. WDYT?

torch_geometric/data/data.py
rusty1s, line 846: Also set `size` of `edge_attr` (if not specified). WDYT?

torch_geometric/data/data.py
```diff
@@ -842,6 +842,9 @@ def _put_edge_index(self, edge_index: EdgeTensorType,
         attr_val = edge_tensor_type_to_adj_type(edge_attr, edge_index)
         setattr(self, attr_name, attr_val)
 
+        # Set edge attributes:
```
rusty1s, line 845: Won't we need to have special logic in `_get_edge_index` as well?

rusty1s (approved):
Approving to unblock, but I have some questions in the comments below on handling `edge_attr` in (`Data` + `HeteroData`).

torch_geometric/data/data.py
mananshah99: `attr_name` should indicate what the layout is (either `edge_index`, `adj`, or `adj_t`), right?

mananshah99: Not sure I understand: why would it be useful to set the size of `EdgeAttr` here? We set the size of the `Data` object if it's provided in `edge_attr`, and when we `get_all_edge_attrs` we return this size as well.

torch_geometric/data/data.py
mananshah99: What kind of special logic? `get_edge_index` just returns the edge corresponding to the requested layout. If we want to handle `is_sorted`, we can handle it at a higher level in `GraphStore`, but I left a TODO to do that later - wdyt?

torch_geometric/data/data.py
rusty1s: You are right, missed that. Shall we use `edge_attr.layout.value` for this? + Also make it private.

rusty1s: Then I think we need some clarity on what we actually want to save in `EdgeAttr` - I was under the impression that all its information is useful to maintain. If the `EdgeAttr` does not define `size`, we can set it based on `self.num_nodes`.

torch_geometric/data/data.py
rusty1s: Got it.

rusty1s:
@mananshah99 My main confusion is probably that `EdgeAttr` defines both attributes to query in `get_edge_index`, but also information that describe the `EdgeTensorType`. Wouldn't it make more sense to push information of `size` and `is_sorted` to `EdgeTensorType` rather than `EdgeAttr`?

torch_geometric/data/data.py
mananshah99: Sure, sounds good

mananshah99:
@rusty1s that's a fair point, you are right that currently `EdgeAttr` contains a mix of information that relate to both query information (e.g. edge type, edge layout) and edge-specific information (e.g. is_sorted, size). I don't think it's easy to store this information with `EdgeTensorType`, as that should just be the returned edge index imo. Open to an alternative refactor here in the future, but for now I think this solution is reasonable.

PR #4929, opened 2022-07-06
`NeighborLoader`: support temporal sampling with `(FeatureStore, GraphStore)`

mananshah99:
Supports temporal sampling by performing the equivalent of a `collect()` operation on `FeatureStore` to construct `node_time_dict`. Also fixes a minor bug in `HeteroData.put_tensor`.


test/loader/test_neighbor_loader.py
```diff
@@ -350,3 +351,52 @@ def test_custom_neighbor_loader(FeatureStore, GraphStore):
...
+
+    num_nodes = data.x.size(dim=0)
+    graph_store.put_edge_index(edge_index=data.edge_index,
+                               edge_type=('paper', 'to', 'paper'),
+                               layout='coo', size=(num_nodes, num_nodes))
+    hetero_data['paper', 'to', 'paper'].edge_index = data.edge_index
+
+    loader1 = NeighborLoader(hetero_data, num_neighbors=[-1, -1],
+                             input_nodes='paper', time_attr='time',
+                             batch_size=1)
```
rusty1s, line 386: Let's add a larger `batch_size` and drop `num_iter` later.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -95,9 +95,21 @@ def __init__(
...
+                # do so, we make an explicit feature store GET call here with
+                # the relevant 'TensorAttr's
+                time_attrs = [
+                    attr for attr in feature_store.get_all_tensor_attrs()
+                    if attr.attr_name == time_attr
+                ]
+                for attr in time_attrs:
+                    attr.index = None
+                time_tensors = feature_store.multi_get_tensor(time_attrs)
+                self.node_time_dict = {
```
rusty1s, line 109:
```python
self.node_time_dict = {
    time.attr.group_name: time_tensor
    for time_attr, time_tensor in zip(time_attrs, time_tensors)
}
```

torch_geometric/loader/neighbor_loader.py
```diff
@@ -475,18 +487,23 @@ def to_index(tensor):
         if isinstance(input_nodes, Tensor):
             return None, to_index(input_nodes)
 
+        # Can't infer number of nodes from a group_name; need an attr_name
         if isinstance(input_nodes, str):
-            num_nodes = feature_store.get_tensor_size(input_nodes)[0]
-            return input_nodes, range(num_nodes)
+            raise NotImplementedError(
```
rusty1s, line 492: Why do we raise here instead of fix it? Shouldn't `feature_store.get_tensor_size(TensorAttr(group_name=input_nodes))[0]` fix this?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -475,18 +487,23 @@ def to_index(tensor):
...
 
         if isinstance(input_nodes, (list, tuple)):
             assert len(input_nodes) == 2
             assert isinstance(input_nodes[0], str)
 
             node_type, input_nodes = input_nodes
             if input_nodes is None:
-                num_nodes = feature_store.get_tensor_size(input_nodes)[0]
-                return input_nodes[0], range(num_nodes)
+                raise NotImplementedError(
```
rusty1s, line 503:
```python
feature_store.get_tensor_size(TensorAttr(group_name=None))[0]
```

rusty1s (approved):

torch_geometric/loader/neighbor_loader.py
mananshah99: That get tensor size call wouldn't work as intended since get tensor rise needs both a group name and are name, hence the raise

rusty1s: So what we could do is to select the first `TensorAttr` that has `group_name==input_nodes`, right? But yeah, not ideal.

PR #4943, opened 2022-07-07
Fixed `BasicGNN` for `num_layers=1` and `out_channels` given

fjulian:
If a model based on `nn.BasicGNN` is created with a single layer and `out_channels` given, the existing code added two layers, where the last layer is a convolution bringing the channel size to `out_channels`. However, this second layer was never reached since in `forward`, the `for` loop only iterates up to `num_layers`. Therefore, the output graph has (counterintuitively) `hidden_channels` channels, instead of `out_channels`.

This PR adds a test for the 1 layer case, which fails with the existing code. Furthermore, the code for `nn.BasicGNN` is extended to properly handle the 1 layer case, making the test pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for graph-level outputs in `to_hetero` ([#4582](https://github.com/pyg-team/pytorch_geometric/pull/4582))
 - Added `CHANGELOG.md` ([#4581](https://github.com/pyg-team/pytorch_geometric/pull/4581))
 ### Changed
+- Fixed `BasicGNN` for 1 layer case, which now respects a desired number of `out_channels`([#4943](https://github.com/pyg-team/pytorch_geometric/pull/4943))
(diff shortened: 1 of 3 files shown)


Padarn:
Interesting, it seems to me that 
```
        out_channels (int, optional): If not set to :obj:`None`, will apply a
            final linear transformation to convert hidden node embeddings to
            output size :obj:`out_channels`. (default: :obj:`None`)
```
is not true - the 'final linear transformation' doesn't exist, it is done via the convolution layer.

Do you think this docstring should be changed?

rusty1s (approved):
Thank you!

rusty1s:
@Padarn Can you clarify? The final transformation `self.lin` is defined whenever `self.jk is not None`.

Padarn:
From the docstring, I would expect a linear transform even when `jk` is None, as long as `out_channels` is not None.  Not a huge deal of course.

rusty1s:
Yeah, I understand. Ultimately, we wanted to support both options, since some models perform better on some datasets when the final classifier is performed inside the convolution.

Padarn:
Makes sense. I think explaining that in the docstring would be more confusing than what is already there.

PR #4944, opened 2022-07-07
Add node-wise normalization mode in `LayerNorm`

lightaime:
This pull request adds the node-wise normalization mode in `LayerNorm`. If `"graph"` mode is used (by default), each graph will be considered as an element to be normalized. If `"node"` mode is used, each node will be considered as an element to be normalized.


torch_geometric/nn/norm/layer_norm.py
```diff
@@ -29,12 +30,18 @@ class LayerNorm(torch.nn.Module):
         affine (bool, optional): If set to :obj:`True`, this module has
             learnable affine parameters :math:`\gamma` and :math:`\beta`.
             (default: :obj:`True`)
+        mode (str, optinal): The normalization mode to use in layer norm
+            (`"graph"` or `"node"`). If `"graph"` is used, each graph will
```
rusty1s, line 34:
```suggestion
            (:obj:`"graph"` or :obj:`"node"`). If :obj:`"graph"` is used, each graph will
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -29,12 +30,18 @@ class LayerNorm(torch.nn.Module):
...
+            be considered as an element to be normalized. If `"node"` is used,
```
rusty1s, line 35:
```suggestion
            be considered as an element to be normalized. If :obj:`"node"` is used,
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -29,12 +30,18 @@ class LayerNorm(torch.nn.Module):
         affine (bool, optional): If set to :obj:`True`, this module has
             learnable affine parameters :math:`\gamma` and :math:`\beta`.
             (default: :obj:`True`)
+        mode (str, optinal): The normalization mode to use in layer norm
```
rusty1s, line 33:
```suggestion
        mode (str, optinal): The normalization mode to use for layer normalization.
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -29,12 +30,18 @@ class LayerNorm(torch.nn.Module):
...
+            (`"graph"` or `"node"`). If `"graph"` is used, each graph will
+            be considered as an element to be normalized. If `"node"` is used,
+            each node will be considered as an element to be normalized.
+            (default: :obj:`"graph"`)
     """
-    def __init__(self, in_channels, eps=1e-5, affine=True):
+    def __init__(self, in_channels, eps=1e-5, affine=True, mode='graph'):
```
rusty1s, line 39:
```suggestion
    def __init__(self, in_channels: int, eps: float=1e-5, affine: bool=True, mode:str='graph'):
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -51,31 +58,40 @@ def reset_parameters(self):
...
+                var = var / norm
 
-            var = scatter(x * x, batch, dim=0, dim_size=batch_size,
-                          reduce='add').sum(dim=-1, keepdim=True)
-            var = var / norm
+                out = x / (var + self.eps).sqrt().index_select(0, batch)
 
-            out = x / (var + self.eps).sqrt().index_select(0, batch)
+            if self.weight is not None and self.bias is not None:
+                out = out * self.weight + self.bias
```
rusty1s, line 84:
```suggestion
                out = out * self.weight + self.bias
                
                return out
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -51,31 +58,40 @@ def reset_parameters(self):
...
 
-        if self.weight is not None and self.bias is not None:
-            out = out * self.weight + self.bias
+        elif self.mode == 'node':
```
rusty1s, line 86:
```suggestion
        if self.mode == 'node':
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -51,31 +58,40 @@ def reset_parameters(self):
...
+            out = F.layer_norm(x, (self.in_channels, ), self.weight, self.bias,
```
rusty1s, line 87:
```suggestion
            return F.layer_norm(x, (self.in_channels, ), self.weight, self.bias,
```

torch_geometric/nn/norm/layer_norm.py
```diff
@@ -51,31 +58,40 @@ def reset_parameters(self):
...
+                               self.eps)
+
+        else:
```
rusty1s, line 90:
```suggestion
```
no need for the `else`.

rusty1s (approved):
Super. Only left nitpicky comments :D

PR #4951, opened 2022-07-09
Support `normalization_resolver` in `MLP`

lightaime:
This pull request added support for the `normalization_resolver` in `MLP`.


torch_geometric/nn/models/mlp.py
```diff
@@ -76,18 +79,25 @@ def __init__(
...
     ):
         super().__init__()
 
-        act_first = act_first or relu_first  # Backward compatibility.
-        batch_norm_kwargs = batch_norm_kwargs or {}
+        # Backward compatibility:
+        act_first = act_first or kwargs.get("relu_first", False)
+        batch_norm = kwargs.get("batch_norm", None)
+        if batch_norm is not None and isinstance(batch_norm, bool):
+            warnings.warn("Argument `batch_norm` will be deprecated, "
```
Padarn, line 96: I'm wondering if we should decide on a version in which it will be depreciated and note that here? WDYT?

torch_geometric/nn/models/mlp.py
```diff
@@ -114,11 +124,15 @@ def __init__(
...
+            if norm is not None:
+                norm_layer = normalization_resolver(
+                    norm,
+                    hidden_channels,
+                    **(norm_kwargs or {}),
+                )
             else:
-                norm = Identity()
-            self.norms.append(norm)
+                norm_layer = Identity()
```
Padarn, line 134: do you need to do this? could we just have no norm layer in the case that none are needed rather than adding the identity?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- Added Support for `normalization_resolver` in `MLP` ([#4951](https://github.com/pyg-team/pytorch_geometric/pull/4951))
```
Padarn, line 8: maybe we should add this to line 10 instead? the changelog is getting very big 😓

torch_geometric/nn/models/mlp.py
```diff
@@ -48,23 +52,22 @@ class MLP(torch.nn.Module):
...
+            respective normalization function defined by :obj:`norm`.
             (default: :obj:`None`)
         plain_last (bool, optional): If set to :obj:`False`, will apply
             non-linearity, batch normalization and dropout to the last layer as
             well. (default: :obj:`True`)
         bias (bool, optional): If set to :obj:`False`, the module will not
             learn additive biases. (default: :obj:`True`)
-        relu_first (bool, optional): Deprecated in favor of :obj:`act_first`.
-            (default: :obj:`False`)
+        **kwargs (optional): Additional arguments of the MLP layer.
```
rusty1s, line 70:
```suggestion
        **kwargs (optional): Additional deprecated arguments of the MLP layer.
```

torch_geometric/nn/models/mlp.py
rusty1s, line 96:
```suggestion
            warnings.warn("Argument `batch_norm` is deprecated, "
```

torch_geometric/nn/models/mlp.py
rusty1s: This shifts the efforts of checking this to the `forward` call. I think either way is fine.

rusty1s (approved):

CHANGELOG.md
lightaime: Good idea!

torch_geometric/nn/models/mlp.py
lightaime: I am not sure when we want to fully deprecate it though. @rusty1s any idea on this?

lightaime:
- [ ] TODO: fix `norm` support in examples. Will make a seperate for it.

Padarn (approved):

rusty1s:
Thank you @lightaime. Looking forward to the follow-up PR.

Issue #4712, opened 2022-05-25
[Roadmap] `torch_geometric.nn.aggr` 🚀 


rusty1s:
### 🚀 The feature, motivation and pitch

The goal of this roadmap is to unify the concepts of aggregation inside GNNs across both [`MessagePassing`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/message_passing.py) and [global readouts](https://github.com/pyg-team/pytorch_geometric/tree/master/torch_geometric/nn/glob). Currently, these concepts are separated, *e.g.*, via `MessagePassing.aggr = "mean"` and `global_mean_pool(...)` while the underlying implementation is the same. In addition, some aggregations are only available as global pooling operators ([`global_sort_pool`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/glob/sort.py), [`Set2Set`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/glob/set2set.py), ...), while, in theory, they are also applicable during `MessagePassing` (and vice versa, *e.g.*, [`SAGEConv.aggr = "lstm"`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/sage_conv.py#L149-L159)). One additional feature is the combination of aggregations, which is a useful feature both in `MessagePassing` ([`PNAConv`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/pna_conv.py), [`EGConv`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/eg_conv.py), ...) and global readouts.

As such, we want to provide **re-usable aggregations** as part of a newly defined `torch_geometric.nn.aggr.*` package. Unifying these concepts also helps us to perform optimization and specialized implementations in a single place (*e.g.*, fused kernels for multiple aggregations). After integration, the following functionality is applicable:

```python
class MyConv(MessagePassing):
    def __init__(self):
        super().__init__(aggr="mean")

class MyConv(MessagePassing):
    def __init__(self):
        super().__init__(aggr=LSTMAggr(channels=...))

class MyConv(MessagePassing):
    def __init__(self):
        super().__init__(aggr=MultiAggr("mean", "max", Set2Set(channels=...))
```

### Roadmap

The general roadmap looks as follows (at best, each implemented in a separate PR):

[...]

PR #4957, opened 2022-07-11
Add `GlobalSortAggr` to `nn.aggr` package.

Padarn:
Add `GlobalSortAggr` to the new `nn.aggr` packages replace existing `global_sort_pool` from `nn.glob.sort`.

Addresses item in https://github.com/pyg-team/pytorch_geometric/issues/4712

WIP items to address:
- [x] input `dim_size` is not checked or used
- [x] input `dim` is not yet respected # skipping as to_dense_batch does not support this yet
- [x] depreciate `nn.glob` version (replace it in existing implementations?)


torch_geometric/nn/aggr/__init__.py
```diff
@@ -14,20 +14,11 @@
...
-    'StdAggregation',
-    'SoftmaxAggregation',
-    'PowerMeanAggregation',
-    'LSTMAggregation',
-    'Set2Set',
-    'DegreeScalerAggregation',
+    'Aggregation', 'MultiAggregation', 'MeanAggregation', 'SumAggregation',
+    'MaxAggregation', 'MinAggregation', 'MulAggregation', 'VarAggregation',
+    'StdAggregation', 'SoftmaxAggregation', 'PowerMeanAggregation',
+    'LSTMAggregation', 'Set2Set', 'DegreeScalerAggregation', 'GlobalSortAggr'
```
rusty1s, line 23:
```suggestion
    'LSTMAggregation', 'Set2Set', 'DegreeScalerAggregation', 'GlobalSortAggr',
```

rusty1s (commented):
Thank you! Looks amazing. If possible, let's deprecate the corresponding function in `torch_geometric.nn.glob` in this PR as well.

Padarn:
Makes sense. Will add it to the list.

Padarn:
Not sure why I'm getting failures in the `glob` tests - do we want to keep them?

torch_geometric/nn/glob/__init__.py
```diff
@@ -18,8 +16,14 @@
...
 
 Set2Set = deprecated(
     details="use 'nn.aggr.Set2Set' instead",
     func_name='nn.glob.Set2Set',
 )(Set2Set)
+
+global_sort_pool = deprecated(
+    details="use 'nn.aggr.GlobalSortAggr' instead",
+    func_name='nn.glob.global_sort_pool',
+)(sort.global_sort_pool)
```
Padarn, line 29: cannot use the new class here as its not a replacement (new class doesn't take `k` in the function call)

Padarn: we could make it a kwarg if we want it to work exactly the same way

torch_geometric/nn/aggr/sort.py
```diff
@@ -0,0 +1,52 @@
+from typing import Optional
+
+import torch
+from torch import Tensor
+
+from torch_geometric.nn.aggr import Aggregation
+
+
+class GlobalSortAggr(Aggregation):
+    r"""The global pooling operator from the `"An End-to-End Deep Learning
```
rusty1s, line 10:
```suggestion
    r"""The sort pooling operator from the `"An End-to-End Deep Learning
```

torch_geometric/nn/aggr/sort.py
```diff
@@ -0,0 +1,52 @@
+from typing import Optional
+
+import torch
+from torch import Tensor
+
+from torch_geometric.nn.aggr import Aggregation
+
+
+class GlobalSortAggr(Aggregation):
```
rusty1s, line 9:
```suggestion
class SortAggr(Aggregation):
```
Since `Aggregation` can be used for both message passing aggregation and global pooling, let's not use `Global` inside the name. WDYT?

torch_geometric/nn/glob/__init__.py
rusty1s: I think you can just define it here:
```python
@deprecated(...)
def global_sort_pool(...):
    module = GlobalSortAggr(...)
    return module(...)
```

rusty1s (commented):
I think the issue with failing tests is that there exists multiple `test_sort.py` files (which is not allowed by `pytest`). Try removing `glob/aggr/test_sort.py`.

torch_geometric/nn/aggr/sort.py
Padarn: Good point

torch_geometric/nn/glob/__init__.py
Padarn: yeah okay this is a better solution, thanks!

Padarn:
Looks okay now. May as well remove the old version at this point. WDYT?

Padarn:
All done :-)

lightaime (approved):
Thanks, @Padarn. LGTM!

lightaime:
nit-pick: I think it may be better to rename `SortAggr` as `SortAggregation` to be consistent with other `Aggregation` classes. WDYT?

Padarn:
I agree @lightaime! So: https://github.com/pyg-team/pytorch_geometric/pull/5085

PR #4958, opened 2022-07-11
Support `normalization_resolver` in `BasicGNN`

lightaime:
This PR added support for `normalization_resolver` in `basic_gnn`.


torch_geometric/nn/models/basic_gnn.py
```diff
@@ -100,9 +106,19 @@ def __init__(
         if norm is not None:
             self.norms = ModuleList()
             for _ in range(num_layers - 1):
-                self.norms.append(copy.deepcopy(norm))
+                norm_layer = normalization_resolver(
```
rusty1s, line 109: I think we still need to deep copy it here.

torch_geometric/nn/models/basic_gnn.py
```diff
@@ -57,7 +63,8 @@ def __init__(
         out_channels: Optional[int] = None,
         dropout: float = 0.0,
         act: Union[str, Callable, None] = "relu",
-        norm: Optional[torch.nn.Module] = None,
+        norm: Optional[str] = None,
```
Padarn, line 66: Should this be a `Union[str, Callable]`?

Padarn (approved):

torch_geometric/nn/models/basic_gnn.py
rusty1s: Good catch. Fixed.

rusty1s (approved):

PR #4962, opened 2022-07-12
`GraphStore`: handle zero-sized edge indices

mananshah99:
Small fix for edge indices of shape `(2, 0)` (decomposed into two tensors of size 0), for which we cannot index into each tensor.


rusty1s (approved):
Thank you!

Issue #583, opened 2019-07-31
DataLoader: Dynamic Batch-size based on num_nodes/num_edges


yg-li:
## ❓ Questions & Help

I was using the RGCNConv layer with num_relations=4, in_channels=out_channels=512. In the forward step, I pass in a graph with 300 nodes and 22768 edges which caused it to raise an CUDA OOM saying it needs 22.23GB memory while I only have 11.17GB.

The line causing the error is
`w = torch.index_select(w, 0, edge_type)`
in the message function of RGCNConv class, which makes sense as it was trying to create a float tensor of size [22768 x 512 x 512].

But it seems copying the weights num_edges times is inefficient and not necessary. Is there another way to implement RGCN without copying the weights this many times?

rusty1s:
I sadly do not know of another way to implement this layer. For each edge, we need to select the appropriate weight matrix and multiply it with the respective feature vector. This is the reason why the authors of the paper train larger graphs on the CPU. Might this be an option to you? If you know of another way to implement this, feel free to reach out :)

yg-li:
> I sadly do not know of another way to implement this layer. For each edge, we need to select the appropriate weight matrix and multiply it with the respective feature vector. This is the reason why the authors of the paper train larger graphs on the CPU. Might this be an option to you? If you know of another way to implement this, feel free to reach out :)

Thank you for your reply @rusty1s . I do not know the way to solve this problem either.

Btw I also have a problem with batching. Since graphs have different shapes/sizes, each batch of 32 graphs are consuming different GPU memory. Hence, some of the batches are capable to be processed but the others cause OOM problem. 

Is there a general way to solve (or mitigate) this problem? I cannot find anything about in the documentation.

rusty1s:
This is a good request and we currently do nor support this. We could add an argument `max_num_nodes` or `max_num_edges` in addition to `batch_size` to the `DataLoader`. However, this requires us to implement our own data loading routine without relying on PyTorch to do this task for us so it can be a bit tricky to get right, especially in combination with `num_workers`.

Zacharias030:
I was about to open the same issue about dynamic batch size.

Could we as a hack create a `collate_fn` which discards the latter part of the batch under the `max_num_nodes` condition (greedily adding samples from the original batch in order and maybe discarding the tail of the batch)?
You probably thought of that before, therefore I'd welcome a hint as to why this is not possible or a pointer for where to start implementing a feature like that!

[...]

PR #4972, opened 2022-07-12
Dynamic batch-size sampler based on `num_nodes`/`num_edges`

GericoVi:
As discussed in #583


torch_geometric/loader/__init__.py
```diff
@@ -12,23 +12,13 @@
...
-    'DenseDataLoader',
-    'TemporalDataLoader',
-    'NeighborSampler',
-    'ImbalancedSampler',
+    'DataLoader', 'NeighborLoader', 'LinkNeighborLoader', 'HGTLoader',
+    'ClusterData', 'ClusterLoader', 'GraphSAINTSampler',
+    'GraphSAINTNodeSampler', 'GraphSAINTEdgeSampler',
+    'GraphSAINTRandomWalkSampler', 'ShaDowKHopSampler', 'RandomNodeSampler',
+    'DataListLoader', 'DenseDataLoader', 'TemporalDataLoader',
+    'NeighborSampler', 'ImbalancedSampler', 'DynamicBatchSampler'
```
rusty1s, line 23:
```suggestion
    'NeighborSampler', 'ImbalancedSampler', 'DynamicBatchSampler',
```
to let `yapf` re-format it.

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
+from typing import Iterator, List, Optional
+
+import torch
+
+from torch_geometric.data import Dataset
+
+
+class DynamicBatchSampler(torch.utils.data.sampler.Sampler[List[int]]):
+    r"""Dynamically adds samples to mini-batch up to a maximum size (number of
```
rusty1s, line 9:
```suggestion
    r"""Dynamically adds samples to a mini-batch up to a maximum size (number of
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+    ``num_steps`` can be supplied to cap the number of mini-batches produced
+    by the sampler.
+
+    **Usage:**
+
+    .. code-block:: python
+
+        from torch_geometric.loader import DataLoader, DynamicBatchSampler
+
+        batch_sampler = DynamicBatchSampler(10000, dataset)
```
rusty1s, line 26:
```suggestion
        batch_sampler = DynamicBatchSampler(dataset, num_nodes=10000)
```
Can we first pass in the data, and also rename `max_num` to `num_nodes`?

GericoVi: Yes, the data argument first makes sense. For the other, I had it as `max_num` instead of `num_nodes` since you can also specify a maximum number of edges in the batch instead of nodes with the `entity` argument. Is there a more generic name? Or do you think `num_nodes` is fine?

test/loader/test_dynamic_batch_sampler.py
```diff
@@ -0,0 +1,31 @@
...
+
+import torch
+
+from torch_geometric.data import Data
+from torch_geometric.loader import DataLoader, DynamicBatchSampler
+
+
+def test_dataloader_with_dynamic_batches():
+    data_list: List[Data] = []
+    for i in range(100, 110):
```
rusty1s, line 11:
```suggestion
    for num_nodes in range(100, 110):
```

test/loader/test_dynamic_batch_sampler.py
```diff
@@ -0,0 +1,31 @@
...
+        data_list.append(Data(num_nodes=i))
+
+    torch.manual_seed(12345)
+    batch_sampler = DynamicBatchSampler(data_list, 300, shuffle=True)
+    loader = DataLoader(data_list, batch_sampler=batch_sampler)
+
+    num_nodes_total = 0
+    for data in loader:
+        num_nodes_total += data.num_nodes
```
rusty1s, line 20: Add an assert that the number of nodes per batch is not larger than 300.

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
+from typing import Iterator, List, Optional
+
+import torch
+
+from torch_geometric.data import Dataset
+
+
+class DynamicBatchSampler(torch.utils.data.sampler.Sampler[List[int]]):
+    r"""Dynamically adds samples to a mini-batch up to a maximum size (number of
+    nodes). When data samples have a wide range in sizes, specifying a
```
rusty1s, line 10: Either based on number of nodes or number of edges.

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+    mini-batch size in terms of number of samples is not ideal and can cause
+    CUDA OOM errors.
+
+    Therefore, the number of steps per epoch is ambiguous, depending on the
```
rusty1s, line 14:
```suggestion
    Within the :class:`DynamicBatchSampler`, the number of steps per epoch is ambiguous, depending on the
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+    order of the samples. By default the ``__len__()`` will be None, fine for
```
rusty1s, line 15:
```suggestion
    order of the samples. By default the :meth:`__len__` will be undefined. This is fine for
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+    most cases but progress bars will be `infinite`. Alternatively,
+    ``num_steps`` can be supplied to cap the number of mini-batches produced
```
rusty1s, line 17:
```suggestion
    :obj:`num_steps` can be supplied to cap the number of mini-batches produced
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+
+    .. code-block:: python
+
+        from torch_geometric.loader import DataLoader, DynamicBatchSampler
+
+        batch_sampler = DynamicBatchSampler(dataset, max_num=10000)
+        loader = DataLoader(dataset, batch_sampler=sampler, ...)
+
+    Args:
+        data_source (Dataset): dataset to sample from, need this so that
```
rusty1s, line 30:
```suggestion
        dataset (Dataset): Dataset to sample from.
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+            potential samples can be sized to create the dynamic batch.
+        max_num (int): size of mini-batch to aim for in number of nodes or
+            edges.
+        mode (str, optional): `nodes` or `edges` to measure batch size.
```
rusty1s, line 34:
```suggestion
        mode (str, optional): obj:`node` or obj:`edge` to measure batch size.
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+            (default: `nodes`)
```
rusty1s, line 35:
```suggestion
            (default: obj:`node`)
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+        shuffle (bool, optional): set to ``True`` to have the data reshuffled
```
rusty1s, line 36: See other occurrences and replace with `:obj:` to let the doc render uniformly.

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+            at every epoch (default: ``False``).
+        skip_too_big (bool, optional): set to ``True`` to skip samples which
+            can't fit in a batch by itself. (default: ``False``).
+        num_steps (int, optional): The number of mini-batches to draw for a
+            single epoch. If set to :obj:`None`, will iterate through all the
+            underlying data, but __len__ will be :obj:`None` since it will be
```
rusty1s, line 42:
```suggestion
            underlying data, but :meth:`__len__` will be :obj:`None` since it will be
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+            ambiguous. (default: :obj:`None`)
+    """
+    def __init__(self, data_source: Dataset, max_num: int, mode: str = 'nodes',
+                 shuffle: bool = False, skip_too_big: bool = False,
+                 num_steps: Optional[int] = None):
+        if not isinstance(max_num, int) or max_num <= 0:
+            raise ValueError("`max_num` should be a positive integer value, "
+                             "but got max_num={max_num}.")
```
rusty1s, line 50:
```suggestion
                             "(got {max_num)}.")
```

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+            num_steps = len(data_source)
+        self.num_steps = num_steps
+
+        self.mode = f"num_{mode}"
+
+    def __iter__(self) -> Iterator[List[int]]:
+        batch = []
+        batch_n = 0
+        num_steps = 0
+        current_place = 0
```
rusty1s, line 69:
```suggestion
        num_processed = 0
```
I find `current_place` a bit misleading due to shuffle option.

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            self.idxs = torch.arange(len(self.data_source), dtype=torch.long)
+
+        # Main iteration loop
+        while (current_place < len(self.data_source)
+               and num_steps < self.num_steps):
+            # Fill batch
+            for idx in self.idxs[current_place:]:
+                # Size of sample
+                n = getattr(self.data_source[idx], self.mode)
```
rusty1s, line 82: Can we make this explicit here and branch out based on mode?

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+
+                if batch_n + n > self.max_num:
+                    if batch_n == 0:
+                        if self.skip_too_big:
+                            continue
+                        else:
+                            raise RuntimeError(
```
rusty1s, line 89: Won‘t we want to skip instead?

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,108 @@
...
+        self.mode = f"num_{mode}"
+
+    def __iter__(self) -> Iterator[List[int]]:
+        batch = []
+        batch_n = 0
+        num_steps = 0
+        current_place = 0
+
+        if self.shuffle:
+            self.idxs = torch.randperm(len(self.data_source), dtype=torch.long)
```
rusty1s, line 72:
```suggestion
            indices = torch.randperm(len(self.data_source), dtype=torch.long)
```
Do we need to add it to `self`?

rusty1s (commented):
Left some final (nitpicky) comments. Thank you!

torch_geometric/loader/dynamic_batch_sampler.py
GericoVi: I don't quite understand this comment. Essentially the purpose of that bit is that if the batch is empty and the next sample will overfill it - either skip that sample or raise an error.
If it's not empty, currently it outputs the batch and then attempts to use that sample in the next batch.

rusty1s: That makes sense! Thanks!

rusty1s: I looked at it again and feel that we should output a warning here and still add that data object in (as a single example per batch). Currently, we only have the option to skip or to error out - and neither of this may be what the user desires.

rusty1s (approved):
Thanks for the updates. Found some minor last issues but happy to merge afterwards. Left one last comment regarding erroring out on too big examples - I think we should print out a warning but still keep the example. WDYT?

rusty1s:
Can you also add this PR to the changelog?

torch_geometric/loader/dynamic_batch_sampler.py
```diff
@@ -0,0 +1,110 @@
...
+    wide range in sizes, specifying a mini-batch size in terms of number of
+    samples is not ideal and can cause CUDA OOM errors.
+
+    Within the :class:`DynamicBatchSampler`, the number of steps per epoch is
+    ambiguous, depending on the order of the samples. By default the
+    :meth:`__len__` will be undefined. This is fine for most cases but
+    progress bars will be `infinite`. Alternatively, :obj:`num_steps` can be
+    supplied to cap the number of mini-batches produced by the sampler.
+
+    **Usage:**
```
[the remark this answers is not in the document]
GericoVi: Which header, sorry? The usage portion?

torch_geometric/loader/dynamic_batch_sampler.py
GericoVi: Yes that probably sounds like a better idea actually. So instead of error, just throw a warning and add the example anyway? To clarify, when this happens there would only be a single example in the mini-batch, and `data.num_nodes` will be greater than the `max_num` specified for that one mini-batch.

That sounds reasonable to me.

rusty1s: Super :)

torch_geometric/loader/dynamic_batch_sampler.py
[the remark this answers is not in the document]
rusty1s: Yes, indeed.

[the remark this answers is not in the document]
GericoVi: Makes sense, will do.

GericoVi:
Discussed updates done. Feel free to merge if happy with changes.

rusty1s (approved):

PR #4978, opened 2022-07-13
Fix `graphgym` activation function dictionary, now not including instances anymore

fjulian:
Until now, the `act_dict` used by the `graphgym` part of this library contained instances of activation functions that were used to build models. This is okay for the activation functions that do not have parameters. However, in the case of PReLU, there is a parameter. This lead to the problem that every PReLU which is built using the `graphgym` infrastructure shared a single parameter, e.g. when PReLUs are used in multiple layers, or when PReLUs are used in several models that are trained alternatingly. The shared parameter caused unexpected changes to a network that was thought to be frozen while another network was trained.

The solution to this problem proposed in this PR is to include activation function constructors in the `act_dict` instead of instances, and create a new instance every time an activation function is used in a model.


rusty1s:
Thanks @fjulian, let's fix/adjust the test.

fjulian:
My pleasure. Just fixed them.

PR #4981, opened 2022-07-14
Added fine grained options for setting `bias` and `dropout` per layer in the `MLP` model

bramton:
I am attampting to replicate the _Dynamic Graph CNN for learning on Point Clouds_ work using the PyTorch geometric library. In order to do so, I needed some fine grained options for setting the biases and dropouts in the MLP model.

For instance, for the [DGCNN classification model](https://github.com/WangYueFt/dgcnn/blob/master/pytorch/model.py), the last MLP uses a bias for the last two layers only. For the [DGCNN segmentation task](https://github.com/antao97/dgcnn.pytorch/blob/master/model.py), in the last MLP a dropout is only added for the second layer.

On a side note, it seems that if `plain_last` is set, the last layer is executed twice. Once in the zipped for loop and then once more in the conditional statement. Not sure if I am correct.


torch_geometric/nn/models/mlp.py
```diff
@@ -77,14 +81,14 @@ def __init__(
         hidden_channels: Optional[int] = None,
         out_channels: Optional[int] = None,
         num_layers: Optional[int] = None,
-        dropout: float = 0.,
+        dropout: Optional[Union[List[float], float]] = 0.,
```
Padarn, line 84:
```suggestion
        dropout: Union[List[float], float] = 0.,
```

torch_geometric/nn/models/mlp.py
```diff
@@ -77,14 +81,14 @@ def __init__(
...
         act: Union[str, Callable, None] = "relu",
         act_first: bool = False,
         act_kwargs: Optional[Dict[str, Any]] = None,
         norm: Union[str, Callable, None] = "batch_norm",
         norm_kwargs: Optional[Dict[str, Any]] = None,
         plain_last: bool = True,
-        bias: bool = True,
+        bias: Optional[Union[List[bool], bool]] = True,
```
Padarn, line 91:
```suggestion
        bias: Union[List[bool], bool] = True,
```
The implementation does not handle these being None so I'd suggest just not marking as optional

Padarn:
In the current implementation `__init__` has 
```
        iterator = channel_list[1:-1] if plain_last else channel_list[1:]
        for hidden_channels in iterator:
            if norm is not None:
                norm_layer = normalization_resolver(
                    norm,
                    hidden_channels,
                    **(norm_kwargs or {}),
                )
            else:
                norm_layer = Identity()
            self.norms.append(norm_layer)
```
so when in the forward pass 
```
for lin, norm in zip(self.lins, self.norms):
```
the `self.norms` is one shorter than the `self.lins` for `plain_last`

Padarn:
Can you check the failing checks? Seems something wrong in the documentation and you haven't updated changelog.md

torch_geometric/nn/models/mlp.py
```diff
@@ -111,14 +115,34 @@ def __init__(
...
         self.plain_last = plain_last
 
+        if isinstance(bias, bool):
+            bias_list = [bias] * (len(channel_list) - 1)
+        else:
+            if len(bias) is not (len(channel_list) - 1):
+                raise ValueError(
+                    f"Number of biases provided ({len(bias)})does not match "
+                    f"the number of layers specified ({len(channel_list)-1})")
+            bias_list = bias
```
Padarn, line 141: could we also add a simple test case that covers  this new input?

rusty1s (approved):

torch_geometric/nn/models/mlp.py
rusty1s: Added one in.

rusty1s:
Thank you! Made some minor modifications, and added a test case. Looks all good to me now.

PR #4986, opened 2022-07-15
Move `nn.glob.attention.GlobalAttention` to `nn.aggr.attention.AttentionalAggregation`

Padarn:
Addresses https://github.com/pyg-team/pytorch_geometric/issues/4712


torch_geometric/nn/aggr/attention.py
```diff
@@ -53,28 +54,21 @@ def reset_parameters(self):
...
-            size (int, optional): The number of graphs in the batch.
-                (default: :obj:`None`)
-        """
-        if batch is None:
-            batch = x.new_zeros(x.size(0), dtype=torch.int64)
+    def forward(self, x: Tensor, index: Optional[Tensor] = None,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+
+        self.assert_index_present(index)
```
rusty1s, line 61: I don't think this is necessarily true. We can replace `scatter_add` with `self.reduce`, and the `softmax` call should support both `index` and `ptr`.

torch_geometric/nn/aggr/attention.py
```diff
@@ -4,13 +4,14 @@
...
 
+from torch_geometric.nn.aggr import Aggregation
 from torch_geometric.utils import softmax
 
 from ..inits import reset
 
 
-class GlobalAttention(torch.nn.Module):
-    r"""Global soft attention layer from the `"Gated Graph Sequence Neural
+class AttentionAggregation(Aggregation):
```
rusty1s, line 13:
```suggestion
class AttentionalAggregation(Aggregation):
```

test/nn/aggr/test_attention.py
```diff
@@ -0,0 +1,38 @@
...
+from torch_geometric.nn.glob import GlobalAttention
+
+
+def test_attention_aggregation():
+    channels, dim_size = (32, 10)
+    gate_nn = Seq(Lin(channels, channels), ReLU(), Lin(channels, 1))
+    nn = Seq(Lin(channels, channels), ReLU(), Lin(channels, channels))
+
+    aggr = AttentionAggregation(gate_nn, nn)
+    assert aggr.__repr__() == (
```
rusty1s, line 16:
```suggestion
    assert str(aggr) == (
```

test/nn/aggr/test_attention.py
```diff
@@ -0,0 +1,38 @@
...
+
+    x = torch.randn((dim_size**2, channels))
+    index = torch.arange(dim_size, dtype=torch.long)
+    index = index.view(-1, 1).repeat(1, dim_size).view(-1)
+
+    assert aggr(x, index).size() == (dim_size, channels)
+    assert aggr(x, index,
+                dim_size=dim_size + 1).size() == (dim_size + 1, channels)
+
+    # test depreciated
```
rusty1s, line 35: No reason to test this IMO.

test/nn/aggr/test_attention.py
```diff
@@ -0,0 +1,38 @@
...
+        '  (1): ReLU()\n'
+        '  (2): Linear(in_features=32, out_features=32, bias=True)\n'
+        '))')
+
+    x = torch.randn((dim_size**2, channels))
+    index = torch.arange(dim_size, dtype=torch.long)
+    index = index.view(-1, 1).repeat(1, dim_size).view(-1)
+
+    assert aggr(x, index).size() == (dim_size, channels)
+    assert aggr(x, index,
```
rusty1s, line 32: Test
```python
out = aggr(x, index)
assert out.size() == (dim_size, channels)
torch.allclose(aggr(x, index, dim_size=dim_size), out)
```
instead.

rusty1s (approved):

Padarn:
Thanks for the comments @rusty1s!

test/nn/aggr/test_attention.py
Padarn: Agree, added this to do some local tests. Thanks for reminding me

torch_geometric/nn/aggr/attention.py
```diff
@@ -53,28 +54,21 @@ def reset_parameters(self):
...
-                respective graph identifier. (default: :obj:`None`)
-            size (int, optional): The number of graphs in the batch.
-                (default: :obj:`None`)
-        """
-        if batch is None:
-            batch = x.new_zeros(x.size(0), dtype=torch.int64)
+    def forward(self, x: Tensor, index: Optional[Tensor] = None,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+
```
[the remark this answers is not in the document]
Padarn: We can do this, but right now I think it would be backwards compatibility breaking because of
```
        x = x.unsqueeze(-1) if x.dim() == 1 else x
```
WDYT?

test/nn/aggr/test_attention.py
Padarn: I added this, but just to confirm you want to make surer that the output does not change if `dim_size` is provided? Maybe it would be better to test
```
out = aggr(x, index)
assert out.size() == (dim_size, channels)
torch.allclose(aggr(x, index, dim_size=dim_size + 1)[:3], out)
```
WDYT?

torch_geometric/nn/aggr/attention.py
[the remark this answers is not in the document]
Padarn: I've changed for now and removed this line. But still interested in your thought here.

torch_geometric/nn/aggr/attention.py
Padarn: you're completely correct, I missed it. Thanks, updated.

test/nn/aggr/test_attention.py
```diff
@@ -0,0 +1,21 @@
+import torch
+
+from torch_geometric.nn import MLP
+from torch_geometric.nn.aggr import AttentionalAggregation
+
+
+def test_attention_aggregation():
```
rusty1s, line 7:
```suggestion
def test_attentional_aggregation():
```

test/nn/aggr/test_attention.py
```diff
@@ -0,0 +1,21 @@
...
+    channels = 16
+    x = torch.randn(6, channels)
+    index = torch.tensor([0, 0, 1, 1, 1, 2])
+    ptr = torch.tensor([0, 2, 5, 6])
+
+    gate_nn = MLP([channels, 1], act='relu')
+    nn = MLP([channels, channels], act='relu')
+    aggr = AttentionalAggregation(gate_nn, nn)
+    assert str(aggr) == (f'AttentionalAggregation(gate_nn=MLP({channels}, 1),'
+                         f' nn=MLP({channels}, {channels}))')
```
rusty1s, line 17:
```suggestion
    assert str(aggr) == (f'AttentionalAggregation(gate_nn=MLP({channels}, 1), '
                         f'nn=MLP({channels}, {channels}))')
```

torch_geometric/nn/aggr/attention.py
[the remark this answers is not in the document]
rusty1s: I personally think this is okay. The change on user-side would be quick, and I assume this case is very rare.

rusty1s (approved):

PR #4995, opened 2022-07-16
Add `semi_grad` and docs to `SoftmaxAggregtion`

lightaime:
This pull request added `semi_grad` to `SoftmaxAggregtion` for saving memory when temperature `t` is not learnable. Also docs are added.

Addresses item in https://github.com/pyg-team/pytorch_geometric/issues/4712.


CHANGELOG.md
```diff
@@ -35,7 +35,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
-- Added the `DimeNet++` model ([#4432](https://github.com/pyg-team/pytorch_geometric/pull/4432), [#4699](https://github.com/pyg-team/pytorch_geometric/pull/4699), [#4700](https://github.com/pyg-team/pytorch_geometric/pull/4700), [#4800](https://github.com/pyg-team/pytorch_geometric/pull/4800))
+- Added the `DimeNet++` model ([#4432](https://github.com/pyg-team/pytorch_geometric/pull/4432), [#4699](https://github.com/pyg-team/pytorch_geometric/pull/4699), [#4700](https://github.com/pyg-team/pytorch_geometric/pull/4700), [#4800](https://github.com/pyg-team/pytorch_geometric/pull/4800), [#4995](https://github.com/pyg-team/pytorch_geometric/pull/4995))
```
Padarn, line 38: I think this is under the wrong bullet point?

torch_geometric/nn/aggr/basic.py
```diff
@@ -67,12 +67,39 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+
+    where :math:`t` controls the softness of softmax aggregation.
+
+    Args:
+        t (float, optional): Initial inverse temperature for softmax
+            aggregation. (default: :obj:`1.0`)
+        learn (bool, optional): If set to :obj:`True`, will learn the value
+            :obj:`t` for softmax aggregation dynamically.
+            (default: :obj:`False`)
+        semi_grad (bool, optional): This is only used to save memory when the
```
Padarn, line 89: Is this any reason to make this optional? couldn't we instead just check learn?

torch_geometric/nn/aggr/basic.py
```diff
@@ -86,7 +113,11 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
         alpha = x
         if not isinstance(self.t, (int, float)) or self.t != 1:
             alpha = x * self.t
-        alpha = softmax(alpha, index, ptr, dim_size, dim)
+        if not self.learn and self.semi_grad:
```
Padarn, line 116:
```suggestion
        if not self.learn:
```
Related to above comment - could we just do this?

Padarn (changes requested):
Looks good to me overall but left a few comments

torch_geometric/nn/aggr/basic.py
```diff
@@ -67,12 +67,39 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
 
 
 class SoftmaxAggregation(Aggregation):
-    def __init__(self, t: float = 1.0, learn: bool = False):
+    r"""The softmax aggregation operator based on a temperature term
+    , as described in the `"DeeperGCN: All You Need to Train Deeper
```
rusty1s, line 71:
```suggestion
    r"""The softmax aggregation operator based on a temperature term,
    as described in the `"DeeperGCN: All You Need to Train Deeper
```

torch_geometric/nn/aggr/basic.py
rusty1s: I think this is correct. Either in case `t` is not learnable, we still may want to backprop through it.

rusty1s (approved):

torch_geometric/nn/aggr/basic.py
Padarn: Right makes sense

CHANGELOG.md
lightaime: Good catch! I may need new eyeglasses... :)

lightaime:
Thanks for the fix and comment @rusty1s @Padarn!

PR #4996, opened 2022-07-17
Deprecate `nn.glob` package

Padarn:
Addresses https://github.com/pyg-team/pytorch_geometric/issues/4712

This depends on merging https://github.com/pyg-team/pytorch_geometric/pull/4957


torch_geometric/nn/glob/__init__.py
```diff
@@ -35,3 +27,48 @@
...
+
+@deprecated(
+    details="use 'nn.aggr.MaxAggregation' instead",
+    func_name='nn.glob.global_max_pool',
+)
+def global_max_pool(x, index, size):
+    return max_aggr(x, index, dim_size=size)
+
+
+@deprecated(
```
rusty1s, line 65: We can safely drop this. It just got recently added to master. I believe we can just replace its usage within the PyG code-base with one of the aggregator modules from `nn.aggr`.

torch_geometric/nn/glob/__init__.py
```diff
@@ -35,3 +27,48 @@
 class GlobalAttention(AttentionalAggregation):
     def __call__(self, x, batch=None, size=None):
         return super().__call__(x, batch, dim_size=size)
+
+
+sum_aggr = SumAggregation()
+
+
+@deprecated(
```
rusty1s, line 35: Let's not deprecate them, they are used quite heavily. Maybe we can just move`glob/glob.py` to `pool/glob.py` instead. Here, we can then just deprecate its usage from the `nn.glob` package.

Padarn: Makes sense will update

rusty1s:
After this, we can also safely remove `nn/glob` folder by simply moving all functionality from `nn/glob/__init__.py` to `nn/glob.py`.

torch_geometric/nn/glob/__init__.py
Padarn: Okay. I'll go through and depreciate the use of this. 👍

Padarn:
I've made the change suggested but we currently have an issue using fx tracer with `Aggregation` because of the conditionals in `__call__`.

I'm not super familiar with `fx` (okay actually I know almost nothing about it), but some quick reading doesn't suggest an obvious solution.

Padarn:
After looking into the `to_hetero_transformer` a bit I think I get the basic idea of what is happening now and why it doesn't work.

I'm thinking maybe one way of solving this would be to rewrite `Aggregation into 
```
class Aggregation:
    def validate():
        ....
    def forward():
         ....
    def __call__(args):
        args = self.validate(args)
        return super().__call__(args)

```
Then when we do the `fx` transform, we could separate these and only trace the forward?

Does that make any sense?

rusty1s:
I think all we need to do is to add support for this here: https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/fx.py#L286

Padarn:
closing this in favour of https://github.com/pyg-team/pytorch_geometric/pull/5016 sorry - something really strange going on with pushing to this branch I can't figure out

Issue #4765, opened 2022-06-05
Support multiple node/edge type sampling using `NeighborLoader` and friends


Padarn:
### 🚀 The feature, motivation and pitch

The goal of this roadmap is to understand how we might support sampling across multiple node or edge types in `NeighborLoader` and `LinkNeighborLoader`. As an example, currently, the constructor of `NeighborLoader` takes a  single `NodeType` with an optional index. For example
```
NeighborLoader(
    data=data, 
    input_nodes=('paper', index)
)
```
However, a user (for example https://github.com/pyg-team/pytorch_geometric/discussions/4707) may reasonably want to instead sample from multiple edge types, so we would like to propose a way to support this directly. Thus we would like to support this for both `NeighborLoader` and `LinkNeighborLoader`.

## Interface

A proposed interface for this could be to extend the supported input types to include a tuple of tuples for the sampling index. This would probably mean also supporting different negative sampling ratios. An example of the new interface would be:
```
NeighborLoader(
    data=data, 
    input_nodes=(('paper', index), ('author', index))
)
```
The output of this for `NeighborLoader` could remain the same, but the sampling would be extended to cover the different input types.

For the case of `LinkNeighborLoader`, the interface change could be similar
```
LinkNeighborLoader(
    data=data, 
    edge_label_index =(
        (('paper', 'author'), index), 
        (('author', 'paper'), index))
)
```
However, in the case of the `LinkNeighborLoader` we store properties `data.edge_label` and `data.edge_label_index` to allow users to keep track of which edges the sampled edges correspond to, and whether or not they are negative samples. So we would need to instead have a dictionary of these.

To avoid confusion with the existing implementation, when using the new functionality the properties `data.edge_label` and `data.edge_label_index` would instead be stored on the edge type themselves:

```
data['edge_type'].edge_label
data['edge_type'].edge_label_index
```

## Risks 

This adds complication in usage, as to effectively predict on a mini-batch you would need to compute the heads for each node/edge type. This needs to explored through some examples and help from the community.

## Roadmap
- [ ] `NeighborLoader`.
- - [ ] Support multiple node types in `NeighborLoader`.
- - [ ] Examples of multiple node types in `NeighborLoader`.
- [ ] `LinkNeighborLoader`.
- - [ ] Support multiple node types in `LinkNeighborLoader`.
- - [ ] Examples of multiple node types in `LinkNeighborLoader`.
- [ ] Support multiple edge types in `LinkNeighborLoader`.
[...]

PR #5013, opened 2022-07-20
Support multiple node type sampling in `NeighborLoader`  

Padarn:
This PR adds functionality to allow for multiple node types to be sampled in `NeighbourLoader`. 

The interface looks as was discussed in the roadmap (https://github.com/pyg-team/pytorch_geometric/issues/4765):
```
NeighbourLoader(
   input_nodes=[
     ('paper', torch.LongTensor([0,1,2])), 
     ('author', torch.LongTensor([0,1,2])) 
   ]
  ...
)
```
Internally, it converts this to a list of tuples. 
```
[('paper', 0), ('paper', 1),....]
```
This is not very efficient, but benchmarks https://github.com/pyg-team/pytorch_geometric/issues/4765#issuecomment-1147105584 showed it to be acceptable.

TODO:
- [x] Add tests 
- [x] Add support for `None` instead of providing specific nodes for some node types

Addresses https://github.com/pyg-team/pytorch_geometric/issues/4765

### DIFF
--- a/test/loader/test_neighbor_loader.py
+++ b/test/loader/test_neighbor_loader.py
@@ -166,6 +166,43 @@ def test_heterogeneous_neighbor_loader(directed):
 
 
 @pytest.mark.parametrize('directed', [True, False])
+def test_heterogeneous_neighbor_loader_mixed_input(directed):
+    torch.manual_seed(12345)
+
+    data = HeteroData()
+
+    data['paper'].x = torch.arange(100)
+    data['author'].x = torch.arange(100, 300)
+
+    data['paper', 'paper'].edge_index = get_edge_index(100, 100, 500)
+    data['paper', 'paper'].edge_attr = torch.arange(500)
+    data['paper', 'author'].edge_index = get_edge_index(100, 200, 1000)
+    data['paper', 'author'].edge_attr = torch.arange(500, 1500)
+    data['author', 'paper'].edge_index = get_edge_index(200, 100, 1000)
+    data['author', 'paper'].edge_attr = torch.arange(1500, 2500)
+
+    batch_size = 20
+
+    loader = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes=[
+        ('paper', None)
+    ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    loader2 = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes=[
+        ('paper', None)
+    ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    loader3 = NeighborLoader(
+        data, num_neighbors=[10] * 2, input_nodes=[
+            ('paper', torch.arange(0, data['paper'].num_nodes,
+                                   dtype=torch.long))
+        ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    for batch1, batch2, batch3 in zip(loader, loader2, loader3):
+        assert torch.equal(batch1['paper'].x, batch2['paper'].x)
+        assert torch.equal(batch2['paper'].x, batch3['paper'].x)
+
+
+@pytest.mark.parametrize('directed', [True, False])
 def test_homogeneous_neighbor_loader_on_cora(get_dataset, directed):
     dataset = get_dataset(name='Cora')
     data = dataset[0]
--- a/test/loader/test_neighbor_loader.py
+++ b/test/loader/test_neighbor_loader.py
@@ -187,9 +187,9 @@ def test_heterogeneous_neighbor_loader_mixed_input(directed):
         ('paper', None)
     ], batch_size=batch_size, directed=directed, shuffle=False)
 
-    loader2 = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes=[
-        ('paper', None)
-    ], batch_size=batch_size, directed=directed, shuffle=False)
+    loader2 = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes='paper',
+                             batch_size=batch_size, directed=directed,
+                             shuffle=False)
 
     loader3 = NeighborLoader(
         data, num_neighbors=[10] * 2, input_nodes=[
--- a/test/loader/test_neighbor_loader.py
+++ b/test/loader/test_neighbor_loader.py
@@ -201,6 +201,17 @@ def test_heterogeneous_neighbor_loader_mixed_input(directed):
         assert torch.equal(batch1['paper'].x, batch2['paper'].x)
         assert torch.equal(batch2['paper'].x, batch3['paper'].x)
 
+    loader = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes=[
+        ('paper', None), ('author', None)
+    ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    for batch in loader:
+        paper_size = (batch['paper'].batch_size if hasattr(
+            batch['paper'], 'batch_size') else 0)
+        author_size = (batch['author'].batch_size if hasattr(
+            batch['author'], 'batch_size') else 0)
+        assert paper_size + author_size == batch_size
+
 
 @pytest.mark.parametrize('directed', [True, False])
 def test_homogeneous_neighbor_loader_on_cora(get_dataset, directed):
--- a/test/loader/test_neighbor_loader.py
+++ b/test/loader/test_neighbor_loader.py
@@ -172,45 +172,36 @@ def test_heterogeneous_neighbor_loader_mixed_input(directed):
     data = HeteroData()
 
     data['paper'].x = torch.arange(100)
-    data['author'].x = torch.arange(100, 300)
+    data['author'].x = torch.arange(100)
+    data['book'].x = torch.arange(100)
+    data['publisher'].x = torch.arange(100)
 
-    data['paper', 'paper'].edge_index = get_edge_index(100, 100, 500)
-    data['paper', 'paper'].edge_attr = torch.arange(500)
-    data['paper', 'author'].edge_index = get_edge_index(100, 200, 1000)
-    data['paper', 'author'].edge_attr = torch.arange(500, 1500)
-    data['author', 'paper'].edge_index = get_edge_index(200, 100, 1000)
-    data['author', 'paper'].edge_attr = torch.arange(1500, 2500)
+    data['paper', 'author'].edge_index = get_edge_index(100, 100, 500)
+    data['book', 'publisher'].edge_index = get_edge_index(100, 100, 500)
 
     batch_size = 20
 
     loader = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes=[
-        ('paper', None)
-    ], batch_size=batch_size, directed=directed, shuffle=False)
-
-    loader2 = NeighborLoader(data, num_neighbors=[10] * 2, input_nodes='paper',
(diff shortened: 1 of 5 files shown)


torch_geometric/loader/neighbor_loader.py
```diff
@@ -197,17 +211,13 @@ def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor]):
             )
         return node_dict, row_dict, col_dict, edge_dict
 
-    def __call__(self, index: Union[List[int], Tensor]):
-        if not isinstance(index, torch.LongTensor):
-            index = torch.LongTensor(index)
+    def __call__(self, index: Union[List[int], Tensor, HeteroNodeList]):
```
Padarn, line 214: @wsad1 Can I get your opinion on this (note the PR is overall still somewhat WIP).

I was finding the logic here getting quite complicated. Do you have any suggestions for simplification? One thing I was considering was providing a custom collate_fn that handles the conversion to some container like 
```
class HeteroNodes:
    ...
```
instead of doing all the type inference here.

wsad1: Thanks for the update. I'll check it out once I am back from a short vacation on Tuesday morning.

Padarn: No rush, thank you.

test/loader/test_neighbor_loader.py
```diff
@@ -166,6 +166,54 @@ def test_heterogeneous_neighbor_loader(directed):
...
+                             shuffle=False)
+
+    loader3 = NeighborLoader(
+        data, num_neighbors=[10] * 2, input_nodes=[
+            ('paper', torch.arange(0, data['paper'].num_nodes,
+                                   dtype=torch.long))
+        ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    for batch1, batch2, batch3 in zip(loader, loader2, loader3):
+        assert torch.allclose(batch1['paper'].x, batch2['paper'].x)
```
Padarn, line 202: This test is flakey - works locally but not in the CI. Any suggestions for what might be better?

wsad1 (commented):
Thanks at @Padarn for the update. I am still looking at your code to see if we can simply things. Added some initial comments.

test/loader/test_neighbor_loader.py
```diff
@@ -166,6 +166,54 @@ def test_heterogeneous_neighbor_loader(directed):
...
+    ], batch_size=batch_size, directed=directed, shuffle=False)
+
+    for batch in loader:
+        paper_size = (batch['paper'].batch_size if hasattr(
+            batch['paper'], 'batch_size') else 0)
+        author_size = (batch['author'].batch_size if hasattr(
+            batch['author'], 'batch_size') else 0)
+        assert paper_size + author_size == batch_size
+
+
```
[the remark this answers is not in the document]
Padarn: Good idea 👍

Padarn:
Thanks. FYI I'll also do this for the link loader and add examples in separate PRs, so we can iterate on the complexity later too if it seems okay but not perfect.

Padarn:
Hey @mananshah99 do you also want to take a look at this one? May need to merge it with https://github.com/pyg-team/pytorch_geometric/pull/5312

torch_geometric/loader/neighbor_loader.py
```diff
@@ -512,3 +538,25 @@ def to_index(tensor):
...
+        input_nodes[i] = get_sampling_nodes(data, input_nodes[i])
+    return to_hetero_types(input_nodes), to_hetero_list(input_nodes)
+
+
+def to_hetero_types(input_nodes: List[Tuple[str, Tensor]]) -> List[str]:
+    return [node_type for node_type, index in input_nodes]
+
+
+def to_hetero_list(input_nodes: List[Tuple[str, Tensor]]) -> HeteroNodeList:
+    return [(node_type, i) for node_type, index in input_nodes for i in index]
```
wsad1, line 555: There might be cases where we return nodes of only one type. Should we mention that this can't be used to train a model which predicts on multiple node types?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -512,3 +538,25 @@ def to_index(tensor):
...
+
+
+def get_mixed_sampling_nodes(data: HeteroData,
+                             input_nodes: List[InputNodes]) -> SamplingNodes:
+    for i in range(len(input_nodes)):
+        input_nodes[i] = get_sampling_nodes(data, input_nodes[i])
+    return to_hetero_types(input_nodes), to_hetero_list(input_nodes)
+
+
+def to_hetero_types(input_nodes: List[Tuple[str, Tensor]]) -> List[str]:
```
wsad1, line 550:
```suggestion
def get_node_types(input_nodes: List[Tuple[str, Tensor]]) -> List[str]:
```

wsad1 (commented):
Added some more comments will take a look again.

torch_geometric/loader/neighbor_loader.py
Padarn: Hmm I didn't fully understand this - do you mean some samples from the neighbor sampler will only contain a single type?

wsad1: Yeah it might contain samples from only one type. 
I am not sure what the use case of this multiple node sampling is. If an envisioned use case is is predicting on multiple node types in a hetero graph, that might not work.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -147,7 +152,8 @@ def _set_num_neighbors_and_num_hops(self, num_neighbors):
         # Add at least one element to the list to ensure `max` is well-defined
         self.num_hops = max([0] + [len(v) for v in num_neighbors.values()])
 
-    def _sparse_neighbor_sample(self, index: Tensor):
+    def _sparse_neighbor_sample(self, index: Union[List[int], Tensor]):
```
mananshah99, line 155: Why? Don't we convert to tensors beforehand?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -159,8 +165,10 @@ def _sparse_neighbor_sample(self, index: Tensor):
         )
         return node, row, col, edge
 
-    def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor],
+    def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str,
+                                                              torch.Tensor],
                                        **kwargs):
```
mananshah99, line 170:
```suggestion
                                       **kwargs,):
```
+ let yapf auto-format :)

torch_geometric/loader/neighbor_loader.py
```diff
@@ -196,19 +204,28 @@ def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor],
...
+
+            batch_sizes = {
+                node_type: index.numel()
+                for node_type, index in index_dict.items()
+            }
 
-        if self.data_cls != 'custom' and issubclass(self.data_cls, Data):
-            return self._sparse_neighbor_sample(index) + (index.numel(), )
+            return self._hetero_sparse_neighbor_sample(index_dict) + (
+                batch_sizes, )
```
mananshah99, line 225: This makes sense, and you are right that it somewhat conflicts with the interface in #5312. I think this is totally okay; that interface will likely change significantly over the coming week or two (we also need to support link-level neighbor sampling, etc.).

If it is alright with you, I would propose first merging 5312, and then adapting that interface as part of this PR to support returning a dict of batch sizes for each node type. Wdyt?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -494,3 +520,34 @@ def to_index(tensor):
...
+
+        node_type = getattr(input_nodes, 'group_name', None)
+        if not input_nodes.is_set('index') or input_nodes.index is None:
+            num_nodes = feature_store.get_tensor_size(input_nodes)[0]
+            return node_type, range(num_nodes)
+        return node_type, input_nodes.index
+
+
+def get_mixed_sampling_nodes(data: HeteroData,
+                             input_nodes: List[InputNodes]) -> SamplingNodes:
```
mananshah99, line 535: Document or perhaps rename? `mixed` isn't super clear to me (this is just getting sampling nodes for different node types, right?)

torch_geometric/loader/neighbor_loader.py
```diff
@@ -494,3 +520,34 @@ def to_index(tensor):
...
+        input_nodes[i] = get_sampling_nodes(data, input_nodes[i])
+    return get_node_types(input_nodes), get_node_list(input_nodes)
+
+
+def get_node_types(input_nodes: List[Tuple[str, Tensor]]) -> List[str]:
+    return [node_type for node_type, index in input_nodes]
+
+
+def get_node_list(input_nodes: List[Tuple[str, Tensor]]) -> HeteroNodeList:
+    return [(node_type, i) for node_type, index in input_nodes for i in index]
```
mananshah99, line 546: Do these need to be separate functions?

mananshah99 (commented):
Thanks for the ping @Padarn. Left a few comments, happy to chat further as well.

Padarn:
Thanks for the comments @mananshah99. Sorry I've been busy with my day job, will review and address over the weekend.

torch_geometric/loader/neighbor_loader.py
Padarn: Agreed! I see its merged now so let me rethink this a little based on what you've added.

torch_geometric/loader/neighbor_loader.py
Padarn: Yes good point, and I think you're right we could merge these two functions. I originally spit them because it was being used in the sampler init, but I've removed this. Let me clean it up.

torch_geometric/loader/neighbor_loader.py
Padarn: 👍 good point

torch_geometric/loader/neighbor_loader.py
Padarn: Ah because that made the logic in `__call__` quite confusing. You would have first to check whether or not it was hetero and then if it was handling the conversion separately if it was a mixed input. To me it seemed simpler to handle these cases in the individual functions, as it was actually less code.

Padarn:
On second look I think I'll wait for you to finish your current refactoring PRs, the code has changed a lot and I'll have to fit into the new interface. Will focus on helping review you PRs first and then rework this one.

mananshah99:
> On second look I think I'll wait for you to finish your current refactoring PRs, the code has changed a lot and I'll have to fit into the new interface. Will focus on helping review you PRs first and then rework this one.

Thank you for accommodating :) The refactoring PRs are complete now, and the interface is mostly stable. Happy to help move this implementation over behind the new interface, it's pretty cool.

Padarn:
Great! I can refactor later this week. I'll probably start a new PR as I think most of the code need to move, will definitely ask you for a review.

mayabechlerspeicher:
@Padarn Hi! Is this supported eventually? I am trying to pass multiple edge types for LinkNeighborLoader in edge_label_index, and can't find any working code example. 

Thanks!

ashrielbrian:
@mayabechlerspeicher Were you able to find a recourse? Only other option I'm leaning towards is to write a custom data loader

Issue #4543, opened 2022-04-26
Interface for datasets that are too large to use `InMemoryDataset`


hatemhelal:
### 🚀 The feature, motivation and pitch

There are several examples of datasets for molecular property prediction where each individual graph example easily fits in memory but there are too many examples to fit within the `InMemoryDataset` interface.  One solution is to save each example in its own `.pt` file but this introduces a significant filesystem overhead to access each example.  

A better solution is to partition the data such that there are many graphs serialised within a single `.pt` file.  The number of graphs can be considered a `chunk_size` parameter which is independent from the training `batch_size`. This `ChunkedDataset` interface would be expected to scale to as large a dataset as desired while avoiding the significant overhead of having one graph per file.  

The design idea is roughly:

* `ChunkedDataset` inherits from the PyG `Dataset` interface
* Accepts a `chunk_size` argument
* Has an abstract method `process_chunk` that accepts a list of data objects that can be processed and saved as a single `.pt` file. 

Other considerations:
* The training batch size should not depend on the `chunk_size` so the dataset
* `ChunkedDataset` should support splitting to read from parallel workers as well as random shuffling

rusty1s:
These are all pretty good thoughts. The `ChunkedDataset` API looks good to me. One additional consideration is that a `ChunkedDataset` likely requires modification to `torch_geometric.data.DataLoader` as well. We have two options here:
* `chunked_dataset[0]` returns a batch of graphs, and `DataLoader(chunked_dataset, batch_size=4)` determines how many chunks we want to batch together. This requires additional logic during `collate_fn`.
* `chunked_dataset` acts as a regular dataset where `chunked_dataset[0]` returns a single graph, but caches remaining data of that chunk in memory for faster access. This likely requires implementing our own `sampler` logic during data loading to guaranteee that consecutive examples are accessed (independent of `shuffle=True`)

Padarn:
Interesting idea for a feature: Seems like something a lot of people could possibly use.

Just wanted to query a bit on 
> One solution is to save each example in its own .pt file but this introduces a significant filesystem overhead to access each example.
Have you run any benchmarks on this? I don't know how big the graphs are usually, but I'm just curious how big of a difference the IO operations make to the overall processing rate.

Reason I ask is that chunking in files will make `shuffle=True` not actually shuffle randomly, and I'm wondering if there is a way we can retain this functionality. But maybe this is not important at all, just a thought :-)

rusty1s:
[...]

PR #5020, opened 2022-07-20
`IterableDataset`

josiahbjorgaard:
Add iterable dataset, inhereting from the pytorch object of the same name. Follow up to [issue 4543](https://github.com/pyg-team/pytorch_geometric/issues/4543).


josiahbjorgaard:
Ready for review

josiahbjorgaard:
Any recommendations on how best to get test coverage for download/process functions?

rusty1s:
Please don't worry too much about code coverage. I think the current tests look fine. Thanks!

torch_geometric/data/iterable_dataset.py
```diff
@@ -0,0 +1,170 @@
...
+import torch.utils.data
+
+from torch_geometric.data import Data
+from torch_geometric.data.makedirs import makedirs
+
+# IndexType = Union[slice, Tensor, np.ndarray, Sequence]
+
+
+class IterableDataset(torch.utils.data.IterableDataset):
+    r"""Dataset base class for creating graph datasets which are loaded as
```
Padarn, line 17: Could we add a bit more detail here about why you would want to do this?

torch_geometric/data/iterable_dataset.py
```diff
@@ -0,0 +1,170 @@
...
+        be present in order to skip downloading."""
+        raise NotImplementedError
+
+    @property
+    def processed_file_names(self) -> Union[str, List[str], Tuple]:
+        r"""The name of the files in the :obj:`self.processed_dir` folder that
+        must be present in order to skip processing."""
+        raise NotImplementedError
+
+    def download(self):
```
Padarn, line 48: I'm personally finding it a bit confusing why we want to add all these unimplemented functions to `IterableDataset`.

I feel it may be more meaningful to either 
- create a more concrete version of this, which takes a list of file paths and processes them as required
- remove the file/remote storage parts of this for now

WDTY?

test/data/test_dataset.py
```diff
@@ -1,7 +1,55 @@
...
+from torch_geometric.data import (
+    Data,
+    Dataset,
+    HeteroData,
+    InMemoryDataset,
+    IterableDataset,
+)
+
+
+class MyTestIterableDataset(IterableDataset):
```
Padarn, line 13: Could we add a test that also uses the `transform` functionality?

Padarn (changes requested):

torch_geometric/data/iterable_dataset.py
josiahbjorgaard: First off, I'm happy to implement whatever the maintainers find to be the most meaningful functionality here. Let me provide some more background on this PR.

I noted that the issue linked in the description of this PR describes having the processing of raw files and transforms as desired functionality from a 'chunked' or grouped data loader.

I use this contributed base class to implement a loader which processes a raw data set, save it in groups of samples (of heterodata graphs) according to the number of samples I'd like to store in a group, then loads the groups into memory several at a time, asynchronously. This allows us to tune the amount memory that our dataset uses for training with large amounts of data that would be prohibitive to load at once, while preventing performance issues associated with loading samples one-by-one from disk and the overhead of having very many small files. The full implementation is something that I can contribute. An idea is to provide this as an example implementation of the IterableDataset in documentation. I would like to know how best to 

I think there is a reason to retain the IterableDataset as a base class - other applications of a Dataset as an Iterable are  loading data from a database, from a stream, or from other sources which require code to execute at the time that a sample is requested by the training code.

Would be great to know what you think about this and how best to contribute.

CHANGELOG.md
```diff
@@ -3,8 +3,9 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-## [2.2.0] - 2022-MM-DD
+## [2.2.0] - 2022-08-10
```
Padarn, line 6: I think this is intentionally not specified until we decide on the release date?

torch_geometric/data/iterable_dataset.py
Padarn: With some of the updates, it's more clear to me. I like the idea of adding such a base class, but in the earlier version of this, the reimplemented methods didn't look to be used by any of the base class-provided functionality.

Padarn: Sorry, I didn't mean to suggest that the idea of having a base class for this was bad, just that I couldn't quite follow what the user was expected to implement to make the `iter` work. It looks like this could be provided in the base class, as (given the current implementation) you do need to work with files?

As it is now, we could subclass `Dataset` and just add an `__iter__`.

I don't have a strong opinion here, the design just isn't clear to me.

josiahbjorgaard: Yes, one would need to work with files during both processing and data loading in some implementations of an iterable dataset.

Pytorch's IterableDataset is a subclass of Pytorch's Dataset, as well - and those are both data containers in PyTorch Lightning https://pytorch-lightning.readthedocs.io/en/stable/guides/data.html

I think both ways work - either subclassing from Pytorch's IterableDataset or subclassing PyG's Dataset, just let me know. It seemed the most straightforward to inherit from PyTorch and parallel the PyTorch Lightning data container structure in PyG when I implemented a chunked dataloader with this.

Padarn: Well I'm okay either way. Leave it to others to follow up. Thanks for explaining your thoughts.

PR #5021, opened 2022-07-21
Fix `fx` problem for `Aggregation`

Padarn:
This PR addresses a problem we had when using `torch.fx` on our new `nn.aggr.Aggregation` class. This class redefines the `__call__` function to contain:

```python
     def __call__(self, x: Tensor, index: Optional[Tensor] = None,
                 dim: int = -2, **kwargs) -> Tensor:

        if dim >= x.dim() or dim < -x.dim():
            raise ValueError(f"Encountered invalid dimension '{dim}' of "
                             f"source tensor with {x.dim()} dimensions")
        ...
```

During a `torch.fx._symbolic_trace` pytorch modules `torch.nn.Module` calls are excluded from symbolic trancing intentionally using `is_leaf_module`; this allows for avoiding errors on symbolic tracing of for example, conditions statements involving tensors (which are not handled). 

However, this is implemented by patching a specific function within `torch.fx._symbolic_trace`:
```python
patcher.patch_method(torch.nn.Module, "__call__", module_call_wrapper, deduplicate=False)
```
from [here](https://github.com/pytorch/pytorch/blob/e68583b4d180066b8e4f108e0d23176a2676421c/torch/fx/_symbolic_trace.py#L702).

This doesn't patch our `__call__` as it no longer lives in `torch.nn.Module`.

The fix here is to reimplemnt the `trace` function to add an extra wrapper around our code.

Note that this should be replaced by a better solution once one is available, by for example:
- Allowing for `kwargs` in `pre_forward_hooks`: https://github.com/pytorch/pytorch/issues/35643#issuecomment-1191258877
- Getting a better way to extend the patched functions from the `torch.fx` team.


torch_geometric/nn/fx.py
```diff
@@ -284,7 +283,9 @@ def is_message_passing_op(module: Module, op: str, target: str) -> bool:
 
 
 def is_global_pooling_op(module: Module, op: str, target: str) -> bool:
-    from torch_geometric.nn import GlobalPooling
+    from torch_geometric.nn import Aggregation, GlobalPooling
     if op == 'call_module':
-        return isinstance(get_submodule(module, target), GlobalPooling)
+        return isinstance(get_submodule(module, target),
+                          GlobalPooling) or isinstance(
```
rusty1s, line 289: Let‘s just replace it here.

torch_geometric/nn/aggr/base.py
```diff
@@ -7,13 +7,53 @@
...
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2, **kwargs) -> Tensor:
+        x, index, ptr, dim_size, dim = validate(x, index, ptr, dim_size, dim)
+        return self._forward_sub(x, index, ptr, dim_size, dim, **kwargs)
 
     # @abstractmethod
     def forward(self, x: Tensor, index: Optional[Tensor] = None,
                 ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
-                dim: int = -2) -> Tensor:
+                dim: int = -2, **kwargs) -> Tensor:
```
rusty1s, line 56: Why do we need to add kwargs here in the first place?

torch_geometric/nn/aggr/base.py
```diff
@@ -7,13 +7,53 @@
...
+
+    return x, index, ptr, dim_size, dim
+
+
 class Aggregation(torch.nn.Module):
     r"""An abstract base class for implementing custom aggregations."""
+    def __init__(self):
+        super().__init__()
+        self._forward_sub = self.forward
+        self.forward = self._foward
```
rusty1s, line 45: I think this is overly complicated. How about we just do
```
def forward:
    self.validate()
    self._forward()
```
And overwrite `_forward` in child modules?

torch_geometric/nn/aggr/base.py
```diff
@@ -35,35 +75,6 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
     def reset_parameters(self):
         pass
 
-    def __call__(self, x: Tensor, index: Optional[Tensor] = None,
```
rusty1s, line 38: Just so I understand better: what was the issue with this code to begin with? It looks like the tracer has problems when overwriting __call__. Is that correct? Is there any way to fix this in the tracer?

torch_geometric/nn/aggr/base.py
Padarn: because of the `GraphMultisetTransformer` - which also takes `edge_index` optionally

torch_geometric/nn/aggr/base.py
Padarn: yeah I was thinking that too - just providing this option incase we want to keep the  current abstract interface

torch_geometric/nn/aggr/base.py
Padarn: yes, its because of this: https://github.com/pytorch/pytorch/blob/e68583b4d180066b8e4f108e0d23176a2676421c/torch/fx/_symbolic_trace.py#L702
the tracer only looks for a leave module in the patched methods - but since we overwrite call its no longer patched (the `super().__call__`  is still patched)

torch_geometric/nn/aggr/base.py
Padarn: If we  can just add this to the default args then the problem is solved, but I don't see it as a common need

torch_geometric/nn/aggr/base.py
rusty1s: Can we patch it ourselves?

Padarn: hmm, yes... but we'll have to copy quite a bit of code from their implementation. I can implement that and we can see what it looks like?

rusty1s: Ok, depends on how easy it would be to try out and how large the code to patch would be. Do not burn too much time into it if it is not easy to fix :)

Padarn: I'll give it a go, and abandon if I am up too late 😁

Padarn: I pushed a working version (its a bit messy - I can cleanup what I can if we want to use it, so no need to review for style yet)

Additionally, I raised a question on the PyTorch slack channel to see if anyone had suggestions but nothing came back yet. However there is an open issue I can try and help with in pytorch to support kwargs in pre-forward-hooks (which I think would be a nice solution): https://github.com/pytorch/pytorch/issues/35643#issuecomment-1191258877

So I'd say the options now are:
- change the interface so subclasses implement `_forward`
- use this nasty approach in `fx` until we have a better solution (like the pre-forward-hooks)
- depreciate global pooling support in the hetero conversion for now and then come back to it after the above

WDYT?

rusty1s: I actually think the
approach of patching isn‘t too bad, thanks for digging into this. You rock!

i was wondering if we really need to duplicate the full method here or can work with `super` instead. It looks like we only really need to do:
```python
            @st.functools.wraps(st._orig_module_call)
             def module_call_wrapper(mod, *args, **kwargs):
                 def forward(*args, **kwargs):
                     return st._orig_module_call(mod, *args, **kwargs)

                 st._autowrap_check(
                     patcher,
                     getattr(getattr(mod, "forward", mod), "__globals__", {}),
                     self._autowrap_function_ids)
                 return self.call_module(mod, forward, args, kwargs)

             with st._Patcher() as patcher:
                  patcher.patch_method(Aggregation, "__call__",
                                      module_call_wrapper, deduplicate=False)
```

Padarn: I thought about this, but I don't know how the `_Patcher` context manager works, so I was afraid to suggest it even if it happened to pass tests. 

Could certainly invest time in understanding if you think its a big upgrade thouh.

rusty1s: Can we quickly test if it works? I am always in favor of reducing number of lines :)

Padarn: Hmm, no doesn't work 'out of the box'.

rusty1s: Ok, then leave it as it is. Might be good to add some exhaustive documentation here, and also link to the PyTorch file this code is coming from.

Padarn: Agreed. Will keep an eye open for better solutions too.

Padarn:
@rusty1s I've added a comment linking to this issue and added details in the description here. I think it is most useful for people looking at this later to be able to come back to this issue if they have any concerns.

torch_geometric/nn/fx.py
```diff
@@ -1,13 +1,17 @@
...
-from torch.nn import Module, ModuleDict, ModuleList, Sequential
 
 try:
-    from torch.fx import Graph, GraphModule, Node
+    import torch.fx._symbolic_trace as st
+    from torch.nn import Module, ModuleDict, ModuleList, Sequential
 except (ImportError, ModuleNotFoundError, AttributeError):
     GraphModule, Graph, Node = 'GraphModule', 'Graph', 'Node'
+    st = None
+    warnings.warn("`torch.fx._symbolic_trace` could not be imported:"
```
Padarn, line 13: This is a bit nasty. I'd like to require it really but the documentation build failed.

rusty1s: What do you mean? Why print the warning here?

torch_geometric/nn/fx.py
```diff
@@ -261,11 +265,99 @@ def maybe_replace_node(n: Node) -> Node:
 def symbolic_trace(
         module: Module,
         concrete_args: Optional[Dict[str, Any]] = None) -> GraphModule:
+
+    from torch_geometric.nn import Aggregation
```
rusty1s, line 269:
```suggestion
    from torch_geometric.nn import Aggregation
    import torch.fx._symbolic_trace as st
```
How about importing it here?

rusty1s (approved):
Would like to remove the warning before merging - otherwise LGTM. Thanks!

Padarn:
Will do. Thanks!

torch_geometric/nn/fx.py
Padarn: I don't like imports inside classes.. but it makes a lot of sense here. Thanks :-)

PR #5026, opened 2022-07-21
Adding `SKLearnPreprocessTransform`

snknitin:
Based on discussion in https://github.com/pyg-team/pytorch_geometric/discussions/5022


torch_geometric/transforms/SKLearnPreprocessTransform.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, attrs: List[str] = ["edge_attr"]):
+        self.attrs = attrs
+
+    def __call__(self, data: Union[Data, HeteroData]):
+        for edge, store in list(zip(data.edge_types, data.edge_stores)):
+            for key, value in store.items(*self.attrs):
+                if "edge_scaler" in store.keys():
+                    scaler = store["edge_scaler"]
+                    # needs to be a float tensor since this is the edge_attr
+                    value = torch.tensor(scaler.transform(value),
```
wsad1, line 33: `scaler.transform(value)` what data type are we assuming `value` is here?

torch_geometric/transforms/SKLearnPreprocessTransform.py
```diff
@@ -0,0 +1,40 @@
...
+
+    where the scaler can be StandardScaler or MinMaxScaler from scikit-learn that
+    has been fit to the total dataset
+
+    Args:
+        attrs (List[str]): The names of attributes to normalize.
+            (default: :obj:`["edge_attr"]`)
+    """
+    def __init__(self, attrs: List[str] = ["edge_attr"]):
+        self.attrs = attrs
```
wsad1, line 25: From `__call__` it looks like only edge attributes are transformed. Do we want to support modifying node attrs too.

wsad1 (changes requested):
Thanks for adding this. Left a few comments.

torch_geometric/transforms/SKLearnPreprocessTransform.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __call__(self, data: Union[Data, HeteroData]):
+        for edge, store in list(zip(data.edge_types, data.edge_stores)):
```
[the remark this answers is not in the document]
snknitin: In my current implementation, I had 2 scalers for 2 edge_types , so i add them like `data[edge].edge_scaler = ...` similar to how we have edge_index, edge_attr and edge_label.  Then I access it in the call, since the scaler needs to be fit on the complete data(multiple graphs like Dataset(30)) and not just data[0] which is just one graph. However, I think this can be a post processing step, where the transform takes in a Data/HeteroData object and calculates the scaler within the transform call instead of it being added to the data object

torch_geometric/transforms/SKLearnPreprocessTransform.py
snknitin: The value is going to be Float Tensor, since it is an edge_attr, but the scaler.transform will convert it into a numpy array. So i mapped it to tensor again, because if the edge_store is modified to to numpy array, some of the other transforms in the Compose pipeline would skip the edge_attr . I noticed this in case of T.ToUndireceted where we add reverse edges but only edge_index was present while edge_attr was ignored since that failed a conditional statement.

torch_geometric/transforms/SKLearnPreprocessTransform.py
snknitin: Yes, I think moving forward we should have support to scale `data[node].x` too. For this, i think we will need to consider all the `unique node features`. For example, if a Dataset contained 10 graphs in it, Dataset(10) , where each graph had ~100 nodes that were a sampled subset from a total collection 300 nodes , there will be repeats if we concatenate them all, to fit a scaler to it,  and some missing nodes that should be considered for min,std,max as well that could be held out in test. Not entirely clear on how to approach this yet. 

Once we have a scaler for each element of `data.metadata`(node_types and edge_types), we can transform individual graphs' nodes and edges accordingly.

torch_geometric/transforms/SKLearnPreprocessTransform.py
```diff
@@ -0,0 +1,40 @@
...
+
+import torch
+
+from torch_geometric.data import Data, HeteroData
+from torch_geometric.data.datapipes import functional_transform
+from torch_geometric.transforms import BaseTransform
+
+
+@functional_transform('sklearn_preprocess')
+class SKLearnPreprocessTransform(BaseTransform):
```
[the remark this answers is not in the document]
snknitin: Yup, will do. I'll add the test. I'm thinking of modifying the approach in this transform, so that it doesn't need the hack to send in an edge scaler and is agnostic to any sklearn method or pipeline. I got busy with another bug i was dealing with, but now i can work on this

PR #5027, opened 2022-07-21
Bug Fix: `collate` method requires keys in dict to be strings, but if `Dict[int:Tensor]`, `_collate` fails

ItamarChinn:
`_collate` in `torch_geometric/data/collate.py` expects all keys to be strings.
In the case where you have a dictionary of Tensors, line 183 will bring you to line 127. Then `__cat_dim__(key, elem, stores[0])` expects the key from your original dict to be a string. For many dictionaries with keys which are hashable non-strings (such as ints) and mapped to Tensors `__cat_dim__` will then fail, as will `get_incs(key, values, data_list, stores)` on line 132. This is because many internal pyg methods expect keys to be strings. 

Instead of changing many internal methods (such as `__cat_dim__` and `__inc__` in `Data` objects) simply make a copy of the `key` as a string (`str_key`), since it is used as a check in this method (in `__cat_dim__` and `__inc__`) there is no reason to mutate `key`. Making a copy is safer. 

The first commit in this pull request is reverted by the second (I originally thought of checking if `key` is a `str` in `__cat_dim__` but opted changing `_collate`

### DIFF
--- a/torch_geometric/data/collate.py
+++ b/torch_geometric/data/collate.py
@@ -125,15 +125,12 @@ def _collate(
         # Concatenate a list of `torch.Tensor` along the `cat_dim`.
         # NOTE: We need to take care of incrementing elements appropriately.
         key = str(key)
-            str_key = key
-        else:
-            str_key = str(key)
-        cat_dim = data_list[0].__cat_dim__(str_key, elem, stores[0])
+        cat_dim = data_list[0].__cat_dim__(key, elem, stores[0])
         if cat_dim is None or elem.dim() == 0:
             values = [value.unsqueeze(0) for value in values]
         slices = cumsum([value.size(cat_dim or 0) for value in values])
         if increment:
-            incs = get_incs(str_key, values, data_list, stores)
+            incs = get_incs(key, values, data_list, stores)
             if incs.dim() > 1 or int(incs[-1]) != 0:
                 values = [
                     value + inc.to(value.device)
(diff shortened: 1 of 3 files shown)


torch_geometric/data/collate.py
```diff
@@ -124,12 +124,16 @@ def _collate(
     if isinstance(elem, Tensor):
         # Concatenate a list of `torch.Tensor` along the `cat_dim`.
         # NOTE: We need to take care of incrementing elements appropriately.
-        cat_dim = data_list[0].__cat_dim__(key, elem, stores[0])
+        if isinstance(key, str):
```
rusty1s, line 127:
```suggestion
        key = str(key)
```
Can‘t we move this to a one-liner?

rusty1s:
I think this may need fixing in `separate` as well. WDYT?

ItamarChinn:
> I think this may need fixing in `separate` as well. WDYT?

Done. I cast all keys to strings (one line).

torch_geometric/data/collate.py
ItamarChinn: As long as this doesn't mutate key from the outer frame this should be fine. If it does mutate it, then this may have side effects (such as making all the keys in any collated dictionaries strings instead of other hashable types, ie ints)

rusty1s: I think this should be fine :)

Issue #5009, opened 2022-07-19
Support for multiple aggregations in `SAGEConv`


lightaime:

lightaime:
The `SAGEConv` should support multiple aggregations already by using `combine`(https://github.com/pyg-team/pytorch_geometric/pull/5000) by setting the correct `in_channels`. Do we still need a specific fix for this? @rusty1s

rusty1s:
I always thought it might be cool to add support for this without the user needing to specify a different number of input channels (so that `SAGEConv` is tunable across a variety of different aggregations). WDYT?

lightaime:
Sure, that sounds good! Will open a PR later to use `combine` to support this.

PR #5033, opened 2022-07-22
Support for multiple aggregations in `SAGEConv`

lightaime:
This PR supports auto-infer the number of `channels` for multiple aggregations in `SAGEConv`.

Addresses item https://github.com/pyg-team/pytorch_geometric/issues/5009 in https://github.com/pyg-team/pytorch_geometric/issues/4712.


torch_geometric/nn/conv/sage_conv.py
```diff
@@ -83,8 +84,28 @@ def __init__(
...
-            kwargs['aggr_kwargs'] = dict(in_channels=in_channels[0],
-                                         out_channels=in_channels[0])
+            kwargs.setdefault('aggr_kwargs', {})
+            kwargs['aggr_kwargs'].setdefault('in_channels', in_channels[0])
+            kwargs['aggr_kwargs'].setdefault('out_channels', in_channels[0])
+
+        aggr_out_channels = in_channels[0]
+        if isinstance(aggr, List) and len(aggr) > 1:
+            kwargs.setdefault('aggr_kwargs', {})
+            kwargs['aggr_kwargs'].setdefault('mode', 'cat')
```
rusty1s, line 94: No reason to set them since they are the default.

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -83,8 +84,28 @@ def __init__(
...
+            kwargs['aggr_kwargs'].setdefault('mode_kwargs', {})
+            combine_mode = kwargs['aggr_kwargs']['mode']
+            if combine_mode == 'cat':
```
rusty1s, line 97: Can we rather integrate this to something like `MultiAggregation.get_out_channels(in_channels)`?

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -83,8 +84,28 @@ def __init__(
...
+                aggr_out_channels = in_channels[0] * len(aggr)
+            if combine_mode == 'proj' or combine_mode == 'attn':
+                kwargs['aggr_kwargs']['mode_kwargs'].setdefault(
+                    'in_channels', in_channels[0])
+                kwargs['aggr_kwargs']['mode_kwargs'].setdefault(
+                    'out_channels', in_channels[0])
+                aggr_out_channels = kwargs['aggr_kwargs']['mode_kwargs'][
+                    'out_channels']
+                if combine_mode == 'attn':
```
rusty1s, line 106: Let’s drop since it is the default anyway.

torch_geometric/nn/conv/sage_conv.py
```diff
@@ -83,8 +84,28 @@ def __init__(
...
+
+        aggr_out_channels = in_channels[0]
+        if isinstance(aggr, List) and len(aggr) > 1:
+            kwargs.setdefault('aggr_kwargs', {})
+            kwargs['aggr_kwargs'].setdefault('mode', 'cat')
+            kwargs['aggr_kwargs'].setdefault('mode_kwargs', {})
+            combine_mode = kwargs['aggr_kwargs']['mode']
+            if combine_mode == 'cat':
+                aggr_out_channels = in_channels[0] * len(aggr)
+            if combine_mode == 'proj' or combine_mode == 'attn':
```
rusty1s, line 99: Here again. Let‘s try to not add any custom combine mode logic here. If the user sets a custom combine mode. They will be responsible for setting mode_kwargs as well.

torch_geometric/nn/conv/sage_conv.py
lightaime: Maka sense

torch_geometric/nn/aggr/multi.py
```diff
@@ -114,6 +116,14 @@ def reset_parameters(self):
         if hasattr(self, 'multihead_attn'):
             self.multihead_attn._reset_parameters()
 
+    def get_out_channels(self, in_channels):
+        if self.out_channels is not None:
+            return self.out_channels
+        if self.mode == 'cat':
```
lightaime, line 122: This may not work. If each aggregator has customized `out_channels` such as `LSTMAggregation`. Do we want to add `get_out_channels` to all the `Aggregation`?

rusty1s: We can maybe add it as a TODO? I think having support for standard aggregators such as mean, max and mean is sufficient for now.

lightaime: Sounds good

torch_geometric/nn/aggr/multi.py
```diff
@@ -114,6 +116,16 @@ def reset_parameters(self):
         if hasattr(self, 'multihead_attn'):
             self.multihead_attn._reset_parameters()
 
+    def get_out_channels(self, in_channels):
```
rusty1s, line 119:
```suggestion
    def get_out_channels(self, in_channels: int) -> int:
```

torch_geometric/nn/aggr/multi.py
```diff
@@ -114,6 +116,16 @@ def reset_parameters(self):
...
+        if self.out_channels is not None:
+            return self.out_channels
+
+        # TODO: Support having customized `out_channels` in each aggregation
+        if self.mode == 'cat':
+            return in_channels * len(self.aggrs)
+        else:
```
rusty1s, line 126: The else can be dropped.

rusty1s (approved):
I like it. Thank you!

PR #5034, opened 2022-07-22
Raise error for one single aggregation when `proj` or `attn` mode is used in `MultiAggregation`

lightaime:
This PR raises error for one single aggregation when `proj` or `attn` mode is used in `MultiAggregation` to avoid unwanted behaviors.


torch_geometric/nn/aggr/multi.py
```diff
@@ -66,6 +66,9 @@ def __init__(
         self.mode = mode
         mode_kwargs = mode_kwargs or {}
         if mode == 'proj' or mode == 'attn':
+            if len(aggrs) == 1:
+                raise ValueError("One single aggregation does not support "
```
Padarn, line 70:
```suggestion
                raise ValueError("Multiple aggregations required for "
```
I feel is slightly more clear

Padarn (approved):

rusty1s (approved):

torch_geometric/nn/aggr/multi.py
rusty1s: Sounds good.

lightaime: Thanks!

PR #5046, opened 2022-07-25
Support `SparseTensor` label in `LightGCN`

hzg0601:
fix the bug of models/lightgcn when edge_indx is an instance of SparseTensor in forward method


torch_geometric/nn/models/lightgcn.py
```diff
@@ -110,7 +111,21 @@ def forward(self, edge_index: Adj,
                 :obj:`edge_index` will be used instead. (default: :obj:`None`)
         """
         if edge_label_index is None:
-            edge_label_index = edge_index
+            if isinstance(edge_index, SparseTensor):
+                edge_label_index = edge_index.to_torch_sparse_coo_tensor(
```
rusty1s, line 115:
```suggestion
                edge_label_index = torch.stack(edge_index.coo()[:2], dim=0)
```

torch_geometric/nn/models/lightgcn.py
```diff
@@ -110,7 +111,21 @@ def forward(self, edge_index: Adj,
...
+                ).coalesce().indices()
+            elif isinstance(edge_index, torch.Tensor) and edge_index.is_sparse(
```
rusty1s, line 117: I don‘t think we need to support this case right now. Any `tensor.is_sparse` logic should be dropped here.

rusty1s (approved):
Thank you!

PR #5049, opened 2022-07-25
Support for weighted metapaths

kpstesla:
This PR adds support for weighted metapaths, by adding two keyword arguments to `AddMetapaths()` in `torch_geometric/transforms/add_metapaths.py`.  The computed weights are stored in the `edge_attr` attribute of the edge.

There are two different ways to compute weights for metapaths.  

First,  by setting `weighted=True`, the computed weights will be the number of actual paths along the metapath for each metapath edge.  For example, if we had the metagraph A -> B -> C, with edges a_0 -> b_0 , a_0 -> b_1, b_0 -> c_0, and b_1 -> c_0, then the metapath A -> C would have the edge a_0 -> c_0, with weight 2, as there are two paths from a_0 to c_0 (one through b_0 and one through b_1).

Second, by setting `weighted=True` and `ues_edge_attr_as_weights=True`, the computation will use the existing edge_attr for each edge type along the metapath, and the resulting edge weights for each metapath edge will be the weighted sum of all paths between the start and end of that edge, where the weight of a path is the product of the edge_attr values for each edge along the path.  This option was added to allow for efficient, weighted computations for metapaths of metapaths.


test/transforms/test_add_metapaths.py
```diff
@@ -87,3 +87,48 @@ def test_add_metapaths_max_sample():
...
+    assert (meta['a', 'a'].edge_attr == torch.Tensor([1, 2, 2,
+                                                      4])).all().item()
+    del meta['a', 'c']
+    del meta['a', 'd']
+    del meta['a', 'a']
+
+    # Compute intra-table metapath efficiently
+    metapaths = [[('a', 'b'), ('b', 'c'), ('c', 'd')]]
+    meta = AddMetaPaths(metapaths, weighted=True)(data)
+    meta['d', 'a'].edge_index = meta['a', 'd'].edge_index.flip([0])
```
RexYing, line 126: perhaps it will be better named as graph or metapath_graph since this variable is GraphData / HeteroData

torch_geometric/transforms/add_metapaths.py
```diff
@@ -87,15 +87,20 @@ class AddMetaPaths(BaseTransform):
         max_sample (int, optional): If set, will sample at maximum
             :obj:`max_sample` neighbors within metapaths. Useful in order to
             tackle very dense metapath edges. (default: :obj:`None`)
+        weighted (bool, optional): If set to :obj:`True` compute weights for
+            each metapath and store them in :obj:`edge_attr`.
+            (default :obj:`False)
+        use_edge_attr_as_weights (bool, optional): If set to :obj:`True`
```
RexYing, line 93: is this variable only useful conditioned on the previous argument? Maybe we could document a bit more about the relations with the weighted argument

torch_geometric/transforms/add_metapaths.py
```diff
@@ -168,3 +180,16 @@ def sample_adj(self, adj: SparseTensor) -> SparseTensor:
...
         mask = torch.rand_like(prob) < prob
         return adj.masked_select_nnz(mask, layout='coo')
+
+    def _get_edge_attr(self, data: HeteroData,
+                       edge_type: EdgeType) -> torch.Tensor:
+        if self.weighted:
+            if self.use_edge_attr_as_weights and \
+                    edge_type in data.edge_attr_dict.keys():
+                edge_attr = data[edge_type].edge_attr
+                assert len(edge_attr.shape) == 1
```
RexYing, line 190:
```suggestion
                assert edge_attr.ndim == 1
```

torch_geometric/transforms/add_metapaths.py
```diff
@@ -119,26 +126,31 @@ def __call__(self, data: HeteroData) -> HeteroData:
...
+                    sparse_sizes=data[edge_type].size(),
+                    edge_attr=adj2_edge_attr)
 
                 adj1 = adj1 @ adj2
 
                 if self.max_sample is not None:
                     adj1 = self.sample_adj(adj1)
 
-            row, col, _ = adj1.coo()
+            row, col, value = adj1.coo()
```
RexYing, line 149: how about calling it `weights` to be consistent with the argument name `weighted`

torch_geometric/transforms/add_metapaths.py
```diff
@@ -168,3 +180,16 @@ def sample_adj(self, adj: SparseTensor) -> SparseTensor:
         prob = (self.max_sample * (1. / deg))[row]
         mask = torch.rand_like(prob) < prob
         return adj.masked_select_nnz(mask, layout='coo')
+
+    def _get_edge_attr(self, data: HeteroData,
+                       edge_type: EdgeType) -> torch.Tensor:
+        if self.weighted:
+            if self.use_edge_attr_as_weights and \
+                    edge_type in data.edge_attr_dict.keys():
+                edge_attr = data[edge_type].edge_attr
```
RexYing, line 189: I'm thinking if we could have users to have a custom edge attribute name that can be used to specify weights.
E.g. the user might want to call it `edge_weights` or `edge_multiplicity` or whatever, so that `edge_attr` can still be used for other purposes (such as in message function of the `MessagePassing` class)`.

For example, [here](https://github.com/pyg-team/pytorch_geometric/blob/575611f4f5e2209c7923dba977a1eebc207bd2e2/torch_geometric/loader/neighbor_loader.py#L97), `time_attr` is defined as a string (we could name it as `time_col` or other names.

RexYing (commented):
Overall looks good! Minor suggestions. We could also add @mananshah99 here since Matthias is on vacation.

test/transforms/test_add_metapaths.py
kpstesla: Yeah.  It's an in place operation so I'll just call it data.

torch_geometric/transforms/add_metapaths.py
kpstesla: Yes.  Clarified.

torch_geometric/transforms/add_metapaths.py
kpstesla: Good idea. I think I can get rid of the second extra argument and just assume that edge_multiplicity is defined then it should be used.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -87,15 +87,23 @@ class AddMetaPaths(BaseTransform):
...
             :obj:`max_sample` neighbors within metapaths. Useful in order to
             tackle very dense metapath edges. (default: :obj:`None`)
+        weighted (bool, optional): If set to :obj:`True` compute weights for
+            each metapath and store them in :obj:`edge_attr`. If
+            :obj:`use_edge_attr_as_weights` is :obj:`True`, then weights
+            will be computed from the existing :obj:`edge_attr`, otherwise
+            weights will be computed with a starting value of one for each
+            edge.
+            (default :obj:`False`)
+        use_edge_attr_as_weights (bool, optional): If set to :obj:`True`
```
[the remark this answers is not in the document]
kpstesla: whoops yes I forgot to update the documentation after removing that kwarg.

test/transforms/test_add_metapaths.py
```diff
@@ -87,3 +87,51 @@ def test_add_metapaths_max_sample():
...
+
+    metapaths = [[('a', 'b'), ('b', 'c')], [('a', 'b'), ('b', 'c'),
+                                            ('c', 'd')],
+                 [('a', 'b'), ('b', 'c'), ('c', 'd'), ('d', 'c'), ('c', 'b'),
+                  ('b', 'a')]]
+    transform = AddMetaPaths(metapaths, max_sample=None, weighted=True)
+    data = transform(data)
+
+    # Make sure manually added metapaths compute the correct number of edges
+    assert (data['a',
```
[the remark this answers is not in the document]
kpstesla: Yes this is the same thing, but the suggestion is a lot cleaner.

kpstesla:
Thanks @RexYing @rusty1s for the review!  Most of the changes I made were cosmetic, so let me know if there's anything else I need to change.

test/transforms/test_add_metapaths.py
```diff
@@ -87,3 +87,46 @@ def test_add_metapaths_max_sample():
...
+
+    # Make sure manually added metapaths compute the correct number of edges
+    assert metapath_data['a', 'c'].edge_weight.tolist() == [1, 2]
+    assert metapath_data['a', 'd'].edge_weight.tolist() == [1, 2]
+    assert metapath_data['a', 'a'].edge_weight.tolist() == [1, 2, 2, 4]
+
+    # Compute intra-table metapaths efficiently
+    metapaths = [[('a', 'b'), ('b', 'c'), ('c', 'd')]]
+    metapath_data = AddMetaPaths(metapaths, weighted=True)(copy.copy(data))
+    metapath_data['d',
```
rusty1s, line 124: This has weird auto-formatting. Maybe just rename to `out`?

test/transforms/test_add_metapaths.py
```diff
@@ -87,3 +87,46 @@ def test_add_metapaths_max_sample():
...
+                  'a'].edge_index = metapath_data['a',
+                                                  'd'].edge_index.flip([0])
+    metapath_data['d', 'a'].edge_weight = metapath_data['a', 'd'].edge_weight
+    metapaths = [[('a', 'd'), ('d', 'a')]]
+    metapath_data = AddMetaPaths(metapaths, weighted=True)(metapath_data)
+    del metapath_data['a', 'd']
```
rusty1s, line 130: Can be dropped?

rusty1s (approved):
Beautiful, thank you!

test/transforms/test_add_metapaths.py
kpstesla: done.

test/transforms/test_add_metapaths.py
kpstesla: I have this here because this is similar to what we do for `add_intra_table_metapaths` in the order that we do it.  The deletes aren't strictly necessary but I think it'd be good to keep them.

PR #5051, opened 2022-07-25
Properly support dynamic inheritance in `Batch`

### LINKED ISSUE #4848: Data Batch problem in PyG
### 🐛 Describe the bug

Hi. I am a computational physics researcher and was using PyG very well.
my pyg code was working well a few weeks ago, but now that I run my code, it is not working anymore without any changes.

the problem is like below.
I have many material structures and in my "custom_dataset" class, these are preprocessed and all graph informations (node features, edge features, edge index etc) are inserted into "Data" object in PyTorch geometric.
You can see that each preprocessed sample with index $i$ was printed normal "Data" object in pyg

캡처2

But When I insert my custom dataset class into pyg DataLoader and I did like below,

``` Python
sample = next(iter(train_loader)) # batch sample
```

batch sample is denoted by "DataDataBatch". I didn't see this kind of object name.
and i can't use "sample.x' or "sample.edge_index" command. Instead I need to do like this

캡처3

I want to use expressions like "sample.x", "sample.edge_index" or "sample.edge_attr" as like before. 
I expect your kind explanations. Thank you.

### Environment

* PyG version: `2.0.5`
* PyTorch version: `1.11.0+cu113`
* OS: `GoogleColab Pro Plus`
* Python version: `Python 3.7.13 in colab`
* CUDA/cuDNN version:
* How you installed PyTorch and PyG (`conda`, `pip`, source): 
``` python
# Install required packages.
import os
import torch
os.environ['TORCH'] = torch.__version__
print(torch.__version__)
!pip install -q torch-scatter -f https://data.pyg.org/whl/torch-${TORCH}.html
!pip install -q torch-sparse -f https://data.pyg.org/whl/torch-${TORCH}.html
!pip install -q git+https://github.com/pyg-team/pytorch_geometric.git
!pip install -q pymatgen==2020.11.11  
```
* Any other relevant information (*e.g.*, version of `torch-scatter`):

mananshah99:
This diff properly supports dynamic inheritance with the `DynamicInheritance` metaclass in `Batch`. The initial issue was that of conflicting metaclasses: `abc.ABCMeta` and `DynamicInheritance` for `Data` and `HeteroData` objects. The resolution required two steps:
* Metaclass conflict resolution of the aforementioned metaclasses
* Patching `pin_memory` in PyTorch which makes assumptions about `MutableMapping` before checking whether `pin_memory` exists in the class. This in turn had adverse effects of calling the metaclass of the class again when constructing the new class as part of `pin_memory`, which resulted in the `DataDataBatch` problem observed in the filed issue.

Hopefully the latter fix can be deprecated in the future, with a patch to PyTorch (see https://github.com/pytorch/pytorch/issues/82185)

Closes #4848


torch_geometric/loader/dataloader.py
```diff
@@ -39,6 +40,29 @@ def collate(self, batch):  # Deprecated...
...
+# desired behavior if 'Data' has a `pin_memory` function. We patch
+# this behavior here by monkeypatching `pin_memory`, but can hopefully patch
+# this in PyTorch in the future:
+__torch_pin_memory = torch.utils.data._utils.pin_memory.pin_memory
+
+
+def pin_memory(data, device=None):
+    if hasattr(data, "pin_memory"):
+        return data.pin_memory()
+    else:
```
rusty1s, line 56:
```suggestion
```
No need for the else?

torch_geometric/loader/dataloader.py
```diff
@@ -39,6 +40,29 @@ def collate(self, batch):  # Deprecated...
...
+        if len(signature(__torch_pin_memory).parameters) > 1:
```
rusty1s, line 57: Can we pre-compute this? AFAIK, `signature` is a pretty slow call.

rusty1s (approved):
Amazing, didn’t know it is due to `pin_memory`. Can we add a test which might have failed without this patch?

torch_geometric/loader/dataloader.py
mananshah99: nice catch

mananshah99:
> Can we add a test which might have failed without this patch?

Since this is an issue with `pin_memory`, it will be hard to verify this on CPU since torch `_BaseDataLoaderIter` internally sets `pin_memory` to False without CUDA. We could define a custom dataloader iterator that overrides this functionality and write a test for this, but that might be overkill :)

PR #5052, opened 2022-07-26
Added `pyg_lib.segment_matmul` integration within `RGCNConv`

puririshi98:


Padarn:
curious why we wouldn't want to make this a dependency of pytorch geometric?

torch_geometric/nn/conv/rgcn_conv.py
```diff
@@ -209,14 +215,23 @@ def forward(self, x: Union[OptTensor, Tuple[OptTensor, Tensor]],
...
+                for i in range(self.num_relations):
+                    tmp = masked_edge_index(edge_index, edge_type == i)
+
+                    if x_l.dtype == torch.long:
+                        out += self.propagate(tmp, x=weight[i, x_l], size=size)
+                    else:
+                        h = self.propagate(tmp, x=x_l, size=size)
+                        out = out + (h @ weight[i])
+            else:
+                edge_type, sort_by_edge_type = torch.sort(edge_type)
```
rusty1s, line 228: We should only sorry if necessary, e.g., either via a is_sorted flag or via `edge_type[1:] >= edge_type[:-1].

torch_geometric/nn/conv/rgcn_conv.py
```diff
@@ -209,14 +215,23 @@ def forward(self, x: Union[OptTensor, Tuple[OptTensor, Tensor]],
...
+                edge_index = edge_index[:, sort_by_edge_type]
+                self.edge_ptr = torch.cumsum(
```
rusty1s, line 230: Let‘s use ind2ptr for this from torch-sparse.

rusty1s (approved):
Thanks!

rusty1s:
@Padarn Indeed. Added installation of `pyg_lib` to our CI.

PR #5057, opened 2022-07-26
`LinkNeighborLoader`: Support `data` with no edges.

wsad1:
Sometimes we have `HeteroData` objects where there's just one `node_type` and we'd like to predict links between nodes of the same type. With this PR `LinkNeighborLoader` supports this.

1. defines `LinkSampler` which is used instead of `LinkNeighborSampler` when `data` doesn't have edges. The `__call__` simply returns the indices to select in the `filter_fn`.
2. defines `LinkSamplerMixin` which contains common function between `LinkSampler` and `LinkNeighborSampler`.


ar104 (approved):

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -258,7 +305,12 @@ def __init__(
         edge_type, edge_label_index = get_edge_label_index(
             data, edge_label_index)
 
-        if neighbor_sampler is None:
+        # Corner Case: When data doesn't have any edges
+        # return a subset of nodes.
+        if not has_edge_index(data):
+            self.neighbor_sampler = LinkSampler(data, neg_sampling_ratio,
```
Padarn, line 311:
```suggestion
            self.neighbor_sampler = LinkSampler(data, self.neg_sampling_ratio,
```
or modify it below, just for consistency

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -52,43 +54,88 @@ def _create_label(self, edge_label_index, edge_label):
...
+
+class LinkSampler(LinkSamplerMixin):
+    """This sampler is user when :obj:`data` has no
+    :obj:`edge_index`."""
+    def __init__(self, data, neg_sampling_ratio: float = 0.0,
+                 input_type: InputEdges = None):
+        super().__init__()
+        self.data_cls = data.__class__ if isinstance(
+            data, (Data, HeteroData)) else 'custom'
+        self.neg_sampling_ratio = neg_sampling_ratio
```
Padarn, line 91: is this used in the `LinkSampler`?

torch_geometric/loader/utils.py
```diff
@@ -224,3 +224,21 @@ def filter_feature_store(
         data[attr.group_name][attr.attr_name] = tensors[i]
 
     return data
+
+
+def has_edge_index(data: Union[Data, HeteroData]) -> bool:
```
Padarn, line 229: Could this be defined on `Data`?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -258,7 +305,12 @@ def __init__(
         edge_type, edge_label_index = get_edge_label_index(
```
Padarn, line 305: I think this might break if your graph has no edges?

Padarn:
@wsad1 I like the refactoring a lot, thanks!

I don't really understand the use case though sorry (I'm just not familiar enough) do you have a dataset/task you can refer me to so I can understand better?

wsad1:
> @wsad1 I like the refactoring a lot, thanks!
> 
> I don't really understand the use case though sorry (I'm just not familiar enough) do you have a dataset/task you can refer me to so I can understand better?

@Padarn the test I added ` test_hetero_no_edge_graph_link_neighbor_loader` is this PR contains a use case. In the test,  we have a graph with `paper` nodes and no edges and we'd like to predict links between `paper` nodes. Previously the `LinkNeighborLoader` wouldn't work as there aren't any edges (neighbors). This PR is kind of a hack to support this.

torch_geometric/loader/link_neighbor_loader.py
wsad1: I do have tests for graphs without edges, it works fine for them. Do you have some specific case in mind

Padarn: sorry I meant if `edge_label_index = None` (default)

Padarn:
thanks... after looking at the the example its more clear!

torch_geometric/loader/link_neighbor_loader.py
wsad1: Do you mean the `neg_samling_ratio` or `data_cls`?
`data_cls` is used in the `__call__`.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Good point, I think we should force the user the pass an `edge_label_index` if the graph doesn't have edges. Maybe with an assert statement.

torch_geometric/loader/utils.py
wsad1: Yeah we could add it to `HeteroData`, `Data` and probably `GraphStore`. That's much cleaner. But I doubt it would be used outside of `LinkNeighborLoader`.

wsad1:
Sorry I have to close this PR as its very hard to resolve some conflicts. Closing in favour of #5061.
@Padarn, I have addressed some of your comments in the new PR.

PR #5061, opened 2022-07-27
`LinkNeighborLoader`: Support `data` with no edges.

wsad1:
Sometimes we have HeteroData objects where there's just one node_type and we'd like to predict links between nodes of the same type. With this PR LinkNeighborLoader supports this.  Check out `test_hetero_no_edge_graph_link_neighbor_loader` added in this PR for an example of such a graph.

defines LinkSampler which is used instead of LinkNeighborSampler when data doesn't have edges. The __call__ simply returns the indices to select in the filter_fn.
defines LinkSamplerMixin which contains common function between LinkSampler and LinkNeighborSampler.

### DIFF
--- a/test/loader/test_link_neighbor_loader.py
+++ b/test/loader/test_link_neighbor_loader.py
@@ -287,3 +287,92 @@ def test_custom_heterogeneous_link_neighbor_loader(FeatureStore, GraphStore):
             'paper', 'to', 'author'].edge_index.size())
         assert (batch1['author', 'to', 'paper'].edge_index.size() == batch1[
             'author', 'to', 'paper'].edge_index.size())
+
+
+@pytest.mark.parametrize('directed', [True, False])
+@pytest.mark.parametrize('neg_sampling_ratio', [0.0, 1.0])
+def test_hetero_no_edge_graph_link_neighbor_loader(directed,
+                                                   neg_sampling_ratio):
+    data = HeteroData()
+
+    data['paper'].x = torch.arange(100)
+    edge_label_index = (('paper', 'paper'), get_edge_index(100, 100, 100))
+
+    loader = LinkNeighborLoader(
+        data,
+        num_neighbors=[],
+        edge_label_index=edge_label_index,
+        batch_size=20,
+        directed=directed,
+        neg_sampling_ratio=neg_sampling_ratio,
+        shuffle=True,
+    )
+
+    assert str(loader) == 'LinkNeighborLoader()'
+    assert len(loader) == 100 / 20
+
+    for batch in loader:
+        assert isinstance(batch, HeteroData)
+
+        if neg_sampling_ratio == 0.0:
+            assert len(batch) == 2
+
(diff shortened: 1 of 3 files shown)


rusty1s:
@wsad1 What‘s the root cause of the issue here? I don‘t think we want to necessarily branch here depending on edge count but instead should fix the real issue.

Padarn:
It seems in the case of literally no edges you just want to take a random subset of `nodes x nodes`?

wsad1:
> @wsad1 What‘s the root cause of the issue here? I don‘t think we want to necessarily branch here depending on edge count but instead should fix the real issue.

@rusty1s Consider this graph and the links we want to predict.  `LinkNeighborSampler` will simply not work in such cases as `data` doesn't have edges at all.
```
  data = HeteroData()
  data['paper'].x = torch.arange(100)
  edge_label_index = (('paper', 'paper'), get_edge_index(100, 100, 100))
```
To resolve this edge case, I added a `LinkSampler` which simply samples nodes based on `edge_label_index` in a given batch.

rusty1s:
Why does it not work? Is it because our C++ routine cannot handle empt graphs?

wsad1:
> Why does it not work? Is it because our C++ routine cannot handle empt graphs?

The `num_neighbors` variable simply doesn't have any meaning any more, as there aren't any edges.

At the Python level: 
  The error that prompted me to make this change was caused in this (line)[https://github.com/pyg- 
team/pytorch_geometric/blob/333d3d367a1a5b78e28c3ca7629cd68dcda8a325/torch_geometric/loader/neighbor_loader.py#L75] in `NeighborSampler`. 

At the C++ level:
Doesn't look like the sampler can handle `num_neighbors_dict`, `row_dict` being empty at the moment. In fact it should ideally not ned to handle such cases since its a "neighbor" sampler.

rusty1s:
Having `num_neighbors` being an empty dictionary should be fine, right? Similarly, sampling should just be a no-op in this case, see https://github.com/rusty1s/pytorch_sparse/blob/master/csrc/cpu/neighbor_sample_cpu.cpp#L196. In which line does the test crash?

wsad1:
> Having `num_neighbors` being an empty dictionary should be fine, right? Similarly, sampling should just be a no-op in this case, see https://github.com/rusty1s/pytorch_sparse/blob/master/csrc/cpu/neighbor_sample_cpu.cpp#L196. In which line does the test crash?

The test doesn't crash at the C++ level. it was my guess that it would do so. It crashes in neighbor_loader in this [line](https://github.com/pyg-team/pytorch_geometric/blob/333d3d367a1a5b78e28c3ca7629cd68dcda8a325/torch_geometric/loader/neighbor_loader.py#L75). We could probably make this fix in the `NeighborSampler` but i felt it'll add too many if conditions.

rusty1s:
Ol, but what is the error message? I don‘t see a reason why it should fail in this line.

PR #5063, opened 2022-07-27
Bugfix: Set `best_val_auc` correctly

pancan21:
I guess the best_val_auc used in the if condition should be updated inside.


rusty1s (approved):
Thanks!

PR #5067, opened 2022-07-27
Fix `HGTLoader` bug producing output dictionaries missing edge types

mananshah99:
The fix is actually located at https://github.com/rusty1s/pytorch_sparse/pull/263, this PR includes a test case that will be resolved with the `torch_sparse` fix. However, this test case will fail even after the `torch_sparse` PR is merged since tests build from the latest wheel. @rusty1s what's the recommended approach here?


rusty1s:
We can use a decorator that tests if `torch-sparse` has version > 0.6.14. It might require to upgrade `torch-sparse` version.

mananshah99:
@rusty1s updated `withPackage` to respect `>=` versions, so the test will run for `pytorch_sparse>=0.6.15`. Lmkwyt

rusty1s (approved):
Thank you! I further extended it a bit by allowing any version specification.

Issue #5056, opened 2022-07-26
Addition of Mean Substraction Operator from "Revisiting “Over-smoothing” in Deep GCNs"


Alvaro-Ciudad:
### 🚀 The feature, motivation and pitch

I have been working on graph autoencoders, and I had some problems with oversmoothing. I ve tried a few alternatives, and this is one of the best working ones. 
I also believe that PyG could benefit of more alternative methods of tackling oversmoothing. The code of the layer is already done, so it would just be a simple pull request with a few tests.

```
from torch_scatter import scatter
class MeanSubstraction(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x, batch = None):

        if batch is None:
            x = x - x.mean(dim=0, keepdim=True)
            return x
        else:
            mean = scatter(x, batch, dim=0, reduce='mean')
            x = x - mean.index_select(0, batch)
            return x
```

### Alternatives

It could also be added as a flag inside of the PairNorm implementation, as it is an special case of this normalization.

### Additional context

The paper in question: https://arxiv.org/pdf/2003.13663v1.pdf

rusty1s:
Thanks for sharing. This looks super easy to integrate within our new `nn.aggr` package (see #4712 and https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/aggr/basic.py). Let me know if you want to work on this. @lightaime and @Padarn can help you further.

Padarn:
Sure I'll pick it up!

Alvaro-Ciudad:
Great, if you pick it up works for me :)

lightaime:
Do we want to add it to the `nn.aggr`. It looks more like something we should add to `nn.norm` to me. It is a part of `PairNorm` as @Alvaro-Ciudad mentioned.

rusty1s:
Oh, you are right. I totally misread.

Padarn:
I also misread this (sorry I was on my phone)... I made a PR here https://github.com/pyg-team/pytorch_geometric/pull/5068, but happy for @Alvaro-Ciudad to take over and make any changes that make sense to you.

I added it to the `nn.norm` package, but added support for using any an aggregator from the new `nn.aggr`.

PR #5068, opened 2022-07-28
Added `MeanSubtractionNorm` layer

Padarn:
Implements a generic version of the layer suggested here: https://github.com/pyg-team/pytorch_geometric/issues/5056


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- Agged `AggSubtraction` norm layer ([#5068](https://github.com/pyg-team/pytorch_geometric/pull/5068))
```
rusty1s, line 8:
```suggestion
- Added the `AggSubtraction` norm layer ([#5068](https://github.com/pyg-team/pytorch_geometric/pull/5068))
```

test/nn/norm/test_agg_norm.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+import torch
+
+from torch_geometric.nn import AggSubtraction
+
+
+def test_agg_norm():
```
rusty1s, line 7: Why don‘t we call it MeanSubtractionNorm?

test/nn/norm/test_agg_norm.py
```diff
@@ -0,0 +1,19 @@
...
+    x = torch.randn(6, 16)
+    index = torch.tensor([0, 0, 1, 1, 1, 2])
+
+    norm = AggSubtraction()
+    out = norm(x)
+    assert out.shape == (6, 16)
+    assert out.mean().item() == pytest.approx(0, abs=1e-6)
```
rusty1s, line 14:
```suggestion
    assert torch.allclose(out.mean(), torch.tensor(0.))
```

torch_geometric/nn/norm/agg_norm.py
```diff
@@ -0,0 +1,43 @@
...
+
+import torch
+from torch import Tensor
+
+from torch_geometric.nn.aggr import Aggregation
+from torch_geometric.nn.resolver import aggregation_resolver
+
+
+class AggSubtraction(torch.nn.Module):
+    r"""Applies layer normalization by subtracting an aggregation
```
rusty1s, line 11: Link to the paper.

torch_geometric/nn/norm/agg_norm.py
```diff
@@ -0,0 +1,43 @@
...
+            expected to implement its own aggregation logic via
+            :meth:`aggregate`. (default: :obj:`"mean"`)
+        aggr_kwargs (Dict[str, Any], optional): Arguments passed to the
+            respective aggregation function in case it gets automatically
+            resolved. (default: :obj:`None`)
+    """
+    def __init__(
+        self,
+        aggr: Optional[Union[str, List[str], Aggregation]] = "mean",
+        aggr_kwargs: Optional[Dict[str, Any]] = None,
```
rusty1s, line 34: I am not sure if any other option than mean makes any sense. It might also be that adding support for other aggregations yields to more confusion here as it suggests that this is a hyperparameter that can be actively tuned. I am in favor of just calling this module `MeanSubtractionNorm` which internally instantiates `MeanAggregator`. If you want to keep the flexible implementation, I will defer to you.

test/nn/norm/test_agg_norm.py
Padarn: just because I added the flexibility

torch_geometric/nn/norm/agg_norm.py
Padarn: I do agree with @rusty1s - and had the same concern once I'd implemented it. It might be more user-friendly to just have a `MeanAggregationNorm` and a `MedianAggregationNorm`.

@Alvaro-Ciudad WDTY? We could add it later if it ends up being useful, I agree with @rusty1s that user experience suffers with my approach.

Alvaro-Ciudad: @Padarn I defer to your judgement. It is probably a better idea to remove it and modify it later if needed.

Padarn: Okay I'll update tomorrow morning. Thanks for the discussion.

torch_geometric/nn/norm/agg_norm.py
```diff
@@ -0,0 +1,43 @@
...
+    ):
+        super().__init__()
+        self.aggr_module = aggregation_resolver(aggr, **(aggr_kwargs or {}))
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None):
+        agg = self.aggr_module(x, index)
```
lightaime, line 40: Will this lead to suboptimal implementation if `index` is absent? Should we use dense reduce `torch.mean()` when `index` is absent?

lightaime (commented):
Nice discussion! Another option is to add a flag to `PairNorm` to support `MeanSubtractionNorm` or reimplement `PairNorm` with `MeanSubtractionNorm` so that we reduce the lines of code.

torch_geometric/nn/norm/agg_norm.py
Padarn: It will, but I was wondering if we should fix this inside of the Àggregation` implementation - WDTY?

Padarn:
Nice suggestion @lightaime! I agree, we should do one of those two.

torch_geometric/nn/norm/agg_norm.py
Padarn: Does it even make sense to "fix" inside `Aggregation` actually? Maybe I'm wrong here.

rusty1s: Mh, supporting this in `Aggregation` is tricky since we will need to implement a dense reduction for every new aggregator with little to no use. Let‘s add this to the norm layer for now until we have a clear understanding on how to implement that within `Aggregation`.

rusty1s:
We can re-implement `PairNorm` with `MeanSubtractionNorm`. Okay to do it here or in a follow-up PR. WDYT?

torch_geometric/nn/norm/agg_norm.py
lightaime: Agree. Fixing it in the norm layer should be enough for now.

Padarn: Alright fair enough.

My thought was we don't need to support dense for all aggregations, but I agree we can at least address it elsewhere.

Padarn:
Agree @rusty1s. Just fixed up the main comments here. Lets do that elsewhere.

torch_geometric/nn/norm/agg_norm.py
Padarn: I reverted to the original implementation: Could have used `global_mean_pool` but we've gotten ourselves into a bit of a twist with that package and some circular imports. In this case because pool itself is importing other modules.

We  can leave this here, but we might want to add a solution to the nn.aggr roadmap.

rusty1s (approved):
Thank for your work @Padarn. Looks great as always!

PR #5072, opened 2022-07-28
`NeighborSampler`: Support sampling graphs without edges

wsad1:
Sometimes we have HeteroData objects where there's just one node_type and we'd like to predict links between nodes of the same type. With this PR LinkNeighborLoader supports this. Check out test_hetero_no_edge_graph_link_neighbor_loader added in this PR for an example of such a graph.

Adds some conditions to `NeighborSampler` to work on graphs without edges.


torch_geometric/loader/neighbor_loader.py
```diff
@@ -50,11 +51,15 @@ def __init__(
...
                     f"'{data.__class__.__name__}' object")
 
-            # Convert the graph data into a suitable format for sampling.
-            out = to_csc(data, device='cpu', share_memory=share_memory,
-                         is_sorted=is_sorted)
-            self.colptr, self.row, self.perm = out
-            assert isinstance(num_neighbors, (list, tuple))
+            if not has_edges(data):
+                self.num_neighbors = []
+                self.colptr = self.row = self.perm = None
```
rusty1s, line 56: Why do we need the branching here? If the graph has no edges, than all these dictionaries will just be empty, correct?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -152,6 +165,8 @@ def __init__(
             raise TypeError(f'NeighborLoader found invalid type: {type(data)}')
 
     def _sparse_neighbor_sample(self, index: Tensor):
+        if len(self.num_neighbors) == 0:
```
rusty1s, line 168: What happens if we run the sampler on empty dictionaries?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -72,15 +77,23 @@ def __init__(
...
+                        for key in self.edge_types
+                    }
+                assert isinstance(num_neighbors, dict)
+                self.num_neighbors = {
+                    edge_type_to_str(key): value
+                    for key, value in num_neighbors.items()
+                }
 
-            self.num_hops = max([len(v) for v in self.num_neighbors.values()])
+                self.num_hops = max(
```
rusty1s, line 95:
```suggestion
                # Add at least one element to the list to ensure that max is well-defined
                self.num_hops = max([0] +
```
to avoid the branching

torch_geometric/loader/neighbor_loader.py
wsad1: We can pass an emtpy list of `num_neighbors` to `torch_sparse.neighbor_sample`. But the issue is that `colptr` and `row` can't be empty tensors, that throws a segfault.

torch_geometric/loader/neighbor_loader.py
wsad1: By avoiding branching do you mean remove the `if not has_edges(data)` condition?

torch_geometric/loader/neighbor_loader.py
wsad1: `to_csc` doesn't work if `data` doesn't have edges. Also `colptr` and `row` aren't dictionaris.

torch_geometric/loader/neighbor_loader.py
rusty1s: Yes

torch_geometric/loader/neighbor_loader.py
rusty1s: Ah yes, I think we could just create `edge_index = torch.empty((2, 0), dtype=torch.long)` in `to_csc` in case it does not exist?

torch_geometric/loader/neighbor_loader.py
rusty1s: Yeah, see my reply above.

torch_geometric/loader/neighbor_loader.py
wsad1: Yep made that fix, it works. Thanks @rusty1s.

torch_geometric/loader/neighbor_loader.py
wsad1: I like to keep the branching statement to make it explicit that we do this when the graph has no edges.

torch_geometric/loader/utils.py
```diff
@@ -56,7 +56,11 @@ def to_csc(
         colptr, row, _ = data.adj_t.csr()
 
     elif hasattr(data, 'edge_index'):
-        (row, col) = data.edge_index
+        if data.edge_index is None:
+            edge_index = torch.empty((2, 0), dtype=torch.long)
```
mananshah99, line 60: In this case, can't we just return `row, col, None`?

mananshah99 (commented):
We'll also need to support this for the (FeatureStore, GraphStore) case. Do you want to handle this? If not, I can handle it in a quick follow-up PR.

torch_geometric/loader/neighbor_loader.py
Padarn: I prefer the comment too, makes it more clear why we even need to set it to be zero.

torch_geometric/loader/utils.py
```diff
@@ -224,3 +228,22 @@ def filter_feature_store(
...
+                or (hasattr(edge_store, 'adj') and edge_store.adj is not None))
+
+    if isinstance(data, Data):
+        return True if edge_present(data) else False
+    else:
+        for store in data.edge_stores:
+            if edge_present(store):
+                return True
+        return False
+    # TODO: check if graph store has edges
```
Padarn, line 249: should we raise an error here for now?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -54,7 +55,10 @@ def __init__(
             out = to_csc(data, device='cpu', share_memory=share_memory,
                          is_sorted=is_sorted)
             self.colptr, self.row, self.perm = out
-            assert isinstance(num_neighbors, (list, tuple))
+            if not has_edges(data):
+                self.num_neighbors = []
```
Padarn, line 59: Or we could just set this to `[0]` here?

torch_geometric/loader/utils.py
wsad1: Done.

torch_geometric/loader/utils.py
wsad1: @mananshah99 I added support for GraphStore without edges. it simply checks if `get_all_edge_attrs` returns a non-empty list. WDYT?

wsad1:
> We'll also need to support this for the (FeatureStore, GraphStore) case. Do you want to handle this? If not, I can handle it in a quick follow-up PR.

@mananshah99 I added support for GraphStore without edges. it simply checks if `get_all_edge_attrs` returns a non-empty list. WDYT?

torch_geometric/loader/utils.py
wsad1: Do you mean set `row` and `col` to None and return that?

wsad1:
Refactored it a bit based on comments, and to avoid repeated code.

torch_geometric/loader/utils.py
mananshah99: No, I just meant that we avoid trying to sort here

wsad1: Good point. I went ahead and skipped the `to_csc` if the graph doesn't have edges. I feel like this is more readable than setting empty tensors in `to_csc`

Padarn (approved):
This looks good to me now! Added a few very small nits.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -121,14 +119,7 @@ def __init__(
...
             # Set other required parameters:
-            if isinstance(num_neighbors, (list, tuple)):
-                num_neighbors = {key: num_neighbors for key in self.edge_types}
-            assert isinstance(num_neighbors, dict)
-            self.num_neighbors = {
-                edge_type_to_str(key): value
-                for key, value in num_neighbors.items()
-            }
-            self.num_hops = max([len(v) for v in self.num_neighbors.values()])
+            self._set_num_neighbors_num_hops(num_neighbors)
```
Padarn, line 122: Can you do this once outside of the condition instead of twice?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- `NeighborLoader` supports graphs without edges ([#5072](https://github.com/pyg-team/pytorch_geometric/pull/5072))
```
Padarn, line 8:
```suggestion
- `NeighborSampler ` supports graphs without edges ([#5072](https://github.com/pyg-team/pytorch_geometric/pull/5072))
```
does this make more sense?

torch_geometric/loader/neighbor_loader.py
wsad1: Ah that would be nice. At the moment we can't because `Data` doesn't have `edge_types` and `_set_num_neighbors_num_hops` needs that. We could add another if condition to check if `edge_type` is an attribute and then call `_set_num_neighbors_num_hops`.

rusty1s (approved):
Thanks!

PR #4892, opened 2022-06-30
[Benchmark] Add inference test for current benchmarking and record performance

yanbing-j:
Currently, PyG benchmark only contains training test. It is not convenient for us to optimize starting from inference. In this PR, we add inference test for citation benchmark and to_hetero_mag, and record end-to-end time in one epoch.

PR #5073, opened 2022-07-28
Add profiling support for benchmark/kernel and benchmark/inference

yanbing-j:
Adding inference test in benchmark/kernel.
Add profile in benchmark/inference.


benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,203 @@
...
+import torch.nn.functional as F
+from ogb.nodeproppred import Evaluator, PygNodePropPredDataset
+from torch.profiler import ProfilerActivity, profile
+
+import torch_geometric.transforms as T
+from torch_geometric.nn import GCNConv, SAGEConv
+from torch_geometric.profile import rename_profile_file, trace_handler
+
+
+class GCN(torch.nn.Module):
```
lightaime, line 14: Is it possible to merge the `GCN` and `SAGE` into one class since they share the same structure and forward passes.

benchmark/kernel/main_performance.py
```diff
@@ -37,40 +47,90 @@
...
-            loss, stats = train(model, optimizer, train_loader)
-            val_acc, val_time = eval_acc(model, val_loader)
-            test_acc, test_time = eval_acc(model, test_loader)
-
-            if epoch >= args.warmup_profile:
-                stats_list.append(stats)
-
-        stats_summary = get_stats_summary(stats_list)
-        print(stats_summary)
+def run_train():
```
lightaime, line 51: Same here. Is it possbile to merge the `run_train` and `run_inference` into one function?

benchmark/kernel/main_performance.py
```diff
@@ -37,40 +47,90 @@
...
+                                    str(num_layers), str(hidden))
+
+
+if not args.inference:
+    # Decorate train and eval functions:
+    train = profileit(print_layer_stats=False)(train)
+    eval_acc = timeit()(eval_acc)
+    run_train()
+else:
+    run_inference()
```
lightaime, line 136: Is it possible to also use a decorator for the `inference_run` for profiling?

lightaime (commented):
Thanks! Looks good! Left some minor comments.

benchmark/kernel/gcn-ogbn.py
yanbing-j: Yes. I have merged them into one, by adding one more argument `use_sage`.

benchmark/kernel/main_performance.py
yanbing-j: Hi, I think `run_train` and `run_inference` can be seperate.
1. Merging `run_train` and `run_inference` will make the function too large, and need `if` to distinguish `train` and `inference`. 
2. Most of the code in `run_inference` is not same with `run_train`.
3. Matthias asked to split into two functions in https://github.com/pyg-team/pytorch_geometric/pull/4892#discussion_r916699601 in benchmark/citation.

What do you think about this?

benchmark/kernel/main_performance.py
yanbing-j: I use `torch.profile` for profiling. And currently, `profileit()` and `timeit()` are for `cuda`, I don't know how to record the event for CPU, like for CUDA `start = torch.cuda.Event(enable_timing=True)`  and `start.record()`.
Could you please give some advices?

benchmark/kernel/main_performance.py
lightaime: I prefer to move the dataset and data loader part out of the `run_train` and `run_inference` functions but no objection to this.

benchmark/kernel/main_performance.py
lightaime: I mean create a new decorator for the `torch.profile` logic. So we can do something like 
```
your_timeit()your_profileit()(inference_run)
run_inference()
```
BTW, a newbie question: why do we only record the inference time at the last inference epoch?

benchmark/kernel/gcn-ogbn.py
lightaime: Thanks. Looks good!

benchmark/kernel/main_performance.py
yanbing-j: Done. I move the dataset and dataloader part to `prepare_dataloader`, and call in `run_train` and `run_inference`.

benchmark/kernel/main_performance.py
yanbing-j: Thank you. I create two decorators `torch_profile` and `e2e_time` to decorate `inference_run`.
For inference, the last inference epoch can be regarded as a well-warmup epoch. And profile in one epoch is reasonable because it is easy for `torch.profile` to record all ops and time used. In some dataset, recording all epochs takes too much time, and the trace will be too large to open.

benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,180 @@
...
+        for epoch in range(1, 1 + args.epochs):
+            print("Epoch ", epoch)
+            if epoch == args.epochs:
+                if torch.cuda.is_available():
+                    torch.cuda.synchronize()
+                t_start = time.time()
+
+            inference(model, data)
+
+            if epoch == args.epochs:
```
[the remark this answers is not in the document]
yanbing-j: Done.

benchmark/kernel/main_performance.py
```diff
@@ -56,21 +61,63 @@
...
```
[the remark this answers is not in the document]
yanbing-j: Done.

lightaime:
Thanks @yanbing-j. LGTM!

yanbing-j:
> Thanks @yanbing-j. LGTM!

Thanks @lightaime . Could you please help merge this PR?

benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,167 @@
...
+                          f'Loss: {loss:.4f}, '
+                          f'Train: {100 * train_acc:.2f}%, '
+                          f'Valid: {100 * valid_acc:.2f}% '
+                          f'Test: {100 * test_acc:.2f}%')
+    else:
+        model.reset_parameters()
+        for epoch in range(1, 1 + args.epochs):
+            print("Epoch ", epoch)
+            if epoch == args.epochs:
+                e2e_inference = e2e_time()(inference)
```
rusty1s, line 154: Using a decorator here is somewhat non-pythonic. Ideally, we need a context manager instead:
```python
with e2e_time:
    inference(model, data)

File: benchmark/kernel/gcn-ogbn.py [hunk 7]
```diff
@@ -0,0 +1,167 @@
...
+                e2e_inference(model, data)
+            else:
+                inference(model, data)
+
+        if args.profile:
+            profile_inference = torch_profile()(inference)
```
rusty1s, line 160: Here again.

torch_geometric/profile/profile.py
```diff
@@ -195,3 +197,37 @@ def rename_profile_file(*args):
         timeline_file += '-' + arg
     timeline_file += '.json'
     os.rename('timeline.json', timeline_file)
+
+
+def e2e_time():
```
rusty1s, line 202: No need for the double nested function here.

rusty1s: How is this different from the `timeit` function?

torch_geometric/profile/profile.py
```diff
@@ -195,3 +197,37 @@ def rename_profile_file(*args):
...
+                torch.cuda.synchronize()
+            t_start = time.time()
+
+            func(*args, **kwargs)
+
+            if torch.cuda.is_available():
+                torch.cuda.synchronize()
+            t_end = time.time()
+            duration = t_end - t_start
+            print(f'End-to-End Inference time: {duration:.8f}s', flush=True)
```
rusty1s, line 215:
```suggestion
            print(f'Time: {duration:.8f}s', flush=True)
```

benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,167 @@
...
+        adj_t = data.adj_t.set_diag()
+        deg = adj_t.sum(dim=1).to(torch.float)
+        deg_inv_sqrt = deg.pow(-0.5)
+        deg_inv_sqrt[deg_inv_sqrt == float('inf')] = 0
+        adj_t = deg_inv_sqrt.view(-1, 1) * adj_t * deg_inv_sqrt.view(1, -1)
+        data.adj_t = adj_t
+
+    data = data.to(device)
+
+    if not args.inference:
```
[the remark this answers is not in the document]
yanbing-j: Currently, we focus on inference benchmark, and also want to do preparation for training benchmark. So we keep the logic of training.
And we can seperate inference and training in different directories, like benchmark/inference and benchmark/training. And for this `gcn-ogbn.py`, I can move it to benchmark/inference, and update it only for inference.

benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,167 @@
...
+                          f'Epoch: {epoch:02d}, '
+                          f'Loss: {loss:.4f}, '
+                          f'Train: {100 * train_acc:.2f}%, '
+                          f'Valid: {100 * valid_acc:.2f}% '
+                          f'Test: {100 * test_acc:.2f}%')
+    else:
+        model.reset_parameters()
+        for epoch in range(1, 1 + args.epochs):
+            print("Epoch ", epoch)
+            if epoch == args.epochs:
```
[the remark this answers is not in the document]
yanbing-j: After updating this script only for inference, the epoch number can be adjusted to have a shorter warm-up.

benchmark/kernel/gcn-ogbn.py
yanbing-j: Hi @rusty1s , @lightaime suggested to write the decorators to align with training part. Since we need to seperate the inference and training, I think the decorators are not necessary.

torch_geometric/profile/profile.py
yanbing-j: `timeit` is only for CUDA, I don't know how to use the event in CPU. And lightaime suggested to write a new one. Now, `e2e_time` is not necessary.

benchmark/kernel/gcn-ogbn.py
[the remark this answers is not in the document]
rusty1s: Sounds good. Thanks!

benchmark/kernel/gcn-ogbn.py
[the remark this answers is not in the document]
rusty1s: 👍

benchmark/kernel/gcn-ogbn.py
rusty1s: What is the alternative? At best, we have `e2e_time` be both a decorator and context manager. WDYT?

torch_geometric/profile/profile.py
rusty1s: We can probably adjust `timeit` for both CPU and GPU?

benchmark/kernel/train_eval.py
```diff
@@ -143,3 +143,12 @@ def eval_loss(model, loader):
...
         loss += F.nll_loss(out, data.y.view(-1), reduction='sum').item()
     return loss / len(loader.dataset)
+
+
+@torch.no_grad()
+def inference_run(model, loader):
+    model.eval()
+    for data in loader:
+        data = data.to(device)
+        with torch.no_grad():
```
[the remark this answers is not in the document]
yanbing-j: Done.

benchmark/kernel/gcn-ogbn.py
```diff
@@ -0,0 +1,167 @@
+import argparse
```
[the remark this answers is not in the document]
yanbing-j: Have removed from `kernel/`.

benchmark/kernel/gcn-ogbn.py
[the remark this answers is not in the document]
yanbing-j: `gcn-ogbn.py` has been removed from `benchmark/kernel`, and `benchmark/inference` has test for GCN+ogbn-products already.

benchmark/kernel/gcn-ogbn.py
yanbing-j: Yes. `e2e_time` can be both a decorator and context manager, by using `contextmanager`. Current implementation meets this requirement. However, if `e2e_time` is a context manager, it has no return value, and cannot merge with `timeit`. `timeit` needs to return `out` and `time`. Do you have any ideas about this?

benchmark/kernel/gcn-ogbn.py
yanbing-j: Current `torch_profile` is both a decorator and context manager.

torch_geometric/profile/profile.py
yanbing-j: Yes. The function of `timeit` and `e2e_time` is similar. I'm not quite sure about the difference between `torch.cuda.Event` and  `time.time()` both with `torch.cuda.synchronize()`. If they are the same, the only difference is return values.

torch_geometric/profile/profile.py
yanbing-j: Done.

torch_geometric/profile/profile.py
```diff
@@ -195,3 +198,26 @@ def rename_profile_file(*args):
         timeline_file += '-' + arg
     timeline_file += '.json'
     os.rename('timeline.json', timeline_file)
+
+
+@contextmanager
+def e2e_time():
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
```
Padarn, line 206: I don't think this will do anything if you're not using CUDA events?

Padarn (changes requested):

torch_geometric/profile/profile.py
Padarn: I think you can use `torch.cuda.Event` in the CPU case too ... but not if there is no CUDA device.  So we could fallback on https://docs.python.org/3/library/timeit.html when there is no CUDA device?

yanbing-j: Yes. We need to cover the case without CUDA device. So we can use timeit.default_timer to record e2e time, right?

torch_geometric/profile/profile.py
yanbing-j: Here is to record the time for both CPU and GPU. In GPU, it needs `torch.cuda.synchronize()` to guarantee the accuracy of time.

benchmark/kernel/gcn-ogbn.py
rusty1s: I think it is fine to have no return value when used as a context manager, right? Let me know if I am missing something.

benchmark/citation/train_eval.py
```diff
@@ -102,24 +101,14 @@ def run_inference(dataset, model, epochs, profiling, permute_masks=None,
...
-
-        inference(model, data)
-
-        if epoch == epochs:
-            if torch.cuda.is_available():
-                torch.cuda.synchronize()
-            t_end = time.time()
-            duration = t_end - t_start
-            print(f'End-to-End Inference Time: {duration:.8f}s', flush=True)
+            with timeit():
```
rusty1s, line 104: This is beautiful now, thank you!

benchmark/inference/utils.py
```diff
@@ -6,13 +6,20 @@
...
+    EdgeCNN,
+    GraphSAGE,
+)
 
 models_dict = {
     'edge_cnn': EdgeCNN,
     'gat': GAT,
     'gcn': GCN,
     'pna': PNA,
+    'graph_sage': GraphSAGE,
```
rusty1s, line 22:
```suggestion
    'sage': GraphSAGE,
```
nit: prefer `sage` but this is personal taste :)

benchmark/kernel/main_performance.py
```diff
@@ -56,21 +61,68 @@
...
+                                                 num_layers, hidden))
+
+            model = Net(dataset, num_layers, hidden).to(device)
+            optimizer = torch.optim.Adam(model.parameters(), lr=args.lr)
+
+            stats_list = []
+            acc_list = []
+            for epoch in range(1, args.epochs + 1):
+                loss, stats = train(model, optimizer, train_loader)
+                with timeit() as t:
```
rusty1s, line 83: Currently, the runtimes will be added to a list but also printed out. Can we add some `timeit(log=False)` argument to disable logging here?

benchmark/kernel/main_performance.py
```diff
@@ -56,21 +61,68 @@
...
+                    val_acc = eval_acc(model, val_loader)
+                val_time = t.duration
+                with timeit() as t:
+                    test_acc = eval_acc(model, test_loader)
+                test_time = t.duration
+
+                if epoch >= args.warmup_profile:
+                    stats_list.append(stats)
+                    acc_list.append([val_acc, val_time, test_acc, test_time])
```
rusty1s, line 92: There seems to be some logic change here since we no longer add `val_acc` and `test_acc` to `stats_list`.

benchmark/kernel/main_performance.py
```diff
@@ -56,21 +61,68 @@
...
+    return dataset, train_loader, val_loader, test_loader
+
+
+def run_train():
+    for dataset_name, Net in product(datasets, nets):
+        dataset, train_loader, val_loader, test_loader = prepare_dataloader(
+            dataset_name)
+
+        for num_layers, hidden in product(layers, hiddens):
+            print("--\n{} - {} - {} - {}".format(dataset_name, Net.__name__,
```
rusty1s, line 73: Any reason the `f`-string syntax was dropped here?
```python
print(f'--\n{dataset_name} - {Net.__name__} - '
         f'{num_layers} - {hidden}')
```

benchmark/kernel/train_eval.py
```diff
@@ -143,3 +143,11 @@ def eval_loss(model, loader):
             out = model(data)
         loss += F.nll_loss(out, data.y.view(-1), reduction='sum').item()
     return loss / len(loader.dataset)
+
+
+def inference_run(model, loader):
```
rusty1s, line 148:
```suggestion
@torch.no_grad()
def inference_run(model, loader):
```

torch_geometric/profile/profile.py
```diff
@@ -92,36 +95,32 @@ def wrapper(*args, **kwargs) -> Tuple[Any, Stats]:
     return decorator
 
 
-def timeit():
-    r"""A decorator to facilitate timing a function, *e.g.*, obtaining the
-    runtime of a specific model on a specific dataset.
+class timeit(ContextDecorator):
+    r"""A ContextDecorator to facilitate timing a function, *e.g.*, obtaining
```
rusty1s, line 99:
```suggestion
    r"""A context decorator to facilitate timing a function, *e.g.*, obtaining
```

torch_geometric/profile/profile.py
```diff
@@ -92,36 +95,32 @@ def wrapper(*args, **kwargs) -> Tuple[Any, Stats]:
...
 
-        @timeit()
         @torch.no_grad()
         def test(model, x, edge_index):
             return model(x, edge_index)
 
-        z, time = test(model, x, edge_index)
+        with timeit() as timeit:
+            z = test(model, x, edge_index)
+        time = timeit.duration
```
rusty1s, line 110:
```suggestion
        with timeit() as t:
            z = test(model, x, edge_index)
        time = t.duration
```
otherwise, we are overriding `timeit` I assume.

yanbing-j: Done.

torch_geometric/profile/profile.py
yanbing-j: Done.

benchmark/kernel/main_performance.py
yanbing-j: Change back to the original one. I don't remember why it changes here.

benchmark/kernel/main_performance.py
yanbing-j: Done.

benchmark/inference/utils.py
yanbing-j: Done.

benchmark/inference/utils.py
```diff
@@ -6,13 +6,20 @@
 
 import torch_geometric.transforms as T
 from torch_geometric.datasets import OGB_MAG, Reddit
-from torch_geometric.nn.models.basic_gnn import GAT, GCN, PNA, EdgeCNN
+from torch_geometric.nn.models.basic_gnn import (
```
rusty1s, line 9:
```suggestion
from torch_geometric.nn import (
```

benchmark/kernel/train_eval.py
```diff
@@ -143,3 +143,11 @@ def eval_loss(model, loader):
...
+    model.eval()
+    for data in loader:
+        data = data.to(device)
+        with torch.no_grad():
+            model(data)
```
rusty1s, line 153:
```suggestion
@torch.no_grad()
def inference_run(model, loader):
    model.eval()
    for data in loader:
        data = data.to(device)
        model(data)
```

rusty1s (approved):

benchmark/inference/utils.py
rusty1s, line 9:
```suggestion
from torch_geometric.nn import (
```

benchmark/kernel/train_eval.py
rusty1s, line 153:
```suggestion
@torch.no_grad()
def inference_run(model, loader):
    model.eval()
    for data in loader:
        data = data.to(device)
        model(data)
```

PR #5078, opened 2022-07-28
Add default splits to `MalNetTiny` dataset

flandolfi:
Hi,

I modified the `MalNetTiny` so that it now downloads also the [official split information](http://malnet.cc.gatech.edu/split-info/) given by the authors. I also removed a previously added `remove_isolated_nodes()` from the process (added in #3592), that I believe it should be done as a transform/pre-transform (with `RemoveIsolatedNodes`), if needed by the user. This may be reverted, if you think otherwise!

The code now add the graphs not in alphabetical order, but in the order defined by the split files. One should be able to easily extend the `MalNetTiny` class to the full dataset `MalNet` just by modifying `self.data_url`, `self.split_url`, `self.raw_file_names()` and `self.download()`. I could not try it since I do not have enough space at the moment (the full data is about 130GB).

Francesco


Padarn (commented):
this looks good to me

Padarn (approved):

PR #5084, opened 2022-07-28
Add `reverse` support in `aggregation_resolver`

lightaime:
This PR added the `reverse` support in `aggregation_resolver` to resolve `Aggregation` to strings. This will keep `message_and_aggregate` functionality intact if an aggr is `FUSE_AGGRS` but passed with `Aggregation` instance.

Addresses https://github.com/pyg-team/pytorch_geometric/issues/4712.


Padarn:
This seems important, but I wonder if we could just define the equivalent aggregation on the aggregation class itself.
```
class SomeAggregation:
    @property
    def resolve_str():
        return "some_aggregation"
```
Not sure if its better. Just a thoguht.

lightaime:
> This seems important, but I wonder if we could just define the equivalent aggregation on the aggregation class itself.
> 
> ```
> class SomeAggregation:
>     @property
>     def resolve_str():
>         return "some_aggregation"
> ```
> 
> Not sure if its better. Just a thoguht.

Thanks for the suggestion. I am open to this solution as well.

rusty1s:
Sorry for the delay in review.

torch_geometric/nn/resolver.py
```diff
@@ -36,10 +45,43 @@ def resolver(classes: List[Any], class_dict: Dict[str, Any],
...
+    assert callable(query)
+    query_cls_repr = camel_to_snake(query.__class__.__name__)
+
+    if not base_cls:
+        return query_cls_repr
+
+    base_cls_repr = camel_to_snake(base_cls.__name__)
+
+    if not isinstance(query, base_cls):
+        choices = {base_cls_repr} | set(
```
Padarn, line 70: maybe we should have a test case covering this? the logic is not trivial

torch_geometric/nn/resolver.py
```diff
@@ -80,14 +122,21 @@ def normalization_resolver(query: Union[Any, str], *args, **kwargs):
 # Aggregation Resolver ########################################################
 
 
-def aggregation_resolver(query: Union[Any, str], *args, **kwargs):
+def aggregation_resolver(query: Union[Any, str], *args, reverse: bool = False,
```
Padarn, line 125: I'm not certain I like having a 'reverse' flag because it makes the logic for 'query' a bit complicated. What do you think about splitting this into `reverse_aggregation_resolver`?

Padarn (approved):
I have a few nits about the design here, but I think its probably more important to get things merged, overall looks okay.

PR #5088, opened 2022-07-29
Fix: properly filter custom feature/graph stores

mananshah99:
The current method of filtering relied on the buggy `str_to_edge_type`, which would fail if a node type had `__` in its name.


torch_geometric/loader/utils.py
```diff
@@ -203,14 +201,17 @@ def filter_feature_store(
 
     # Filter edge storage:
     # TODO support edge attributes
-    for key in edge_dict:
-        edge_index = torch.stack([row_dict[key], col_dict[key]], dim=0)
-        data[str_to_edge_type(key)].edge_index = edge_index
+    edge_attrs = graph_store.get_all_edge_attrs()
+    for attr in edge_attrs:
+        edge_type_str = edge_type_to_str(attr.edge_type)
```
wsad1, line 206: don't you need to check if `edge_type_str` is in `edge_dict`?

rusty1s: As far as I can tell, I feel we want to error out here. Resolving this but happy to address it later.

rusty1s: Yeah, this is a good suggestion, especially since this is what the old implementation does. Added it in.

rusty1s (approved):
LGTM. Thanks!

PR #5098, opened 2022-07-30
Add `MedianAggregation` and `QuantileAggregation`

### LINKED ISSUE #5054: Median Aggregation Support

I'm working on a noise graph, some papers show that the interference of noise can be largely reduced using median aggregation.

Papers:
Robustness of Graph Neural Networks at Scale [NIPS'21]
Understanding Structural Vulnerability in Graph Convolutional Networks [IJCAI'21]

### Alternatives

Using torch.argsort like https://github.com/EdisonLeeeee/MedianGCN in the large-scale graph is time costing.
The CUDA-accelerated version of median aggregation is implemented here,https://github.com/sigeisler/robustness_of_gnns_at_scale. However, it requires a specific software version and seems to have some inexplicable errors, which makes debugging difficult for many people who do not know CUDA programming.
If Pyg could support this kind of aggregation, it would be a great joy for the community! Thanks in advance.

### Additional context

_No response_

flandolfi:
This should fix #5054. I also implemented a `QuantileAggregation`, since it could be generalized easily. 
I followed the [`median`](https://pytorch.org/docs/stable/generated/torch.median.html) and [`quantile`](https://pytorch.org/docs/stable/generated/torch.quantile.html) behavior in pytorch, so the first one returns the lowest value in case of even number of elements, while the second has more interpolation options.


test/nn/aggr/test_quantile.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.parametrize('interpolation', QuantileAggregation.interpolations)
+@pytest.mark.parametrize('dim', [0, 1])
+def test_quantile_aggregation(q, interpolation, dim):
+    aggr = QuantileAggregation(q=q, interpolation=interpolation)
+    exp = torch.quantile(x, q=q, interpolation=interpolation, dim=dim,
+                         keepdims=True)
+    obs = aggr(x, torch.zeros(x.size(dim), dtype=torch.long, device=x.device),
+               dim=dim)
+
+    assert torch.allclose(exp, obs), f"EXP: {exp},\n OBS: {obs}"
```
Padarn, line 59:
```suggestion
    assert torch.allclose(exp, obs)
```

test/nn/aggr/test_quantile.py
```diff
@@ -0,0 +1,59 @@
...
+    [2, 3, 4],
+    [5, 6, 7],
+    [8, 9, 0],
+    [1, 2, 3],
+    [4, 5, 6],
+    [7, 8, 9],
+]).float()
+
+
+@pytest.mark.parametrize('index,result,dim', [
```
Padarn, line 20: [nit] I feel this is pretty verbose, personally would prefer just defining in the function body

torch_geometric/nn/aggr/basic.py
```diff
@@ -66,6 +66,130 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+
+        super().__init__()
+        self.q = q
+        self.interpolation = interpolation
+        self.fill_value = fill_value
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+        # we could also expand `ptr` to an index
```
Padarn, line 113: We only support `index` for many of the aggregations already,  seems fine to me.

torch_geometric/nn/aggr/basic.py
```diff
@@ -66,6 +66,130 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+        self.assert_index_present(index)
+
+        if dim_size is None:
```
Padarn, line 116: Not needed, this is done in the `__call__` of `Aggregation`

torch_geometric/nn/aggr/basic.py
```diff
@@ -66,6 +66,130 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+                median = 0.5 * l_med + 0.5 * r_med
+
+        # if the number of elements is 0, return 'nan'/fill_value
+        out_mask = (count > 0).view(*shape)
+        return torch.where(
+            out_mask, median,
+            torch.tensor([self.fill_value], dtype=median.dtype,
+                         device=median.device))
+
+    def __repr__(self) -> str:
```
Padarn, line 166: We do have test cases  for most of the custom representations

torch_geometric/nn/aggr/basic.py
```diff
@@ -66,6 +66,130 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+    .. note::
+        If the median lies between two values, the lowest one is returned.
+        To compute the midpoint (or other kind of interpolation) of the two
+        values, use :class:`QuantileAggregation` instead.
+
+    Args:
+        fill_value (float, optional): Default value in the case no entry is
+            found for a given index (default: :obj:`NaN`).
+    """
+    def __init__(self, fill_value: float = float('nan')):
```
Padarn, line 186: why not include interpolation as an option here too?

torch_geometric/nn/aggr/basic.py
```diff
@@ -66,6 +66,130 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+        x, x_perm = torch.sort(x, dim=dim)
+        index = index.view(*shape).expand_as(x)
+        index = index.take_along_dim(x_perm, dim=dim)
+        index, index_perm = torch.sort(index, dim=dim, stable=True)
+        x = x.take_along_dim(index_perm, dim=dim)
+
+        # compute the quantile interpolations
+        if self.interpolation == 'lower':
+            idx = torch.floor(q_point).long()
+            median = x.index_select(dim, idx)
```
Padarn, line 140:
```suggestion
            q = x.index_select(dim, idx)
```
or `quantile`?

test/nn/aggr/test_quantile.py
flandolfi: I did it as you suggested in the first place, but then switched to `parametrize` to know which test case failed, while using a for loop inside the function body would only show a single `F` during `pytest`. Any way, I could change it, if you prefer the other way!

torch_geometric/nn/aggr/basic.py
flandolfi: Ok, I'll remove the comment!

torch_geometric/nn/aggr/basic.py
flandolfi: You are right, I started implementing `median` and then swithced to `quantile`, I forgot to change the naming

torch_geometric/nn/aggr/basic.py
flandolfi: I'll make one also for this!

torch_geometric/nn/aggr/basic.py
flandolfi: I followed the [`median` implementation](https://pytorch.org/docs/stable/generated/torch.median.html) in PyTorch (also they suggest to use `quantile` for different interpolations). Also, in the future `median` could also return the index (if this is implemented in torch-scatter), and this can be done only for `"lower"`, `"higher"`, and `"nearest"` interpolations. This is merely a desing choice, so I'll stick to your preferences!

Padarn: That's a good point. Then it makes sense to me.

test/nn/aggr/test_quantile.py
Padarn: Ah yes that is true. I don't really have a strong opinion here, have just seen the other way in most of the codebase. Probably not worth changing.

flandolfi:
@rusty1s I updated the docstrings according to your latest PR (#5104), although they're probably too verbose. Feel free to change them!

torch_geometric/nn/aggr/basic.py
```diff
@@ -112,6 +112,150 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+        self.assert_index_present(index)
+        assert index is not None  # required for TorchScript
+
+        # compute the quantile points
+        if dim_size is None:
+            dim_size = 0  # default value for `bincount`
```
flandolfi, line 179: These assertion/if seem unavoidable for TorchScript (even if `self.assert_index_present(index)` already checks for `index`)

Padarn: Can you give a bit more detail on where you're seeing the problem?

flandolfi: Sure. This is what happens without the first assertion:
```
RuntimeError: 

aten::bincount(Tensor self, Tensor? weights=None, int minlength=0) -> (Tensor):
Expected a value of type 'Tensor' for argument 'self' but instead found type 'Optional[Tensor]'.
:
  File "/opt/Workspace/pytorch_geometric/torch_geometric/nn/aggr/basic.py", line 180
        # if dim_size is None:
        #     dim_size = 0  # default value for `bincount`
        count = torch.bincount(index, minlength=dim_size)
                ~~~~~~~~~~~~~~ <--- HERE
        cumsum = torch.cumsum(count, dim=0)
        q_point = self.q * (count - 1) + cumsum - count
```
and this instead without the `if` statement:
```
RuntimeError: 

aten::bincount(Tensor self, Tensor? weights=None, int minlength=0) -> (Tensor):
Expected a value of type 'int' for argument 'minlength' but instead found type 'Optional[int]'.
:
  File "/opt/Workspace/pytorch_geometric/torch_geometric/nn/aggr/basic.py", line 180
        # if dim_size is None:
        #     dim_size = 0  # default value for `bincount`
        count = torch.bincount(index, minlength=dim_size)
                ~~~~~~~~~~~~~~ <--- HERE
        cumsum = torch.cumsum(count, dim=0)
        q_point = self.q * (count - 1) + cumsum - count
```
All the above were run with `torch.__version__ == 1.11.0`.

By the way, the same assertion is also present in the other aggregations, as it is done by [`self.reduce`](https://github.com/pyg-team/pytorch_geometric/blob/e805f0ca606ad67564d1c2de9e4a893243ca5d78/torch_geometric/nn/aggr/base.py#L112), while the `dim_size` check is done by [`scatter`](https://github.com/rusty1s/pytorch_scatter/blob/fc1b13945c8021c446aabd7b9207286b2239a899/torch_scatter/scatter.py#L76), since it also has an `Optional[int]` type hint.

Padarn: Sorry I missed your reply, I will take a look in detail tomorow

Padarn: Hmm yeah I see the issue... I can't see a good workaround for this right now other than including the check as you have here, but we're going to have this with all `Aggregation` modules.

I think lets leave the assert here and maybe we can add an issue to address this later? WDYT?

cc @rusty1s

Padarn: Summay of my understanding:
- issue 1: `jit` by default looks ad 'forward', and we'd want it to look at '__call__'
- issue 2: `jit` cannot handle `**kwargs` that we have in `__call__`
- issue 3: `jit` cannot handle `super().` that we have in `__call__`

flandolfi: @Padarn I agree, this should become a separate issue, as it may lead to some refactoring (maybe the `assert_*_present` can forward/return `index`/`ptr`/etc? This should solve the `Optional[Tensor]` issue, but not the `Optional[int]` one of `dim_size`).

flandolfi:
Just want to let you know that I'll probably extend this PR, as I believe `QuantileAggregation` should allow a list of quantiles: calling multiple instances, e.g., `[QuantileAggregation(q=0.25), QuantileAggregation(q=0.5), QuantileAggregation(q=0.75)]`, would cause repeated sorts, while once we have computed the first one we should be able to retrieve all quantiles needed (something like this `QuantileAggregation(q=[0.25, 0.5, 0.75])`).

So, leave this PR open for a while! :) (Feel free to add a [WIP] if you prefer.)

Padarn:
I was thinking of suggesting that, but didn't know if it would be useful or not :-) Thanks!

flandolfi:
I completed also the multi-quantile option. I don't think I have any more changes to apply, I'll wait for your reviews/comments!

Padarn:
Don't forget to updte the changelog :)

rusty1s (approved):
Amazing!

PR #5113, opened 2022-08-02
Add learnable channel wise parameter in `nn.aggr.SoftmaxAggregation` and `nn.aggr.PowerMeanAggregation`

Padarn:
This addresses an outstanding TODO in the codebase.

I was just experimenting. Interested in feedback on what people think about this.


torch_geometric/nn/aggr/basic.py
```diff
@@ -160,9 +164,14 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
                 ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
                 dim: int = -2) -> Tensor:
 
+        if self.channels > 1:
```
rusty1s, line 167: This does not seem like a requirement.

torch_geometric/nn/aggr/basic.py
```diff
@@ -135,10 +135,13 @@ class SoftmaxAggregation(Aggregation):
             gradient calculation during softmax computation. Therefore, only
             semi-gradient is used during backpropagation. Useful for saving
             memory when :obj:`t` is not learnable. (default: :obj:`False`)
+        channels (int, optional): Number of channels to learn from
+            :math:`t`. If set, :math:`t` will be a learnable tensor of shape
```
rusty1s, line 139:
```suggestion
            :math:`t`. If set to a value greater than :obj:`1`, :math:`t` will be learned per input feature channel.
```

torch_geometric/nn/aggr/basic.py
```diff
@@ -160,9 +164,14 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
                 ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
                 dim: int = -2) -> Tensor:
 
+        if self.channels > 1:
+            self.assert_two_dimensional_input(x, dim)
+
         alpha = x
         if not isinstance(self.t, (int, float)) or self.t != 1:
```
rusty1s, line 171: Is this `if` bugged? It looks like `self.t` is never int or float. Maybe condition based on
```python
if not self.learn and self._init_t !=  1.:::
```

torch_geometric/nn/aggr/basic.py
```diff
@@ -160,9 +164,14 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
-            alpha = x * self.t
+            shape = [1] * x.dim()
+            shape[dim + 1] = -1
```
rusty1s, line 173:
```suggestion
            shape[dim + 1] = self.channels
```

rusty1s (approved):
Super, thanks a lot!

rusty1s:
I am fine with adding something similar to `PowerMeanAggregation` within this PR as well. Please go ahead with this one (if you have time).

torch_geometric/nn/aggr/basic.py
Padarn: You mean we should always check for two dimensional input, or we should never? 

I added this because in the case of a per channel `t` we need to infer which dimension in the input is the channel dimension. I thought supporting two dimensions only was simpler than taking also a channel dim input.

WDYT?

torch_geometric/nn/aggr/basic.py
Padarn: I think it is a float if the `learn` parameter is set to `False`
```
self.t = Parameter(torch.Tensor(self.channels)) if learn else t
```
but I think it could be simplified for clarity to 
```
t = self.t
if is tensor:
   manipulate shape
alpha = x * t
```
Not sure why we  want to avoid the multiplication by 1

torch_geometric/nn/aggr/basic.py
```diff
@@ -209,10 +224,20 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
...
+        p = self.p
+        if isinstance(p, Tensor):
+            shape = [1] * x.dim()
+            shape[dim + 1] = self.channels
+            p = p.view(*shape)
+
         out = self.reduce(x, index, ptr, dim_size, dim, reduce='mean')
-        if isinstance(self.p, (int, float)) and self.p == 1:
+
+        if isinstance(p, (int, float)) and p == 1:
```
Padarn, line 238: this logic is a bit odd to me... we skip the logic when `p = 1`? I  can see that its not very useful in that case, but I think it would be more consistent to still clamp personally

torch_geometric/nn/aggr/basic.py
```diff
@@ -135,10 +135,13 @@ class SoftmaxAggregation(Aggregation):
             gradient calculation during softmax computation. Therefore, only
             semi-gradient is used during backpropagation. Useful for saving
             memory when :obj:`t` is not learnable. (default: :obj:`False`)
+        channels (int, optional): Number of channels to learn from
+            :math:`t`. If set to a value greater than :obj:`1`, :math:`t` will
+            be learned per input feature channel. This requires compatible
+            shapes for the input to the forward calculation.
```
rusty1s, line 141:
```suggestion
            shapes for the input to the forward calculation.
            (default: :obj:`1`)
```

torch_geometric/nn/aggr/basic.py
rusty1s: Ah, you are right, thanks for clarifying.

torch_geometric/nn/aggr/basic.py
```diff
@@ -192,12 +203,16 @@ class PowerMeanAggregation(Aggregation):
         learn (bool, optional): If set to :obj:`True`, will learn the value
             :obj:`p` for powermean aggregation dynamically.
             (default: :obj:`False`)
+        channels (int, optional): Number of channels to learn from
+            :math:`p`. If set to a value greater than :obj:`1`, :math:`p` will
+            be learned per input feature channel. This requires compatible
+            shapes for the input to the forward calculation.
```
rusty1s, line 209:
```suggestion
            shapes for the input to the forward calculation.
            (default: :obj:`1`)
```

rusty1s (approved):

torch_geometric/nn/aggr/basic.py
rusty1s: I think the `clamp` is only needed to prevent overflows in `pow`. Not really sure what's the best approach here though so I left it as it is for now. Hope that works for you.

Padarn: Fair enough. Thanks

PR #5117, opened 2022-08-02
`RandLA-Net` example

CharlesGaydon:
The paper: [RandLA-Net: Efficient Semantic Segmentation of Large-Scale Point Clouds](https://arxiv.org/abs/1911.11236)

# Context
There lacks a good pytorch implementation of RandLa-Net that leverages pytorch geometric standards and modules. 
In torch-points3d, the current modules are outdated leading to [some confusion among users](https://github.com/torch-points3d/torch-points3d/issues?q=is%3Aissue+randlanet+is%3Aclosed). 

The implementation with the most stars on github is [aRI0U/RandLA-Net-pytorch](https://github.com/aRI0U/RandLA-Net-pytorch/blob/master/model.py), which has nasty dependencies (torch_points or torch_points_kernels), makes slow back-and-forth between cpu and gpu when calling knns, and only accepts fixed size point clouds. 

# Proposal
I would like to implement RandLA-Net as part of pyg's examples. For now I would tackle the ModelNet classification task, and would follow the structure of other examples (pointnet2_classification in particular). 

The RandLa-Net paper focuses on segmentation, but for classification I would simply add a MLP+Global Max Pooling after the first DilatedResidualBlocks.

RandLa-Net architecture is conceptually close to PointNet++, augmented with different tricks to speed things up (random sampling instead of fps), use more context (with a sort of dilated KNN), and encode local information better (by explicitly calculating positions, distances, and euclidian distance between points in a neighborhood, and by using self-attention on these features).

If I have some success, I will take on the segmentation task as well (which is what interests me anyway for [my own project](https://github.com/IGNF/myria3d))

# Where I am at
I have a working implementation at `examples/randlanet_classification.py`. I still have to review it to make sure that I am following the paper as closely as possible, but I think I am on the right track.

I would love some guidance on how to move forward. In particular:
- Am I using MessagePassing modules correctly?
- What should I aim for in term of accuracy on ModelNet? 
- Should I stick strictly to the paper? Or adapt the architecture to ModelNet.

Indeed the hyperparameters were not chosen by the author for small objects but rather for large scale Lidar data, which could make convergence way longer that needed.

With 4 DilatedResidualBlocks (like in the paper), we reach ~57% accuracy at epoch 200.

With 3 DilatedResidualBlocks, we reach up to 75% accuracy at the 20th epoch

With only 2 DilatedResidualBlocks, we reach 90% accuracy at the 81st epoch, getting closer to the [leaderboard](https://modelnet.cs.princeton.edu/) for the ModelNet10 challenge.


CharlesGaydon:
I implemented RandLa-Net for segmentation as well, and made some small refactor. 
The model seems to learn quite well, and reaches 70% accuracy after 3 epochs. It takes ~1s to run on CPU.

Unfortunately, I am not able to fully test it out on ShapeNet's Airplane task due to pytorch conflicts that prevent me to use CUDA :/

I need to install the `master` branch of pyg to run segmentation training. When I follow [ instructions to build pytorch geometric from source](https://pytorch-geometric.readthedocs.io/en/latest/notes/installation.html#installation-from-source), I see that I am working on a machine with CUDA 11.4 installed. From what I understand, I have to build pytorch against CUDA 11.4 before installing dependencies (`torch_scatter`, etc), and then `pytorch_geometric` from master. But it seems that pytorch + CUDA 11.4 is not really supported - I could not find how to build it from source, and using `cudatoolkit=11.4` in pytorch's conda install does not work. 
Any help would be appreciated :)

rusty1s:
I suggest to simply install the wheels with CUDA 11.3, this should work even for CUDS 11.4.

CharlesGaydon:
Thank you, this worked like a charm. I think this is ready for review.  :)

Right now both scripts follow the paper's architecture (in terms of hyperparameters, depth and number of channels in MLPs). Those were chosen by authors for large scale aerial lidar, not ModelNet and ShapeNet. For ModelNet, removing a few layer enables to reach good accuracy. I was not able to replicate this for ShapeNet, which quickly plateaus around 70% train accuracy (vs. 90% train accuracy / 79% test IoU for PointNet++).

I think it is cleaner to keep everything as it is to follow the paper. We could also change the benchmark for this model (with e.g. S3DIS), but I am not really sure that this is worth the extra work for an example.

CharlesGaydon:
I identified some differences between my implementation and the original paper (in particular in terms of batch norms, activations, and number of channels). Will come back with fixes and modifications!

rusty1s:
Thanks! Sorry for the delay in review. Please keep me posted.

CharlesGaydon:
Here comes the update!
I made a few changes:
- Set the BatchNorm and LeakyReLU with parameters from the paper (in particular: 0.99 momentum for BN, 0.2 alpha for lrelu, like in tensorflow).
- Made a default_MLP class with `plain_last=False`, because by default class `torch_geometric.nn.MLP` disables BN, activation, and dropout for the last layer. 
- Added a missing Linear layer at the beginning of the network.
- Changed the meaning of "d_out" in DilatedResidualBlock so that it is the actual number of channel that this block produces. This makes the dimensionnality of successive layers easier to read

For the classification task, I commented out two of the DilatedResidualBlock, because, as mentionned earlier, performances are way better for ModelNet with a smaller network.

CharlesGaydon:
With these modifications : 

- ModelNet : 85%-88% accuracy after the ~50th epoch, with only two DilatedResidualBlocks instead of 4. (target in [leaderbord](https://modelnet.cs.princeton.edu/): >=90%)
- ShapeNet : Loss: 0.6704 Train Acc: 0.7500 Test IoU: 0.4746 at epoch 30 (target:  90% train accuracy / 79% test IoU for PointNet++)

There may be an issue with my implementation for the segmentation task, but maybe it is simply a matter of model size / configuration vs. the task (as it is the case for classification where a smaller model works way better).

@rusty1s Ready for review. :)

CharlesGaydon:
I realized that the flow direction of the MessagePassing class responsible for the summarization of local neighborhood was inverted. I got immediate improvement for the segmentation task once fixed.
ShapeNet : Loss: 0.3035 Train Acc: 0.8881Test IoU: 0.6739  at epoch 30 (target: 90% train accuracy / 79% test IoU for PointNet++)

examples/randlanet_classification.py
```diff
@@ -0,0 +1,243 @@
...
+        relative_infos = torch.cat(
+            [pos_i, pos_j, dist, euclidian_dist], dim=1
+        )  # N//4 * K, d
+        local_spatial_encoding = self.mlp_encoder(relative_infos)  # N//4 * K, d
+        local_features = torch.cat([x_j, local_spatial_encoding], dim=1)  # N//4 * K, 2d
+
+        # attention will weight the different features of x
+        attention_scores = torch.softmax(
+            self.mlp_attention(local_features), dim=-1
+        )  # N//4 * K, d_out
```
saedrna, line 79: Here you are making attention scores from the channel direction. But RandLANet should aggregate from the K points. In fact, the message function does not have any information about KNN neighbors.

saedrna: See here for a reference implementation with 3D Tensors.

https://github.com/aRI0U/RandLA-Net-pytorch/blob/057035adf587a4e377e431f54654a090e53740f2/model.py#L107-L135

The softmax is applied to the K Dim.

CharlesGaydon: Good catch! Thank you very much for this review :100: 

To react to your comment in https://github.com/torch-points3d/torch-points3d/issues/782

> Because the inputs of the message function do not have information about the KNN at all. I am wondering how to implement such spatial pooling across the KNN points.

To my understanding, the "sum" pooling operation is applied internally by the `self.propagate(...)` call. Under the hood, the indices in edge_index are used to pool the results `message(...)`. That's where param `flow` comes into action.

I reverted back `flow` to the default `source_to_target`, because I want a sum-pooling to occur on targets i.e. the centers of neighborhoods. It is somewhat counter-intuitive, but I build `edge_index` following many examples with this line : 
```
row, col = knn(pos, pos[idx], self.num_neighbors, batch_x=batch, batch_y=batch[idx])
edge_index = torch.stack([col, row], dim=0)
```
And as a result, the targets are the rows i.e. the centroids of the neighborhoods.

CharlesGaydon:
@ saedrna 
I think one last thing that I have to change is linked to the last upsampling.
In the paper, we see that the output of the first FC layer is used for the last skip connection.
image
But on the [author's reference tensorflow implementation](https://github.com/QingyongHu/RandLA-Net/blob/6b5445f5f279d33d2335e85ed39ca8b68cb1c57e/RandLANet.py#L119-L131t) as well as in [aRI0U's implementation](https://github.com/aRI0U/RandLA-Net-pytorch/blob/057035adf587a4e377e431f54654a090e53740f2/model.py#L266-L288) (they are structurally very similar), it is different: the unsampled output of the first dilated residual block is used instead. This makes sense, but is not reflected in the paper.

CharlesGaydon:
And voilà :100: ! 
We reach PointNet++-comparable performances on ShapeNet's Airplane class after a few epochs :)
ShapeNet :  Loss: 0.2036 **Train Acc: 92%, Test IoU: 82.7%** at epoch 30 (target: 90% train accuracy / 79% test IoU for PointNet++)

saedrna:
I also made an implementation and tested for 60 epochs. Reach similar results. The highest is at epoch 58 with IoU of 0.83.
```
Epoch: 58, Test IoU: 0.8301
[10/40] Loss: 0.2139 Train Acc: 0.9232
[20/40] Loss: 0.2094 Train Acc: 0.9241
[30/40] Loss: 0.2041 Train Acc: 0.9247
[40/40] Loss: 0.2030 Train Acc: 0.9235
```
I just put it here for simplicity. The code is relatively easier thanks to the point_transformer_segmentation example.

```
import os.path as osp
from typing import Optional, Union

import torch
import torch.nn.functional as F
from torch import Tensor
from torch.nn import Linear, Sequential, ReLU, Dropout, LeakyReLU

from torch_geometric.datasets import ShapeNet
from torch_geometric.loader import DataLoader
from torch_geometric.nn import Linear, knn_interpolate, MessagePassing, knn_graph
from torch_geometric.nn.inits import reset
from torch_geometric.typing import Adj, OptTensor, PairTensor
from torch_geometric.utils import intersection_and_union as i_and_u
from torch_geometric.utils import remove_self_loops, add_self_loops, softmax


class SharedMLP(torch.nn.Module):
    def __init__(self, in_channels, out_channels, bn=False, act=None):
        super(SharedMLP, self).__init__()
        self.lin = Linear(in_channels, out_channels)
        if bn and bn is not None:
            self.bn = torch.nn.BatchNorm1d(out_channels, eps=1e-6, momentum=0.99)
        else:
            self.bn = None

        self.act = act

        self.lin.reset_parameters()
        if self.bn:
            self.bn.reset_parameters()

    def forward(self, x):
        x = self.lin(x)
        if self.bn is not None:
            x = self.bn(x)
        if self.act is not None:
            x = self.act(x)
        return x


class RandlaConv(MessagePassing):
    def __init__(self, d, d_slash, add_self_loops: bool = True, **kwargs):
        kwargs.setdefault('aggr', 'add')
        super(RandlaConv, self).__init__(**kwargs)
        self.mlp_rppe = SharedMLP(10, d, bn=True, act=ReLU())  # Relative Point Position Encoding
        self.mlp_att = Linear(2 * d, 2 * d, bias=False)
        self.mlp_post_pool = SharedMLP(2 * d, d_slash, bn=True, act=ReLU())
        self.add_self_loops = add_self_loops

        self.reset_parameters()

    def reset_parameters(self):
        reset(self.mlp_att)

    def forward(self,
                x: Union[Tensor, PairTensor],
                pos: Union[Tensor, PairTensor],
                edge_index: Adj
                ) -> Tensor:

        if isinstance(x, Tensor):
            x: PairTensor = (x, x)

        if isinstance(pos, Tensor):
            pos: PairTensor = (pos, pos)

        if self.add_self_loops:
                edge_index, _ = remove_self_loops(edge_index)
                edge_index, _ = add_self_loops(
                    edge_index, num_nodes=min(pos[0].size(0), pos[1].size(0)))

        # propagate_type: (x: PairTensor, pos: PairTensor)
        out = self.propagate(edge_index, x=x, pos=pos, size=None)
        f_tilde = self.mlp_post_pool(out)
        return f_tilde

    def message(self, x_j: Tensor, pos_i: Tensor, pos_j: Tensor,
                index: Tensor, ptr: OptTensor, size_i: Optional[int]) -> Tensor:
        delta = pos_i - pos_j
        dist = delta.norm(dim=-1, keepdim=True)
        rp = torch.cat([pos_i, pos_j, delta, dist], dim=-1)
        r = self.mlp_rppe(rp)

        # makesure the dim of r and xj are both d and the size are the same
        assert x_j.shape == r.shape
        f_hat = torch.cat([x_j, r], dim=-1)

        # g(f,W)
        f_hat = self.mlp_att(f_hat)
        # do softmax along the KNN
        s = softmax(f_hat, index, ptr, size_i)
        f_hat = s * f_hat
        # propagate will do sum aggregation after attentive pooling between KNN
        return f_hat


class DilatedResidualBlock(torch.nn.Module):
    def __init__(self, d_in, d_out, k=16):
        super(DilatedResidualBlock, self).__init__()
        self.mlp_start = SharedMLP(d_in, d_out // 2, act=LeakyReLU(0.2))
        self.mlp_end = SharedMLP(d_out, 2 * d_out)
        self.mlp_skip = SharedMLP(d_in, 2 * d_out, bn=True)

        self.locse_ap1 = RandlaConv(d_out // 2, d_out // 2)
        self.locse_ap2 = RandlaConv(d_out // 2, d_out)

        self.lrelu = torch.nn.LeakyReLU()
        self.k = k

    def reset_parameters(self):
        reset(self.lrelu)

    def forward(self, x, pos, batch):
        edge_index = knn_graph(pos, k=self.k, batch=batch)
        x_skip = self.mlp_skip(x)  # 2*dout

        x = self.mlp_start(x)  # dout / 2
        x = self.locse_ap1(x, pos, edge_index)  # dout / 2
        x = self.locse_ap2(x, pos, edge_index)  # dout
        x = self.mlp_end(x)  # 2*dout
        x = x + x_skip
        x = self.lrelu(x)  # 2 * dout
        return x


class FPModule(torch.nn.Module):
    """Upsampling with a skip connection."""

    def __init__(self, k, nn):
        super().__init__()
        self.k = k
        self.nn = nn

    def forward(self, x, pos, batch, x_skip, pos_skip, batch_skip):
        x = knn_interpolate(x, pos, pos_skip, batch, batch_skip, k=self.k)
        if x_skip is not None:
            x = torch.cat([x, x_skip], dim=1)
        x = self.nn(x)
        return x, pos_skip, batch_skip


def inverse_permutation(perm):
    inv = torch.empty_like(perm)
    inv[perm] = torch.arange(perm.size(0), device=perm.device)
    return inv


class Net(torch.nn.Module):
    def __init__(self, num_features, num_classes, k=16, decimation=4):
        super(Net, self).__init__()
        self.k = k  # knn
        self.decimation = decimation  # decimation ratio

        self.fc_start = SharedMLP(num_features, 8, bn=True, act=LeakyReLU(0.2))

        # encoder
        self.module_down = torch.nn.ModuleList([
            DilatedResidualBlock(8, 16, self.k),
            DilatedResidualBlock(32, 64, self.k),
            DilatedResidualBlock(128, 128, self.k),
            DilatedResidualBlock(256, 256, self.k)
        ])

        self.mlp_summit = SharedMLP(512, 512, act=ReLU())

        self.module_up = torch.nn.ModuleList([
            FPModule(1, SharedMLP(512 + 512, 256, bn=True, act=ReLU())),
            FPModule(1, SharedMLP(256 + 256, 128, bn=True, act=ReLU())),
            FPModule(1, SharedMLP(128 + 128, 32, bn=True, act=ReLU())),
            FPModule(1, SharedMLP(32 + 32, 8, bn=True, act=ReLU()))
        ])
        self.mlp_cls = Sequential(
            SharedMLP(8, 64, bn=True, act=ReLU()),
            SharedMLP(64, 32, bn=True, act=ReLU()),
            Dropout(),
            SharedMLP(32, num_classes)
        )

    def forward(self, x, pos, batch, ptr):
        # create random permutation for each batch
        B = ptr.size()[0] - 1
        batch_sizes = torch.Tensor([ptr[i + 1] - ptr[i] for i in range(B)]).long()

        indices = [torch.randperm(batch_sizes[i], dtype=torch.int64, device=ptr.device) + ptr[i] for i in range(B)]
        indices = torch.cat(indices)
        inverse_indices = inverse_permutation(indices)

        x = x[indices, :]
        pos = pos[indices, :]
        batch = batch[indices,]

        out_x = []
        out_pos = []
        out_batch = []

        x = self.fc_start(x)

        for encoder in self.module_down:
            # KNN aggregation
            x = encoder(x, pos, batch)
            out_x.append(x)
            out_pos.append(pos)
            out_batch.append(batch)

            # select first part of ::decimated indices in each batch
            batch_sizes = batch_sizes // self.decimation
            indices_decimated = torch.cat([
                torch.arange(batch_sizes[i], device=ptr.device) + ptr[i] for i in range(B)
            ])
            # update ptr
            for i in range(B):
                ptr[i + 1] = ptr[i] + batch_sizes[i]
            x, pos, batch = x[indices_decimated, :], pos[indices_decimated, :], batch[indices_decimated,]

        x = self.mlp_summit(x)

        for i, decoder in enumerate(self.module_up):
            skip_x = out_x.pop()
            skip_pos = out_pos.pop()
            skip_batch = out_batch.pop()

            x, pos, batch = decoder(x, pos, batch, skip_x, skip_pos, skip_batch)

        x = self.mlp_cls(x)
        x = x[inverse_indices, :]
        return x.log_softmax(dim=-1)


if __name__ == '__main__':

    category = 'Airplane'  # Pass in `None` to train on all categories.
    path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'ShapeNet')
    transform = T.Compose([
        T.RandomTranslate(0.01),
        T.RandomRotate(15, axis=0),
        T.RandomRotate(15, axis=1),
        T.RandomRotate(15, axis=2)
    ])
    pre_transform = T.NormalizeScale()
    train_dataset = ShapeNet(path, category, split='trainval', transform=transform,
                             pre_transform=pre_transform)
    test_dataset = ShapeNet(path, category, split='test',
                            pre_transform=pre_transform)
    train_loader = DataLoader(train_dataset, batch_size=60, shuffle=True,
                              num_workers=6)
    test_loader = DataLoader(test_dataset, batch_size=60, shuffle=False,
                             num_workers=6)

    model = Net(3, train_dataset.num_classes).to(device)
    optimizer = torch.optim.Adam(model.parameters(), lr=0.002)
    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.5)


    def train():
        model.train()

        total_loss = correct_nodes = total_nodes = 0
        for i, data in enumerate(train_loader):
            data = data.to(device)
            optimizer.zero_grad()
            out = model(data.x, data.pos, data.batch, data.ptr)
            loss = F.nll_loss(out, data.y)
            loss.backward()
            optimizer.step()
            total_loss += loss.item()
            correct_nodes += out.argmax(dim=1).eq(data.y).sum().item()
            total_nodes += data.num_nodes

            if (i + 1) % 10 == 0:
                print(f'[{i + 1}/{len(train_loader)}] Loss: {total_loss / 10:.4f} '
                      f'Train Acc: {correct_nodes / total_nodes:.4f}')
                total_loss = correct_nodes = total_nodes = 0


    @torch.no_grad()
    def test(loader):
        model.eval()

        y_mask = loader.dataset.y_mask
        ious = [[] for _ in range(len(loader.dataset.categories))]

        for data in loader:
            data = data.to(device)
            pred = model(data.x, data.pos, data.batch, data.ptr).argmax(dim=1)

            i, u = i_and_u(pred, data.y, loader.dataset.num_classes, data.batch)
            iou = i.cpu().to(torch.float) / u.cpu().to(torch.float)
            iou[torch.isnan(iou)] = 1

            # Find and filter the relevant classes for each category.
            for iou, category in zip(iou.unbind(), data.category.unbind()):
                ious[category.item()].append(iou[y_mask[category]])

        # Compute mean IoU.
        ious = [torch.stack(iou).mean(0).mean(0) for iou in ious]
        return torch.tensor(ious).mean().item()


    for epoch in range(1, 61):
        train()
        iou = test(test_loader)
        print(f'Epoch: {epoch:02d}, Test IoU: {iou:.4f}')
        scheduler.step()

```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,256 @@
...
+        )  # N//decim * K, d
+        local_spatial_encoding = self.mlp_encoder(relative_infos)  # N//decim * K, d
+        local_features = torch.cat(
+            [x_j, local_spatial_encoding], dim=1
+        )  # N//decim * K, 2d
+
+        # Attention will weight the different features of x
+        att_features = self.mlp_attention(local_features)  # N//decim * K, d_out
+        d_out = att_features.size(1)
+        att_scores = torch.softmax(
```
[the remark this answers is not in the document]
CharlesGaydon: That's a nice idea. Very convenient that an `index` object is accessible in `message`. Will replace it!

examples/randlanet_classification.py
```diff
@@ -0,0 +1,256 @@
...
+        x = self.lrelu(x + shortcut_of_x)  # N, d_out
+
+        sampled = (x[idx], pos[idx], batch[idx])
+        if return_unsampled_as_well:
+            # Needed for skip connection in final upsampling
+            return sampled, (x, pos, batch)
+        return sampled  # N//decim, d_out
+
+
+def subsample_by_decimation(batch, decimation):
```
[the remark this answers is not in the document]
CharlesGaydon: Good idea as well. This makes the main network forward somewhat more complex, but with the advantage to avoid an edge case where we return the unsampled block1 output, and also a simpler decimation function.
I cherry-picked some lines in your implementation, thank you very much for sharing it :)

CharlesGaydon:
Thank you very much for sharing your implementation @saedrna. That was really helpful in order to give final touches of simplification, better readibility. I cherry-picked the following elements: 
- SharedMLP as a class instead of a function
- knn_graph instead of knn
- Use of pyg's softmax instead of torch's softmax
- Use of ptr for decimation
- Decimation that happens in main forward loop, which removes a special case where we needed non-decimated block output.

I prefer to avoid loops for encoders/decoders in order to keep everything easy to read, which is important in an example - I thus keep the structure of the PointNet++ example.

I think we are a go :)

CharlesGaydon:
@saedrna As a side comment, if you still wand to use your implementation with a loop, I have the following comments:
- Authors of RandLA-Net used `LeakyReLU(negative_slope=0.2)` instead of standard `ReLU()` activations.
- I think you would always want to have self_loops in your knn_graph, so the features of the neighborhood centroid are also considered as part of the local feature encoding. You can do this by specifying `loop=True` in `knn_graph`
:)

saedrna:
Looks good to me~
Just one more thing: one of the beautiful things for RandLA is that, although it's called the selection of random decimation points during the transition-down, in implementation we don't really need actually to permute every time. Just permute the input and choosing every decimated slice in the first parts should be more elegant (like `x=x[perm], x=x[:N/d], x=x[inv_perm]`). Just don't forget to permute back.
I learned this trick from the Pytorch implementation. But he uses sort to permute back, which is O(nlogn), the O(n) solution exists with a temporary array.

https://github.com/aRI0U/RandLA-Net-pytorch/blob/057035adf587a4e377e431f54654a090e53740f2/model.py#L233-L300

CharlesGaydon:
> Looks good to me

Nice. :)

> ~ Just one more thing: one of the beautiful things for RandLA is that, although it's called the selection of random decimation points during the transition-down, in implementation we don't really need actually to permute every time. Just permute the input and choosing every decimated slice in the first parts should be more elegant (like `x=x[perm], x=x[:N/d], x=x[inv_perm]`). Just don't forget to permute back. I learned this trick from the Pytorch implementation. But he uses sort to permute back, which is O(nlogn), the O(n) solution exists with a temporary array.
> 
> https://github.com/aRI0U/RandLA-Net-pytorch/blob/057035adf587a4e377e431f54654a090e53740f2/model.py#L233-L300

I see te appeal as well. I actually encountered empty clouds on my data with this solution: even though I made sure that num_nodes in each cloud was about decimation**4,`x=x[:N/d]` would not decimate clouds homogeneously, and would result in errors. As a result, I prefer to resort to either your concatenation i.e.
```
            indices_decimated = torch.cat([
                torch.arange(batch_sizes[i], device=ptr.device) + ptr[i] for i in range(B)
            ])
```
or mine, which leverages a randperm:
```
    idx_decim = torch.cat(
        [
            (ptr[i] + torch.randperm(decimated_num_nodes[i], device=ptr.device))
            for i in range(batch_size)
        ],
        dim=0,
    )
```

And if we have to cat things, might as well avoid a confusing permutation + inverse permutation. 
I am actually thinking of adding a `+1` to decimated_num_nodes[i] to always have at least one node even if the input cloud is too small. That would be a nice trick to avoid any unforeseen error and more complex preprocessing. Any objection to that strategy? :)

saedrna:
I think everyone has his own flavor for coding and I'm fine with this.

saedrna:
I found an issue, the `momentum` parameter in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html) and [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization) is different. So it should be 0.01 rather than 0.99.

CharlesGaydon:
> I found an issue, the `momentum` parameter in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html) and [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization) is different. So it should be 0.01 rather than 0.99.

Indeed. Good catch, thank you, I'll update it soon.
It is a shame because I just got my best model so far on my own data with this implementation. Hope it last with more momentum ^^

rusty1s:
Super, let me do a final review tomorrow:) Thanks for the amazing progress!

CHANGELOG.md
```diff
@@ -114,6 +114,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `nn.aggr.EquilibrumAggregation` implicit global layer ([#4522](https://github.com/pyg-team/pytorch_geometric/pull/4522))
 - Added support for graph-level outputs in `to_hetero` ([#4582](https://github.com/pyg-team/pytorch_geometric/pull/4582))
 - Added `CHANGELOG.md` ([#4581](https://github.com/pyg-team/pytorch_geometric/pull/4581))
+- Added the RandLA-Net architecture as an example ([#5117](https://github.com/pyg-team/pytorch_geometric/pull/5117))
```
rusty1s, line 117: Move up to PyG 2.2 changelog notes.

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        # LeakyRelu with 0.2 slope by default.
+        kwargs["act"] = kwargs.get("act", "LeakyReLU")
+        kwargs["act_kwargs"] = kwargs.get("act_kwargs", lrelu02_kwargs)
+        # BatchNorm with 1 - 0.99 = 0.01 momentum
+        # and 1e-6 eps by defaut (tensorflow momentum != pytorch momentum)
+        kwargs["norm_kwargs"] = kwargs.get("norm_kwargs", bn099_kwargs)
+        super().__init__(*args, **kwargs)
+
+
+class GlobalPooling(torch.nn.Module):
```
rusty1s, line 38: it looks like the output `pos` and `batch` are never used. Should be fine to drop this module altogether.

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Attention will weight the different features of x
+        # along the neighborhood dimension.
+        att_features = self.mlp_attention(local_features)  # N * K, d_out
+        att_scores = softmax(att_features, index=index)  # N * K, d_out
+
+        return att_scores * local_features  # N * K, d_out
+
+
+class DilatedResidualBlock(MessagePassing):
```
rusty1s, line 94:
```suggestion
class DilatedResidualBlock(torch.nn.Module):
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+class LocalFeatureAggregation(MessagePassing):
+    """Positional encoding of points in a neighborhood."""
+    def __init__(self, d_out, num_neighbors):
+        super().__init__(aggr="add")
+        self.mlp_encoder = SharedMLP([10, d_out // 2])
+        self.mlp_attention = SharedMLP([d_out, d_out], bias=False, act=None,
+                                       norm=None)
+        self.mlp_post_attention = SharedMLP([d_out, d_out])
+        self.num_neighbors = num_neighbors
```
rusty1s, line 55: `num_neighbors` seems to not be used.

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
+import os.path as osp
```
rusty1s, line 1: Can we add a reference to the paper here, and expected output model performance if you have it?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # MLP on input
+        self.mlp1 = SharedMLP([d_in, d_out // 8])
+        # MLP on input, and the result is summed with the output of mlp2
+        self.shortcut = SharedMLP([d_in, d_out], act=None)
+        # MLP on output
+        self.mlp2 = SharedMLP([d_out // 2, d_out], act=None)
+
+        self.lfa1 = LocalFeatureAggregation(d_out // 4, num_neighbors)
+        self.lfa2 = LocalFeatureAggregation(d_out // 2, num_neighbors)
```
rusty1s, line 114:
```suggestion
        self.lfa1 = LocalFeatureAggregation(d_out // 4)
        self.lfa2 = LocalFeatureAggregation(d_out // 2)
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+    def forward(self, x, pos, batch):
+        x = global_max_pool(x, batch)
+        pos = pos.new_zeros((x.size(0), 3))
+        batch = torch.arange(x.size(0), device=batch.device)
+        return x, pos, batch
+
+
+class LocalFeatureAggregation(MessagePassing):
+    """Positional encoding of points in a neighborhood."""
+    def __init__(self, d_out, num_neighbors):
```
rusty1s, line 49:
```suggestion
    def __init__(self, channels, num_neighbors):
```
nit: would prefer `channels` to keep naming consistent but definitely non-blocking :)

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        x = self.mlp1(x)  # N, d_out//8
+        x = self.lfa1(edge_index, x, pos)  # N, d_out//2
+        x = self.lfa2(edge_index, x, pos)  # N, d_out//2
+        x = self.mlp2(x)  # N, d_out
+        x = self.lrelu(x + shortcut_of_x)  # N, d_out
+
+        return x, pos, batch
+
+
+def get_decimation_idx(ptr, decimation):
```
rusty1s, line 131: Would be great to move this to `nn.pool` IMO.

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+    """Subsamples each point cloud by a decimation factor.
+
+    Decimation happens separately for each cloud to prevent
+    emptying smaller point clouds.
+
+    """
+    batch_size = ptr.size(0) - 1
+    num_nodes = torch.Tensor([ptr[i + 1] - ptr[i]
+                              for i in range(batch_size)]).long()
```
rusty1s, line 140:
```suggestion
    bincount = ptr[1:] - ptr[:-1]
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+    decimated_num_nodes = num_nodes // decimation
+    idx_decim = torch.cat(
+        [(ptr[i] + torch.randperm(decimated_num_nodes[i], device=ptr.device))
```
rusty1s, line 143: Won't you want to do something like `randperm(num_nodes)[:decimated_num_nodes]`?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+         for i in range(batch_size)],
```
rusty1s, line 144: Anyway we can avoid the for-loop here?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        num_features,
+        num_classes,
+        decimation: int = 4,
+        num_neighboors: int = 16,
+        return_logits: bool = False,
+    ):
+        super().__init__()
+        self.decimation = decimation
+        self.return_logits = return_logits
+        self.fc0 = Sequential(Linear(in_features=num_features, out_features=8))
```
rusty1s, line 172:
```suggestion
        self.fc0 = Linear(in_features=num_features, out_features=8)
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        # 2 DilatedResidualBlock converges better than 4 on ModelNet.
+        self.block1 = DilatedResidualBlock(num_neighboors, 8, 32)
+        self.block2 = DilatedResidualBlock(num_neighboors, 32, 128)
+        self.mlp1 = SharedMLP([128, 128])
+        self.pool = GlobalPooling()
```
rusty1s, line 177:
```suggestion
        self.pool = MaxAggregation()
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        self.mlp_end = Sequential(SharedMLP([128, 32], dropout=[0.5]),
```
rusty1s, line 178: This reads a bit weird. Wouldn't `[128, 32, num_classes]` work as well?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+                                  Linear(32, num_classes))
+
+    def forward(self, batch):
```
rusty1s, line 181:
```suggestion
    def forward(self, x, pos, batch, ptr):
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        ptr = batch.ptr.clone()
```
rusty1s, line 182: Why do we need to clone it? Will it be modified in-place?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+        in_0 = (self.fc0(batch.x), batch.pos, batch.batch)
+
+        b1 = self.block1(*in_0)
```
rusty1s, line 186:
```suggestion
        b1 = self.block1(self.fc0(x), pos, batch)
```
is more readable, isn't it?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+        x, _, _ = self.pool(x, b2_decimated[1], b2_decimated[2])
+
+        logits = self.mlp_end(x)
+        if self.return_logits:
+            return logits
+        probas = logits.log_softmax(dim=-1)
+        return probas
+
+
+class SetPosAsXIfNoX(BaseTransform):
```
rusty1s, line 202: Let's move this to the model forward pass?
```python
x = x if x is not None else pos
```

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+def train(epoch):
+    model.train()
+
+    for data in tqdm(train_loader):
+        data = data.to(device)
+        optimizer.zero_grad()
+        loss = F.nll_loss(model(data), data.y)
+        loss.backward()
+        optimizer.step()
```
rusty1s, line 218: Return globally averaged loss?

examples/randlanet_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def test(loader):
+    model.eval()
+
+    correct = 0
+    for data in loader:
+        data = data.to(device)
+        with torch.no_grad():
+            pred = model(data).max(1)[1]
```
rusty1s, line 228:
```suggestion
            pred = model(data).argmax(dim=-1)
```

rusty1s (commented):
Thanks for adding this. This looks great already! Left some comments. Let me know if you have any questions.

CharlesGaydon:
> Thanks for adding this. This looks great already! Left some comments. Let me know if you have any questions.

My pleasure. Thanks for the great review, I'll be able to look into it next week. :)

rusty1s:
Super, please ping me :)

CHANGELOG.md
CharlesGaydon: done.

examples/randlanet_classification.py
CharlesGaydon: Indeed. No need to group here contrary to the segmentation task.

examples/randlanet_classification.py
CharlesGaydon: Indeed. Removed.

examples/randlanet_classification.py
CharlesGaydon: Removed and replaced with global_max_pool directly.

examples/randlanet_classification.py
CharlesGaydon: Done.

examples/randlanet_classification.py
CharlesGaydon: Done.

examples/randlanet_classification.py
CharlesGaydon: By default nll_loss does returns the mean loss :+1: 

```
reduction (string, optional): Specifies the reduction to apply to the output:
    ``'none'`` | ``'mean'`` | ``'sum'``. ``'none'``: no reduction will be applied,
    ``'mean'``: the sum of the output will be divided by the number of
    elements in the output, ``'sum'``: the output will be summed. Note: :attr:`size_average`
    and :attr:`reduce` are in the process of being deprecated, and in the meantime,
    specifying either of those two args will override :attr:`reduction`. Default: ``'mean'``
```

CharlesGaydon: Are you suggesting that maybe it should average along the batch dimension to have equal weight to each sample ?

examples/randlanet_classification.py
CharlesGaydon: done.

examples/randlanet_classification.py
CharlesGaydon: Yes! Done.

examples/randlanet_classification.py
CharlesGaydon: Done.

examples/randlanet_classification.py
CharlesGaydon: Done ! :+1:

examples/randlanet_classification.py
CharlesGaydon: Snap, you are right, good catch. At the first layer the N // 4 first elements were always selected. Corrected.

examples/randlanet_classification.py
CharlesGaydon: It would not without some additional modifications: I use a `Linear` layer to remove BatchNorm & Activation, which are used by defaut in the shared MLP.
To have something less weird I just removed the use of `Sequential`, let me know what you think.

examples/randlanet_classification.py
CharlesGaydon: Added some reference to the paper. Do you mean expected performances based on my own experiment? Or by the authors ? (they have not evaluated their model on ModelNet nor ShapeNet)

examples/randlanet_classification.py
CharlesGaydon: I'm not sure, I don't think so, torch.randperm is not very flexible and I do not see a way to deal with the variables sizes at hand.

examples/randlanet_classification.py
CharlesGaydon: Good idea. I'll make the refactor. :)

CharlesGaydon:
Hi @rusty1s, I took most of your comment into account. I made a few replies as well for clarification.
Last item on the list would be moving the decimation pooling function into `nn.pool` as you suggested. I will have the time to do so mid-October and hopefully finalize this PR :100:

examples/randlanet_classification.py
CharlesGaydon: Done :)

CharlesGaydon:
@rusty1s Hi! 
Back to the office! I moved the function to get decimation indices to its own `decimation` module under `torch.nn.pool`, along with a few simple tests.
On a side note, I always wonder what `ptr` stands for (maybe pointer?). Feel free to edit its docstring if needed (currenly: ` ptr (LongTensor): indices of samples in the batch.`).

CharlesGaydon:
@rusty1s @saedrna The gentlest bump on this :)

rusty1s:
Yes, I will merge this over the weekend. Sorry for the delay.

CharlesGaydon:
@rusty1s Resolved the conflict in Changelog :)

examples/randlanet_classification.py
rusty1s: Thanks! I meant expected performance when running this script, but not super urgent if you don't have these numbers.

examples/randlanet_classification.py
rusty1s: Yeah, I mean something like
```
total_loss = total_examples = 0
for ...
    total_loss += float(loss) * loss.numel()
    total_examples += loss.numel()
return total_loss / total_examples
```

rusty1s (approved):
Looks all good to me, just the remaining two comments.

examples/randlanet_classification.py
CharlesGaydon: I see your point. Will add that soon.

PR #5120, opened 2022-08-03
Move to PyTorch `scatter_reduce` implementation

dszwicht:
Switch to PyTorch scatter_reduce/add implementation.

PyTorch `scatter_reduce` implementation is ready for some time now and it appears to perform with higher performance than `torch_scatter`. The goal of this PR is to provide a thin wrapper around PyTorch `scatter_reduce`, making its API aligned with torch_scatter (`out` is optional, `src` and `index` do broadcast). Two functions were provided inside the `utils` module, `scatter` and `scatter_composite` - to differentiate where we only wrapping API and where we are creating composite (in the future, when PyTorch API will support broadcasting, we may leave only `scatter_composite`). This approach gives users a bit cleaner API and all performance advantages related to the PyTorch implementation.

Some [inference](https://github.com/pyg-team/pytorch_geometric/blob/master/benchmark/inference/inference_benchmark.py) benchmark results:
| config*                  | before [S] | now [S] | speedup [before / now] |
|--------------------------|------------|---------|------------------------|
| edge_cnn + ogbn-products | 537.082    | 190.359 | 2.821                  |
| edge_cnn + Reddit        | 398.456    | 156.316 | 2.549                  |
| pna + ogbn-products      | 1840.234   | 662.290 | 2.779                  |
| pna + Reddit             | 2843.046   | 656.774 | 4.329                  |
| gcn + ogbn-products      | 211.345    | 208.555 | 1.013                  |
| gcn + Reddit             | 83.070     | 82.307  | 1.009                  |
| gat + ogbn-products      | 260.519    | 247.975 | 1.051                  |
| gat + Reddit             | 101.863    | 96.378  | 1.057                  |
| rgat + ogbn-mag     | 25039.162 | 24072.808 | 1.040 |

*All configs ran with following arguments: `--num-hidden-channels 128 --num-layers 3 --eval-batch-size 1024 --num-workers 2` .  Speedup is not that impressive in all benchmarks, as `torch_scatter` for `sum` algorithm already uses PyTorch version under the hood.

Considerations:
- `min` and `max` algorithms inside `torch_scatter` return a pair consisting of values and corresponding indices. The latter is used right now inside [`LastAggregator`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/tgn.py#L186), so I left `torch_scatter` there, but we need to cover this case somehow with PyTorch scatter.
- after making more tests with big workloads it was found that `mean` version from [`torch_scatter`](https://github.com/rusty1s/pytorch_scatter/blob/master/torch_scatter/scatter.py#L38) (uses `sum` under the hood), on average is faster than PyTorch version - hence used here as well.
- PyTorch `scatter_reduce` is marked as a beta feature and produces a warning. I'm not sure whether it should be propagated to PyG users, maybe we suppress this warning?
- another thing to consider is [`include_self`](https://pytorch.org/docs/1.12/generated/torch.Tensor.scatter_reduce_.html#torch.Tensor.scatter_reduce_) attribute which appeared in `torch==1.12`. In prior versions of `scatter_reduce`, `self` tensor value is included in calculations. For some algorithms (e.g. `mean`) this effect cannot be omitted. This means that changes introduced here are valid only for `torch>=1.12`.


torch_geometric/graphgym/contrib/layer/generalconv.py
```diff
@@ -54,7 +53,7 @@ def norm(edge_index, num_nodes, edge_weight=None, improved=False,
             edge_index, edge_weight, fill_value, num_nodes)
 
         row, col = edge_index
-        deg = scatter_add(edge_weight, row, dim=0, dim_size=num_nodes)
+        deg = scatter(edge_weight, row, dim=0, dim_size=num_nodes)
```
Padarn, line 56: Should we leave this as `scatter_sum` or add it as the `reduce` argument? I feel its quite nice for us to be explicit about the reduce method?

rusty1s:
Hey @dszwicht, thank you so much! Instead of making the full switch within this PR, can we first add the `scatter_reduce` functionality to PyG and solely add it to the `nn.aggr.Aggregation.reduce` method? This will limit the number of files changed and can ease reviewing.

dszwicht:
> Hey @dszwicht, thank you so much! Instead of making the full switch within this PR, can we first add the `scatter_reduce` functionality to PyG and solely add it to the `nn.aggr.Aggregation.reduce` method? This will limit the number of files changed and can ease reviewing.

Hi, sure 😉

torch_geometric/graphgym/contrib/layer/generalconv.py
dszwicht: In general, I think it's better to have a single function in API (`scatter`) - more options raise confusion. I didn't have a strong preference for setting or not the `reduce='sum'` (default, most common option), but now after giving it a 2nd thought, I agree with you - we should set the reduction algorithm explicitly throughout the code.

Padarn: I agree having a single function is better 👍

lightaime (commented):
Thanks @dszwicht. The improvement is awesome! Left some small comments

torch_geometric/utils/scatter_reduce/composite/scatter.py
```diff
@@ -0,0 +1,49 @@
...
+
+import torch
+from torch import Tensor
+
+from .logsumexp import scatter_logsumexp
+from .softmax import scatter_log_softmax, scatter_softmax
+from .std import scatter_std
+
+
+def scatter_composite(src: Tensor, index: Tensor, dim: int = -1,
```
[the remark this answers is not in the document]
dszwicht: Once PyTorch support for broadcasting is in place, we can slowly get users used to using PyTorch `scatter` (instead of the wrapper). Atter that procedure, only `scatter_composite` would be left in PyG API.

torch_geometric/utils/scatter_reduce/composite/scatter.py
```diff
@@ -0,0 +1,49 @@
...
+        dim_size (int, optional): If :attr:`out` is not given, automatically
+            create output with size :attr:`dim_size` at dimension :attr:`dim`.
+            If :attr:`dim_size` is not given, a minimal sized output tensor
+            according to :obj:`index.max() + 1` is returned.
+        reduce (str, optional): The reduce operation (:obj:`"std"`, :obj:`"logsumexp"`,
+            :obj:`"logsoftmax"` or :obj:`"softmax"`). (default: :obj:`"std"`)
+
+    :rtype: :class:`Tensor`
+    """
+    if reduce == 'std':
```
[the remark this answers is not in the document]
dszwicht: I'm not sure about that. Complex aggregations do not use composite scatter right now. Taking `std` as an example, `StdAggregation` always computes `std` for the "entire population" using a simplified formula, while `scatter_composite(..., 'std')` can calculate `std` either for the "entire population" or for some sample of it (controlled by `unbiased` argument). A bit more complexity results in worse performance. Not sure whether there is a place for both versions and whether we should add additional reduction algorithms. @rusty1s any feedback on that?

dszwicht:
@rusty1s waiting for some feedback. 😉

rusty1s:
Yes, merging this tomorrow :)

rusty1s:
@dszwicht Thank you! I did some clean-up and only expose `scatter` for now - while doing that I noticed that most of the examples started failing due to backpropagation issues due to in-place modifications. Can you go through the files as well to see if I did everything correctly? A second set of eyes may help :)

rusty1s (approved):

dszwicht:
> @dszwicht Thank you! I did some clean-up and only expose `scatter` for now - while doing that I noticed that most of the examples started to fail due to backpropagation issues with in-place modifications. As such, I modified all neural networks to not make use of in-place ops. Can you go through the files as well to see if I did everything correctly? A second set of eyes may help :)

Hi @rusty1s. I saw in logs that it may be caused by some OP versioning issues: `(...) ScatterReduceBackward0, is at version 3; expected version 2 instead (...)`. Using [`scatter_add_`](https://github.com/pyg-team/pytorch_geometric/pull/5120/files#diff-ae201e703b2eec86d66d31cde515892dd7ea4bcdaa5bd601c56c3f1b51f78a9bR42) for `sum` algorithm instead of `scatter_reduce_` seems to fix that.

test/utils/test_scatter.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+import torch
+import torch_scatter
+
+from torch_geometric.utils import scatter
+
+torch.manual_seed(12345)
```
dszwicht, line 7: If "random" `index` below would lack some number from the requested range, for `mul` algorithm default value in PyTorch would be `0`, while in `torch_scatter` it would be `1` (I don't think that's correct). Setting a fixed seed will prevent sporadical errors.

rusty1s:
@dszwicht Thanks for the updates. I think this only fixes it for `scatter_add` though, `scatter_max`, etc will still fail :(

dszwicht:
@rusty1s which specific case do you mean? 😉 Also maybe you see fail because you pushed again Convolution changes. I removed them before.

rusty1s:
@dszwicht I added an additional test to show-case:
```python
@pytest.mark.parametrize('reduce', ['sum', 'add', 'min', 'max'])
def test_scatter_backward(reduce):
    torch.manual_seed(12345)

    src = torch.randn(8, 100, 32).requires_grad_(True)
    index = torch.randint(0, 10, (100, ), dtype=torch.long)

    out = scatter(src, index, dim=1, reduce=reduce).relu_()
    assert src.grad is None
    out.mean().backward()
    assert src.grad is not None


@pytest.mark.parametrize('reduce', ['sum', 'add', 'min', 'max'])
def test_old_scatter_backward(reduce):
    torch.manual_seed(12345)

    src = torch.randn(8, 100, 32).requires_grad_(True)
    index = torch.randint(0, 10, (100, ), dtype=torch.long)

    out = torch_scatter.scatter(src, index, dim=1, reduce=reduce).relu_()
    assert src.grad is None
    out.mean().backward()
    assert src.grad is not None
```
I don't think there is a workaround yet despite removing any in-place ops here.

rusty1s:
CI passes for PyTorch 1.11 but fails for PyTorch 1.12. Let me know if you can think of a fix.

dszwicht:
> CI passes for PyTorch 1.11 but fails for PyTorch 1.12. Let me know if you can think of a fix.

Ok, will let you know.

dszwicht:
@rusty1s [Calling `clone`](https://github.com/pyg-team/pytorch_geometric/pull/5120/files#diff-ae201e703b2eec86d66d31cde515892dd7ea4bcdaa5bd601c56c3f1b51f78a9bR43) on returned tensor fixes error. It looks to me now that the version mentioned in the warning was describing a number of modifications, there was one additional made in place, so a single state was missing. With `clone` it is now traceable back. I made some basic performance measurements, looks like there is no negative impact, but I'll make some more tests.

rusty1s:
Yeah, that is one option. It will increase the memory footprint though, right?

rusty1s:
@dszwicht How about we merge this but make this feature experimental, e.g., via a flag?

dszwicht:
> Yeah, that is one option. It will increase the memory footprint though, right?

Yes, that's true.

> @dszwicht How about we merge this but make this feature experimental, e.g., via a flag?

We could do that. As making a thin wrapper is troublesome, maybe we can consider only some helper functions like `make_broadcastable`, and use torch directly inside the aggregation module?

dszwicht:
> > Yeah, that is one option. It will increase the memory footprint though, right?
> 
> Yes, that's true.
> 
> > @dszwicht How about we merge this but make this feature experimental, e.g., via a flag?
> 
> We could do that. As making a thin wrapper is troublesome, maybe we can consider only some helper functions like `make_broadcastable`, and use torch directly inside the aggregation module?

@rusty1s any feedback on my last proposition?

rusty1s:
Can you clarify what you mean? I am not sure I understand. My idea would be to introduce some form of `experimental_mode` in PyG, e.g.,

```python
torch_geometric.set_experimental_mode(True)
```
I can add this to this PR if that sounds good to you.

dszwicht:
> Can you clarify what you mean? I am not sure I understand. My idea would be to introduce some form of `experimental_mode` in PyG, e.g.,
> 
> ```python
> torch_geometric.set_experimental_mode(True)
> ```
> 
> I can add this to this PR if that sounds good to you.

@rusty1s to clarify, my proposition will not fix the issue with in-place modifications. It's just another solution for "experimental" API. Saying that with a single API it will call either `torch.scatter_reduce` or `torch_scatter.reduce` may be not as clear as saying, if you want to use PyTorch here are some utility functions for you.

`experimental_mode` sounds good to me. 😉

rusty1s (approved):

PR #5131, opened 2022-08-04
`LinkNeighborLoader` Refactoring

wsad1:
With this `edge_label` is set to `torch.zeros` if `None`. This removes a lot of `if` conditions from the code and simplifies it.


torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -298,7 +289,9 @@ def __init__(
 
         edge_type, edge_label_index = get_edge_label_index(
             data, edge_label_index)
-
+        edge_label = (torch.zeros(
+            edge_label_index.size(1), device=edge_label_index.device,
+            dtype=torch.long) if edge_label is None else edge_label)
```
rusty1s, line 294:
```suggestion
        if edge_label is None:
            edge_label = edge_label_index.new_zeros(edge_label_index.size(1))
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -288,7 +280,6 @@ def __init__(
         # Save for PyTorch Lightning < 1.6:
         self.num_neighbors = num_neighbors
         self.edge_label_index = edge_label_index
-        self.edge_label = edge_label
```
rusty1s, line 291: Why remove?

test/loader/test_link_neighbor_loader.py
```diff
@@ -166,7 +164,7 @@ def test_link_neighbor_loader_edge_label():
     )
 
     for batch in loader:
-        assert batch.edge_label.dtype == torch.float
+        assert batch.edge_label.dtype == torch.long
```
rusty1s, line 167: Can we make this a `float`? It simplifies usage of `BCELoss`.

rusty1s (approved):

torch_geometric/loader/link_neighbor_loader.py
wsad1: `self.edge_label` isn't used anywhere.

test/loader/test_link_neighbor_loader.py
wsad1: Ah i see, then maybe we want to convert `torch.float` in the `init` function.

torch_geometric/loader/link_neighbor_loader.py
rusty1s: Yes, but PyTorch Lightning still expects that it is present.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -211,7 +203,7 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             (default: :obj:`None`)
         edge_label (Tensor): The labels of edge indices for which neighbors are
             sampled. Must be the same length as the :obj:`edge_label_index`.
-            If set to :obj:`None` then no labels are returned in the batch.
+            If set to :obj:`None` its set to `troch.zeros(...)` internally.
```
Padarn, line 206:
```suggestion
            If set to :obj:`None` its set to `torch.zeros(...)` internally.
```

torch_geometric/loader/link_neighbor_loader.py
wsad1: Ah good point. added it back.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -211,7 +203,7 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             (default: :obj:`None`)
         edge_label (Tensor): The labels of edge indices for which neighbors are
```
rusty1s, line 204:
```suggestion
        edge_label (Tensor, optional): The labels of edge indices for which neighbors are
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -211,7 +203,7 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
...
             sampled. Must be the same length as the :obj:`edge_label_index`.
-            If set to :obj:`None` then no labels are returned in the batch.
+            If set to :obj:`None` its set to `torch.zeros(...)` internally.
```
rusty1s, line 206:
```suggestion
            If set to :obj:`None` its set to `torch.zeros(...)` internally.
            (default: :obj:`None`)
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -298,7 +289,9 @@ def __init__(
 
         edge_type, edge_label_index = get_edge_label_index(
             data, edge_label_index)
-
+        if edge_label is None:
+            edge_label = edge_label_index.new_zeros(edge_label_index.size(1))
+        self.edge_label = edge_label.to(torch.float)
```
rusty1s, line 294:
```suggestion
            edge_label = torch.new_zeros(edge_label_index.size(1), device=edge_label_index.device)
        self.edge_label = edge_label
```

wsad1: if the user passes a `torch.long` `edge_label` we need to convert to `torch.float`.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -298,7 +290,9 @@ def __init__(
 
         edge_type, edge_label_index = get_edge_label_index(
             data, edge_label_index)
-
+        if edge_label is None:
```
Padarn, line 293: Will this also affect the behavior of negative sampling? Because edge_label exists, the documentation is no longer true I believe.
```
 neg_sampling_ratio (float, optional): The ratio of sampled negative
            edges to the number of positive edges.
            If :obj:`edge_label` does not exist, it will be automatically
            created and represents a binary classification task
            (:obj:`1` = edge, :obj:`0` = no edge).
            If :obj:`edge_label` exists, it has to be a categorical label from
            :obj:`0` to :obj:`num_classes - 1`.
```

wsad1: Good catch :) . maybe we set edge_label to `1` if `neg_sampling_ration` is 0 in `init`, would that resolve this?

Padarn: sounds good, thanks!

Padarn (approved):

PR #5136, opened 2022-08-04
`LinkNeighborLoader` Refactoring (part 2)

rusty1s:
We only want to use `torch.float` for binary classification. If `edge_label` is given, we do not need to cast its data type.

### DIFF
--- a/torch_geometric/loader/link_neighbor_loader.py
+++ b/torch_geometric/loader/link_neighbor_loader.py
@@ -58,29 +58,26 @@ class LinkNeighborSampler(NeighborSampler):
             self.num_dst_nodes = data[self.input_type[-1]].num_nodes
 
     def _create_label(self, edge_label_index, edge_label):
-
         num_pos_edges = edge_label_index.size(1)
         num_neg_edges = int(num_pos_edges * self.neg_sampling_ratio)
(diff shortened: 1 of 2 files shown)


torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -291,10 +288,8 @@ def __init__(
         edge_type, edge_label_index = get_edge_label_index(
             data, edge_label_index)
         if edge_label is None:
-            edge_label = edge_label_index.new_zeros(edge_label_index.size(1))
-            edge_label = (edge_label +
-                          1 if self.neg_sampling_ratio == 0 else edge_label)
```
wsad1, line 296: The documentation says that `If edge_label does not exist, it will be automatically created and represents a binary classification task (1 = edge, 0 = no edge).`. Either update the doc to say what happens when `neg_sampling_ratio=0` and `edge_label=None` or keep this line.

wsad1 (approved):

torch_geometric/loader/link_neighbor_loader.py
rusty1s: Good point. Thanks!

PR #5137, opened 2022-08-04
`LinkNeighborLoader`: Support `edge_label_time`

wsad1:
With this PR users can pass an `edge_time` to `LinkNeighborLoader`, and sampling will respect the `edge_time`.


Padarn:
FYI the linked issue is in a private repo

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -291,6 +342,11 @@ def __init__(
         if edge_label is None:
             edge_label = torch.zeros(edge_label_index.size(1),
                                      device=edge_label_index.device)
+        self.edge_label = edge_label
+        if edge_time is not None and time_attr is None:
```
Padarn, line 346: Maybe in this case we could assume that the edge times are all larger than node times?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -256,6 +306,7 @@ def __init__(
         num_neighbors: NumNeighbors,
         edge_label_index: InputEdges = None,
         edge_label: OptTensor = None,
+        edge_time: OptTensor = None,
```
Padarn, line 309: Why for edge_time do we pass it as a tensor rather than passing it as an attribute name as we do for nodes?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -21,19 +22,30 @@ def __init__(
         self,
         data,
         *args,
+        edge_time=None,
```
Padarn, line 25: wow I finally realise the advantage of adding `*args` here 😅

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -81,10 +93,33 @@ def _create_label(self, edge_label_index, edge_label):
...
+        def update_time(input_type):
+            index = query_dict[input_type]
+            new_node_time, _ = scatter_max(edge_time, index,
+                                           dim_size=self.num_src_nodes)
+            index_unique = index.unique()
+            self.node_time_dict[input_type][index_unique] = new_node_time[
+                index_unique]
+
+        update_time(self.input_type[0])
+        # TODO(jinu) input_type[0] = input_type[1]
```
Padarn, line 106: I don't mind that much, but the TODO doesn't make it very clear to me what is actually pending here.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -178,6 +218,10 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
         :obj:`neg_sampling_ratio` is currently implemented in an approximate
         way, *i.e.* negative edges may contain false negatives.
 
+        if :obj:`edge_time` is not :obj:`None` then :obj:`time_attr` needs
+        to be specified.
+
+        if :obj:`edge_time` is :obj:`None`
```
Padarn, line 224: incompete?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -202,6 +246,12 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             which neighbors are sampled. Must be the same length as
             the :obj:`edge_label_index`. If set to :obj:`None` its set to
             `torch.zeros(...)` internally. (default: :obj:`None`)
+        edge_time (Tensor, optional): The timestamps for edge indicies for
+            which neighbors are sampled. Must be the same length as
+            the :obj:`edge_label_index`. If set, temporal sampling will be
+            used such that neighbors are guaranteed to fulfill temporal
+            constraints, *i.e.* neighbors have an earlier timestamp than
```
Padarn, line 253:
```suggestion
            constraints, *i.e.*, neighbors have an earlier timestamp than
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -202,6 +246,12 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
...
+            the edge. (default: :obj:`None`)
```
Padarn, line 254:
```suggestion
            the ouput edge. (default: :obj:`None`)
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -291,6 +342,11 @@ def __init__(
         if edge_label is None:
             edge_label = torch.zeros(edge_label_index.size(1),
                                      device=edge_label_index.device)
+        self.edge_label = edge_label
```
Padarn, line 345: Do we need this?

Padarn (changes requested):
This is cool functionality; nice to see we can get it with so few changes. Left a few comments/questions. Mostly just nits, feel free to ignore.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Good question. My thinking was that `edge_label_index` and `edge_label` are `Tensor` so it makes sense to have `edge_time` as a `Tensor` too.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Will update that. When the `node_type`s we should take `max` across both `src` and `dst` edges.

torch_geometric/loader/link_neighbor_loader.py
wsad1: its needed for pytorch_lightning <1.6, will add a comment.

torch_geometric/loader/link_neighbor_loader.py
wsad1: We don't do temporal sampling if `time_attr` is `None`. So this if condition raises a warning now.

torch_geometric/loader/link_neighbor_loader.py
Padarn: Hmm yeah I see that logic too. I feel it would be nice to be consistent, but no strong feeling on this.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Removed that comment.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -202,6 +252,12 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             which neighbors are sampled. Must be the same length as
             the :obj:`edge_label_index`. If set to :obj:`None` its set to
             `torch.zeros(...)` internally. (default: :obj:`None`)
+        edge_time (Tensor, optional): The timestamps for edge indicies for
+            which neighbors are sampled. Must be the same length as
+            the :obj:`edge_label_index`. If set, temporal sampling will be
+            used such that neighbors are guaranteed to fulfill temporal
+            constraints, *i.e.*, neighbors have an earlier timestamp than
+            the ouput edge. (default: :obj:`None`)
```
rusty1s, line 260: Can we add some documentation here that `time_attr` needs to be set for this as well?

torch_geometric/loader/link_neighbor_loader.py
rusty1s: I think we need to pass it as a tensor here since the labels we are interested in are not necessarily part of the graph.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -276,7 +333,19 @@ def __init__(
...
 
+        edge_type, edge_label_index = get_edge_label_index(
+            data, edge_label_index)
+        if edge_label is None:
+            edge_label = torch.zeros(edge_label_index.size(1),
+                                     device=edge_label_index.device)
+
+        if edge_time is not None and time_attr is None:
+            edge_time = None
+            warnings.warn("`edge_time` is specified but `time_attr` is `None`."
```
rusty1s, line 344:
```suggestion
            raise ValueError("`edge_time` is specified but `time_attr` is `None`")
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -299,14 +363,15 @@ def __init__(
...
+                edge_time=edge_time,
                 neg_sampling_ratio=self.neg_sampling_ratio,
                 num_src_nodes=num_src_nodes,
                 num_dst_nodes=num_dst_nodes,
                 time_attr=time_attr,
                 share_memory=kwargs.get('num_workers', 0) > 0,
             )
 
-        super().__init__(Dataset(edge_label_index, edge_label),
+        super().__init__(Dataset(edge_label_index, self.edge_label, edge_time),
```
rusty1s, line 374:
```suggestion
        super().__init__(Dataset(edge_label_index, edge_label, edge_time),
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -21,14 +23,28 @@ def __init__(
...
         **kwargs,
     ):
         super().__init__(data, *args, **kwargs)
         self.neg_sampling_ratio = neg_sampling_ratio
-
+        self.edge_time = edge_time
+
+        # TODO if self.edge_time is not None and
+        # input_type doesn't have time attribute
+        # set it to float(inf).
```
rusty1s, line 38: This won't work as `node_time_dict` is of type integer.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -21,14 +23,28 @@ def __init__(
...
+
+        if self.edge_time is not None:
```
rusty1s, line 40: Not really sure why we need to do this in `__init__`. Shouldn't we want to do this within `__call__`? I assume for this we can modify `_hetero_sparse_neighbor_sample` to take in additional attributes.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -79,16 +101,43 @@ def _create_label(self, edge_label_index, edge_label):
...
+                                           dim_size=self.num_src_nodes)
+            index_unique = index.unique()
+            self.node_time_dict[input_type][index_unique] = torch.min(
+                new_node_time[index_unique],
+                self.node_time_dict[input_type][index_unique])
+
+        update_time(edge_label_index[0], self.input_type[0])
+        update_time(edge_label_index[1], self.input_type[-1])
+
+    def _reset_node_time(self, edge_label_index):
```
rusty1s, line 120: No need to reset when we simply pass in a temporary dictionary to `hetero_sparse_neighbor_sample` without modifying the underlying attribute. Sounds cleaner to me.

torch_geometric/loader/link_neighbor_loader.py
Padarn: That's true, but presumably the same could have been true for the node times?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -79,16 +101,43 @@ def _create_label(self, edge_label_index, edge_label):
 
         edge_label = torch.cat([pos_edge_label, neg_edge_label], dim=0)
 
-        return edge_label_index, edge_label
+        return edge_label_index, edge_label, edge_time
+
+    def _modify_node_time(self, edge_label_index, edge_time):
+        """For edges in a batch replace `src` and `dst`
```
Padarn, line 107: Hmm curious if this makes sense? Doesn't it mean that any node that is used in a future edge cannot be sampled? I.e if a node has an old and newer edge, we cannot sample the old edge?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -21,14 +23,28 @@ def __init__(
         self,
         data,
         *args,
+        edge_time=None,
```
Padarn, line 26:
```suggestion
        edge_time: OptTensor = None,
```
But actually this is not used below except for checking if it is `None`, so maybe it should be a bool?

torch_geometric/loader/link_neighbor_loader.py
Padarn: I agree with this, we are having to infer during `__call__` whether there is an `edge_time` present in the data sampled from the dataset anyway.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Nodes are a part of the graph (`HeteroData` object), so their time will always be a node attribute. 
But the edge we want to predict might not be in the graph.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Good point, can we set it to the integer representable `torch.long` then?

rusty1s: I think we should just remove the TODO here and expect that `node_time_dict` is present in case `edge_label_time` is present (and vice versa).

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -228,9 +284,9 @@ class LinkNeighborLoader(torch.utils.data.DataLoader):
             ease-of-use of :meth:`F.cross_entropy`). (default: :obj:`0.0`).
         time_attr (str, optional): The name of the attribute that denotes
             timestamps for the nodes in the graph.
-            If set, temporal sampling will be used such that neighbors are
-            guaranteed to fulfill temporal constraints, *i.e.* neighbors have
-            an earlier timestamp than the center node. (default: :obj:`None`)
+            If set, and :obj:`edge_time` is :obj:`None` temporal sampling will
```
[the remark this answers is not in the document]
wsad1: Disabled it. Now either both `time_attr` and `edge_label_time` need to be set, for temporal sampling.

torch_geometric/loader/link_neighbor_loader.py
wsad1: I didn't fully get you here. 
Do you mean we  create a copy of `self.node_time_dict[input_type]` and update that tensor in `update_time`.

torch_geometric/loader/link_neighbor_loader.py
wsad1: `node_time_dict` needs to exist, but there might be cases where `src` and `dst` nodes might not have a `time_attr` *i.e* `node_time_dict[input_type[0]]` might not exist.

rusty1s: Ah, I see, then I misread. In that case we don't want to fill the `node_time_dict` of that node type with any temporal information, right?

torch_geometric/loader/link_neighbor_loader.py
wsad1: Sorry I don't get your question fully. 
`I.e if a node has an old and newer edge, we cannot sample the old edge`
The temporal sampler only makes sure nodes obey temporal constraints, as in we don't check if the edge obeys the constraint.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Well with this hacky implementation we do, because the `edge_label_time` is "passed" to the `src`(input_type[0]) and `dst`(input_type[-1]) nodes which is then passed to the temporal sampler.

rusty1s: got it. You are right, we cannot simply not set it.

torch_geometric/loader/link_neighbor_loader.py
Padarn: Oh good point. That makes sense thanks!

torch_geometric/loader/link_neighbor_loader.py
Padarn: Okay right. Do you think we should consider adding support for this in the temporal sampler directly?

wsad1: Yeah in the long term we should we should support any edge in the graph with a timestamp or `edge_label_time`  directly in the `torch_sprase` sampler.

torch_geometric/loader/link_neighbor_loader.py
rusty1s: I don't think we should modify the `node_time_dict` property here. Then there is also no need to modify reset the property at all. Instead, let us shallow-copy `node_time_dict`, modify it, and input the modified copy to the sampler:
```python
def __call__(...):
    # Modify node_time_dict
    node_time_dict = copy.copy(node_time_dict)
    node_time_dict[...] = ... # make changes
    self._hetero_sample(..., node_time_dict=node_time_dict)
```

wsad1: Thanks for this suggestions. As you mentioned this is needed for correct behaviour when `num_workers>0`.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Removed that check in `__init__`.

torch_geometric/loader/neighbor_loader.py
```diff
@@ -159,8 +159,12 @@ def _sparse_neighbor_sample(self, index: Tensor):
         )
         return node, row, col, edge
 
-    def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor]):
-        if self.node_time_dict is None:
+    def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor],
```
rusty1s, line 162: What we could do instead is define this as
```suggestion
    def _hetero_sparse_neighbor_sample(self, index_dict: Dict[str, Tensor], **kwargs):
```
and then later do
```python
kwargs.get('num_neighbors', self.num_neighbors),
kwargs.get('node_time_dict', self.node_time_dict),
...
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -79,16 +94,46 @@ def _create_label(self, edge_label_index, edge_label):
...
         edge_label = torch.cat([pos_edge_label, neg_edge_label], dim=0)
 
-        return edge_label_index, edge_label
+        return edge_label_index, edge_label, edge_label_time
+
+    def _modify_node_time(self, edge_label_index, edge_label_time):
+        """If `edge_label_time` is `None` return `self.node_time_dict`.
+        Else, for edges in a batch replace `src` and `dst`
+        node times by the min across all edge times."""
+        if edge_label_time is None:
```
rusty1s, line 103: Let's just call this function in case the task is temporal?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -79,16 +94,46 @@ def _create_label(self, edge_label_index, edge_label):
...
+            index_unique = index.unique()
+            node_time_dict[input_type][index_unique] = torch.min(
+                new_node_time[index_unique],
+                self.node_time_dict[input_type][index_unique])
+
+        node_time_dict = copy.copy(self.node_time_dict)
+        node_time_dict[self.input_type[0]] = node_time_dict[
+            self.input_type[0]].clone()
+        node_time_dict[self.input_type[-1]] = node_time_dict[
+            self.input_type[-1]].clone()
```
rusty1s, line 118:
```suggestion
        src, _, dst = self.input_type
        node_time_dict[src] = node_time_dict[src].clone()
        node_time_dict[dst] = node_time_dict[dst].clone()
```
looks cleaner IMO.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -79,16 +94,46 @@ def _create_label(self, edge_label_index, edge_label):
...
+        return edge_label_index, edge_label, edge_label_time
+
+    def _modify_node_time(self, edge_label_index, edge_label_time):
+        """If `edge_label_time` is `None` return `self.node_time_dict`.
+        Else, for edges in a batch replace `src` and `dst`
+        node times by the min across all edge times."""
+        if edge_label_time is None:
+            return self.node_time_dict
+
+        def update_time(node_time_dict, index, input_type, num_nodes):
```
rusty1s, line 106: I think this is overly complicated. I think
```python
scatter_min(edge_label_time, index, out=node_time_dict[input_type])
```
should do the tricky already.

Also: Don't we want to use max here? I thought the goal is to increase the timestamp so that we can sample neighbors that appeared after the seed node.

test/loader/test_link_neighbor_loader.py
```diff
@@ -189,20 +189,33 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
     data['paper', 'author'].edge_index = get_edge_index(100, 200, 1000)
     data['author', 'paper'].edge_index = get_edge_index(200, 100, 1000)
 
+    with pytest.raises(ValueError, match=r'`edge_label_time` is specified .*'):
+        loader = LinkNeighborLoader(data, num_neighbors=[-1] * 2,
+                                    edge_label_index=('paper', 'paper'),
+                                    batch_size=32, time_attr='time')
+
+    # With edge_time:
+    edge_time = torch.arange(data['paper', 'paper'].edge_index.size(1))
```
rusty1s, line 204: I am not sure how easy it is to add, but would be great to have a test that would fail without properly handling of `edge_label_time` and overriding `node_time_dict`.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -276,7 +327,21 @@ def __init__(
...
         self.data = data
 
+        edge_type, edge_label_index = get_edge_label_index(
+            data, edge_label_index)
+        if edge_label is None:
+            edge_label = torch.zeros(edge_label_index.size(1),
+                                     device=edge_label_index.device)
+
+        if ((edge_label_time is not None and time_attr is None)
+                or (edge_label_time is None and time_attr is not None)):
```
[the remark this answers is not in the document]
wsad1: Nice. Much cleaner.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Done.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Thanks that simplified things a lot.
If a node is a part of multiple links, then the nodes time should be update to the min of all those links, that way we'll only sample nodes before any of those links.

test/loader/test_link_neighbor_loader.py
wsad1: with `num_neighbors>0` this test used to fail before this commit https://github.com/pyg-team/pytorch_geometric/pull/5137/commits/fab08082bda9785223c7d65dc8aa1bcb9c65ea58.

torch_geometric/loader/link_neighbor_loader.py
rusty1s: Ok, that means we use the min for `edge_label_index`, but if we update `node_time_dict` we want to take the `max`, right?

test/loader/test_link_neighbor_loader.py
rusty1s: Yes, not referring to `num_workers > 0`, but to the previous implementation of how we handled temporal sampling in link prediction (without `edge_label_time`).

wsad1: Ah, I see. Now `LinkNeighborLoader` can't do temporal sampling without `edge_label_time`.

wsad1: Modified line 190 in the test. This checks if the `author` nodes have time stamp smaller than the `edge_label_time` and not necessarily smaller than `paper` nodes.

torch_geometric/loader/link_neighbor_loader.py
wsad1: Yes your right. Updated that.

rusty1s (approved):

Issue #5108, opened 2022-08-01
AttributeError: 'str' object has no attribute 'y' when using  torch_geometric.loader.ImbalancedSampler


EdisonLeeeee:
### 🐛 Describe the bug

I came across a bug when using `torch_geometric.loader.ImbalancedSampler` on a Heterogeneous graph. Here is my code:
```python
from torch_geometric.datasets import DBLP
from torch_geometric.loader import ImbalancedSampler, NeighborLoader

dataset = DBLP(root='~/data/pygdata/dblp')
data = dataset[0]

sampler = ImbalancedSampler(data['author'], input_nodes=data['author'].train_mask)

train_loader = NeighborLoader(data, input_nodes=('author', data['author'].train_mask),
                              num_neighbors=[10, 10], 
                              sampler=sampler, batch_size=20)
```

The output messages are:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [4], in <cell line: 7>()
      4 dataset = DBLP(root='~/data/pygdata/dblp')
      5 data = dataset[0]
----> 7 sampler = ImbalancedSampler(data['author'], input_nodes=data['author'].train_mask)
      9 train_loader = NeighborLoader(data, input_nodes=('author', data['author'].train_mask),
     10                               num_neighbors=[10, 10], 
     11                               sampler=sampler, batch_size=20)

File ~/Documents/Jupyter/pytorch_geometric/torch_geometric/loader/imbalanced_sampler.py:68, in ImbalancedSampler.__init__(self, dataset, input_nodes, num_samples)
     65     assert len(dataset) == y.numel()
     67 else:
---> 68     ys = [data.y for data in dataset]
     69     if isinstance(ys[0], Tensor):
     70         y = torch.cat(ys, dim=0).view(-1)

File ~/Documents/Jupyter/pytorch_geometric/torch_geometric/loader/imbalanced_sampler.py:68, in <listcomp>(.0)
     65     assert len(dataset) == y.numel()
     67 else:
---> 68     ys = [data.y for data in dataset]
     69     if isinstance(ys[0], Tensor):
     70         y = torch.cat(ys, dim=0).view(-1)

AttributeError: 'str' object has no attribute 'y'
```
This seems attributed to the data['author'] that is not recognized as a `Data` instance.

### Environment

* PyG version: 2.0.4
* PyTorch version: 1.12
* OS: Mac OS
* Python version: 3.9
[...]

PR #5138, opened 2022-08-04
Let `ImbalancedSampler` accept `torch.Tensor` as input

EdisonLeeeee:
Hi, I modified `torch_geometric.loader.ImbalancedSampler` to accept `torch.Tensor` object, i.e., the class distribution as input. Now it is supported to use the class labels to initialize the sampler, as like:
```python
from torch_geometric.loader import NeighborLoader, ImbalancedSampler

sampler = ImbalancedSampler(data.y, input_nodes=data.train_mask)
loader = NeighborLoader(data.y, input_nodes=data.train_mask,
                        batch_size=64, num_neighbors=[-1, -1],
                        sampler=sampler, ...)    
                                sampler=sampler, ...)  
```
And the issue in #5108 should be fixed with minimal adaptation:
```python
from torch_geometric.datasets import DBLP
from torch_geometric.loader import ImbalancedSampler, NeighborLoader

dataset = DBLP(root='~/data/pygdata/dblp')
data = dataset[0]

# NOTE: input the node labels instead of the dataset `data['author']`
sampler = ImbalancedSampler(data['author'].y, input_nodes=data['author'].train_mask) 

train_loader = NeighborLoader(data, input_nodes=('author', data['author'].train_mask),
                              num_neighbors=[10, 10], 
                              sampler=sampler, batch_size=20)
```


EdisonLeeeee (commented):
Fix several typos

torch_geometric/loader/imbalanced_sampler.py
```diff
@@ -60,6 +72,10 @@ def __init__(
             assert dataset.num_nodes == y.numel()
             y = y[input_nodes] if input_nodes is not None else y
 
+        elif isinstance(dataset, Tensor):
+            y = dataset.view(-1)
+            y = y[input_nodes] if input_nodes is not None else y
```
Padarn, line 77: I find this condition a bit confusing - why would someone want to provide both?

test/loader/test_imbalanced_sampler.py
```diff
@@ -41,6 +41,7 @@ def test_neighbor_loader_with_imbalanced_sampler():
     edge_index = torch.empty((2, 0), dtype=torch.long)
     data = Data(edge_index=edge_index, y=y, num_nodes=y.size(0))
 
+    # test with `torch_geometric.data.Data` instance as input
```
Padarn, line 44: Do we need this comment?

torch_geometric/loader/imbalanced_sampler.py
```diff
@@ -33,10 +33,22 @@ class distribution.
                                 batch_size=64, num_neighbors=[-1, -1],
                                 sampler=sampler, ...)
 
+    It is also supported to accept a :class:`torch.Tensor` object,
+    which is the node label that represents the class distribution:
```
Padarn, line 37:
```suggestion
    which has the node label distribution that we wish to sample with:
```
Is this accurate? (No need to accept this change, but the current description was confusing for me)

I would have thought you'd directly want to provide weights in this case.

test/loader/test_imbalanced_sampler.py
EdisonLeeeee: Ok, I'll remove this comment.

torch_geometric/loader/imbalanced_sampler.py
EdisonLeeeee: I'm also a bit confused about this. Indeed, is it really necessary to support so many types of input for this variable `dataset`? I would suggest to support only `torch.Tensor` (labels) or `List[torch.Tensor]` ( a list of labels) as input.

```python
class ImbalancedSampler(torch.utils.data.WeightedRandomSampler):
    def __init__(
        self,
        y: Union[Tensor, List[Tensor]],
        num_samples: Optional[int] = None,
    ):

        if isinstance(y, Tensor):
            y = y.view(-1)
        else:
            if isinstance(y[0], Tensor):
                y = torch.cat(y, dim=0).view(-1)
            else:
                y = torch.tensor(y).view(-1)

    assert y.dtype == torch.long  # Require classification.

    num_samples = y.numel() if num_samples is None else num_samples

    class_weight = 1. / y.bincount()
    weight = class_weight[y]

    return super().__init__(weight, num_samples, replacement=True)
```
Users need to specify themselves which labels should be inputted
```python
from torch_geometric.loader import ImbalancedSampler

sampler = ImbalancedSampler(data.y[data.train_mask]) 
```

This may result in some changes for relevant users. I'm not so sure. Could you give me some suggestions?

Padarn: Yeah, seems fairly heavily used so maybe best not to break the current contract. You could define a new class for  this approach... probably need @rusty1s for a better opinion on this

rusty1s: I personally would like the change of only inputting a `Tensor` which results in s cleaner interface but has the trade-off that users need to first collect their label tensor. I think we can leave the interface as it is for now, and add a deprecation warning for other inputs than tensors. We would also need to update examples. Happy to also do it in a follow-up PR if this is too complicated right now.

rusty1s (approved):
Thanks!

PR #5140, opened 2022-08-05
Add `dgNN` support

HenryChang213:
- run `examples/fused_gat.py` (need to install dgNN first, [https://github.com/dgSPARSE/dgNN](url))
- a `FusedGATConv` layer is added in `nn/fused_gat_conv.py`, which is equivalent to the standard GAT model mathmatically with less memory consumption and faster speed, due to kernel fusion techniques, as in paper 'Understanding GNN Computational Graph: A Coordinated Computation, IO, and Memory Perspective'([https://proceedings.mlsys.org/paper/2022/hash/9a1158154dfa42caddbd0694a4e9bdc8-Abstract.html](url))
- a `to_dgnn` converting function is added in `utils/to_dgnn.py` to convert the `edge_index` format used by PyG into the CSR and CSC format used by dgNN


Padarn:
Hey @HenryChang213 could we add some tests?

torch_geometric/nn/conv/fused_gat_conv.py
```diff
@@ -0,0 +1,132 @@
...
+        kwargs.setdefault('aggr', 'add')
+        super().__init__(node_dim=0, **kwargs)
+
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.heads = heads
+        self.concat = concat
+        self.negative_slope = negative_slope
+        self.dropout = dropout
+        # self.add_self_loops = add_self_loops
```
Padarn, line 47: Maybe want to remove these?

torch_geometric/nn/conv/fused_gat_conv.py
```diff
@@ -0,0 +1,132 @@
...
+            assert x.dim() == 2, "Static graphs not supported in 'GATConv'"
+            x_src = x_dst = self.lin_src(x).view(-1, H, C)
+        else:  # Tuple of source and target node features:
+            x_src, x_dst = x
+            assert x_src.dim() == 2, "Static graphs not supported in 'GATConv'"
+            x_src = self.lin_src(x_src).view(-1, H, C)
+            if x_dst is not None:
+                x_dst = self.lin_dst(x_dst).view(-1, H, C)
+
+        x = (x_src, x_dst)
```
Padarn, line 102: This is not used below?

torch_geometric/nn/conv/fused_gat_conv.py
```diff
@@ -0,0 +1,132 @@
...
+    def forward(self, x: Union[Tensor, OptPairTensor], edge_index: Adj):
+
+        H, C = self.heads, self.out_channels
+
+        # We first transform the input node features. If a tuple is passed, we
+        # transform source and target node features via separate weights:
+        if isinstance(x, Tensor):
+            assert x.dim() == 2, "Static graphs not supported in 'GATConv'"
+            x_src = x_dst = self.lin_src(x).view(-1, H, C)
+        else:  # Tuple of source and target node features:
```
Padarn, line 95: Curious why you would want to pass a tuple if one of them were `None`? Would it make more sense to have 
`Union[Tensor, PairTensor]` or `OptPairTensor` as the input types?

torch_geometric/nn/conv/fused_gat_conv.py
```diff
@@ -0,0 +1,132 @@
+from typing import Tuple, Union
```
rusty1s, line 1: I think we should be able to inherit from `GATConv` here, and just override its `forward` method. Correct?

examples/fused_gat.py
```diff
@@ -0,0 +1,88 @@
+import argparse
```
rusty1s, line 1: Maybe we can add a `benchmark/fuse` repo instead and compare runtimes between `GATConv` and `FusedGATConv`?

torch_geometric/utils/to_dgnn.py
```diff
@@ -0,0 +1,29 @@
+import scipy.sparse as sp
+import torch
+
+
+def to_dgnn(edge_index):
```
rusty1s, line 5: Can we add some tests to this function?

torch_geometric/nn/conv/fused_gat_conv.py
fuvty: Yeah, I think so. We can inherit from GATConv instead.

torch_geometric/nn/conv/fused_gat_conv.py
fuvty: Yeah. Thanks for pointing it out, we deleted it.

torch_geometric/utils/to_dgnn.py
```diff
@@ -0,0 +1,31 @@
...
```
rusty1s, line 5: Would be also good to name it more generally, e.g., `compute_all_graph_representations`.

torch_geometric/utils/to_dgnn.py
```diff
@@ -0,0 +1,31 @@
...
+    device = edge_index.device
+
+    col, row = edge_index.cpu()
+    numlist = torch.arange(col.size(0), dtype=torch.int32)
+    num_nodes = torch.max(col) - torch.min(col) + 1
+
+    adj_csr = sp.csr_matrix((numlist.numpy(), (row, col)),
```
rusty1s, line 12: Can we implement this via PyTorch manually? Current implementation is CPU-only :(

torch_geometric/nn/conv/fused_gat_conv.py
```diff
@@ -0,0 +1,85 @@
+from typing import Optional, Tuple, Union
+
+from dgNN.operators import GATConvFuse
```
rusty1s, line 3: Can we make this an optional dependency? Similar to [here](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/spline_conv.py#L15)?

MathsCode:
The test errors in the above check are the same ('Trainer' object has no attribute 'lr_schedulers' in test_graphgym.py). The test_graphgym.py uses the class "LoggerCallback()" in line 187. And the LoggerCallback() is in logger.py (torch_geomtric/graphgym/logger.py). There is a function called _get_stats using the "lr_schedulers". I think this is the root of the problem.

fuvty:
> The test errors in the above check are the same ('Trainer' object has no attribute 'lr_schedulers' in test_graphgym.py). The test_graphgym.py uses the class "LoggerCallback()" in line 187. And the LoggerCallback() is in logger.py (torch_geomtric/graphgym/logger.py). There is a function called _get_stats using the "lr_schedulers". I think this is the root of the problem.

It is caused by another recent commit to the master branch of PyG, so we did not fix this issue in this PR. All the tests related to this PR have passed :)

fuvty:
The code looks good to me now :)

rusty1s (approved):

PR #5149, opened 2022-08-06
Respect `flow` argument in `GCN` normalization

gau-nernst:
Add flow to `gcn_norm` to correctly normalize adjacency matrix when `flow=target_to_source` for directed graphs. This is discussed in #5100.

A simple test is added to verify this new behavior.

@rusty1s Let me know if you need me to edit any thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- Added `flow` argument to `gcn_norm` to correctly normalize adjacency matrix for `GCNConv` when input is a directed graph ([#5149](https://github.com/pyg-team/pytorch_geometric/pull/5149))
 - `NeighborSampler` supports graphs without edges ([#5072](https://github.com/pyg-team/pytorch_geometric/pull/5072))
 - Added the `MeanSubtractionNorm` layer ([#5068](https://github.com/pyg-team/pytorch_geometric/pull/5068))
 - Added `pyg_lib.segment_matmul` integration within `RGCNConv` ([#5052](https://github.com/pyg-team/pytorch_geometric/pull/5052), [#5096](https://github.com/pyg-team/pytorch_geometric/pull/5096))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
-- Added `flow` argument to `gcn_norm` to correctly normalize adjacency matrix for `GCNConv`. ([#5149](https://github.com/pyg-team/pytorch_geometric/pull/5149))
+- Added `flow` argument to `gcn_norm` to correctly normalize the adjacency matrix in `GCNConv` ([#5149](https://github.com/pyg-team/pytorch_geometric/pull/5149))
 - `NeighborSampler` supports graphs without edges ([#5072](https://github.com/pyg-team/pytorch_geometric/pull/5072))
 - Added the `MeanSubtractionNorm` layer ([#5068](https://github.com/pyg-team/pytorch_geometric/pull/5068))
 - Added `pyg_lib.segment_matmul` integration within `RGCNConv` ([#5052](https://github.com/pyg-team/pytorch_geometric/pull/5052), [#5096](https://github.com/pyg-team/pytorch_geometric/pull/5096))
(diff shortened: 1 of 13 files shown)


torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -17,21 +17,22 @@
 
 @torch.jit._overload
 def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
-             add_self_loops=True, dtype=None):
-    # type: (Tensor, OptTensor, Optional[int], bool, bool, Optional[int]) -> PairTensor  # noqa
+             add_self_loops=True, flow="source_to_target", dtype=None):
+    # type: (Tensor, OptTensor, Optional[int], bool, bool, str, Optional[int]) -> PairTensor  # noqa
```
wsad1, line 21:
```suggestion
    # type: (Tensor, OptTensor, Optional[int], bool, bool, Optional[str], Optional[int]) -> PairTensor  # noqa
```

torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -17,21 +17,22 @@
...
     pass
 
 
 @torch.jit._overload
 def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
-             add_self_loops=True, dtype=None):
-    # type: (SparseTensor, OptTensor, Optional[int], bool, bool, Optional[int]) -> SparseTensor  # noqa
+             add_self_loops=True, flow="source_to_target", dtype=None):
+    # type: (SparseTensor, OptTensor, Optional[int], bool, bool, str, Optional[int]) -> SparseTensor  # noqa
```
wsad1, line 28:
```suggestion
    # type: (SparseTensor, OptTensor, Optional[int], bool, bool, Optional[str], Optional[int]) -> SparseTensor  # noqa
```

torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -61,7 +62,8 @@ def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
             edge_weight = tmp_edge_weight
 
         row, col = edge_index[0], edge_index[1]
-        deg = scatter_add(edge_weight, col, dim=0, dim_size=num_nodes)
+        idx = col if flow == "source_to_target" else row
```
wsad1, line 65: We need to do something similar if `edge_index` is a sparse tensor, probably to `dim` variable in `sparsesum`.

wsad1 (changes requested):
Thanks for the fix.

torch_geometric/nn/conv/gcn_conv.py
gau-nernst: I'm not 100% sure about PyG internal workings, but from what I understand, when `edge_index` is a sparse tensor, only `flow="source_to_target"` is allowed (it will raise error otherwise), even though technically it is doing aggregation from target to source (assuming first dim of `edge_index` is source, second dim is target). This is also reflected in the documentation (see [here](https://pytorch-geometric.readthedocs.io/en/latest/notes/sparse_tensor.html), "transpose sparse matrix") and unit tests.
Therefore, technically `flow="target_to_source"` is not possible in `gcn_norm()` function, and we don't need to handle that for sparse tensor. Always calculating the degree by summing across `dim=1` should always be correct. Changing the flow for sparse tensor should be by transposing the matrix I believe, not by changing the `flow` argument.

torch_geometric/nn/conv/gcn_conv.py
gau-nernst: Optional[str] = Union[str, None], which is not what we want here I believe (passing `flow=None` should be wrong). It does not necessarily mean optional argument. Just like `add_self_loops=True`, the signature is still just `bool`, not `Optional[bool]`.

torch_geometric/nn/conv/gcn_conv.py
wsad1: Ah thanks for pointing that out.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.0.5] - 2022-MM-DD
 ### Added
+- Added `flow` argument to `gcn_norm` to correctly normalize adjacency matrix for `GCNConv` when input is a directed graph ([#5149](https://github.com/pyg-team/pytorch_geometric/pull/5149))
```
wsad1, line 8:
```suggestion
- Added `flow` argument to `gcn_norm` to correctly normalize adjacency matrix for `GCNConv`. ([#5149](https://github.com/pyg-team/pytorch_geometric/pull/5149))
```

wsad1 (approved):

gau-nernst:
@wsad1 I have updated the changelog and merged from master. 1 unrelated test is failing (`test_copy_linear[True]`), probably due to nan values.

rusty1s (approved):

rusty1s:
Thanks @gau-nernst. Also added the `flow` argument to other layers that utilize `gcn_norm` :)

Issue #5143, opened 2022-08-05
GINEConv assumes nn is subscriptable


oiao:
### 🐛 Describe the bug

There is an issue with the GINEConv class when providing the `edge_dim` parameter.
Line https://github.com/pyg-team/pytorch_geometric/blob/e805f0ca606ad67564d1c2de9e4a893243ca5d78/torch_geometric/nn/conv/gin_conv.py#L144
assumes that the passed `nn` is subscriptable, which is not the case when it is directly derived from `nn.Module`:

``` python
from torch import nn
from torch_geometric import nn as gnn

class Net(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin = nn.Linear(1, 2)
    def forward(self, x):
        return self.lin(x)

gnn.GINEConv(Net(), edge_dim=3)
```
```
File torch_geometric/nn/conv/gin_conv.py:144, in GINEConv.__init__(self, nn, eps, train_eps, edge_dim, **kwargs)
    [142]     self.register_buffer('eps', torch.Tensor([eps]))
    [143] if edge_dim is not None:
--> [144]     if hasattr(self.nn[0], 'in_features'):
    [145]         in_channels = self.nn[0].in_features
    [146]     else:

TypeError: 'Net' object is not subscriptable
```

### Environment

* PyG version: 2.0.4
* PyTorch version: 1.12
* OS: Ubuntu
* Python version: 3.8
* CUDA/cuDNN version: 11.3
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip

rusty1s:
Thanks for reporting. Will look into it.

Padarn:
I had some spare time, so raised a PR to address this.

PR #5154, opened 2022-08-07
Fix bug for input module in `GINEConv`

Padarn:
See https://github.com/pyg-team/pytorch_geometric/issues/5143

### DIFF
--- a/torch_geometric/nn/conv/gin_conv.py
+++ b/torch_geometric/nn/conv/gin_conv.py
@@ -149,9 +149,12 @@ class GINEConv(MessagePassing):
                 nn = self.nn
             if hasattr(nn, 'in_features'):
                 in_channels = self.nn.in_features
-            else:
+            elif hasattr(nn, 'in_channels'):
                 in_channels = self.nn.in_channels
+            else:
+                raise ValueError("Could not infer input channels from `nn`.")
             self.lin = Linear(edge_dim, in_channels)
+
         else:
             self.lin = None
         self.reset_parameters()
--- a/torch_geometric/nn/conv/gin_conv.py
+++ b/torch_geometric/nn/conv/gin_conv.py
@@ -142,8 +142,7 @@ class GINEConv(MessagePassing):
         else:
             self.register_buffer('eps', torch.Tensor([eps]))
         if edge_dim is not None:
-            if (isinstance(self.nn, torch.nn.Sequential)
-                    or isinstance(self.nn, torch.nn.ModuleList)):
+            if isinstance(self.nn, torch.nn.Sequential):
                 nn = self.nn[0]
             else:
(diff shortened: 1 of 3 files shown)


torch_geometric/nn/conv/gin_conv.py
```diff
@@ -141,10 +142,15 @@ def __init__(self, nn: Callable, eps: float = 0., train_eps: bool = False,
         else:
             self.register_buffer('eps', torch.Tensor([eps]))
         if edge_dim is not None:
-            if hasattr(self.nn[0], 'in_features'):
-                in_channels = self.nn[0].in_features
+            if (isinstance(self.nn, torch.nn.Sequential)
```
Padarn, line 145: I don't actually like this logic much, I think it might be nicer to just require the input channels to be provided. But it at least solves the bug.

torch_geometric/nn/conv/gin_conv.py
```diff
@@ -141,11 +142,19 @@ def __init__(self, nn: Callable, eps: float = 0., train_eps: bool = False,
...
+                    or isinstance(self.nn, torch.nn.ModuleList)):
```
rusty1s, line 146: I am pretty sure a ModuleList wouldn‘t work here.

torch_geometric/nn/conv/gin_conv.py
rusty1s: LazyLinear does not work since it can only infer the input channels, not the output channels.

Padarn: Ah, good point

torch_geometric/nn/conv/gin_conv.py
Padarn: You are very right. Thanks. Fixed

torch_geometric/nn/conv/gin_conv.py
```diff
@@ -141,11 +142,18 @@ def __init__(self, nn: Callable, eps: float = 0., train_eps: bool = False,
         else:
             self.register_buffer('eps', torch.Tensor([eps]))
         if edge_dim is not None:
-            if hasattr(self.nn[0], 'in_features'):
-                in_channels = self.nn[0].in_features
+            if isinstance(self.nn, torch.nn.Sequential):
+                nn = self.nn[0]
             else:
-                in_channels = self.nn[0].in_channels
+                nn = self.nn
```
rusty1s, line 148: I think we can safely drop the else here.

rusty1s (approved):
Thanks!

torch_geometric/nn/conv/gin_conv.py
Padarn: agree

PR #5158, opened 2022-08-07
GraphGym: Fix parsing of scientific notation when aggregating results

MoetaYuko:
Fix parsing strings like foo-a=3-b=1e-4-c=1e-5


rusty1s (approved):
Thanks!

Issue #5190, opened 2022-08-11
`RandomLinkSplit` enhancement for undirected graphs


saiden89:
### 🚀 The feature, motivation and pitch

As briefly discussed in #5169, `RandomLinkSplit` currently returns `edge_label_index` in directed form, regardless of the input graph being undirected or the arguments passxfed to the transform. We can extend to possibility to return appropriate supervision edges according to the usecase.
As far as I can tell we can do so in various ways:
- Let PyG guess the directiveness of the input graph  and perform the split accordingly: this way undirected graphs will have only undirected edge indices and viceversa. Additionally, give the user the possibility to override the settings.
- Keep the current defaults (without guessing), but give the user the possibility to override.

As far as my (quite limited) experience goes, if the usecase is centered on undirected network(s) we will probably stick to those throughout all the steps. The same goes for the directed counterpart. Even when dealing with heterogenous network, there is rarely the need to have different types of message passing (MP) and supervision edges within the same edge types. So I guess that for most cases keeping in sync MP and validation indices is preferrable.

### Alternatives

Currently, it is possible to manually carry out the necessary transformations after the split, on a per use basis.

rusty1s:
I have no strong opinion here. As discussed in #5169, I think that if one wants to predict undirected edges, one should use a symmetric encoder or take care of both directions on its own. We can make this more clear in the documentation.

Wondering if other people have some thoughts here (@Padarn @lightaime @wsad1) - if this feels indeed overly complicated, then let's change it :)

Padarn:
> I think that if one wants to predict undirected edges, one should use a symmetric encoder or take care of both directions on its own.

I tend to agree with this. But I do think the current interface makes this a bit unclear (its a little confusing that the `undirected` flag does not make the labels undirected). I'd prefer to provide clear documentation and leave it as is.

But also not a very strong opinion.

Tigerrr07:
# Problem background
When applying `RandomLinkSplit` in an undirected HeteroData `data`, in figure1 (I already store reverse edges in original `edge_index`, see figure 2 as an example), I find  in splitted data, such as `train_data`, `val_data` or `test_data`, the `edge_index` of one `edge_type` is not undirected , I test it using `is_undirected()` method.
![image](https://user-images.githubusercontent.com/86994637/184830074-edfb2210-7e02-4d1d-9ae7-6ec860bfe47e.png)
[...]

PR #5193, opened 2022-08-13
Updated docstring for `RandomLinkSplit`

saiden89:
Update doctring as discussed in #5190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Removed
 
 ## [2.1.0] - 2022-08-17
-###
+### Added
 - Added `edge_label_time` argument to `LinkNeighborLoader` ([#5137](https://github.com/pyg-team/pytorch_geometric/pull/5137), [#5173](https://github.com/pyg-team/pytorch_geometric/pull/5173))
 - Let `ImbalancedSampler` accept `torch.Tensor` as input ([#5138](https://github.com/pyg-team/pytorch_geometric/pull/5138))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Removed
 
 ## [2.1.0] - 2022-08-17
-### Added
```
Padarn, line 12: Is this intentional

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+###
```
Padarn, line 12:
```suggestion
### Added
```

torch_geometric/transforms/random_link_split.py
```diff
@@ -40,7 +40,8 @@ class RandomLinkSplit(BaseTransform):
             (default: :obj:`0.2`)
         is_undirected (bool): If set to :obj:`True`, the graph is assumed to be
             undirected, and positive and negative samples will not leak
-            (reverse) edge connectivity across different splits.
+            (reverse) edge connectivity across different splits. Note that
```
Padarn, line 43: Maybe something like "Note that this only affects the graph split, label data will not be returned undirected". I think its good to be clear about what isn't done.

Padarn (changes requested):
Looks good but I think maybe one mistake in the Changelog?

CHANGELOG.md
saiden89: It's not. First time editing a chagelog and didn't notice that the header had also a title, sorry.

CHANGELOG.md
saiden89: It's not, first time editing a changelog and I didn't notice that header, sorry.

Padarn (approved):

rusty1s (approved):

rusty1s:
Thank you!

PR #5207, opened 2022-08-16
fix device mismatch bug in metapath2vec example

EdisonLeeeee:
This PR fixes the device mismatch bug in [metapath2vec example](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/hetero/metapath2vec.py). This bug was also mentioned in [this discussion](https://torchgeometricco.slack.com/archives/CRAR9HK2P/p1660639673925579).


rusty1s (approved):
Thank you!

PR #5210, opened 2022-08-16
Remove `MutableMapping` base class in `FeatureStore`

rusty1s:
Ideally, we want to let `FeatureStore` inherit from`MutableMapping` to clearly indicate its behavior and usage to the user. However, having `MutableMapping` as a base class leads to strange behavior in combination with PyTorch and PyTorch Lightning, in particular since these libraries use customized logic during mini-batch for `Mapping` base classes.

### DIFF
--- a/torch_geometric/data/feature_store.py
+++ b/torch_geometric/data/feature_store.py
@@ -240,6 +240,13 @@ class AttrView(CastMixin):
                 f'attr={self._attr})')
 
 
+# TODO (manan, matthias) Ideally, we want to let `FeatureStore` inherit from
+# `MutableMapping` to clearly indicate its behaviour and usage to the user.
+# However, having `MutableMapping` as a base class leads to strange behaviour
+# in combination with PyTorch and PyTorch Lightning, in particular since these
+# libraries use customized logic during mini-batch for `Mapping` base classes.
+
+
 class FeatureStore:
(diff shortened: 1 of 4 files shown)


torch_geometric/data/feature_store.py
```diff
@@ -241,7 +240,7 @@ def __repr__(self) -> str:
                 f'attr={self._attr})')
 
 
-class FeatureStore(MutableMapping):
```
Padarn, line 244: Might be good to add a one-liner in the PR about why (otherwise if I found this later while trying to understand the code I'd have no idea - actually I have no idea now :-))

mananshah99 (approved):
Let's leave a comment about the rationale here (in particular, that there seems to be some strange interaction with torch DDP); but we can revert for now. :(

torch_geometric/data/feature_store.py
rusty1s: Updated the description - sorry about missing it.

Ideally, we want to let `FeatureStore` inherit from`MutableMapping` to clearly indicate its behavior and usage to the user. However, having `MutableMapping` as a base class leads to strange behavior in combination with PyTorch and PyTorch Lightning, in particular since these libraries use customized logic during mini-batch for `Mapping` base classes.

PR #5227, opened 2022-08-18
Replace `torch.nn.ModuleDict` with `CustomModuleDict`

kiriti-kumo:
`torch.nn.ModuleDict` doesn't allow `.` to be used in the keys. This prevents using node types with `.` in the name.
`CustomModuleDict` will support it by converting the `.` to `#` in the internal representation and converts it back to `.` in the external representation.


yaoyaowd (commented):
LGTM. Do we know any legacy reasons that `.` is not supported?

torch_geometric/utils/module_dict.py
```diff
@@ -0,0 +1,44 @@
+from typing import Iterable, Mapping, Optional
+
+from torch.nn import Module, ModuleDict
+
+
+# `torch.nn.ModuleDict` doesn't allow `.` to be used in the keys.
+# `CustomModuleDict` will support it by converting the `.` to `#` in the
+# internal representation and converts it back to `.` in the external
+# representation.
+class CustomModuleDict(ModuleDict):
```
rusty1s, line 10:
```suggestion
class ModuleDict(torch.nn.ModuleDict):
```

torch_geometric/utils/module_dict.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, modules: Optional[Mapping[str, Module]] = None) -> None:
+        # Replace the keys in modules.
+        if modules:
+            modules = {
+                CustomModuleDict.to_internal_key(key): module
```
rusty1s, line 15:
```suggestion
                self.to_internal_key(key): module
```

torch_geometric/utils/module_dict.py
```diff
@@ -0,0 +1,44 @@
...
+    @staticmethod
+    def to_internal_key(key: str) -> str:
+        return key.replace(".", "#")
+
+    @staticmethod
+    def to_external_key(key: str) -> str:
+        return key.replace("#", ".")
+
+    def __getitem__(self, key: str) -> Module:
+        return super().__getitem__(CustomModuleDict.to_internal_key(key))
```
rusty1s, line 29:
```suggestion
        return super().__getitem__(self.to_internal_key(key))
```
and in the remaining ones.

torch_geometric/utils/module_dict.py
```diff
@@ -0,0 +1,44 @@
+from typing import Iterable, Mapping, Optional
```
rusty1s, line 1: I think we should move this to `nn/module_dict.py`.

torch_geometric/utils/__init__.py
```diff
@@ -36,6 +37,7 @@
 from .scatter import scatter
 
 __all__ = [
+    'CustomModuleDict',
```
rusty1s, line 40: No reason to expose this. I see this as internal only.

torch_geometric/utils/module_dict.py
rusty1s, line 10: Can we add a basic test for this module?

torch_geometric/nn/fx.py
```diff
@@ -192,7 +194,7 @@ def _init_submodule(self, module: Module, target: str) -> Module:
                 for i, submodule in enumerate(module)
             ])
         elif isinstance(module, ModuleDict):
-            return ModuleDict({
+            return CustomModuleDict({
```
rusty1s, line 197: This will likely not work in `torch.fx` - I would be in favor of just doing the adjustment in `HeteroConv` for now. Would that work for you?

test/nn/conv/test_hetero_conv.py
```diff
@@ -21,19 +21,21 @@ def get_edge_index(num_src_nodes, num_dst_nodes, num_edges):
 @pytest.mark.parametrize('aggr', ['sum', 'mean', 'min', 'max', None])
 def test_hetero_conv(aggr):
     data = HeteroData()
-    data['paper'].x = torch.randn(50, 32)
+    data['src.paper'].x = torch.randn(50, 32)
```
rusty1s, line 24: Move this to its own test?

rusty1s (commented):
Thanks! Looks good. Importantly, I don't think we have to replace every occurence of `ModuleDict` for now - let's keep it simple and start with `HeteroConv`.

rusty1s:
@yaoyaowd I believe PyTorch provides nested module access via dot-syntax in `ModuleDict`.

rusty1s:
Tests failure does not seem related to this PR. I will look into this separately.

kiriti-kumo:
@yaoyaowd  As per https://github.com/pytorch/pytorch/pull/6639 - seems like `torch` uses `.` as a reserved delimiter.

torch_geometric/nn/fx.py
kiriti-kumo: Yes, will update the PR to just limit to `HeteroConv` as of now.

mananshah99 (approved):
LGTM after comments are addressed, thanks :)

test/nn/conv/test_hetero_conv.py
kiriti-kumo: Done

rusty1s (approved):
This is amazing. Thank you!

PR #5232, opened 2022-08-18
Add experimental mode for `utils.scatter`

dszwicht:
Add experimental mode for `utils.scatter`.

Introduces several functions that allow us to control experimental options (currently only `pytorch_scatter`).
Implementation of `utils.scatter` is selected now based on experimental option value instead of environment variable. Selection process also happens on the fly instead at the import time - for greater usage flexibility.

Related PR: #5120

### DIFF
--- a/torch_geometric/experimental.py
+++ b/torch_geometric/experimental.py
@@ -1,51 +1,48 @@
-from typing import Dict
+from typing import List, Optional, Union
 
-__experimental_options__ = {'pytorch_scatter': False}
-
-
-def is_experimental_option_enabled(option: str) -> bool:
-    r"""Returns :obj:`True`, if the experimental :attr:`option` is enabled.
-    See :class:`~torch_geometric.experimental.set_experimental_options` for
-    a list of available options."""
-    return __experimental_options__.get(option, False)
+__experimental_flag__ = {'scatter_reduce': False}
 
 
-def disable_all_experimental_options() -> None:
-    r"""Disables all experimental options."""
-    for key in __experimental_options__:
-        __experimental_options__[key] = False
+def is_experimental_mode_enabled(
+        options: Optional[Union[str, List[str]]] = None) -> bool:
+    r"""Returns :obj:`True` if the experimental mode is enabled."""
+    if options is None:
+        options = list(__experimental_flag__.keys())
+    if isinstance(options, str):
+        options = [options]
+    return all([__experimental_flag__[option] for option in options])
 
 
-class set_experimental_options:
-    r"""Context-manager that sets experimental options on or off.
-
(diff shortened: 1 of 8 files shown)


dszwicht:
@rusty1s Please take a look.

torch_geometric/experimental.py
```diff
@@ -0,0 +1,51 @@
...
+    return __experimental_options__.get(option, False)
+
+
+def disable_all_experimental_options() -> None:
+    r"""Disables all experimental options."""
+    for key in __experimental_options__:
+        __experimental_options__[key] = False
+
+
+class set_experimental_options:
```
Padarn, line 19: Just a thought (if we want to avoid breaking naming conventions. We can have a class 
```
class Options:
     # does what your context manager does
```
and a function that wraps that so it can be used as a context manager
```
def option_context
     ....
     return Options(...)
```
This is how some other libraries deal with this, for example https://github.com/mwaskom/seaborn/blob/master/seaborn/rcmod.py

torch_geometric/utils/scatter.py
```diff
@@ -1,64 +1,79 @@
...
-        out = out.clone()
-
-        return out
-
-else:
-    import torch_scatter
-
-    def scatter(src: Tensor, index: Tensor, dim: int = -1,
+
+class ScatterImpl(torch.nn.Module):
```
Padarn, line 61: Sorry stupid question - why do we want this to be a module?

torch_geometric/experimental.py
dszwicht: Hi @Padarn, thanks for your suggestion. I don't think we break any convention here, PEP8 says this about class naming ([source](https://peps.python.org/pep-0008/#class-names)): `The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.` Approach taken here is also aligned with existing in [debug.py](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/debug.py).

torch_geometric/utils/scatter.py
dszwicht: The short answer is that we need this for `torch.jit.script` to work with `scatter` function. I just followed JIT troubleshooting scheme.

torch_geometric/experimental.py
Padarn: 🆗

torch_geometric/utils/scatter.py
Padarn: Got it, thanks for the details

rusty1s (approved):
Pretty cool! Thanks a lot :)

PR #5233, opened 2022-08-18
Add missing test labels in `HGBDataset`

EdisonLeeeee:
Hi, I found a bug when using `torch_geometric.dataset.HGBDataset`. Here are my codes to reproduce the bug:
```python

import torch
from torch_geometric.datasets import HGBDataset
from collections import Counter

datasets = ['acm', 'dblp', 'freebase', 'imdb']
labeled_types = ['paper', 'author', 'book', 'movie']

for name, types in zip(datasets, labeled_types):
    data = HGBDataset("~/data/pygdata/HGB", name)[0]
    if name != 'imdb':
        print(name, Counter(data[types].y[data[types].test_mask].tolist()))
    else:
        print(name, torch.any(data[types].y[data[types].test_mask]))
    
```
And the outputs are:
```bash
acm Counter({-1: 1059})
dblp Counter({-1: 1420})
freebase Counter({-1: 2784})
imdb tensor(False)
```
This indicates that all test labels are missing on four datasets.

I checked the code in `process` and found that it seems to have forgotten to assign values to the test labels. After adding these codes, the bug has been fixed. The output becomes:

```bash
acm Counter({2: 364, 0: 355, 1: 340})
dblp Counter({2: 404, 0: 404, 3: 354, 1: 258})
freebase Counter({1: 1372, 2: 591, 4: 347, 0: 146, 6: 141, 5: 127, 3: 60})
imdb tensor(True)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.2.0] - 2022-MM-DD
 ### Added
+- Added missing test labels in `HGBDataset` ([#5233](https://github.com/pyg-team/pytorch_geometric/pull/5233))
 - Added `BaseStorage.get()` functionality ([#5240](https://github.com/pyg-team/pytorch_geometric/pull/5240))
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
Thank you!

PR #5250, opened 2022-08-20
Rewrite tests to not depend on currently broken `dblp` dataset

Padarn:
This PR changes tests that were using the `dblp` as it seems to be currently broken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `BaseStorage.get()` functionality ([#5240](https://github.com/pyg-team/pytorch_geometric/pull/5240))
 - Added a test to confirm that `to_hetero` works with `SparseTensor` ([#5222](https://github.com/pyg-team/pytorch_geometric/pull/5222))
 ### Changed
-- Changed tests relying on `dblp` datasets to instead se  synthetic data. ([#5250](https://github.com/pyg-team/pytorch_geometric/pull/5250))
+- Changed tests relying on `dblp` datasets to instead use  synthetic data. ([#5250](https://github.com/pyg-team/pytorch_geometric/pull/5250))
 ### Removed
 
 ## [2.1.0] - 2022-08-17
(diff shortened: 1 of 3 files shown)


test/loader/test_hgt_loader.py
```diff
@@ -175,13 +175,3 @@ def forward(self, x, edge_index, edge_weight):
     out2 = hetero_model(hetero_batch.x_dict, hetero_batch.edge_index_dict,
                         hetero_batch.edge_weight_dict)['paper'][:batch_size]
     assert torch.allclose(out1, out2, atol=1e-6)
-
-
-@withPackage('torch_sparse>=0.6.15')
```
Padarn, line 180: Couldn't see what this was testing that wasn't tested by adding the above line, so have removed it

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `BaseStorage.get()` functionality ([#5240](https://github.com/pyg-team/pytorch_geometric/pull/5240))
 - Added a test to confirm that `to_hetero` works with `SparseTensor` ([#5222](https://github.com/pyg-team/pytorch_geometric/pull/5222))
 ### Changed
+- Changed tests relying on `dblp` datasets to instead se  synthetic data. ([#5250](https://github.com/pyg-team/pytorch_geometric/pull/5250))
```
rusty1s, line 14:
```suggestion
- Changed tests relying on `dblp` datasets to instead use  synthetic data. ([#5250](https://github.com/pyg-team/pytorch_geometric/pull/5250))
```

test/data/test_lightning_datamodule.py
```diff
@@ -274,9 +281,17 @@ def test_lightning_hetero_node_data(get_dataset):
 @onlyFullTest
 @withPackage('pytorch_lightning')
 def test_lightning_hetero_link_data(get_dataset):
```
rusty1s, line 283:
```suggestion
def test_lightning_hetero_link_data():
```

test/loader/test_hgt_loader.py
rusty1s: It tests that `HGTLoader` always returns the full set of node and edge types. We can disable this test for via `pytest.skipif` until the dataset is fixed.

Padarn: Okay. I've also added the same test in the above
```assert set(batch.edge_types) == set(data.edge_types)```
but makes sense to keep it

Padarn: I added just a skip - it didn't seem worth giving a condition. I'll instead open an issue now and note  this as a todo? WDTY?

rusty1s (approved):

test/loader/test_hgt_loader.py
rusty1s: Yeah, that‘s what I meant. Thanks!

Padarn: Done here https://github.com/pyg-team/pytorch_geometric/issues/5251

PR #5253, opened 2022-08-20
Parameterize the `run_batch.sh` script for GraphGym

jprorama:
Add bash defaults for the variables in the script.
Enables overriding the variable values on the commandline so the script
can easily reference different experiment, grid, etc values at runtime.

### DIFF
--- a/graphgym/run_batch.sh
+++ b/graphgym/run_batch.sh
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
-CONFIG=example_node
-GRID=example
-REPEAT=3
-MAX_JOBS=8
-SLEEP=1
-MAIN=main
+CONFIG=${CONFIG:="example_node"}
+GRID=${GRID:="example"}
+REPEAT=${REPEAT:="3"}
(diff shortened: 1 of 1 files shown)


graphgym/run_batch.sh
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
-CONFIG=example_node
-GRID=example
-REPEAT=3
-MAX_JOBS=8
-SLEEP=1
-MAIN=main
+CONFIG=${CONFIG:="example_node"}
+GRID=${GRID:="example"}
```
Padarn, line 4:
```suggestion
GRID=${GRID:-example}
```
No need for the quality here as you are setting it anyway, the equality is used to do an assignment, see https://unix.stackexchange.com/questions/122845/using-a-b-for-variable-assignment-in-scripts
Also you don't need the quotes?

Padarn (commented):
Looks sensible, but left comments on the style used

Padarn (approved):

Padarn:
Can you add a note to the changelog?

rusty1s (approved):
Thanks!

PR #5254, opened 2022-08-21
Enable experimental mode in inference benchmarks

dszwicht:
Enable experimental mode in inference benchmarks.

Related PRs: #5120, #5232

### DIFF
--- a/benchmark/inference/inference_benchmark.py
+++ b/benchmark/inference/inference_benchmark.py
@@ -4,6 +4,7 @@ from timeit import default_timer
 import torch
 from utils import get_dataset, get_model
 
+import torch_geometric
 from torch_geometric.loader import NeighborLoader
 from torch_geometric.nn import PNAConv
 
@@ -93,8 +94,13 @@ def run(args: argparse.ArgumentParser) -> None:
                         model.eval()
 
                         start = default_timer()
-                        model.inference(subgraph_loader, device,
-                                        progress_bar=True)
+                        if args.experimental_mode:
+                            with torch_geometric.experimental_mode():
+                                model.inference(subgraph_loader, device,
+                                                progress_bar=True)
+                        else:
+                            model.inference(subgraph_loader, device,
+                                            progress_bar=True)
                         stop = default_timer()
                         print(f'Inference time={stop-start:.3f} seconds\n')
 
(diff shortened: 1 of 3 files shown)


benchmark/inference/inference_benchmark.py
```diff
@@ -93,8 +94,14 @@ def run(args: argparse.ArgumentParser) -> None:
                         model.eval()
 
                         start = default_timer()
-                        model.inference(subgraph_loader, device,
-                                        progress_bar=True)
+                        if args.maybe_with_experimental:
```
Padarn, line 97: When would this be False? The default is an empty list.

benchmark/inference/inference_benchmark.py
```diff
@@ -121,6 +128,11 @@ def run(args: argparse.ArgumentParser) -> None:
         '--hetero-num-neighbors', default=-1, type=int,
         help='number of neighbors to sample per layer for hetero workloads')
     argparser.add_argument('--num-workers', default=2, type=int)
+    argparser.add_argument(
+        '--maybe-with-experimental', nargs='+', default=[], type=str,
```
Padarn, line 132: Why `maybe_with`? Isn't it just `experimental-options`?

Padarn (changes requested):

benchmark/inference/inference_benchmark.py
dszwicht: Empty list evaluates as `False`.

benchmark/inference/inference_benchmark.py
dszwicht: Please see the description of this argument: `(...) It may happen that it will be not possible to use some of them (warning will be produced at such a scenario)`. For example, `pytorch_scatter` experimental requires `torch>=1.12`, so it may happen that the user of this benchmark will use `--maybe-with-experimental` and it will not have any effect (in case of `torch<1.12`).

rusty1s: Can't we use `action="store_true"` here, see [here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/gcn.py#L17)? Also rename to `args.experimental_node`.

dszwicht: Yes, we could, but that makes sense only for an argument like `with_pytorch_scatter`. I assumed here, that there might be more experimental options in the future

rusty1s: Oh, I see. I think we should not over-complicate this for now - we can revisit this once we add more experimental options. Does that work for you?

dszwicht: It's ok for me.

benchmark/inference/inference_benchmark.py
Padarn: Oh haha understood, thanks.

Padarn (approved):

PR #5270, opened 2022-08-24
feat(graphstore): support `num_nodes`, enabling `Tuple[FeatureStore, GraphStore]` in `LightningLinkData`

mananshah99:
There is still cleanup work to be done around `LightningNodeData`, `LightningLinkData`, and how we fetch `num_nodes`. This is simply a hotfix to have functionality working, the cleanup will happen shortly.


torch_geometric/data/lightning_datamodule.py
```diff
@@ -481,11 +487,36 @@ def __init__(
...
+            if isinstance(data, (Data, HeteroData)):
+                if input_type is not None:
+                    num_src_nodes = data[input_type[0]].num_nodes
+                    num_dst_nodes = data[input_type[-1]].num_nodes
+                else:
+                    num_src_nodes = num_dst_nodes = data.num_nodes
             else:
-                num_src_nodes = num_dst_nodes = data.num_nodes
+                # TODO improve num_nodes logic, currently we can only do this
+                # by explicitly querying a feature tensor
```
yaoyaowd, line 499: why? I thought that we still expose num_nodes in table data.

mananshah99: this is the case of a `FeatureStore` generic object, which does not currently support `num_nodes`

torch_geometric/data/lightning_datamodule.py
```diff
@@ -481,11 +487,36 @@ def __init__(
 
         if loader in ['neighbor', 'link_neighbor']:
             input_type = get_edge_label_index(data, input_train_edges)[0]
-            if input_type is not None:
-                num_src_nodes = data[input_type[0]].num_nodes
-                num_dst_nodes = data[input_type[-1]].num_nodes
+
+            if isinstance(data, (Data, HeteroData)):
+                if input_type is not None:
```
rusty1s, line 492: I think we can drop this logic here.

torch_geometric/data/lightning_datamodule.py
rusty1s: Can we add this to `FeatureStore`? I feel we are replicating this logic all over PyG.

torch_geometric/testing/feature_store.py
```diff
@@ -37,6 +37,9 @@ def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
         if attr.index is None:
             return tensor
 
+        if isinstance(attr.index, torch.Tensor) and attr.index.numel() == 0:
+            return []
```
rusty1s, line 41: This does not match the return type annotation. Why do we want to return an empty list?

torch_geometric/data/lightning_datamodule.py
mananshah99: I think we still need this, if we want to support `Data`.

torch_geometric/data/lightning_datamodule.py
mananshah99: I think we can, but I would prefer to do that in a separate PR. The best way to incorporate this would be to have a RemoteBackend class that performs operations on a Tuple[FeatureStore, GraphStore], one of which is `RemoteBackend.num_nodes(FeatureStore, GraphStore)`. 

I can implement this in a separate PR and get rid of all the cases in PyG where we introduce this logic, but it may be a good idea to implement this first for correctness. Wdyt?

torch_geometric/testing/feature_store.py
mananshah99: This is in the case that the index has zero elements, which may happen if we are trying to get features for a group name that has no indexes after sampling.

mananshah99:
Completely agree that we need to stop replicating this logic across PyG. My preference is to handle this in a separate PR, and merge this one so we maintain correctness for link prediction with remote backends. The separate PR will be out soon.

test/data/test_lightning_datamodule.py
```diff
@@ -327,3 +329,41 @@ def test_lightning_hetero_link_data():
...
+        loader='neighbor',
+        num_neighbors=[5],
+        batch_size=32,
+        num_workers=0,
+    )
+
+    for batch in datamodule.train_dataloader():
+        assert 'edge_label' in batch['author', 'paper']
+        assert 'edge_label_index' in batch['author', 'paper']
+        break
```
rusty1s, line 369:
```suggestion
    batch = next(iter(datamodule.train_dataloader()))
    assert 'edge_label' in batch['author', 'paper']
    assert 'edge_label_index' in batch['author', 'paper']
```

torch_geometric/data/lightning_datamodule.py
rusty1s: I don't think so - for `Data` and `HeteroData`, `num_src_nodes` and `num_dst_nodes` are automatically inferred, see https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/loader/link_neighbor_loader.py#L61-L65.

torch_geometric/data/lightning_datamodule.py
rusty1s: Does it need to be so complex? I think it makes sense to just support `GraphStore.num_src_nodes(edge_type)` and `GraphStore.num_dst_nodes(edge_type)`. I am not 100% convinced we should use the `FeatureStore` for this at all.

torch_geometric/testing/feature_store.py
```diff
@@ -37,6 +37,9 @@ def _get_tensor(self, attr: TensorAttr) -> Optional[FeatureTensorType]:
         if attr.index is None:
             return tensor
 
+        if isinstance(attr.index, torch.Tensor) and attr.index.numel() == 0:
```
rusty1s, line 40: Can't we simply index with an empty index tensor here? Using `torch.tensor([])` is dangerous, in particular because the shape of the output should be `(0, num_features)` rather than `(0, )`.

torch_geometric/data/lightning_datamodule.py
rusty1s: We also have `EdgeAttr.size` we can leverage for this, right?

yaoyaowd: Yes, looks like `EdgeAttr` is easier.

yaoyaowd (approved):
Minor comment.

torch_geometric/data/lightning_datamodule.py
mananshah99: Yes, this is true in `LinkNeighborLoader`, but not in `LightningLinkData`.

rusty1s: Not sure I understand, just let them be `None` instead?

mananshah99: Sorry, I misunderstood. Yes, we can remove this.

torch_geometric/data/lightning_datamodule.py
mananshah99: Yes, this interface works. It may still be worth considering inferring based on features as well for link prediction tasks where the input edge type is not a part of the graph. For this case, we are passing-through kwargs as before.

torch_geometric/data/hetero_data.py
```diff
@@ -846,6 +846,16 @@ def get_all_edge_attrs(self) -> List[EdgeAttr]:
 
         return out
 
+    def _num_src_nodes(self, edge_attr: EdgeAttr) -> int:
+        key = self._to_canonical(edge_attr.edge_type)
+        src, _, _ = key
+        return self[src].num_nodes
```
rusty1s, line 852:
```suggestion
        src, _, _ = self._to_canonical(edge_attr.edge_type)
        return self[src].num_nodes
```

torch_geometric/data/hetero_data.py
```diff
@@ -846,6 +846,15 @@ def get_all_edge_attrs(self) -> List[EdgeAttr]:
...
         return out
 
+    def _num_src_nodes(self, edge_attr: EdgeAttr) -> int:
+        src, _, _ = self._to_canonical(edge_attr.edge_type)
+        return self[src].num_nodes
+
+    def _num_dst_nodes(self, edge_attr: EdgeAttr) -> int:
+        key = self._to_canonical(edge_attr.edge_type)
+        _, _, dst = key
+        return self[dst].num_nodes
```
rusty1s, line 856:
```suggestion
        _, _, dst = self._to_canonical(edge_attr.edge_type)
        return self[dst].num_nodes
```

rusty1s (approved):

PR #5271, opened 2022-08-24
Refactor `ARMAConv` example

rfdavid:
This PR is a small change on arma example to follow similar standards as the other examples. 
The `dataset` and `data` are moved out from Net class, so it is a bit cleaner and easier for other people who are implementing this example in a different codebase.


rusty1s (approved):
Thank you!

Issue #5087, opened 2022-07-29
Support for other integer types by MessagePassing


DomInvivo:
### 🐛 Describe the bug

Why does PyG enforces `edge_index` to be of type long? Certain graphs can work properly with `int16` such as molecules, which will rarely surpass the 32767, unless batch sizes larger than 1000 are used. They have an average of 20 atoms. Instead, we could simply ensure that there are no negative numbers (which happen when the edge index overflows), or that the maximum of a specific datatype is not reached.

Also, some hardwares do not support `long`, such as TPU and IPU, which are limited to `int32`.

https://github.com/pyg-team/pytorch_geometric/blob/412ae53d0897660a2968283ed2cc60b5928c1229/torch_geometric/nn/conv/message_passing.py#L183

### Environment

* PyG version: All
* PyTorch version: All
* OS: All
* Python version: All
* CUDA/cuDNN version: All
* How you installed PyTorch and PyG (`conda`, `pip`, source): Any
* Any other relevant information (*e.g.*, version of `torch-scatter`): Any

rusty1s:
This is mostly due to how PyTorch works, e.g., `index_select` only works with indices of `dtype=torch.long`. Not much we can do about it currently, sorry!

DomInvivo:
Since [Pytorch 1.8.0](https://pytorch.org/docs/1.8.0/generated/torch.index_select.html), `index_select` supports IntTensor as well as LongTensor, according to the docs.

rusty1s:
Oh, my bad. I just tested with `torch.short` and was convinced it is only working for `torch.long`. I guess we can then start looking into supporting both `torch.long` and `torch.int`.

hatemhelal:
I can take a look at this: as a first attempt I'll try to relax the check for `edge_index.dtype == torch.long` and add some additional tests in `test/nn/conv/test_message_passing.py` and see what breaks.

rusty1s:
Sounds good. Thanks! We also added support to make use of `torch.scatter_reduce` (see `utils/scatter.py`) which will help in this transition.

EdisonLeeeee:
> I can take a look at this: as a first attempt I'll try to relax the check for edge_index.dtype == torch.long and add some additional tests in test/nn/conv/test_message_passing.py and see what breaks.

Hi @howardjp, I've tried and it breaks at `torch_scatter.scatter_add`. 

[...]

PR #5281, opened 2022-08-26
Allow any integer tensors when checking edge_index input to message passing

hatemhelal:
Partial fix for #5087 

This relaxes the `__check_input__` method of the message passing interface to allow any integer types for the `edge_index`.  This is necessary to allow PyTorch to dispatch to different execution backends (see [this tutorial](https://pytorch.org/tutorials/advanced/extend_dispatcher.html#testing-your-backend-against-native-pytorch-backends)).

This effectively defers the question of which types are supported to the underlying device backend.

### DIFF
--- a/test/nn/conv/test_message_passing.py
+++ b/test/nn/conv/test_message_passing.py
@@ -522,3 +522,23 @@ def test_message_passing_with_aggr_module(aggr_module):
     out = conv(x, edge_index)
     assert out.size(0) == 4 and out.size(1) in {8, 16}
     assert torch.allclose(conv(x, adj.t()), out)
+
+
+def test_message_passing_int32_edge_index():
+    # Check that we can dispatch an int32 edge_index up to aggregation
+    x = torch.randn(4, 8)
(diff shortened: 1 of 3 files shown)


test/nn/conv/test_message_passing.py
```diff
@@ -522,3 +522,23 @@ def test_message_passing_with_aggr_module(aggr_module):
...
+    def cast_index_hook(module, inputs):
+        input_dict = inputs[-1]
+        input_dict['index'] = input_dict['index'].long()
+        return (input_dict, )
+
+    conv = MyConv(8, 32)
+    conv.register_aggregate_forward_pre_hook(cast_index_hook)
+
+    out = conv(x, edge_index, edge_weight)
+    assert out.size() == (4, 32)
```
rusty1s, line 544:
```suggestion
    assert conv(x, edge_index, edge_weight).size() == (4, 32)
```

rusty1s (approved):
Can we also add this change to the `CHANGELOG.md`?

Issue #5289, opened 2022-08-27
Make validation of `dim_size` argument in `Aggregation` interface optional


hatemhelal:
### 🚀 The feature, motivation and pitch

The new `Aggregation` interface supports providing the `dim_size` argument for the expected size in the aggregation dimension.  This is validated as follows:

https://github.com/pyg-team/pytorch_geometric/blob/615d0b609aa0aceff2eb3813d1d802f250410919/torch_geometric/nn/aggr/base.py#L109-L113

This helps ensure correctness when first developing a model but can incur an additional performance cost when porting the model to other PyTorch execution backends.  For example when using the [LazyTensor](https://pytorch.org/blog/understanding-lazytensor-system-performance-with-pytorch-xla-on-cloud-tpu/) abstraction developed for XLA, this validation would introduce a sync point to return the result of `max` back to the main python process on the CPU.  

Making this validation optional should help make it easier to port a model from running on the default CPU to other supported device types.  This also may enable further whole graph optimisations that would otherwise not be possible to support.

The proposal is to a similar same pattern as established in the PyTorch `Distribution` interface:

https://github.com/pytorch/pytorch/blob/3bcc19b29ae1a41db8cabef6154041765a8e7b03/torch/distributions/distribution.py#L16-L33

### Alternatives

We could also integrate with the PyG `is_debug_enabled` but this may be lead us to default to not running this validation.

PR #5290, opened 2022-08-27
Add `Aggregation.set_validate_args` method to toggle validation of `dim_size`

hatemhelal:
Proposal for #5289

This change should default to running the eager validation but allows a user to override this behaviour when using different PyTorch execution backend.


rusty1s (approved):
Wow, this is super cool! Thanks for making PyG better :)

PR #5307, opened 2022-08-29
refactor(data): simplify remote backend `num_nodes` computation

mananshah99:
This PR introduces `remote_backend_utils` (currently just a set of functions, open to making it a static class as well) that helps define common utilities to be used across remote backends. The first (and perhaps most useful) function included is `num_nodes`, which allows one to infer the number of nodes in a (node type, edge type) by leveraging attributes in a feature store and graph store. This significantly simplifies internal code and also simplifies some external interfaces as well.


torch_geometric/data/remote_backend.py
```diff
@@ -0,0 +1,72 @@
+# This file defines a set of utilities for remote backends (backends that are
+# characterize as Tuple[FeatureStore, GraphStore]).
+from typing import Tuple, Union
+
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
+from torch_geometric.typing import EdgeType, NodeType
+
+
+def num_nodes(
```
yaoyaowd, line 10: I actually like the previous implementation for each feature store and graph store better. That is more intuitive to me compare with get all stores and query the nodes with a query (which sounds more heavy weight).
Could we just expose num_nodes and sizes for features and edges ?

mananshah99: Sure, we can expose num nodes for features and edges, but there are instances where an edge index is not part of a graph (e.g. link prediction) or we just want to get the number of nodes from a group name, not a full edge index type. I also don't think it's too heavyweight to query all the edge attributes and node attributes, since these are small data structures and are somewhat bounded. 

However, if you think this overhead would be significant, we can expose methods in the feature store and graph store to get a TensorAttr/EdgeAttr from a partial specification, and directly query these (instead of listing all of them and iterating). This can be done as an improvement in a separate PR. Wdyt?

yaoyaowd: After a second look, I think the diverge comes from whether we want to introduce this new `remote_backend` or consolidate logic in `GraphStore`, i.e we can consider `GraphStore` to be a backend that can be `remote`. And then have an api like
```
class GraphStore:
  def num_nodes(Union[NodeType, EdgeType]):
```
We can also hide `FeatureStore` behind `GraphStore` or `FeatureStore` metadata as part of information in `GraphStore`.

If we want to introduce `remote_backend`, you may want to think more about whether it is a `Backend` object or it is a helper provide util functions talk to `FeatureStore` and `GraphStore`.

mananshah99: A few clarifications
* Yes, a GraphStore can be remote.
* We do not want to hide FeatureStore behind GraphStore; ideally, a feature store owns features, and a graph store owns the graph. The complications come because there can be edges in the graph store with no corresponding features in the feature store, etc. as mentioned above.
* `remote_backend` is not a `Backend` object; it's really a helper providing utility functions to talk to `FeatureStore` and `GraphStore` (as in the comment at the top of `remote_backend.py`). However, I agree that the name is confusing. I can change it to `remote_backend_util.py`.

torch_geometric/data/remote_backend.py
```diff
@@ -0,0 +1,72 @@
...
+    feature_store: FeatureStore,
+    graph_store: GraphStore,
+    query: Union[NodeType, EdgeType],
+) -> Union[int, Tuple[int, int]]:
+    r"""Returns the number of nodes in the node type or the number of source
+    and destination nodes in an edge type by sequentially accessing attributes
+    in the feature and graph stores that reveal this number. Returns a single
+    integer in the case of a feature store and a tuple of integers in the case
+    of a graph store."""
```
wsad1, line 19: I am confused by the part that says `Returns a single integer...`. Don't the number of integers returned depend only on the query.
```suggestion
    in the feature and graph stores that reveal this number. """
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -34,37 +32,17 @@ def __init__(
...
+        if self.data_cls == 'custom':
+            feature_store, graph_store = data
+            self.num_src_nodes, self.num_dst_nodes = remote_backend.num_nodes(
+                feature_store, graph_store, self.input_type)
+
+        elif issubclass(self.data_cls, Data):
+            self.num_src_nodes = self.num_dst_nodes = data.num_nodes
+        else:  # issubclass(self.data_cls, HeteroData):
+            self.num_src_nodes = data[self.input_type[0]].num_nodes
+            self.num_dst_nodes = data[self.input_type[-1]].num_nodes
```
wsad1, line 45: This can also use the `remote_backend`, right?
Ok if its refactored in another PR.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -34,37 +32,17 @@ def __init__(
...
-
-            elif issubclass(self.data_cls, Data):
-                self.num_src_nodes = self.num_dst_nodes = data.num_nodes
-            else:  # issubclass(self.data_cls, HeteroData):
-                self.num_src_nodes = data[self.input_type[0]].num_nodes
-                self.num_dst_nodes = data[self.input_type[-1]].num_nodes
+        # set it to largest representable torch.long.
+        if self.data_cls == 'custom':
+            feature_store, graph_store = data
+            self.num_src_nodes, self.num_dst_nodes = remote_backend.num_nodes(
```
wsad1, line 38: Nitpick: consider calling it someting like `remote_utils` or `remote_backend_utils`.

torch_geometric/data/remote_backend.py
```diff
@@ -0,0 +1,72 @@
...
+    for edge_attr in edge_attrs:
+        if (_matches_edge_type(query, edge_attr.edge_type)
+                and edge_attr.size is not None):
+            if node_query:
+                return edge_attr.size[0] if query == edge_attr.edge_type[
+                    0] else edge_attr.size[1]
+            else:
+                return edge_attr.size
+
+    # 2. Check FeatureStore:
```
wsad1, line 47: Check `FeatureStore` first and then `GraphStore`. When the last node is an isolated node the `FeatureStore` will have a more accurate size.

torch_geometric/data/remote_backend.py
mananshah99: Thanks for the catch, will update

torch_geometric/loader/link_neighbor_loader.py
mananshah99: In theory, yes - but I think it's best to start consolidating Data/HeteroData logic behind FeatureStore/GraphStore logic in a separate PR.

torch_geometric/loader/link_neighbor_loader.py
mananshah99: Sounds good, will change to `remote_backend_utils`. Agreed that this naming is confusing.

torch_geometric/data/remote_backend.py
mananshah99: I think we want to check GraphStore first, since the sizes of the edge indices are guaranteed to be accurate. We can then back off to features if we have to. If the last node is an isolated node, we just won't find it in the GraphStore, so we will back off to the FeatureStore anyway, right?

wsad1 (approved):

torch_geometric/data/remote_backend_utils.py
```diff
@@ -0,0 +1,70 @@
...
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
+from torch_geometric.typing import EdgeType, NodeType
+
+
+def num_nodes(
+    feature_store: FeatureStore,
+    graph_store: GraphStore,
+    query: Union[NodeType, EdgeType],
+) -> Union[int, Tuple[int, int]]:
```
rusty1s, line 14: I am not liking this union here too much. It looks like separating the interface into
```python
num_nodes(feature_store, graph_store, query: NodeType) -> int
```

```python
size(feature_store, graph_store, query: EdgeType) -> Tuple[int, int]
```
would make more sense?

torch_geometric/data/remote_backend_utils.py
```diff
@@ -0,0 +1,70 @@
...
+        if isinstance(query, (list, tuple)):  # EdgeType
+            return query[0] == node_type or query[-1] == node_type
+        else:
+            return query == node_type
+
+    node_query = isinstance(query, NodeType)
+
+    # 1. Check GraphStore:
+    edge_attrs = graph_store.get_all_edge_attrs()
+    for edge_attr in edge_attrs:
```
rusty1s, line 36: Any opinion on avoiding the for-loop here? Isn't this an implementation detail how the `GraphStore`/`FeatureStore` save the edge attributes? If it is done as part of some hash map, we should be able to leverage this.

torch_geometric/data/remote_backend_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
+from torch_geometric.typing import EdgeType, NodeType
+
+
+def num_nodes(
+    feature_store: FeatureStore,
+    graph_store: GraphStore,
+    query: Union[NodeType, EdgeType],
```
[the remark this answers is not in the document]
mananshah99: Good point, introduced in those data structures.

torch_geometric/data/remote_backend_utils.py
mananshah99: Sgtm

torch_geometric/data/remote_backend_utils.py
```diff
@@ -0,0 +1,70 @@
...
+            return query == edge_type[0] or query == edge_type[-1]
+
+    def _matches_node_type(query: Union[NodeType, EdgeType],
+                           node_type: NodeType) -> bool:
+        if isinstance(query, (list, tuple)):  # EdgeType
+            return query[0] == node_type or query[-1] == node_type
+        else:
+            return query == node_type
+
+    node_query = isinstance(query, NodeType)
```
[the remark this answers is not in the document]
mananshah99: Gotcha, left a note.

torch_geometric/data/remote_backend_utils.py
mananshah99: Yeah, I think graph store and feature store should expose (optional) methods to obtain a tensorattr/edgeattr from the first member of the corresponding dataclass. But wanted to leave that for another PR.

PR #5312, opened 2022-08-30
refactor(sampler): consolidate sampling interface, part 1

mananshah99:
This PR begins the effort to consolidate PyG's sampling interface in preparation for moving `sample(...)` behind the `GraphStore` interface. This effort is somewhat large in scope and will be broken into multiple PRs for ease of review. There are many TODOs left in this PR; all will be resolved in forthcoming PRs (the code will become cleaner in short order).

This first change moves `NeighborSampler` behind a common base class with clearly defined inputs and outputs. Many more changes will be required before this base class defines a stable interface (e.g. `LinkNeighborSampler` has different semantics, `NeighborSampler.perm_dict` is used in `NeighborLoader`, etc.). But for now, it will act as a first step towards this goal. 

It further refactors `loader/utils.py` into those that are required for sampling (moved into `sampler/utils.py`) and those that are required for filtering (remain in `loader/utils.py`). 

**Note about integration with GraphStore.** Moving sampling behind the graph store requires defining a clear interface for sampling across PyG. Since a graph store should not inherently be coupled with a sampling method used (e.g. one can imagine using a NeighborLoader and LinkNeighborLoader on the same graph store), we must define this interface independently from the GraphStore. One can then imagine the GraphStore defining supported samplers for different loaders (along with the defaults, which sample in-memory), and those samplers are then seamlessly used in the loaders.


torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class BaseSampler(ABC):
+    r"""A base class that initializes a graph sampler and provides a `sample`
+    routine that performs sampling on an input list or tensor of node indices.
+    """
+    @abstractmethod
+    def __init__(
+        self,
+        data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]],
```
Padarn, line 47: These are going to get painful to continue to support.. do we have "data to graphstore"? Can we assume that you will always convert to the new representation? Maybe that doesn't make sense...

mananshah99: Yeah, this is a great point. We do plan to consolidate behind `Tuple[FeatureStore, GraphStore]` in the near future, but probably in a separate PR.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.2.0] - 2022-MM-DD
 ### Added
+- Consolidated sampler routines behind `torch_geometric.sampler`, enabling ease of extensibility in the future ([#5312](https://github.com/pyg-team/pytorch_geometric/pull/5312))
+- Enabled `bf16` support in benchmark scripts ([#5293](https://github.com/pyg-team/pytorch_geometric/pull/5293))
```
rusty1s, line 9: Merge conflict.

torch_geometric/sampler/__init__.py
```diff
@@ -0,0 +1,5 @@
+from torch_geometric.sampler.neighbor_sampler import NeighborSampler
+
+__all__ = [
+    'NeighborSampler',
```
rusty1s, line 4: We can consider to also expose `utils`.

torch_geometric/sampler/__init__.py
```diff
@@ -0,0 +1,5 @@
+from torch_geometric.sampler.neighbor_sampler import NeighborSampler
+
+__all__ = [
```
rusty1s, line 3:
```python
from .base import Sampler
```
also expose `Sampler` base class.

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
+from torch_geometric.typing import EdgeType, NodeType
+
+# An input to a sampler a tensor of node indices:
+SamplerInput = torch.Tensor
+
+
+# A sampler output contains the following information.
+#   * input_size: the number of input nodes to begin sampling from.
```
rusty1s, line 16: This is very specific to neighbor sampling. Consider dropping.

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+#       the associated output tensors.
+#   * row: a tensor of edge indices that correspond to the column values of the
+#       edges in the sampled subgraph. Note that these indices must be
+#       re-indexed from 0..n-1 corresponding to the nodes in the 'node' tensor.
+#       In the heterogeneous case, this is a dictionary mapping edge types to
+#       the associated output tensors.
+#   * edge: a tensor of the indices in the original graph; e.g. to be used to
+#       obtain edge attributes.
+class SamplerOutput(NamedTuple):
+    node: Union[torch.Tensor, Dict[NodeType, torch.Tensor]]
```
rusty1s, line 33: We might want to add an optional `batch` attribute which assigns each node to an example. This will be necessary to integrate with the new `pyg-lib` implementation.

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+#       In the heterogeneous case, this is a dictionary mapping edge types to
+#       the associated output tensors.
+#   * row: a tensor of edge indices that correspond to the column values of the
+#       edges in the sampled subgraph. Note that these indices must be
+#       re-indexed from 0..n-1 corresponding to the nodes in the 'node' tensor.
+#       In the heterogeneous case, this is a dictionary mapping edge types to
+#       the associated output tensors.
+#   * edge: a tensor of the indices in the original graph; e.g. to be used to
+#       obtain edge attributes.
+class SamplerOutput(NamedTuple):
```
rusty1s, line 32: I am in favor of adding `SamplerOutput` and `HeteroSamplerOutput` and support `sample(index) -> Union[SamplerOutput, HeteroSamplerOutput]`.

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+    r"""A base class that initializes a graph sampler and provides a `sample`
+    routine that performs sampling on an input list or tensor of node indices.
+    """
+    @abstractmethod
+    def __init__(
+        self,
+        data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]],
+        **kwargs,
+    ):
+        r"""Initializes a sampler with common data it will need to perform a
```
rusty1s, line 50: Move the doc-string up to the base doc-string. The note should be marked via `.. warning ::`.

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+        so it is best to limit the amount of information stored here."""
+        pass
+
+    @abstractmethod
+    def sample(self, index: SamplerInput, **kwargs) -> SamplerOutput:
+        pass
+
+    def __call__(self, index: SamplerInput, **kwargs) -> SamplerOutput:
+        if not isinstance(index, torch.LongTensor):
+            index = torch.LongTensor(index)
```
rusty1s, line 61:
```suggestion
        if isinstance(index, (list, tuple)):
            index = torch.tensor(index)
```

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+    edge: Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]
+    input_size: int
+
+
+class BaseSampler(ABC):
+    r"""A base class that initializes a graph sampler and provides a `sample`
+    routine that performs sampling on an input list or tensor of node indices.
+    """
+    @abstractmethod
+    def __init__(
```
rusty1s, line 45: Why do we need to provide an `abstractmethod` for `__init__` here? Not sure what `super().__init__(data)` means in the `NeighborSampler`.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -0,0 +1,225 @@
...
+        data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]],
+        num_neighbors: NumNeighbors,
+        replace: bool = False,
+        directed: bool = True,
+        input_type: Optional[Any] = None,
+        time_attr: Optional[str] = None,
+        is_sorted: bool = False,
+        share_memory: bool = False,
+    ):
+        super().__init__(data)
```
rusty1s, line 36:
```suggestion
```

torch_geometric/sampler/utils.py
```diff
@@ -0,0 +1,106 @@
...
+        return edge_type
+    return EDGE_TYPE_SPLIT_STR.join(edge_type)
+
+
+def edge_type_to_str_dict(
+        edge_type_dict: Dict[EdgeType, Any]) -> Dict[str, Any]:
+    return {edge_type_to_str(k): v for k, v in edge_type_dict.items()}
+
+
+def edge_type_from_str(edge_type: Union[EdgeType, str]) -> EdgeType:
```
rusty1s, line 29: Do we need this function? This is generally dangerous since it means we do not support any types with `__` in its name.

rusty1s (approved):
LGTM.

torch_geometric/sampler/base.py
mananshah99: Yeah, agreed. Can replace with a vague `'metadata'` for now, that we can potentially refine later.

torch_geometric/sampler/utils.py
mananshah99: The reason this is included is that it allows samplers to speak purely with edge types, instead of with the string edge types (currently, loaders filter based on `edge_type_to_str`). I agree that it is dangerous for that reason; the alternative would be to define a custom `to_str` and `from_str` here that are PyG-specific, but that can lead to weird debugging outputs, or we can speak through string edge types (but this leads to using `edge_type_to_str` in lots of places, since GraphStore currently speaks through edge types). Wdyt?

torch_geometric/sampler/base.py
rusty1s: sounds good.

torch_geometric/sampler/utils.py
rusty1s: I think a workaround is to maintain this mapping explicitely, see https://github.com/pyg-team/pyg-lib/blob/master/pyg_lib/sampler/__init__.py#L86-L87

torch_geometric/sampler/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+import torch
+
+from torch_geometric.data import Data, HeteroData
+from torch_geometric.data.feature_store import FeatureStore
+from torch_geometric.data.graph_store import GraphStore
+from torch_geometric.typing import EdgeType, NodeType
+
+# An input to a sampler a tensor of node indices:
+SamplerInput = torch.Tensor
```
wsad1, line 12: nitpick: Do we really need this, why not just use torch.Tensor directly.

torch_geometric/sampler/utils.py
mananshah99: Sure, that works too.

wsad1 (approved):

torch_geometric/sampler/base.py
mananshah99: Soon, this will be split into `NodeSamplerInput` and `EdgeSamplerInput`, so I prefer to leave it as-is in the current PR.

torch_geometric/sampler/base.py
mananshah99: Added a TODO.

torch_geometric/sampler/__init__.py
mananshah99: With the simplification of utils as a result of removing `edge_type_to_str` and `edge_type_from_str`, etc, I am not sure this is necessary at the current moment; perhaps in the future.

torch_geometric/sampler/base.py
mananshah99: Done; although this may become more complicated once we try to simplify `'metadata'`. Can revert later if necessary.

PR #5316, opened 2022-08-30
`WLConv` operation for continuously attributed graphs

cgoliver:
Implements the conv operator from [Wasserstein Weisfeiler Lehman Graph Kernels](https://arxiv.org/abs/1906.01277) paper.
Instead of hash, a degree-weighted mean is taken during the refinement step.


torch_geometric/nn/conv/wl_conv_continuous.py
```diff
@@ -0,0 +1,66 @@
...
+                edge_weight: OptTensor = None, size: Size = None) -> Tensor:
+        """"""
+        if isinstance(x, Tensor):
+            x: OptPairTensor = (x, x)
+
+        # propagate_type: (x: OptPairTensor, edge_weight: OptTensor)
+        out = self.propagate(edge_index, x=x, edge_weight=edge_weight,
+                             size=size)
+        x_r = x[1]
+        deg = degree(edge_index[1]).view(-1, 1)
```
Padarn, line 56: I think this could be written as a custom `Aggregation`?

cgoliver: Thanks for the feedback @Padarn. I'm not sure the `Aggregation` class works here since the convolution operation is defined recursively on the output from previous layers. Do you mean to incorporate a custom aggregation within the convolution class?

Padarn: My thought was to add a new `Aggregation` that could be used inside of message passing when aggregating messages, not as a single `Aggregation`. Does that make sense?

rusty1s: @Padarn you mean a custom aggregation that takes care of weighting based on node degree? Isn‘t that what `mean` is already doing? Why do we actually need to perform normalization twice here?

cgoliver: You are right, @rusty1s , the aggregation in this convolution operation could be done with the mean aggregator, I didn't know about it when I wrote this. Then it just becomes a regular GraphConv with a mean aggregation and an optional scaling on the neighbors.

Padarn: Right, `mean` does it too :-)

rusty1s (approved):
Thank you!

torch_geometric/nn/conv/wl_conv_continuous.py
rusty1s: I think we could do provide a `WeightedAggregation` with custom weights per element, and clean this further up. Happy to take this in a separate PR.

PR #5317, opened 2022-08-30
`GraphSAGE` example with `LinkNeighborLoader`

scigeek72:
Replaced NeighborSampler by LinkNeighborLoader and made the necessary changes


examples/graph_sage_unsup.py
```diff
@@ -78,17 +63,16 @@ def train():
...
+        h = model(loader)
+        src_nodes = loader.edge_label_index[0]
+        dst_nodes = loader.edge_label_index[1]
+        out = h[src_nodes] * h[dst_nodes]
+        loss = F.binary_cross_entropy_with_logits(out.sum(dim=-1),
+                                                  loader.edge_label)
 
-        pos_loss = F.logsigmoid((out * pos_out).sum(-1)).mean()
-        neg_loss = F.logsigmoid(-(out * neg_out).sum(-1)).mean()
-        loss = -pos_loss - neg_loss
```
wsad1, line 91: is the new loss equivalent to the previous loss.

examples/graph_sage_unsup.py
```diff
@@ -49,10 +33,11 @@ def __init__(self, in_channels, hidden_channels, num_layers):
...
 
-    def forward(self, x, adjs):
-        for i, (edge_index, _, size) in enumerate(adjs):
-            x_target = x[:size[1]]  # Target nodes are always placed first.
-            x = self.convs[i]((x, x_target), edge_index)
+    def forward(self, data):
+
+        x, edge_index = data.x, data.edge_index
+        for i in range(self.num_layers):
+            x = self.convs[i]((x, x), edge_index)
```
wsad1, line 40:
```suggestion
            x = self.convs[i](x, edge_index)
```
I guess this should also work.

wsad1 (commented):
Thanks for updating the example. Left a few comments. Will take a look again later.

examples/graph_sage_unsup.py
```diff
@@ -4,40 +4,24 @@
...
 from sklearn.linear_model import LogisticRegression
-from torch_cluster import random_walk
 
 import torch_geometric.transforms as T
 from torch_geometric.datasets import Planetoid
-from torch_geometric.loader import NeighborSampler as RawNeighborSampler
+from torch_geometric.loader import LinkNeighborLoader
 from torch_geometric.nn import SAGEConv
 
+#from torch_cluster import random_walk
```
rusty1s, line 13: Remove.

examples/graph_sage_unsup.py
```diff
@@ -4,40 +4,24 @@
...
-
-        batch = torch.cat([batch, pos_batch, neg_batch], dim=0)
-        return super().sample(batch)
-
-
-train_loader = NeighborSampler(data.edge_index, sizes=[10, 10], batch_size=256,
-                               shuffle=True, num_nodes=data.num_nodes)
+train_loader = LinkNeighborLoader(data, batch_size=256, shuffle=True,
+                                  neg_sampling_ratio=1.0,
+                                  num_neighbors=[10, 10])
```
rusty1s, line 24: Can we also increase `num_workers` to `6` (or similar)?

examples/graph_sage_unsup.py
```diff
@@ -49,10 +33,11 @@ def __init__(self, in_channels, hidden_channels, num_layers):
             in_channels = in_channels if i == 0 else hidden_channels
             self.convs.append(SAGEConv(in_channels, hidden_channels))
 
-    def forward(self, x, adjs):
-        for i, (edge_index, _, size) in enumerate(adjs):
-            x_target = x[:size[1]]  # Target nodes are always placed first.
-            x = self.convs[i]((x, x_target), edge_index)
+    def forward(self, data):
+
+        x, edge_index = data.x, data.edge_index
```
rusty1s, line 38:
```suggestion
    def forward(self, x, edge_index):
```

examples/graph_sage_unsup.py
rusty1s: Yes :)

rusty1s (approved):
I like it. Thanks!

examples/graph_sage_unsup.py
```diff
@@ -108,11 +92,11 @@ def test():
     val_acc = clf.score(out[data.val_mask], data.y[data.val_mask])
     test_acc = clf.score(out[data.test_mask], data.y[data.test_mask])
 
-    return val_acc, test_acc
+    return val_acc, test_acc, out
```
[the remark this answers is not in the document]
scigeek72: `val_acc` and `test_acc` are reduced when I ran it. Also it was little bit slower than the original examples (but that could be because of my machine).

examples/graph_sage_unsup.py
scigeek72: I tried to use `num_workers = 6` but was getting an error ( on a MacBook Pro, no cuda support). So left it unchanged.

examples/graph_sage_unsup.py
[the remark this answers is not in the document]
wsad1: Thanks.

examples/graph_sage_unsup.py
```diff
@@ -108,11 +90,11 @@ def test():
...
```
wsad1, line 93:
```suggestion
    return val_acc, test_acc
```

wsad1 (approved):

PR #5330, opened 2022-08-31
`HeteroLinear` with `segment_matmul`

puririshi98:


torch_geometric/nn/dense/linear.py
```diff
@@ -161,42 +163,46 @@ def __repr__(self) -> str:
...
     to types
-
     .. math::
         \mathbf{x}^{\prime}_{\kappa} = \mathbf{x}_{\kappa}
         \mathbf{W}^{\top}_{\kappa} + \mathbf{b}_{\kappa}
-
     for type :math:`\kappa`.
     It supports lazy initialization and customizable weight and bias
     initialization.
-
```
rusty1s, line 172: Revert line break diff.

torch_geometric/nn/dense/linear.py
```diff
@@ -161,42 +163,46 @@ def __repr__(self) -> str:
...
+        self._WITH_PYG_LIB = torch.cuda.is_available() and _WITH_PYG_LIB
+        self.is_sorted = is_sorted
         self.lins = torch.nn.ModuleList([
             Linear(in_channels, out_channels, **kwargs)
             for _ in range(num_types)
         ])
 
         self.reset_parameters()
+        if self._WITH_PYG_LIB:
```
rusty1s, line 199: Can't we instead initialize a `Parameter` with shape `[num_types, in_channels, out_channels]`?

torch_geometric/nn/dense/linear.py
```diff
@@ -208,10 +214,21 @@ def forward(self, x: Tensor, type_vec: Tensor) -> Tensor:
...
-            out[mask] = lin(x[mask])
+        if self._WITH_PYG_LIB:
+            if not self.is_sorted:
+                if (type_vec[1:] < type_vec[:-1]).any():
+                    type_vec, perm = type_vec.sort()
+                    x = x[:, perm]
+            edge_type_ptr = torch.ops.torch_sparse.ind2ptr(
+                type_vec, self.num_types)
+            out = segment_matmul(x, edge_type_ptr, self.weights)
+            for i in range(self.num_types):
```
rusty1s, line 225: Any chance we can avoid the for-loop here? Something like `out += self.biases[type_vec]` should work as well.

rusty1s (approved):

torch_geometric/nn/dense/linear.py
puririshi98: well i figured this way we preserve the kwargs and initiliazers of linear without having to re-code that specifically for the pyg_lib case

rusty1s (approved):

PR #5332, opened 2022-09-01
Fix `PointTransformerConv` aggregation from `mean` to `sum`

saedrna:
The message in PointTransformerConv is already softmax, which sum to one. The aggregation should be `add` rather `mean`.

### DIFF
--- a/torch_geometric/nn/conv/point_transformer_conv.py
+++ b/torch_geometric/nn/conv/point_transformer_conv.py
@@ -74,7 +74,7 @@ class PointTransformerConv(MessagePassing):
                  out_channels: int, pos_nn: Optional[Callable] = None,
                  attn_nn: Optional[Callable] = None,
                  add_self_loops: bool = True, **kwargs):
-        kwargs.setdefault('aggr', 'mean')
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/conv/point_transformer_conv.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, in_channels: Union[int, Tuple[int, int]],
                  out_channels: int, pos_nn: Optional[Callable] = None,
                  attn_nn: Optional[Callable] = None,
                  add_self_loops: bool = True, **kwargs):
-        kwargs.setdefault('aggr', 'mean')
+        kwargs.setdefault('aggr', 'add')
```
Padarn, line 77: Hmm this makes sense, but I guess its a breaking change for anyone already using it? What about making it an option?

rusty1s: @Padarn I marked this as a breaking change in the changelog. I think we just definitely fix this since this change is for the better.

rusty1s (approved):
Thanks for fixing!

PR #5342, opened 2022-09-02
PyTorch nightly `fx` tracing fix

JakubPietrakIntel:
Fix for inference_benchmark.py failing with an error:
```
return self._module_getattr(attr, attr_val,parameter_proxy_cache)
AttributeError: 'Tracer' object has no attribute '_module_getattr'
```
Replace not implemented Tracer._module_getattr with Tracer.getattr to return the value of nn.Module attr.


Padarn:
Indeed this looks like a mistake that was also recently fixed in `fx` (where we got code from) https://github.com/pytorch/pytorch/pull/84011

rusty1s (approved):

PR #5353, opened 2022-09-05
Replace in-place operations by out-of-place ones

dszwicht:
Goal of this PR is to align with `utils.scatter` (`torch.scatter_reduce` wrapper) operation, which do not allow user to perform in-place operation on its output, in case of training. Current walk-around is made by cloning tensor on return.

During training `autograd` complains about modification of tensor needed for backward step, such behavior can be also observed with `torch_scatter` implementation, when using `mul` algorithm (caused by [this](https://github.com/dszwicht/pytorch_scatter/blob/master/csrc/scatter.cpp#L95)).

Extensive training tests were made (based on example scripts) and here are some thought:
1) `in-place -> out-of-place` without experimental mode:
in most cases memory allocations take no more memory than in default case and maximum rise is up to 4%. In half of tested examples execution time shrinks (which is good).
2) experimental mode with `clone()` or `in-place -> out-of-place`:
between `clone()` and `in-place -> out-of-place` difference is negligible. In most cases both memory allocations and execution time is worse by up to 10% (comparing to the non-experimental mode).
3) `SAGEConv`:
All models based on `SAGEConv` appear to have much worse performance when using experimental mode. `torch_scatter` implementation of `mean` algorithm (based on `torch.scatter_add`) is approximately 2 times faster than `torch.scatter_reduce[mean]` - this issue was mentioned already here: #5120.

All experiments were made on CPU.

### DIFF
--- a/test/utils/test_scatter.py
+++ b/test/utils/test_scatter.py
@@ -27,7 +27,7 @@ def test_scatter_backward(reduce):
     index = torch.randint(0, 10, (100, ), dtype=torch.long)
 
     with torch_geometric.experimental_mode('scatter_reduce'):
-        out = scatter(src, index, dim=1, reduce=reduce).relu_()
+        out = torch.relu(scatter(src, index, dim=1, reduce=reduce))
     assert src.grad is None
     out.mean().backward()
     assert src.grad is not None
--- a/test/utils/test_scatter.py
+++ b/test/utils/test_scatter.py
@@ -27,7 +27,7 @@ def test_scatter_backward(reduce):
     index = torch.randint(0, 10, (100, ), dtype=torch.long)
 
     with torch_geometric.experimental_mode('scatter_reduce'):
-        out = torch.relu(scatter(src, index, dim=1, reduce=reduce))
+        out = scatter(src, index, dim=1, reduce=reduce).relu()
     assert src.grad is None
     out.mean().backward()
     assert src.grad is not None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `BaseStorage.get()` functionality ([#5240](https://github.com/pyg-team/pytorch_geometric/pull/5240))
 - Added a test to confirm that `to_hetero` works with `SparseTensor` ([#5222](https://github.com/pyg-team/pytorch_geometric/pull/5222))
 ### Changed
+- Replace in-place operations with out-of-place ones to align with `torch.scatter_reduce` API ([#5353](https://github.com/pyg-team/pytorch_geometric/pull/5353))
 - Breaking bugfix: `PointTransformerConv` now correctly uses `sum` aggregation ([#5332](https://github.com/pyg-team/pytorch_geometric/pull/5332))
 - Improve out-of-bounds error message in `MessagePassing` ([#5339](https://github.com/pyg-team/pytorch_geometric/pull/5339))
 - Allow file names of a `Dataset` to be specified as either property and method ([#5338](https://github.com/pyg-team/pytorch_geometric/pull/5338))
--- a/torch_geometric/nn/conv/appnp.py
+++ b/torch_geometric/nn/conv/appnp.py
@@ -118,7 +118,7 @@ class APPNP(MessagePassing):
             x = self.propagate(edge_index, x=x, edge_weight=edge_weight,
                                size=None)
             x = x * (1 - self.alpha)
-            x += self.alpha * h
+            x = x + self.alpha * h
 
         return x
 
--- a/torch_geometric/nn/conv/arma_conv.py
+++ b/torch_geometric/nn/conv/arma_conv.py
@@ -128,10 +128,11 @@ class ARMAConv(MessagePassing):
                                  size=None)
 
             root = F.dropout(x, p=self.dropout, training=self.training)
-            out += root @ self.root_weight[0 if self.shared_weights else t]
+            root = root @ self.root_weight[0 if self.shared_weights else t]
+            out = out + root
 
             if self.bias is not None:
-                out += self.bias[0 if self.shared_weights else t]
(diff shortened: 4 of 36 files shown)


test/utils/test_scatter.py
```diff
@@ -27,7 +27,7 @@ def test_scatter_backward(reduce):
     index = torch.randint(0, 10, (100, ), dtype=torch.long)
 
     with torch_geometric.experimental_mode('scatter_reduce'):
-        out = scatter(src, index, dim=1, reduce=reduce).relu_()
+        out = torch.relu(scatter(src, index, dim=1, reduce=reduce))
```
rusty1s, line 30:
```suggestion
        out = scatter(src, index, dim=1, reduce=reduce).relu()
```

rusty1s (approved):
Amazing, thank you!

rusty1s:
Can we add this change to the `CHANGELOG.md`? Seems important to add.

dszwicht:
> Can we add this change to the `CHANGELOG.md`? Seems important to add.

Sure thing, just added.

rusty1s:
Super. Please feel free to merge if you think this change is for the better.

> experimental mode with clone() or in-place -> out-of-place:
> between clone() and in-place -> out-of-place difference is negligible. In most cases both memory allocations and execution time is worse by up to 10% (comparing to the non-experimental mode).

Does this mean that `clone()+in-place` is faster or `out-of-place` is faster?

> SAGEConv:
> All models based on SAGEConv appear to have much worse performance when using experimental mode. torch_scatter implementation of mean algorithm (based on torch.scatter_add) is approximately 2 times faster than torch.scatter_reduce[mean] - this issue was mentioned already here:
https://github.com/pyg-team/pytorch_geometric/pull/5120.

Yeah, I know. Ultimately, I feel this needs to be fixed on PyTorch side, but I am happy to fix this temporarily on our end as well. WDYT?

dszwicht:
> Super. Please feel free to merge if you think this change is for the better.
> 
> > experimental mode with clone() or in-place -> out-of-place:
> > between clone() and in-place -> out-of-place difference is negligible. In most cases both memory allocations and execution time is worse by up to 10% (comparing to the non-experimental mode).
> 
> Does this mean that `clone()+in-place` is faster or `out-of-place` is faster?

I would say in general none is faster. Usually, the difference is below 1% (sometimes in favor of one sometimes in favor of another), but there are a few cases where the difference is more visible, like [dna.py](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/dna.py) where out-of-place performs more than 10% better or [graph_unet.py](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/graph_unet.py) where in-place + clone() gives similarly better performance.
 
> > SAGEConv:
> > All models based on SAGEConv appear to have much worse performance when using experimental mode. torch_scatter implementation of mean algorithm (based on torch.scatter_add) is approximately 2 times faster than torch.scatter_reduce[mean] - this issue was mentioned already here:
> > #5120.
> 
> Yeah, I know. Ultimately, I feel this needs to be fixed on PyTorch side, but I am happy to fix this temporarily on our end as well. WDYT?

PyTorch changes seem to take quite a lot of time... Calling `scatter_add` explicitly instead of `scatter_reduce` also brings some slight speedups (that's why I used both `scatter_add` and `scatter_mean` explicitly in my initial design). I think most users don't dig into `utils.scatter` module, hence they're not overwhelmed by a slightly more complicated implementation. On the other hand, they may appreciate better performance while using models and convolutions provided by PyG. So summarizing I think it may sense, to provide the most performant solution. This also brings another question to my mind, do you think it makes sense to use torch version of scatter if only `torch>=1.12` is available (without experimental mode)?

rusty1s:
> PyTorch changes seem to take quite a lot of time... Calling scatter_add explicitly instead of scatter_reduce also brings some slight speedups (that's why I used both scatter_add and scatter_mean explicitly in my initial design). I think most users don't dig into utils.scatter module, hence they're not overwhelmed by a slightly more complicated implementation. On the other hand, they may appreciate better performance while using models and convolutions provided by PyG. So summarizing I think it may sense, to provide the most performant solution. This also brings another question to my mind, do you think it makes sense to use torch version of scatter if only torch>=1.12 is available (without experimental mode)?

Yes, we can implement `scatter_add` and `scatter_reduce` as done as in `torch-scatter`. My main point of concern is that whenever PyTorch updates and accelerates its `scatter_reduce` implementation, we would have to swap implementation as well.

> This also brings another question to my mind, do you think it makes sense to use torch version of scatter if only torch>=1.12 is available (without experimental mode)?

Yes, happy to make this change.

dszwicht:
> > PyTorch changes seem to take quite a lot of time... Calling scatter_add explicitly instead of scatter_reduce also brings some slight speedups (that's why I used both scatter_add and scatter_mean explicitly in my initial design). I think most users don't dig into utils.scatter module, hence they're not overwhelmed by a slightly more complicated implementation. On the other hand, they may appreciate better performance while using models and convolutions provided by PyG. So summarizing I think it may sense, to provide the most performant solution. This also brings another question to my mind, do you think it makes sense to use torch version of scatter if only torch>=1.12 is available (without experimental mode)?
> 
> Yes, we can implement `scatter_add` and `scatter_reduce` as done as in `torch-scatter`. My main point of concern is that whenever PyTorch updates and accelerates its `scatter_reduce` implementation, we would have to swap implementation as well.
> 
> > This also brings another question to my mind, do you think it makes sense to use torch version of scatter if only torch>=1.12 is available (without experimental mode)?
> 
> Yes, happy to make this change.

Ok, let's then merge this PR and continue work on things mentioned here in the following one. BTW, do I have write access to this repository? And what are recommendations for merging, squash, or some other approach?

rusty1s:
You should have write access and be able to squash merge this PR, right? No need to remove any commits since they are squashed together anyway.

dszwicht:
> You should have write access and be able to squash merge this PR, right? No need to remove any commits since they are squashed together anyway.

Yep, it seems I can merge PRs.

PR #5359, opened 2022-09-06
`Explainer`: bug fix.

wsad1:
Fixes #5356.
In pooling models the `edge_attr` comes before `batch` see [here](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.pool.TopKPooling.forward). The `Explainer.get_model_prediction` assumed `batch` comes right after `x` and `edge_index`, this PR fixes that.


rusty1s (approved):
Thanks!

PR #5363, opened 2022-09-06
Revert PR #5233 for missing test labels in `HBGDataset`

EdisonLeeeee:


torch_geometric/datasets/hgb_dataset.py
```diff
@@ -179,15 +177,6 @@ def process(self):
                 else:
                     data[n_type].y[n_id] = int(y[3])
                 data[n_type].train_mask[n_id] = True
-            for y in test_ys:
```
rusty1s, line 182: The old implementation still keeps the `test_mask`. What do you suggest?

EdisonLeeeee: I would suggest removing the attribute `test_mask` to avoid confusion, even though it is all `False`. Once the HGB team releases the test labels, I'll make a new PR and add it back. WDYT?

EdisonLeeeee: I saw a note in the comments:
```
    .. note::
        Test labels are randomly given to prevent data leakage issues.
        If you want to obtain final test performance, you will need to submit
        your model predictions to the
        `HGB leaderboard <https://www.biendata.xyz/hgb/>`_.
```
Shall we also revise it?

rusty1s: Won't we need the `test_mask` to upload the predictions to the HGB website?
Alternatively, since we already have the note in the doc-string, I feel like the current implementation works as well.

EdisonLeeeee: I agree with you. Will close this PR.

rusty1s: Yeah, sorry about wasted efforts here :(

EdisonLeeeee: No worries :)

PR #5365, opened 2022-09-07
refactor(sampler): consolidate link neighbor sampling interface, part 2

mananshah99:
This PR begins the effort to consolidate PyG's sampling interface in preparation for moving `sample(...)` behind the `GraphStore` interface. This effort is somewhat large in scope and will be broken into multiple PRs for ease of review. It resolves some TODOs from #5312, and introduces others that will be resolved in follow-up PRs. 

The major change removes `LinkNeighborSampler`, consolidating it under a the common `BaseSampler` interface with `NeighborSampler`. In doing so, it modifies the `BaseSampler` interface to support `sample_from_nodes` and `sample_from_edges`, since oftentimes edge-based sampling applies some transformations and then re-uses the same logic as node-based sampling. A sampler need not define both node-based and edge-based sampling, but if it doesn't, an appropriate error will be raised.

Resolved TODOs:
* `LinkNeighborSampler` and `NeighborSampler` interfaces are aligned
* No more special handling of `edge_type_to_str` in `filter_*`
* No more special handling of `perm_dict`, which removes the need for `edge_type_to_str` entirely


torch_geometric/sampler/base.py
```diff
@@ -3,13 +3,20 @@
...
-SamplerInput = torch.Tensor
+# An input to a node-based sampler is a tensor of node indices:
+NodeSamplerInput = torch.Tensor
 
+# An input to an edge-based sampler consists of four tensors:
+#   * The row of the edge index in COO format
+#   * The column of the edge index in COO format
+#   * The labels of the edges
+#   * (Optionally) the time attribute corresponding to the edge label
+EdgeSamplerInput = Union[torch.Tensor, torch.Tensor, torch.Tensor, OptTensor]
```
rusty1s, line 16:
```suggestion
EdgeSamplerInput = Tuple[torch.Tensor, torch.Tensor, torch.Tensor, OptTensor]
```

rusty1s (approved):
This looks perfect :)

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -214,13 +239,19 @@ def _hetero_sparse_neighbor_sample(
             )
         return node_dict, row_dict, col_dict, edge_dict
 
-    def sample(
+    # Node-based sampling #####################################################
+
+    def sample_from_nodes(
```
wsad1, line 244: Add more to doc string like ```samples nodes starting from nodes in index```.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -249,9 +280,71 @@ def sample(
...
                             f'{type(self.data_cls)}')
 
+    # Edge-based sampling #####################################################
 
-###############################################################################
+    def sample_from_edges(
+        self,
+        index: EdgeSamplerInput,
+        **kwargs,
+    ) -> Union[SamplerOutput, HeteroSamplerOutput]:
```
wsad1, line 289:
```suggestion
    ) -> Union[SamplerOutput, HeteroSamplerOutput]:
    r"""Implements neighbor sampling starting from edges in index, 
    by calling 'torch-sparse' sampling routines, conditional on the type of data object."""
```

torch_geometric/sampler/utils.py
```diff
@@ -72,3 +74,84 @@ def to_hetero_csc(
...
+def add_negative_samples(
+    edge_label_index,
+    edge_label,
+    edge_label_time,
+    num_src_nodes: int,
+    num_dst_nodes: int,
+    negative_sampling_ratio: float,
+):
+    """Add negative samples and their `edge_label` and `edge_time`
+    if `self.neg_sampling_ratio > 0`"""
```
wsad1, line 91:
```suggestion
    if `neg_sampling_ratio > 0`"""
```

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -373,7 +246,10 @@ def filter_fn(self, out: Any) -> Union[Data, HeteroData]:
         return data if self.transform is None else self.transform(data)
 
     def collate_fn(self, index: Union[List[int], Tensor]) -> Any:
```
wsad1, line 248:
```suggestion
    def collate_fn(self, index:Tuple[Tensor, Tensor, Tensor, OptTensor[) -> Any:
```

wsad1 (approved):
Thanks. This looks good.

torch_geometric/sampler/neighbor_sampler.py
mananshah99: Added this (and the above comment) in the base class.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -359,8 +232,8 @@ def filter_fn(self, out: Any) -> Union[Data, HeteroData]:
             if edge_label_time is not None:
                 data[edge_type].edge_label_time = edge_label_time
         else:
-            (node_dict, row_dict, col_dict, edge_dict, edge_label_index,
-             edge_label, edge_label_time) = out
+            (node_dict, row_dict, col_dict, edge_dict, metadata) = out
+            edge_label_index, edge_label, edge_label_time = metadata
```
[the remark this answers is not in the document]
mananshah99: We always return `edge_label_time` in the heterogeneous case, even if it is `None`.

Padarn:
Didn't get around to reviewing before merged.. but all looks great to me :-)

PR #5375, opened 2022-09-07
Add `IndexToMask` and `MaskToIndex` transforms

hatemhelal:
Both of these transforms will try to infer the right attributes to convert.

* `MaskToIndex` matches any attributes that end with `_mask`
* `IndexToMask` matches any attributes that end with `_indices`. 

For the latter case, it would be a problem to automatically convert `edge_index` so I went with this suffix.


torch_geometric/transforms/mask.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@functional_transform('index_to_mask')
+class IndexToMask(BaseTransform):
+    r"""Converts indices to a mask representation
+
+    Args:
+        attrs (str, [str], optional): If given, will only perform index to mask
+            conversion for the given attributes. If omitted, will infer the
+            attributes from the suffix `_indices`  (default: :obj:`None`)
```
rusty1s, line 27:
```suggestion
            attributes from the suffix :obj:`*_idx`  (default: :obj:`None`)
```
Can we name this `_idx` rather than `_indices`?

torch_geometric/transforms/mask.py
```diff
@@ -0,0 +1,88 @@
...
+        sizes (int, [int], optional). The size of the mask. If set to
+            :obj:`None`, a minimal sized mask is created.
```
rusty1s, line 29: Interesting. Wouldn't it make more sense to return a mask with the shape of number of nodes?

torch_geometric/transforms/mask.py
```diff
@@ -0,0 +1,88 @@
...
+            (default: :obj:`None`)
+    """
+    def __init__(self, attrs: Optional[Union[str, List[str]]] = None,
+                 sizes: Optional[Union[int, List[int]]] = None):
+        self.attrs = (attrs, ) if isinstance(attrs, str) else attrs
+        self.sizes = sizes
+
+    def __call__(self, data: Data) -> Data:
```
rusty1s, line 37: Can we let this support both `data` and `HeteroData`? We can simply iterate over all `node_stores` and `edge_stores` to support this.

torch_geometric/transforms/mask.py
```diff
@@ -0,0 +1,88 @@
...
+            attributes from the suffix `_mask` (default: :obj:`None`)
+    """
+    def __init__(self, attrs: Optional[Union[str, List[str]]] = None):
+        self.attrs = [attrs] if isinstance(attrs, str) else attrs
+
+    def __call__(self, data: Data) -> Data:
+        suffix = "_mask"
+        attrs = self.attrs if self.attrs else attrs_ends_with(data, suffix)
+
+        for mask_attr in attrs:
```
rusty1s, line 79: Option to replace the attribute?

torch_geometric/transforms/mask.py
```diff
@@ -0,0 +1,88 @@
...
+            mask = getattr(data, mask_attr)
+            index = mask_to_index(mask)
+            new_attr = update_suffix(mask_attr, suffix, "_indices")
+            setattr(data, new_attr, index)
```
rusty1s, line 83:
```suggestion
            index = mask_to_index(data[mask_attr])
            data[update_suffix(mask_attr, suffix, "_indices")] = index
```

test/transforms/test_index_to_mask.py
```diff
@@ -0,0 +1,32 @@
+import pytest
```
rusty1s, line 1: merge into single file `test_mask.py`.

rusty1s (approved):
This mostly looks good, thank you! Left some nitpicky comments. One thing I would like to add is `HeteroData` support. If this is too much to handle, we can add it in a follow-up PR.

torch_geometric/transforms/mask.py
hatemhelal: Yes I wondered about that, how does an option `replace` that accepts a `bool` to the transform sound?

torch_geometric/transforms/mask.py
hatemhelal: thanks I wasn't sure how to do this with `HeteroData` - agree it would be better to support this.

torch_geometric/transforms/mask.py
hatemhelal: Yes that is a good suggestion

torch_geometric/transforms/mask.py
hatemhelal: sure! I didn't really like `indices`

test/transforms/test_index_to_mask.py
hatemhelal: so I tried this initially but had a build failure mentioning that there was a conflict with another `test_mask.py` in `test/utils` 

Let me dig up the log from that run in case it makes more sense to you.

torch_geometric/transforms/mask.py
rusty1s: +1 :)

test/transforms/test_index_to_mask.py
rusty1s: We can name it like `test_mask_transform.py`.

torch_geometric/transforms/mask.py
hatemhelal: I think this is working now for `HeteroData` but still wanted to add a test.  Let me know if you spot anything obviously missing in the implementation while I work on that.

rusty1s:
Hey @hatemhelal. Can you give me right access to your fork/your PR?

rusty1s (approved):
Thank you!

PR #5376, opened 2022-09-07
build: do not run tests in minimal install

mananshah99:


rusty1s:
Any reason for this? It ensures that a non-full installation is still functional.

mananshah99:
Ah, I was just confused because minimal install can also fail if a test somewhere fails (as opposed to install itself being the problem). But totally happy keeping it as-is as well.

rusty1s:
Yeah, maybe a good idea to rename the action.

PR #5377, opened 2022-09-07
build: improve support for `skip-changelog` in changelog enforcer

mananshah99:


rusty1s:
Why do we need this?

mananshah99:
This allows folks to add the skip-changelog tag after a PR is created, upon which the action will automatically re-run.

rusty1s (changes requested):
Got it. Thanks for clarifying!

rusty1s (approved):

PR #5381, opened 2022-09-08
Add `PositionEncoding` module

lightaime:
Add 1D position encoding. The `sin` mode encoding of `out_channels=512`  is like:
image


torch_geometric/nn/__init__.py
```diff
@@ -1,18 +1,20 @@
+from .data_parallel import DataParallel
```
rusty1s, line 1: `isort` is disabled on `__init__.py`, so we do not necessarily need this. Doesn't hurt though.

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
+from typing import Union
+
+import torch
+import torch.nn.functional as F
+
+
+class PositionEncoding1D(torch.nn.Module):
+    r"""The 1D positioning encoding from `"Attention Is All You Need"
+    <https://arxiv.org/pdf/1706.03762.pdf>`_ paper.
```
rusty1s, line 9: Can we add a doc-string for the arguments?

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
+from typing import Union
+
+import torch
+import torch.nn.functional as F
+
+
+class PositionEncoding1D(torch.nn.Module):
```
rusty1s, line 7:
```suggestion
class PositionalEncoding1D(torch.nn.Module):
```

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
...
+
+        self.mode = mode
+        self.num_pos = num_pos
+        self.out_channles = out_channels
+
+        if self.mode == "sin":
+            if out_channels % 2 != 0:
+                raise ValueError(
+                    "Cannot use sinusoidal positional encoding with "
+                    f"odd dim. Got `out_channles`: {out_channels}).")
```
rusty1s, line 23:
```suggestion
                    f"odd 'out_channels' (got {out_channels}).")
```

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
...
+    r"""The 1D positioning encoding from `"Attention Is All You Need"
+    <https://arxiv.org/pdf/1706.03762.pdf>`_ paper.
+    """
+    def __init__(self, num_pos: int, out_channels: int,
+                 max_wave_len: Union[float, int] = 10000., mode: str = "sin"):
+        super().__init__()
+
+        self.mode = mode
+        self.num_pos = num_pos
+        self.out_channles = out_channels
```
rusty1s, line 17:
```suggestion
        self.out_channels = out_channels
```

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if self.mode == "sin":
+            if out_channels % 2 != 0:
+                raise ValueError(
+                    "Cannot use sinusoidal positional encoding with "
+                    f"odd dim. Got `out_channles`: {out_channels}).")
+
+            position_encoding = torch.zeros(num_pos, out_channels)
+
+            position = torch.arange(0, num_pos, dtype=torch.float).unsqueeze(1)
```
rusty1s, line 27: I wonder whether we can support floating-point positions, which would let us support positions of different granularity. For this, we would likely need to drop "learn" support (not sure how tragic this would be), and perform sinus/cosinus computation on the fly. WDYT?

torch_geometric/nn/position_encoding.py
```diff
@@ -0,0 +1,50 @@
...
+            x = position * frequency
+            position_encoding[:, 0::2] = torch.sin(x)
+            position_encoding[:, 1::2] = torch.cos(x)
+
+            self.register_buffer('position_encoding', position_encoding)
+
+        if self.mode == "learn":
+            self.position_encoding = torch.nn.Embedding(num_pos, out_channels)
+
+    def forward(self, pos_ids):
```
rusty1s, line 40:
```suggestion
    def forward(self, pos_ids: Tensor) -> Tensor:
```

rusty1s:
@lightaime Can you push recent changes? Looks like comments are resolved but I don't see the update :)

torch_geometric/nn/position_encoding.py
rusty1s: I am no expert TBH but I felt that the range of positions can still be given via `num_pos` (or better via `max_pos`), right? I don't think we can normalize by max on a per-batch basis.

lightaime: Add a simple `granularity` support. The users can change the granularity if the floating-point positions are scaled or if they want to capture more stable changes in floating-point positions.

torch_geometric/nn/positional_encoding.py
```diff
@@ -0,0 +1,77 @@
...
+                out_channels,
+            )
+
+    def compute_sinus_encoding(self, position: Tensor) -> Tensor:
+        position_encoding = torch.zeros(
+            (position.size(0), self.out_channels),
+            device=position.device,
+        )
+        x = (position / self.granularity).unsqueeze(1) * self.frequency
+        position_encoding[:, 0::2] = torch.sin(x)
```
lightaime, line 64: I believe it is fine to just straight concatenate the `sin` and `cos` instead of this interleaved way. But it is conventional in implementations. WDYT?

ar104: It shouldn't matter because the net can just permute its weights when looking across that vector.

rusty1s: Yes, let‘s just concatenate for extra soeed up.

lightaime: Thanks, @ar104 @rusty1s. Agreed.

Issue #4176, opened 2022-03-01
GNNs with Missing Node Features 🚀


rusty1s:
### 🚀 The feature, motivation and pitch

**Paper:** [On the Unreasonable Effectiveness of Feature propagation in Learning on Graphs with Missing Node Features](https://arxiv.org/abs/2111.12128)
**Code:** [zip](https://openreview.net/attachment?id=tx4qfdJSFvG&name=supplementary_material)

The paper presents simple, fast and scalable approach for handling missing features in graph machine learning applications.
It might be very interesting to add the approach of feature propagation and an example of it to PyG.

 @wsad1 Let me know if you have interest in looking into this.

wsad1:
Sure @rusty1s, I'll check this out and come up with a short plan.

wsad1:
Some thoughts

- Add `FeaturePropagation` to `torch_geometric.nn.models`.
- 
![image](https://user-images.githubusercontent.com/13963626/157891878-47d3f437-3508-49c0-b43f-4d514cca487e.png)
Instead of resetting know features every iteration, we could remove incoming edges to nodes with features from nodes without features, before starting the propagation. This would be more efficient than resetting.

rusty1s:
Since this model has no parameters, I'm not sure we want to put it into `model`. How about providing it as a `transform`? What that work as well? Also note that the code just got released: https://github.com/twitter-research/feature-propagation

rusty1s:
Thanks for chiming in @emalgorithm.

I think resetting features is easier to implement. It should also be more efficient if we utilize `SparseTensor` matrix-multiplication rather than scatter/gather (but this is just a guess). The edge removal proposal also doesn't work if features of a node are only partially missing, right?

emalgorithm:
Yes actually, that's true, the edge removal would work only when a node either has all features or none, but not in the general case of some features missing.  Standard diffusion and then resetting is definitely the cleanest solution then.

wsad1:
Yes it does fit into `transform` better.

For the general case where node features are partially missing the user would input a `mask` of shape num_nodes*num_features, which we would use while resetting the features, right? But at the same time we might need to let the user pass a `mask` of shape num_nodes when a node has all features or none.

EdisonLeeeee:
Is anyone working on this? If not I can take it.

rusty1s:
This would be awesome :)

Praniyendeev:
[...]

PR #5387, opened 2022-09-08
Add `FeaturePropagation` as a `transform`

EdisonLeeeee:
This PR implements the `FeaturePropagation` method proposed in [On the Unreasonable Effectiveness of Feature propagation in Learning on Graphs with Missing Node Features](https://arxiv.org/abs/2111.12128).

Based on the prev discussion in [this issue](https://github.com/pyg-team/pytorch_geometric/issues/4176), I implement `FeaturePropagation` as a `transform`. ~~A minimal example to reproduce the experiments is also provided.~~


torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,90 @@
...
+        missing_mask (Tensor, optional): The mask indicating where
+            node features are missing. (default: :obj:`None`)
+        add_self_loops (bool, optional): If set to :obj:`False`, will not add
+            self-loops to the input graph. (default: :obj:`True`)
+        normalize (bool, optional): Whether to add self-loops and compute
+            symmetric normalization coefficients on the fly.
+            (default: :obj:`True`)
+    """
+    def __init__(self, num_iterations: int = 40,
+                 missing_mask: OptTensor = None, add_self_loops: bool = True,
```
wsad1, line 45: Lets make `missing_mask` a required argument.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,90 @@
...
+        assert missing_mask is None or missing_mask.dtype == torch.bool
+        self.num_iterations = num_iterations
+        self.missing_mask = missing_mask
+        self.normalize = normalize
+        self.add_self_loops = add_self_loops
+
+    def __call__(self, data: Data) -> Data:
+
+        assert 'edge_index' in data or 'adj_t' in data
+
```
wsad1, line 56: Check if the shapes of `missing_mask` and `x` are compatible.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,90 @@
...
+                edge_weight = data.edge_weight
+            edge_index = data.edge_index
+            adj_t = SparseTensor(row=edge_index[1], col=edge_index[0],
+                                 value=edge_weight,
+                                 sparse_sizes=data.size()[::-1],
+                                 is_sorted=False, trust_data=True)
+        else:
+            adj_t = data.adj_t
+
+        if self.normalize:
```
wsad1, line 69: I don't think normalization should be optional. According to the paper the recursion converges when the adjacency matrix is normalized.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,90 @@
...
+            (default: :obj:`True`)
+    """
+    def __init__(self, num_iterations: int = 40,
+                 missing_mask: OptTensor = None, add_self_loops: bool = True,
+                 normalize: bool = True):
+        assert missing_mask is None or missing_mask.dtype == torch.bool
+        self.num_iterations = num_iterations
+        self.missing_mask = missing_mask
+        self.normalize = normalize
+        self.add_self_loops = add_self_loops
```
wsad1, line 51: Lets drop `add_self_loops`.

wsad1 (commented):
Really appreciate you adding this. Please add a test for this feature. Left some initial comments, will take a look again in a few days.

torch_geometric/transforms/feature_prop.py
EdisonLeeeee: Thanks, will add it soon.

torch_geometric/transforms/feature_prop.py
EdisonLeeeee: Thanks for your suggestion. Will make normalization a default behavior.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,87 @@
...
+
+        x = data.x
+        known_feature_mask = None
+        missing_mask = self.missing_mask
+        if missing_mask is not None:
+            out = torch.zeros_like(x)
+            known_feature_mask = ~missing_mask
+            out[known_feature_mask] = x[known_feature_mask]
+        else:
+            out = x.clone()
```
Padarn, line 73: Sorry naive question: Why do we need a clone here?

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,87 @@
...
+
+        for _ in range(self.num_iterations):
+            # Diffuse current features
+            out = matmul(adj_t, out)
```
Padarn, line 77: could we avoid defining an out above by using
```python
if self.missing_mask is not None:
   x[self.missing_mask] = 0

for _ in range(self.num_iterations):
    out = matmul(adj_t, x)
    if known_feature_mask is not None:
        # Reset original known features
        out[known_feature_mask] = x[known_feature_mask]
    x = out
```

maybe its not better

EdisonLeeeee: Looks good to me. Updated in the latest commit, thanks!

torch_geometric/transforms/feature_prop.py
emalgorithm: I think the current behaviour with a `missing_mask` is counterintuitive. According to the algorithm, if no features are missing, then everything is left unchanged. On the other hand, if no features are missing (which is the case if `missing_mask` is None), then all features are diffused and change. 

I think it's easier to make `missing_mask` a required argument, so that there is only one behaviour of this algorithm.

torch_geometric/transforms/feature_prop.py
emalgorithm: Normalization to a symmetrically-normalized adjacency should probably happen inside the method, so that it is possible to call this method with a raw edge index, what do you think?

emalgorithm:
Hi, the author of the paper here 👋 . Thanks a lot for working on this and let me know if you have any question!

torch_geometric/transforms/feature_prop.py
wsad1: Yeah i agree. looks like its in the current implementation in line 63.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,83 @@
...
+
+        For an example of using :class:`~torch_geometric.transforms.
+        FeaturePropagation`, see `examples/feature_propagation.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        feature_propagation.py>`_.
+
+    Args:
+        num_iterations (int): The number of propagations. (default: :obj:`40`)
+        missing_mask (Tensor, optional): The mask indicating where
+            node features are missing. (default: :obj:`None`)
```
wsad1, line 37:
```suggestion
        missing_mask (Tensor, optional): The mask indicating where
            node features are missing, of shape `[num_nodes, channels]`. (default: :obj:`None`)
```
Add shape information.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+    where missing node features are inferred by known features via propagation.
+
+    .. note::
+
+        For an example of using :class:`~torch_geometric.transforms.
+        FeaturePropagation`, see `examples/feature_propagation.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        feature_propagation.py>`_.
```
wsad1, line 32: Not a strong preference. But i don't think we need an example for this, a code snippet with the documentation will do, like done [here](https://pytorch-geometric.readthedocs.io/en/latest/modules/transforms.html#torch_geometric.transforms.RandomLinkSplit).

wsad1 (approved):

torch_geometric/transforms/feature_prop.py
wsad1: Yeah I agree, lets drop the option to set it to `None`.

EdisonLeeeee: Sure,`missing_mask` will be a required argument in the next commit.

torch_geometric/transforms/feature_prop.py
EdisonLeeeee: Indeed. Thanks for your suggestion!

torch_geometric/transforms/feature_prop.py
EdisonLeeeee: Hi @Padarn,  If I understood it correctly, this line of code `x = out` would lead to a different implementation from the original paper. According to the [original implementation](https://github.com/twitter-research/feature-propagation/blob/af2733589eab4023fca67f7e71a3b46ddbbea8cd/src/feature_propagation.py#L25-L31), `x` should always be the original node feature `data.x`, instead of the intermediate result of `out`.

I checked it and kept `x` unchanged in the latest commit.

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,79 @@
...
+
+        from torch_geometric.transforms import FeaturePropagation
+
+        missing_mask = torch.isnan(data.x)
+        transform = FeaturePropagation(missing_mask)
+        data = transform(data)
+
+    Args:
+        missing_mask (Tensor): The mask indicating where
+            node features are missing, of shape `[num_nodes, channels]`.
```
wsad1, line 37:
```suggestion
        missing_mask (BoolTensor): Mask matrix
                :math:`\mathbf{M} \in {\{ 0, 1 \}}^{N\times F}`, which
                indicates where node features are missing.
```

torch_geometric/transforms/feature_prop.py
```diff
@@ -0,0 +1,79 @@
...
+                edge_weight = data.edge_weight
+            edge_index = data.edge_index
+            adj_t = SparseTensor(row=edge_index[1], col=edge_index[0],
+                                 value=edge_weight,
+                                 sparse_sizes=data.size()[::-1],
+                                 is_sorted=False, trust_data=True)
+        else:
+            adj_t = data.adj_t
+
+        adj_t = gcn_norm(adj_t)
```
wsad1, line 62:
```suggestion
        adj_t = gcn_norm(adj_t, add_self_loops=False)
```

wsad1 (approved):
Thanks a lot for adding this. left some final comments.

test/transforms/test_feature_propagation.py
```diff
@@ -0,0 +1,28 @@
+import torch
+
+from torch_geometric.data import Data
+from torch_geometric.transforms import FeaturePropagation, ToSparseTensor
+
+
+def test_feature_propagation():
+    assert str(FeaturePropagation()) == ('FeaturePropagation('
```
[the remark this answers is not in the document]
EdisonLeeeee: My bad. I'll check it again.

EdisonLeeeee:
@wsad1 Thanks for your thorough review. Would it be able to merge now?

rusty1s (approved):
Thank you! Made some minor modifications.

PR #5393, opened 2022-09-09
Link Pred Example: Re-use train embeddings, add negative sampling to evaluation

tomonori-masui:
This PR updates the following two points in link_pred.py:

1. Use node embedding `z` from train() in test() function instead of recreating it on test/valid set
2. Add negative sampling into test() function

These make the code more consistent with the original [GAE repo](https://github.com/tkipf/gae).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed tests relying on `dblp` datasets to instead use  synthetic data ([#5250](https://github.com/pyg-team/pytorch_geometric/pull/5250))
 - Fixed a bug for the initialization of activation function examples in `custom_graphgym` ([#5243](https://github.com/pyg-team/pytorch_geometric/pull/5243))
 - Allow any integer tensors when checking edge_index input to message passing ([5281](https://github.com/pyg-team/pytorch_geometric/pull/5281))
+- Fixed `link_pred.py` to use node embedding from training in `test()` and add negative sampling into `test()`([#5393](https://github.com/pyg-team/pytorch_geometric/pull/5393))
 ### Removed
 
 ## [2.1.0] - 2022-08-17
--- a/examples/link_pred.py
+++ b/examples/link_pred.py
@@ -52,18 +52,7 @@ def train():
     z = model.encode(train_data.x, train_data.edge_index)
 
     # We perform a new round of negative sampling for every training epoch:
-    neg_edge_index = negative_sampling(
-        edge_index=train_data.edge_index, num_nodes=train_data.num_nodes,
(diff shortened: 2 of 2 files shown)


rusty1s:
Thanks for this PR. I don't think we want to re-create negative edges in every test iteration. Instead, we are operating on a single fixed negative sample such that evaluation across different steps are comparable. In the same fashion, we don't want to necessarily re-use `z` from the training phase since this matrix was computed during training, e.g., it potentially contains dropped out values from `dropout`.

tomonori-masui:
I updated the code per the feedback:

- Negative samples are added to valid & test outside the epochs loop
- Node embedding `z` is recreated using `train_data` in `test()` instead of reusing it from `train()` (the last prediction `model.decode_all()` is also the same: using `z` recreated from `train_data`)

tomonori-masui:
@rusty1s I'm following up on the updates by providing some reasonings behind this change:

- The current master branch code use test/valid data to create embeddings when evaluating the model, but this does not make sense as the test/valid edges are what the model is going to predict and should not be used for the model input. And also ignoring all the train edges does not make sense as they are the actual existing edges and want to be used when creating the embeddings.
- The current code does not have negative links when evaluating the model, but this is not optimal as you can easily achieve 100% accuracy if you predict everything positive (meaning the evaluation can be biased toward predicting positives). To have a reliable evaluation, adding negative links to test/valid data is necessary.

As mentioned before, the changes are more consistent with [the original link prediction/graph auto-encoder repo](https://github.com/tkipf/gae).

EdisonLeeeee:
Hi @tomonori-masui thanks for this PR. I'm afraid there might be some misunderstandings.

> The current master branch code uses test/valid data to create embeddings when evaluating the model, but this does not make sense as the test/valid edges are what the model is going to predict and should not be used for the model input. And also ignoring all the train edges does not make sense as they are the actual existing edges and want to be used when creating the embeddings.

The train/valid data in the current implementation share the sample `x` and `edge_index`. For test data, the `edge_index` also includes the edges used for validation, i.e., (`train_data.edge_index + val_edges`). We can use one of them to predict the test edges without any leaking. The major difference among them is `edge_label_index` and `edge_label` which indicates the (train/val/test) links to predict and their corresponding labels (0 or 1).

> The current code does not have negative links when evaluating the model, but this is not optimal as you can easily achieve 100% accuracy if you predict everything positive 

Actually, the negative links for evaluation were added by `T.RandomLinkSplit`. `edge_label_index` includes both positive and negative edges.

WDYT?

tomonori-masui:
@EdisonLeeeee Thanks for the reply. You are totally correct. I was actually misunderstanding the functionality of `T.RandomLinkSplit` and also confusing `edge_index` and `edge_label_index`. The code totally makes sense to me now.

Issue #5390, opened 2022-09-08
A faster implementation of `AddMetaPaths`


EdisonLeeeee:
### 🚀 The feature, motivation and pitch

I've recently come across the method `AddMetaPaths` and I found it quite slow particularly on large-scale heterogeneous graphs. Take an example in [this discussion](https://github.com/pyg-team/pytorch_geometric/discussions/5319):
```python
from torch_geometric.datasets import AMiner
from torch_geometric.transforms import AddMetaPaths
data = AMiner(root='./data/aminer_pyg/')[0]
metapaths = [[("author", "paper"), ("paper", "author")],
[("author", "paper"), ("paper", "venue"), ("venue", "paper"), ("paper", "author")]]
data = AddMetaPaths(metapaths)(data)
```
it takes me hours and much memory to run on the `AMiner` graph, even though I specified `max_sample=1` in `AddMetaPaths`.

I think the implementation can be accelerated via multiple one-step random walks to generate the metapaths. For example, for the metapath `[("author", "paper"), ("paper", "author")]`, we can perform one-step random walk on `("author", "paper")` graph, and then on `("paper", "author")` graph, to generate the metapath efficiently. 

I've tested this method and it run much faster than the current implementation with lower memory overheads. Would it be possible to be added to PyG?

rusty1s:
Definitely. Parts of the slow implementation can also be contributed to the slow CPU-based `spspmm`.
Adding @kpstesla to the conversation as well. Happy to take any contributions here!

kpstesla:
This is a great idea!  I'll take a look at the implementation.  I'm also curious to see how this performance compares with the much faster `spspmm` implementation for the gpu, although that will surely have memory limitations.

EdisonLeeeee:
I've tested previous `AddMetaPaths` (implemented with `spspmm`) on the AMiner graph with ~300k nodes. It cannot be finished in hours on both GPU and CPU. The code snippet is:
```python
from torch_geometric.datasets import AMiner
from torch_geometric.transforms import AddMetaPaths

data = AMiner(root="AMiner")[0]
metapaths = [
    [("author", "paper"), ("paper", "author")],
    [("author", "paper"), ("paper", "venue"), ("venue", "paper"), ("paper", "author")],
]
data = data.to('cuda')
data = AddMetaPaths(metapaths, max_sample=1)(data)
```

PR #5397, opened 2022-09-09
Add random walk based `AddRandomMetaPaths` as a faster alternative

EdisonLeeeee:
# Description
This PR implements `AddRandomMetaPaths` as a faster alternative to generative metapaths. As mentioned in [this issue](https://github.com/pyg-team/pytorch_geometric/issues/5390), metapaths are sampled via multiple one-step random walks (walk length per node is 1). The implementation of a one-step random walk was inspired by that in [Metapath2Vec](https://github.com/pyg-team/pytorch_geometric/blob/0da9e1903c3b8ffd41b9a51b9e481dd1ba616a7f/torch_geometric/nn/models/metapath2vec.py#L234)

# Feature
One can specify `walks_per_node` for the number of random walks per starting node in different metapaths. `walks_per_node` must be an integer or a list of integers. For the later, the length of `walks_per_node` must match the length of `metapaths`.
```python
metapaths = [
    [("author", "paper"), ("paper", "author")],
    [("author", "paper"), ("paper", "venue"), ("venue", "paper"), ("paper", "author")],
]
>>> AddRandomMetaPaths(metapaths, walks_per_node=1)

>>> AddRandomMetaPaths(metapaths, walks_per_node=[10, 20])
```
# Benchmark test
## Time
```python
from torch_geometric.datasets import AMiner
from torch_geometric.transforms import AddMetaPaths, AddRandomMetaPaths

data = AMiner(root="~/data/pygdata/AMiner")[0]
metapaths = [
    [("author", "paper"), ("paper", "author")],
    [("author", "paper"), ("paper", "venue"), ("venue", "paper"), ("paper", "author")],
]
```

+ AddRandomMetaPaths(on CPU and GPU)
```python
data = AddRandomMetaPaths(metapaths)(data)
# cannot be finished in hours
```

+ AddRandomMetaPaths(on CPU)
```python
%%time
data = AddRandomMetaPaths(metapaths)(data)
# 417 ms
```
## Performance
Take the example [here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/hetero/han_imdb.py) for comparison.

+ AddMetapaths

```python
metapaths = [[('movie', 'actor'), ('actor', 'movie')],
             [('movie', 'director'), ('director', 'movie')]]
transform = T.AddMetaPaths(metapaths=metapaths, drop_orig_edges=True,
                           drop_unconnected_nodes=True)
# ACC (at most) 0.59
```

+ AddRandomMetaPaths
```python
metapaths = [[('movie', 'actor'), ('actor', 'movie')],
             [('movie', 'director'), ('director', 'movie')]]
transform = T.AddRandomMetaPaths(metapaths=metapaths, drop_orig_edges=True,
                           drop_unconnected_nodes=True, walks_per_node=[20, 15])
# ACC (at most) 0.58
```
It seems `AddRandomMetaPaths` requires to speficy a larger `walks_per_node` to obtain a better performance.

PS. Sorry for such a straightforward name, but currently I don't have a better idea to name this class.


kpstesla (commented):
@EdisonLeeeee This is a great improvement to see!  I ran a test, and even on a gpu, this implementation is faster than `AddMetaPaths` with `max_sample=1`, so great job!

About the interface, my thought would be to extend `AddMetaPaths` to potentially use this method, by adding a `method` kwarg.  We could have `method=spspmm` and `method=random_walk` to determine which method to use.  You could also change the interface from `walks_per_node` and `sample_ratio` back to `max_sample`, which would ensure the same performance between both methods (as they produce metapaths from the same distribution).

How does that sound?

EdisonLeeeee:
@EdisonLeeeee Sounds great! I think this is the best way to be compatible with the existing implementation while adding new features to `AddMetaPaths`. 

However, I'm still a bit confused about the interface. As you can see, the `spspmm` based implementation can also support weighted sampling (set `weighted=True`) while the random walk based cannot. Besides, I have no idea so far to change the interface from `walks_per_node` and `sample_ratio` back to `max_sample` as they might share different meanings.

Let me think about how to optimize the interface without breaking the compatibility. Really appreciate your suggestion and help!

kpstesla:
@EdisonLeeeee that sounds good!  I had a couple thoughts that may be helpful:

- One way to implement `max_sample` would be a 2 step process.  We could sample the adjacency matrix to restrict the out degree of each node to be `max_sample`.  Then, we could do something similar to what you've done here and compute the columns by array indexing, which would be much faster than spspmm in most cases.
- As for weighted sampling, it may be possible to compute the weights of each path by indexing the value row of the sparse tensor similarly to how you index the column row of the sparse tensor.  This might give you duplicate edges though.  To fix this, you could use `torch_geometric.utils.coalesce`, which reduce adds duplicate edges.

EdisonLeeeee:
Thank you @kpstesla, these are great thoughts. I think I can start working on it.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -192,3 +193,183 @@ def _get_edge_weight(self, data: HeteroData,
         else:
             edge_weight = None
         return edge_weight
+
+
+@functional_transform('add_metapaths2')
+class AddMetaPaths2(BaseTransform):
```
wsad1, line 199: Lets call this function `AddRandomMetaPaths`. I feel like its functionality is different enough, to not merge it with `AddMetaPaths` and this name seems to better indicate how it works.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -192,3 +193,183 @@ def _get_edge_weight(self, data: HeteroData,
...
+            edge types between the same node type are not dropped even in case
+            :obj:`drop_orig_edges` is set to :obj:`True`.
+            (default: :obj:`False`)
+        drop_unconnected_nodes (bool, optional): If set to :obj:`True` drop
+            node types not connected by any edge type. (default: :obj:`False`)
+        walks_per_node (int, List[int], optional): The number of walks to
+            sample for each starting node in a metapth. (default: :obj:`1`)
+        sample_ratio (float, optional): The ratio of the number of sampled
+            nodes from the whole graph, starting for random walks
+            at the beginning of a metapath. (default: :obj:`1.0`)
```
wsad1, line 283:
```suggestion
        sample_ratio (float, optional): The ratio of source nodes to start random
            walks from. (default: :obj:`1.0`)
```

wsad1 (commented):
Thanks for adding this. I think we should keep this as a separate class and call it `AddRandomMetapaths`. Will take a look again soon.

torch_geometric/transforms/add_metapaths.py
EdisonLeeeee: Got it. Will update the code soon.

EdisonLeeeee:
Still need some time to adapt it when `weighted=True`.

torch_geometric/transforms/add_metapaths.py
```diff
@@ -192,3 +193,187 @@ def _get_edge_weight(self, data: HeteroData,
...
+    @staticmethod
+    def sample(src: SparseTensor, subset: Tensor,
+               num_neighbors: int = 1) -> Tensor:
+
+        rowcount = src.storage.rowcount()[subset]
+        mask = rowcount > 0
+        offset = torch.zeros_like(subset)
+        offset[mask] = src.storage.rowptr()[subset[mask]]
+
+        rand = torch.rand((rowcount.size(0), num_neighbors),
```
wsad1, line 368: A bit confused by this. `torch.rand` returns numbers between `0-1`, is this your intention?

torch_geometric/transforms/add_metapaths.py
```diff
@@ -192,3 +193,187 @@ def _get_edge_weight(self, data: HeteroData,
...
+            for i, edge_type in enumerate(metapath):
+                edge_index = data[edge_type].edge_index
+                adj = SparseTensor(row=edge_index[0], col=edge_index[1],
+                                   sparse_sizes=data[edge_type].size())
+                col, mask = self.sample(adj, start)
+                row, col = row[mask], col[mask]
+                start = col
+
+            new_edge_type = (metapath[0][0], f'metapath_{j}', metapath[-1][-1])
+            data[new_edge_type].edge_index = coalesce(torch.vstack([row, col]))
```
wsad1, line 335: Set `edge_weight` as `1`, this will count the number of `metapaths` between two nodes and set that as the `edge_weight` if `weighted` is `True`. I am guessing you were going to do this anyway.

torch_geometric/transforms/add_metapaths.py
EdisonLeeeee: Correct. This is to sample from the node neighborhoods. I followed the way how the random walks are sampled in [MetaPath2Vec](https://github.com/pyg-team/pytorch_geometric/blob/80c6c782edb58a17cbabd54ae69d19cc1ba7785a/torch_geometric/nn/models/metapath2vec.py#L234-L251)

torch_geometric/transforms/add_metapaths.py
EdisonLeeeee: Exactly. Yet, the problem is, `edge_weight` is not really the number of metapaths between two nodes and its value depends on `walks_per_node`. This would potentially bring some inconsistency between `AddMetapaths` and `AddRandomMetaPaths`.

wsad1: In `AddMetapaths` the `edge_weight` is either the sum of weights of all metapaths between two nodes or the number of metapaths between two nodes. So setting in to number of random metapaths isn't inconsistent right?

EdisonLeeeee: Yep. Maybe we should drop the argument `weighted` in `AddRandomMetaPaths`? WDYT?

wsad1: I am ok with not supporting it in this PR, we could add a todo comment to add `weighted` and address it in another PR.

EdisonLeeeee: Thanks for your feedback @wsad1. I think I have nothing else to do for this PR. Could you please check it again at your convenience @rusty1s ?

wsad1 (approved):
Thanks for addressing all the comments till now. This looks good to me mostly, left some last comments 😄 .

EdisonLeeeee:
Thanks again for your comments. Will address these ASAP.

EdisonLeeeee:
Jobs done @wsad1

rusty1s:
Thanks to all for getting this merged:)

PR #5398, opened 2022-09-09
Fix `UninitializedParameter` script bug

tangwxxxxxxx2126:
there is not have attriute named "UninitializedParameter" in torch.nn.parameter(my torch version is 1.7).
when i modify it to "torch.nn.parameter -> torch.nn.parameter()",it works.


rusty1s:
@tangwxxxxxxx2126 Can you clarify the fix? It looks like after the merge with master, there are no more changes within this PR.

rusty1s:
Closing this for now since there seems to be an issue with this PR. Please feel free to re-open.

Issue #5352, opened 2022-09-05
Is there an error in the ASAPooling implementation?


chaolinhuang:
### 🐛 Describe the bug

When ASAPooling calculates the attention score, there is a linear layer missing in front of softmax. I think the formula in the paper has this layer.

### Environment

* PyG version:
* PyTorch version:
* OS:
* Python version:
* CUDA/cuDNN version:
* How you installed PyTorch and PyG (`conda`, `pip`, source):
* Any other relevant information (*e.g.*, version of `torch-scatter`):

rusty1s:
Thanks for reporting. Isn't this already covered by `self.att`, see [here](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/pool/asap.py#L102)?

PR #5399, opened 2022-09-09
Enable `torch.scatter_reduce` implicitly when torch version requirements are met

dszwicht:
Enable `torch.scatter_reduce` implicitly when torch version requirements are met.

The experimental mode was removed, for `torch>=1.12`, `utils.scatter` will automatically choose `torch` implementation.
For `add` reduction, `scatter_add` is used explicitly as it performs slightly better than `scatter_reduce`.
For `mean` reduction, `torch_scatter` implementation is now used (usually about 2 times faster than PyTorch implementation).

Some benchmarks for described approach were already provided here: #5120
A discussion which led to this solution can be found here: #5353


dszwicht:
@rusty1s, waiting for some feedback 😉

test/test_experimental.py
```diff
@@ -1,28 +0,0 @@
-import pytest
-
-from torch_geometric import (
-    experimental_mode,
-    is_experimental_mode_enabled,
-    set_experimental_mode,
-)
-
-
-@pytest.mark.parametrize('options', [None, 'scatter_reduce'])
```
rusty1s, line 10: I don't think we necessarily have to remove `experimental_mode` - there might be future features which might need this. For now, we can simply remove it from the documentation and remove the `scatter_reduce` option. WDYT?

torch_geometric/utils/scatter.py
```diff
@@ -1,66 +1,91 @@
...
-import torch_scatter
 from torch import Tensor
 
-import torch_geometric
-
 major, minor, _ = torch.__version__.split('.', maxsplit=2)
 major, minor = int(major), int(minor)
 has_pytorch112 = major > 1 or (major == 1 and minor >= 12)
 
+if has_pytorch112:  # pragma: no cover
```
rusty1s, line 10: Can we also add a test for https://github.com/pyg-team/pytorch_geometric/issues/5352 to see if it works as expected now?

torch_geometric/utils/scatter.py
```diff
@@ -1,66 +1,91 @@
...
+            index_dim = dim
+            if index_dim < 0:
+                index_dim = index_dim + src.dim()
+            if index.dim() <= index_dim:
+                index_dim = index.dim() - 1
+
+            ones = torch.ones(index.size(), dtype=src.dtype, device=src.device)
+            count = ScatterHelpers.generate_out(ones, index, index_dim,
+                                                dim_size).scatter_add_(
+                                                    index_dim, index, ones)
```
rusty1s, line 50:
```suggestion
            count = ScatterHelpers.generate_out(ones, index, index_dim, dim_size)
             count.scatter_add_(index_dim, index, ones)
```

torch_geometric/utils/scatter.py
```diff
@@ -1,66 +1,91 @@
...
+            else:
+                size[dim] = 0
+            return src.new_zeros(size)
+
+        @staticmethod
+        def scatter_mean(src: Tensor, index: Tensor, dim: int, out: Tensor,
+                         dim_size: Optional[int]) -> Tensor:
+            out.scatter_add_(dim, index, src)
+
+            index_dim = dim
```
rusty1s, line 41: Can we add some comment here what's going on with `index_dim`?

rusty1s (approved):

torch_geometric/utils/scatter.py
dszwicht: Done.

torch_geometric/utils/scatter.py
dszwicht: Done.

test/test_experimental.py
dszwicht: Yeah, good idea 😉. I was also considering this option. Code changed as suggested.

dszwicht:
@rusty1s, all comments resolved. Please check whether the final solution looks good to you 😉

test/nn/pool/test_asap.py
```diff
@@ -1,7 +1,11 @@
+import io
+
+import pytest
 import torch
 
 from torch_geometric.nn import ASAPooling, GCNConv, GraphConv
 from torch_geometric.testing import is_full_test
+from torch_geometric.testing.decorators import onlyFullTest
```
rusty1s, line 8:
```suggestion
from torch_geometric.testing import onlyFullTest
```

rusty1s (approved):
Amazing, thank you!

PR #5401, opened 2022-09-09
Add support for precomputed edges in `SchNet` model

hatemhelal:
The main feature this allows is to call a SchNet model instance with pre-calculated edges.  This can be more efficient if the radius graphs have been computed in a `pre_transform` of a PyG dataset since this calculation will only run once and the result is cached on disk.

### DIFF
--- a/examples/qm9_pretrained_schnet.py
+++ b/examples/qm9_pretrained_schnet.py
@@ -55,7 +55,10 @@ for target in range(12):
     for data in tqdm(loader):
         data = data.to(device)
         with torch.no_grad():
-            pred = predict(model, data)
+            if not args.use_precomputed_edges:
+                 pred = model(data.z, data.pos, data.batch)
+            else:
+                 pred = model(data.z, data.pos, data.batch, data.edge_index)
         mae = (pred.view(-1) - data.y[:, target]).abs()
         maes.append(mae)
 
--- a/examples/qm9_pretrained_schnet.py
+++ b/examples/qm9_pretrained_schnet.py
@@ -8,7 +8,7 @@ from torch_geometric.datasets import QM9
 from torch_geometric.loader import DataLoader
 from torch_geometric.logging import log
 from torch_geometric.nn import SchNet
-from torch_geometric.transforms import Compose, Distance, RadiusGraph
+from torch_geometric.transforms import RadiusGraph
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--cutoff', type=float, default=10.0,
@@ -26,10 +26,7 @@ if not args.use_precomputed_edges:
     dataset = QM9(path)
 else:
     path = osp.join(pwd, '..', 'data', f'QM9_{args.cutoff}')
-    pre_transform = Compose(
-        [RadiusGraph(args.cutoff),
-         Distance(norm=False, cat=False)])
-    dataset = QM9(path, pre_transform=pre_transform)
+    dataset = QM9(path, pre_transform=RadiusGraph(args.cutoff))
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
 
@@ -39,7 +36,7 @@ def default_predict(model, data):
 
 
 def precomputed_predict(model, data):
-    return model(data.z, data.batch, data.edge_index, data.edge_attr)
+    return model(data.z, data.pos, data.batch, data.edge_index)
 
 
 if args.use_precomputed_edges:
--- a/examples/qm9_pretrained_schnet.py
+++ b/examples/qm9_pretrained_schnet.py
@@ -6,7 +6,6 @@ from tqdm import tqdm
 
 from torch_geometric.datasets import QM9
 from torch_geometric.loader import DataLoader
-from torch_geometric.logging import log
 from torch_geometric.nn import SchNet
 from torch_geometric.transforms import RadiusGraph
 
@@ -19,8 +18,6 @@ parser.add_argument(
 args = parser.parse_args()
 pwd = osp.dirname(osp.realpath(__file__))
 
-log(**vars(args))
-
 if not args.use_precomputed_edges:
     path = osp.join(pwd, '..', 'data', 'QM9')
     dataset = QM9(path)
@@ -29,20 +26,6 @@ else:
     dataset = QM9(path, pre_transform=RadiusGraph(args.cutoff))
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-
-
-def default_predict(model, data):
-    return model(data.z, data.pos, data.batch)
-
-
-def precomputed_predict(model, data):
-    return model(data.z, data.pos, data.batch, data.edge_index)
-
-
-if args.use_precomputed_edges:
-    predict = precomputed_predict
-else:
-    predict = default_predict
 
 for target in range(12):
     model, datasets = SchNet.from_qm9_pretrained(path, dataset, target)
@@ -56,9 +39,9 @@ for target in range(12):
         data = data.to(device)
         with torch.no_grad():
             if not args.use_precomputed_edges:
-                 pred = model(data.z, data.pos, data.batch)
+                pred = model(data.z, data.pos, data.batch)
             else:
-                 pred = model(data.z, data.pos, data.batch, data.edge_index)
+                pred = model(data.z, data.pos, data.batch, data.edge_index)
         mae = (pred.view(-1) - data.y[:, target]).abs()
         maes.append(mae)
 
--- a/examples/qm9_pretrained_schnet.py
+++ b/examples/qm9_pretrained_schnet.py
@@ -1,15 +1,51 @@
+import argparse
 import os.path as osp
 
 import torch
+from tqdm import tqdm
 
 from torch_geometric.datasets import QM9
 from torch_geometric.loader import DataLoader
+from torch_geometric.logging import log
 from torch_geometric.nn import SchNet
+from torch_geometric.transforms import Compose, Distance, RadiusGraph
 
-path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'QM9')
-dataset = QM9(path)
+parser = argparse.ArgumentParser()
+parser.add_argument('--cutoff', type=float, default=10.0,
+                    help='Cutoff distance for interatomic interactions')
+parser.add_argument(
+    '--use_precomputed_edges', action='store_true',
+    help='Compute graph edges and weights as a dataset pre_transform')
+args = parser.parse_args()
+pwd = osp.dirname(osp.realpath(__file__))
+
+log(**vars(args))
+
+if not args.use_precomputed_edges:
+    path = osp.join(pwd, '..', 'data', 'QM9')
+    dataset = QM9(path)
+else:
+    path = osp.join(pwd, '..', 'data', f'QM9_{args.cutoff}')
+    pre_transform = Compose(
+        [RadiusGraph(args.cutoff),
+         Distance(norm=False, cat=False)])
+    dataset = QM9(path, pre_transform=pre_transform)
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+
+def default_predict(model, data):
+    return model(data.z, data.pos, data.batch)
+
+
+def precomputed_predict(model, data):
+    return model(data.z, data.batch, data.edge_index, data.edge_attr)
+
+
+if args.use_precomputed_edges:
+    predict = precomputed_predict
+else:
+    predict = default_predict
 
 for target in range(12):
     model, datasets = SchNet.from_qm9_pretrained(path, dataset, target)
@@ -19,10 +55,10 @@ for target in range(12):
     loader = DataLoader(test_dataset, batch_size=256)
 
     maes = []
-    for data in loader:
+    for data in tqdm(loader):
         data = data.to(device)
         with torch.no_grad():
-            pred = model(data.z, data.pos, data.batch)
+            pred = predict(model, data)
         mae = (pred.view(-1) - data.y[:, target]).abs()
         maes.append(mae)
 
--- a/test/nn/models/test_schnet.py
+++ b/test/nn/models/test_schnet.py
@@ -11,7 +11,6 @@ def generate_data():
     return Data(
         z=torch.randint(1, 10, (20, )),
         pos=torch.randn(20, 3),
-        y=torch.tensor([1.]),
     )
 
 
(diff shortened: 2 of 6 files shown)


hatemhelal:
Planning to add a new test in `test/nn/models/test_schnet.py` but possibly using the `@onlyFullTest` marker.  Open to any suggestions though.

torch_geometric/nn/models/schnet.py
```diff
@@ -224,17 +225,34 @@ def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
...
-        """"""
+    @staticmethod
+    def _validate_fwd_args(z, pos, batch, edge_index, edge_weight):
         assert z.dim() == 1 and z.dtype == torch.long
+
+        if batch is not None:
+            assert batch.dim() == 1 and batch.dtype == torch.long
+
+        if pos is not None:
+            assert edge_index is None and edge_weight is None
```
wsad1, line 236: raise maybe a `ValueError`, with a string that explains the error, that will enhance debugability.

torch_geometric/nn/models/schnet.py
```diff
@@ -224,17 +225,34 @@ def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
...
+        else:
+            assert edge_index is not None and edge_weight is not None
+
+    def forward(self, z, pos=None, batch=None, edge_index=None,
+                edge_weight=None):
+        """"""
```
wsad1, line 242: Add a short doc explaining which variables can and cannot be `None` at the same time. Also add that if `edge_index` and `edge_weight` are not given it will be computed in every forward pass.

torch_geometric/nn/models/schnet.py
```diff
@@ -224,17 +225,34 @@ def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
...
+        if batch is not None:
+            assert batch.dim() == 1 and batch.dtype == torch.long
+
+        if pos is not None:
+            assert edge_index is None and edge_weight is None
+        else:
+            assert edge_index is not None and edge_weight is not None
+
+    def forward(self, z, pos=None, batch=None, edge_index=None,
+                edge_weight=None):
```
wsad1, line 241: Lets drop the `edge_weight` argument all together. We can always compute it pretty inexpensively and it simplifies the interface.

wsad1 (commented):
Thanks a lot for this improvement. 
Left a few comments, will take a look again.

torch_geometric/nn/models/schnet.py
hatemhelal: Decided to remove the details assertions, I'm not sure about this but I was weary of adding a check that might raise exceptions in case that impacts model performance.  

Hopefully the more detailed documentation will help.

hatemhelal: I think also with your suggestion to drop providing the `edge_weight` argument it is much simpler and easier to use the model correctly.

hatemhelal:
Many thanks @wsad1 for the helpful feedback.  I think the open questions remaining are:

* Should the forward arguments be validated more strictly?
* There is a usual convention across PyG for having the `batch` argument as the last one. I think if we did this it could break backwards compatibility so I went with putting the `edge_index` last.

Let me know if you have any thoughts on these or anything else that comes to mind on this PR. Thanks again 😄

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,75 @@
...
+    losses = torch.empty(num_steps)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+    for i in range(num_steps):
+        optimizer.zero_grad()
+        out = model(data.z, data.pos)
+        loss = F.mse_loss(out.view(-1), data.y)
+        loss.backward()
+        optimizer.step()
+        losses[i] = loss.detach()
+    assert losses[num_steps - 1] < losses[0]
```
wsad1, line 41:
```suggestion
```
Let's drop this. Tests usually just check the shape of the output.

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,75 @@
...
+    model = SchNet(hidden_channels=16, num_filters=16, num_interactions=2,
+                   num_gaussians=10, cutoff=6.0)
+
+    with torch.no_grad():
+        out = model(batch.z, batch.pos, batch.batch)
+        assert out.size() == (num_graphs, 1)
+
+
+@withPackage('torch_cluster')
+def test_schnet_batch_pre_compute():
```
wsad1, line 60:
```suggestion
def test_schnet_pre_compute_edge_index():
```

wsad1 (commented):
Thanks for the updates @hatemhelal. 
I don't think we need strict validation of inputs. As you suggested it makes sense to  keep `batch` before `edge_index`.

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,64 @@
...
+
+def generate_data():
+    return Data(
+        z=torch.randint(1, 10, (20, )),
+        pos=torch.randn(20, 3),
+        y=torch.tensor([1.]),
+    )
+
+
+@onlyFullTest
```
wsad1, line 19:
```suggestion
```
since its a small test.

wsad1 (approved):
Thanks a lot .

hatemhelal:
Thanks @wsad1!

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,63 @@
...
+
+def generate_data():
+    return Data(
+        z=torch.randint(1, 10, (20, )),
+        pos=torch.randn(20, 3),
+        y=torch.tensor([1.]),
+    )
+
+
+def test_schnet():
```
hatemhelal, line 19:
```suggestion
@withPackage('torch_cluster')
def test_schnet():
```

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,64 @@
+import torch
+import torch.nn.functional as F
```
hatemhelal, line 2:
```suggestion
```

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,64 @@
...
+
+from torch_geometric import seed_everything
+from torch_geometric.data import Batch, Data
+from torch_geometric.nn import SchNet
+from torch_geometric.testing import onlyFullTest, withPackage
```
hatemhelal, line 7:
```suggestion
from torch_geometric.testing import withPackage
```

hatemhelal (commented):
fix lint

hatemhelal:
Looks like the PR is failing on the coverage check - I could add some more tests but if the coverage is more important than the increase in test time and complexity.

@wsad1 what do you think?

hatemhelal:
@wsad1, I tried to update the test to cover the dipole changes but it seems to still not run.  Is it possible to check if the `ase` package is available in the covecov run?

torch_geometric/logging.py
```diff
@@ -25,7 +25,7 @@ def init_wandb(name: str, **kwargs):
 
 def log(**kwargs):
     def _map(value: Any) -> str:
-        if isinstance(value, int):
+        if isinstance(value, int) and not isinstance(value, bool):
```
rusty1s, line 28: Not sure I understand this change.

torch_geometric/nn/models/schnet.py
```diff
@@ -104,10 +102,13 @@ def __init__(self, hidden_channels: int = 128, num_filters: int = 128,
...
-        atomic_mass = torch.from_numpy(ase.data.atomic_masses)
-        self.register_buffer('atomic_mass', atomic_mass)
+        if self.dipole:
+            import ase
 
-        self.embedding = Embedding(100, hidden_channels)
+            atomic_mass = torch.from_numpy(ase.data.atomic_masses)
+            self.register_buffer('atomic_mass', atomic_mass)
+
+        self.embedding = Embedding(100, hidden_channels, padding_idx=0)
```
rusty1s, line 111: Can you clarify why we need `padding_idx=None`?

torch_geometric/nn/models/schnet.py
```diff
@@ -224,15 +226,30 @@ def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
 
         return net, (dataset[train_idx], dataset[val_idx], dataset[test_idx])
 
-    def forward(self, z, pos, batch=None):
-        """"""
+    def forward(self, z, pos, batch=None, edge_index=None):
+        r"""Forward pass of the SchNet model
+
```
rusty1s, line 231:
```suggestion
        r"""
```

torch_geometric/nn/models/schnet.py
```diff
@@ -224,15 +226,30 @@ def from_qm9_pretrained(root: str, dataset: Dataset, target: int):
...
+            z (LongTensor): Atomic number of each atom with shape
+                :obj:`[num_atoms]`
+            pos (Tensor): Coordinates of each atom with shape
+                :obj:`[num_atoms, 3]`
+            batch (LongTensor, optional): Batch indices assigning each atom to
+                a seperate molecule with shape :obj:`[num_atoms]`.
+                (default: :obj:`None`)
+            edge_index (LongTensor, optional): Indices of interacting pairs of
+                atoms with shape :obj:`[2, num_edges]`. Will be computed in
+                every forward pass if not provided. (default: :obj:`None`)
```
rusty1s, line 242:
```suggestion
            z (LongTensor): Atomic number of each atom with shape
                :obj:`[num_atoms]`.
            pos (Tensor): Coordinates of each atom with shape
                :obj:`[num_atoms, 3]`.
            batch (LongTensor, optional): Batch indices assigning each atom to
                a seperate molecule with shape :obj:`[num_atoms]`.
                (default: :obj:`None`)
            edge_index (LongTensor, optional): Indices of interacting pairs of
                atoms with shape :obj:`[2, num_edges]`. Will be computed in
                every forward pass if not provided. (default: :obj:`None`)
```

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,63 @@
...
+from torch_geometric.nn import SchNet
+from torch_geometric.testing import withPackage
+from torch_geometric.transforms import RadiusGraph
+
+
+def generate_data():
+    return Data(
+        z=torch.randint(1, 10, (20, )),
+        pos=torch.randn(20, 3),
+        y=torch.tensor([1.]),
```
rusty1s, line 14:
```suggestion
```

test/nn/models/test_schnet.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+
+@withPackage('torch_cluster')
+@withPackage('ase')
+def test_schnet():
+    seed_everything(0)
```
rusty1s, line 21: Can we drop the `seed_everything` calls here? I don't think they are necessary.

examples/qm9_pretrained_schnet.py
```diff
@@ -1,16 +1,49 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument('--cutoff', type=float, default=10.0,
+                    help='Cutoff distance for interatomic interactions')
+parser.add_argument(
+    '--use_precomputed_edges', action='store_true',
+    help='Compute graph edges and weights as a dataset pre_transform')
+args = parser.parse_args()
+pwd = osp.dirname(osp.realpath(__file__))
+
+log(**vars(args))
```
rusty1s, line 22: Drop.

examples/qm9_pretrained_schnet.py
```diff
@@ -19,10 +52,10 @@
     loader = DataLoader(test_dataset, batch_size=256)
 
     maes = []
-    for data in loader:
+    for data in tqdm(loader):
         data = data.to(device)
         with torch.no_grad():
-            pred = model(data.z, data.pos, data.batch)
+            pred = predict(model, data)
```
rusty1s, line 58:
```suggestion
            if not args.use_precomputed_edges:
                 pred = model(data.z, data.pos, data.batch)
            else:
                 pred = model(data.z, data.pos, data.batch, data.edge_index)
```
and drop `default_predict` and `precomputed_predict`.

rusty1s (approved):

torch_geometric/logging.py
hatemhelal: I think what happens when you pass in a boolean flag is it gets coerced into an int and logs as:

```python
from torch_geometric.logging import log

log(use_opt=True)
```

outputs:
```bash
use_opt: 001
```

with my change the output is:
```bash
use_opt: True
```

Let me know if you'd prefer to handle this in a separate patch or even drop it.

test/nn/models/test_schnet.py
hatemhelal: sure, I did this out of habit.  I've seen pathological cases where a test may fail for a given PRNG state. Arguably overly paranoid though to do this everywhere.

torch_geometric/nn/models/schnet.py
hatemhelal: We use padding atoms with atomic number == 0 to fill partial batches up to a fixed size.  The `padding_idx` option on the embedding could use better documentation but essentially:

* zeros out the embedding vector 
* detaches the corresponding row in the weight matrix from receiving weight updates.

Adding a comment in case that helps.

rusty1s: Ok, but this might break for anyone how uses `0` as a regular atom type, right?

hatemhelal: I'm not aware of anyone using `SchNet` for non-molecular datasets so I thought this would be safe since atomic number zero doesn't map to any elements.

torch_geometric/logging.py
hatemhelal: Very possible this is my bad python, I tried:

```python
vals = [1, True, False, 0, '']

[isinstance(v, int) and not isinstance(v, bool) for v in vals]
```

outputs:
```
[True, False, False, True, False]
```

Removing the `not isinstance(v, bool)` from the condition:
```python
vals = [1, True, False, 0, '']

[isinstance(v, int) for v in vals]
```
outputs:
```
[True, True, True, True, False]
```

hatemhelal: Ah here is the answer:

> The [bool](https://docs.python.org/3/library/functions.html#bool) class is a subclass of [int](https://docs.python.org/3/library/functions.html#int)

from  https://docs.python.org/3/library/functions.html#bool

rusty1s: Oh wow :D Thanks for clarifying :)

PR #5402, opened 2022-09-09
refactor(sampler): clean up sampler attributes, part 3

mananshah99:
_Description TODO._

PR #5404, opened 2022-09-09
refactor(loader): `NodeLoader` and `LinkLoader` as base implementation classes, part 4

mananshah99:
This PR continues the effort to consolidate PyG's sampling interface in preparation for moving `sample(...)` behind the `GraphStore` interface. This effort is somewhat large in scope and will be broken into multiple PRs for ease of review. It builds off of #5402, and makes a significant move to abstract data loading behind a `data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]]` and a `sampler: BaseSampler`.

It does so by introducing two base implementation classes: `NodeLoader` and `LinkLoader`. `NodeLoader` performs sampling from nodes (using `sample_from_nodes`), and `LinkLoader` does the same from edges (using `sample_from_edges`). They both expose parameters in their initializers that are intended for **loading** (that is, the process of using a sampler to get subgraphs, using a feature fetcher to get features, and joining these together to construct a `HeteroData` object to pass downstream). Samplers are intended to expose parameters that are used for **sampling** (that are particular to the sampling method).

The implementations of `NeighborLoader` and `LinkNeighborLoader` are now very simple: they pass the `NeighborSampler` and any necessary initialization parameters directly in `__init__`, with no other change.


torch_geometric/loader/node_loader.py
```diff
@@ -0,0 +1,110 @@
...
+            del kwargs['dataset']
+        if 'collate_fn' in kwargs:
+            del kwargs['collate_fn']
+
+        self.data = data
+
+        # Initialize sampler with keyword arguments:
+        # NOTE sampler is an attribute of 'DataLoader':
+        self.node_sampler = node_sampler
+        if initialize_sampler:
```
Padarn, line 48: Is there a use case for not doing this?

torch_geometric/loader/node_loader.py
```diff
@@ -0,0 +1,110 @@
...
+    SamplerOutput,
+)
+from torch_geometric.typing import InputNodes
+
+
+class NodeLoader(torch.utils.data.DataLoader):
+    r"""A data loader that performs neighborhood sampling from node information."""
+    def __init__(
+        self,
+        data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]],
```
Padarn, line 28: Random question I'll ask here: Do we imagine supporting all three types in the Future? If so:
- should we define a type for `Tuple[FeatureStore, GraphStore]]`, its used in many places
- do you think its possible to make `Data` provide the same interface as `HeteroData` (i.e a dictionary with edge_type) so we can reduce a lot of special case handling?

mananshah99: > should we define a type for Tuple[FeatureStore, GraphStore]], its used in many places

Good point; let me handle this in a follow-up PR (since this one is already a little large).

> do you think its possible to make Data provide the same interface as HeteroData (i.e a dictionary with edge_type) so we can reduce a lot of special case handling?

Another great point, I've been thinking about this a bit as well. I think that would indeed reduce a lot of special case handling; will take a closer look in a follow-up (after all this main refactoring is complete).

torch_geometric/loader/node_loader.py
mananshah99: Yeah, see `LightningNodeData` and `LightningLinkData` for examples of this (we initialize a sampler once, and use it for multiple dataloaders).

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+        return DataLoaderIterator(super()._get_iterator(), self.filter_fn)
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}()'
+
+
+###############################################################################
+
+
+class Dataset(torch.utils.data.Dataset):
```
yaoyaowd, line 194: Could we add some comment to this class? e.g edge_label_index must be COO index etc.
And also when we don't stack edge_label_index, edge_label, edge_label_time into one tensor?

mananshah99: Sure, can add more comments. We can also stack these into one tensor, but I am not sure there would be an efficiency reason to do so; indeed, `NeighborSampler::sample_from_edges` immediately stacks the tensors.

torch_geometric/loader/node_loader.py
Padarn: Cool sounds good - Do we have a roadmap issue for this? If so maybe I can help out with a couple of these issue too. Thanks for your thoughts!

torch_geometric/loader/node_loader.py
Padarn: Ahh I see. `Lightning` always gives caes I hadn't thought of. Thanks

torch_geometric/loader/node_loader.py
```diff
@@ -0,0 +1,154 @@
...
+    NodeSamplerInput,
+    SamplerOutput,
+)
+from torch_geometric.typing import InputNodes
+
+
+class NodeLoader(torch.utils.data.DataLoader):
+    r"""A data loader that performs neighbor sampling from node information,
+    using a generic :class:`~torch_geometric.sampler.BaseSampler`
+    implementation that defines a `sample_from_nodes` function and is supported
```
rusty1s, line 27:
```suggestion
    implementation that defines a :meth:`sample_from_nodes` function and is supported
```

torch_geometric/loader/node_loader.py
```diff
@@ -0,0 +1,154 @@
...
+            The :class:`~torch_geometric.data.Data` or
+            :class:`~torch_geometric.data.HeteroData` graph object.
+        node_sampler (torch_geometric.sampler.BaseSampler, optional): The
+            sampler implementation to be used with this loader. Note that the
+            sampler implementation must be compatible with the input data
+            object.
+        node_sampler_kwargs (Dict[str, Any], optional): the keyword arguments
+            to be passed to the constructor of `node_sampler`; only necessary
+            if `node_sampler` has not already been constructed.
+        initialize_sampler (bool, optional): whether `node_sampler` should be
```
rusty1s, line 41: I don't necessarily understand this argument. Why not just always require a `node_sampler` argument and drop `node_sampler_kwargs` and `initialize_sampler`. I feel like this makes the base class interface more complex than it needs to be.

torch_geometric/loader/utils.py
```diff
@@ -156,3 +157,112 @@ def filter_custom_store(
         data[attr.group_name][attr.attr_name] = tensors[i]
 
     return data
+
+
+# Input Utilities #############################################################
+
+
+def get_input_nodes(
```
rusty1s, line 165: IMO, makes more sense to move this to `node_loader.py`

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+    HeteroSamplerOutput,
+    SamplerOutput,
+)
+from torch_geometric.typing import InputEdges, OptTensor
+
+
+class LinkLoader(torch.utils.data.DataLoader):
+    r"""A data loader that performs neighbor sampling from link information,
+    using a generic :class:`~torch_geometric.sampler.BaseSampler`
+    implementation that defines a `sample_from_edges` function and is supported
```
rusty1s, line 27:
```suggestion
    implementation that defines a :meth:`sample_from_edges` function and is supported
```

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+    on the provided input :obj:`data` object.
+
+    Args:
+        data (torch_geometric.data.Data or torch_geometric.data.HeteroData):
+            The :class:`~torch_geometric.data.Data` or
+            :class:`~torch_geometric.data.HeteroData` graph object.
+        link_sampler (torch_geometric.sampler.BaseSampler, optional): The
```
rusty1s, line 34: Here again.

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+        data: Union[Data, HeteroData, Tuple[FeatureStore, GraphStore]],
+        link_sampler: Optional[BaseSampler] = None,
+        link_sampler_kwargs: Dict[str, Any] = None,
+        initialize_sampler: bool = True,
+        edge_label_index: InputEdges = None,
+        edge_label: OptTensor = None,
+        edge_label_time: OptTensor = None,
+        transform: Callable = None,
+        filter_per_worker: bool = False,
+        neg_sampling_ratio: float = 0.0,
```
rusty1s, line 87: missing in the doc-string. also move right after `edge_label_time`.

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+        # Get input type, or None for homogeneous graphs:
+        edge_type, edge_label_index = get_edge_label_index(
+            data, edge_label_index)
+        if edge_label is None:
+            edge_label = torch.zeros(edge_label_index.size(1),
+                                     device=edge_label_index.device)
+        self.input_type = edge_type
+
+        has_time_attr = link_sampler_kwargs.get('time_attr', None) is not None
+        if ((edge_label_time is not None) != has_time_attr):
```
rusty1s, line 123: nit: Split into two conditionals to make the error message more descriptive.

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}()'
+
+
+###############################################################################
+
+
+class Dataset(torch.utils.data.Dataset):
+    def __init__(self, edge_label_index: torch.Tensor,
+                 edge_label: torch.Tensor, edge_label_time: OptTensor = None):
```
rusty1s, line 196:
```suggestion
                 edge_label: torch.Tensor, edge_label_time: OptTensor = None, ):
```

torch_geometric/loader/link_loader.py
```diff
@@ -0,0 +1,217 @@
...
+        self.edge_label_index = edge_label_index
+        self.edge_label = edge_label
+        self.edge_label_time = edge_label_time
+
+    def __getitem__(self, idx: int) -> Tuple[int]:
```
rusty1s, line 201: Looks like we actually return `Union[Tuple[Tensor, Tensor, Tensor], Tuple[Tensor, Tensor, Tensor, Tensor]]`.

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -168,186 +156,32 @@ def __init__(
...
-        if 'dataset' in kwargs:
-            del kwargs['dataset']
-        if 'collate_fn' in kwargs:
-            del kwargs['collate_fn']
-
-        self.data = data
+        # Get input type:
+        # TODO(manan): this computation is required twice, once here and once
+        # in LinkLoader:
+        edge_type, _ = get_edge_label_index(data, edge_label_index)
```
rusty1s, line 162: Yeah, what are ways to resolve this? Move this to a property of the loader?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -168,186 +156,32 @@ def __init__(
...
 
-        # Save for PyTorch Lightning < 1.6:
+        # Store num_neigbors to retain backwards compatibility:
```
rusty1s, line 164: Why?

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -1,28 +1,16 @@
...
-from torch_geometric.sampler.base import (
-    EdgeSamplerInput,
-    HeteroSamplerOutput,
-    SamplerOutput,
-)
 from torch_geometric.typing import InputEdges, NumNeighbors, OptTensor
 
 
-class LinkNeighborLoader(torch.utils.data.DataLoader):
-    r"""A link-based data loader derived as an extension of the node-based
```
rusty1s, line 25: woot

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -168,186 +156,32 @@ def __init__(
...
+            link_sampler_kwargs={
+                'num_neighbors': num_neighbors,
+                'replace': replace,
+                'directed': directed,
+                'input_type': edge_type,
+                'time_attr': time_attr,
+                'is_sorted': is_sorted,
+                'share_memory': kwargs.get('num_workers', 0) > 0,
+            },
+            initialize_sampler=neighbor_sampler is None,
```
rusty1s, line 179: why not do it as before?

torch_geometric/loader/neighbor_loader.py
```diff
@@ -1,30 +1,17 @@
...
-    SamplerOutput,
-)
 from torch_geometric.typing import InputNodes, NumNeighbors
 
 
-class NeighborLoader(torch.utils.data.DataLoader):
-    r"""A data loader that performs neighbor sampling as introduced in the
-    `"Inductive Representation Learning on Large Graphs"
+class NeighborLoader(NodeLoader):
+    """A data loader that performs neighbor sampling as introduced in the
```
rusty1s, line 13: woot

torch_geometric/loader/neighbor_loader.py
```diff
@@ -1,30 +1,17 @@
...
+    `"Inductive Representation Learning on Large Graphs
```
rusty1s, line 14: re-add "

torch_geometric/loader/neighbor_loader.py
```diff
@@ -177,138 +164,32 @@ def __init__(
...
-            if input_nodes is None:
-                return node_type, range(
-                    remote_backend_utils.num_nodes(feature_store, graph_store,
-                                                   input_nodes))
-            return node_type, to_index(input_nodes)
+        # A NeighborLoader is simply a NodeLoader that uses the NeighborSampler
+        # sampling implementation:
+        super().__init__(
+            data=data,
+            node_sampler=neighbor_sampler or NeighborSampler,
```
rusty1s, line 180: Here again.

torch_geometric/sampler/base.py
```diff
@@ -73,6 +76,19 @@ class BaseSampler(ABC):
         amount of information stored in the sampler, and to initialize all this
         information at `__init__`.
     """
+    @abstractmethod
```
rusty1s, line 79: Still not a big fan of this. Why do we need it?

rusty1s (approved):
Looks mostly good to me. I feel there is some over-complication with `sampler` and `sampler_kwargs` but happy to defer to you if you feel strongly about it.

torch_geometric/loader/neighbor_loader.py
mananshah99: :)

torch_geometric/loader/link_neighbor_loader.py
mananshah99: Some examples explicitly queried `loader.num_neighbors`, so I left this in temporarily in case this was a more general pattern. If not, happy to remove.

torch_geometric/loader/node_loader.py
mananshah99: Agreed, fixed - thanks!

torch_geometric/loader/utils.py
mananshah99: I'm not totally sure, since this is generic enough that if someone wanted to write their own loader, it may also use `get_input_nodes`.

rusty1s: Ok :)

torch_geometric/sampler/base.py
mananshah99: Agreed, and it's no longer important at all :)

torch_geometric/loader/link_neighbor_loader.py
rusty1s: I think this is safe to remove. Let's mark it in the changelog as a potential breaking change.

mananshah99: sgtm

torch_geometric/loader/link_neighbor_loader.py
mananshah99: Yeah, or split up the logic to obtain an edge type and edge label index. Will handle in a follow-up :)

torch_geometric/loader/node_loader.py
mananshah99: I haven't had the time to set up a roadmap issue yet, but will handle this tomorrow and tag you. Thanks for helping out :)

Issue #5409, opened 2022-09-10
ShaDowKHopSampler creates incorrect data.batch when using Batch graph


cvanea:
### 🐛 Describe the bug

Hello! Firstly, thank you for this fantastic library.

I am trying to run various scalable GNNs on custom data but I am having an issue with shaDow.

On this line in the shaDow model definition (as per the example):
`x = torch.cat([x[root_n_id], global_mean_pool(x, batch)], dim=-1)`
This error is thrown:
`RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 1024 but got size 1 for tensor number 1 in the list.`
Where 1024 is the batch size.

I've found this is only a problem when the graph dataset is composed into a disjointed `Batch` graph.

To reproduce the problem, run the code from `examples/shadow.py` but change the data object into a `Batch` object:

```
from torch_geometric.data import Batch
dataset = Flickr(path)
datas = [dataset[0]]
data = Batch.from_data_list(datas)
```

Thank you for any help!

### Environment

* PyG version: 2.1.0.post1
* PyTorch version: 1.11.0+cu113
* OS: Ubuntu 20.04.2 (WSL) and macOS 11.1
* Python version: 3.7.2
* CUDA/cuDNN version: 11.3
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`):

EdisonLeeeee:
Hi @cvanea 

I can indeed reproduce this error and I solved it by deleting the attribute `batch` in `data`:
```python

from torch_geometric.data import Batch
dataset = Flickr(path)
datas = [dataset[0]]
data = Batch.from_data_list(datas)
del data.batch # here
```
I understand that this solution might not be perfect. Will investigate it further and report back.

EdisonLeeeee:
To be fixed in [this PR](https://github.com/pyg-team/pytorch_geometric/pull/5410).

cvanea:
Thanks! This has fixed it on my data as well. 

One additional note, in the `shadow.py` example it may be worth changing `data.num_graphs` as the count of nodes in each batch to `out.size()[0]`. With more than 1 graph in `Batch` this key is overridden as well. That should probably also be added to the excluded keys in the PR change.

EdisonLeeeee:
[...]

PR #5410, opened 2022-09-10
Fix potentially overridden key `batch` and `ptr` in `ShaDowKHopSampler`

EdisonLeeeee:
This PR is to fix the bug mentioned in [this issue](https://github.com/pyg-team/pytorch_geometric/issues/5409) when `ShaDowKHopSampler` accepts `DataBatch` as inputs. This bug was caused by the duplicated key `'batch'` each sampled batch in the `ShaDowKHopSampler` data loader would be potentially overridden by the input `DataBatch`.

Currently, my solution is to exclude the key `'batch'` when iterating `self.data`.
+ Before
```python
        for k, v in self.data:
            if k in ['edge_index', 'adj_t', 'num_nodes']:
                continue
```
+ After
```python
        for k, v in self.data:
            if k in ['edge_index', 'adj_t', 'num_nodes', 'batch', 'ptr']:
                continue
```
I also excluded the key `ptr` to avoid the same problem.

### DIFF
--- a/torch_geometric/loader/shadow.py
+++ b/torch_geometric/loader/shadow.py
@@ -88,7 +88,7 @@ class ShaDowKHopSampler(torch.utils.data.DataLoader):
             batch.edge_index = torch.stack([row, col], dim=0)
 
         for k, v in self.data:
-            if k in ['edge_index', 'adj_t', 'num_nodes']:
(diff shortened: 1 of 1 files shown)


rusty1s (approved):
Thanks!

PR #5425, opened 2022-09-13
Keep `edge_index` with `SparseTensor` data representation for `PNA` model

JakubPietrakIntel:
PNAConv calculates a degree for each node to obtain a scaling factor in [get_degree_histogram()](https://github.com/pyg-team/pytorch_geometric/blob/8bcc77c6f36b44f7be0ca7dc11f4dc3faecd31ed/torch_geometric/nn/conv/pna_conv.py#L191-L203)
In the benchmark using torch_geometric.transforms.ToSparseTensor by default removes edge_index from the transformed tensor, which will lead to test failing on `TypeError: 'NoneType' object is not subscriptable` with input `--models pna --use-sparse-tensor`
Therefore a conditioning is added to keep edge_index in sparse transformed tensor if 'pna' is among tested benchmark models.


benchmark/inference/utils.py
```diff
@@ -20,9 +20,11 @@
 }
 
 
-def get_dataset(name, root, use_sparse_tensor=False, bf16=False):
+def get_dataset(name, root, use_sparse_tensor=False, bf16=False,
```
rusty1s, line 23: Let's just use `remove_edge_index=True`. Not sure if we desperately need this additional argument.

JakubPietrakIntel: I think you meant remove_edge_index=False...  I've set it as default now.

benchmark/inference/inference_benchmark.py
```diff
@@ -25,7 +25,8 @@ def run(args: argparse.ArgumentParser) -> None:
         ), f"Dataset {dataset_name} isn't supported."
         print(f'Dataset: {dataset_name}')
         dataset, num_classes = get_dataset(dataset_name, args.root,
-                                           args.use_sparse_tensor, args.bf16)
+                                           args.use_sparse_tensor, args.bf16,
+                                           remove_edge_index=False)
```
rusty1s, line 29:
```suggestion
                                           args.use_sparse_tensor, args.bf16)
```

rusty1s (approved):

benchmark/inference/utils.py
rusty1s: Yeah, sorry :)

PR #5438, opened 2022-09-14
Add `Summary` for PyG datasets

hatemhelal:
This PR adds the ability to collect summary statistics for PyG datasets.  Initially implemented as a standalone class and posting here for feedback.  Various questions I'm thinking about:

* Would this make more sense as a method on the `torch_geometric.data.Dataset` interface?
* Is it ok to use the `pandas` and `tabulate` packages? I see them in the full requirements and could change the implementation to lazy import / error if those packages aren't available.
* Need to make the summary more useful for single-graph datasets.

Here is a quick demo on QM9:

```python
from torch_geometric.datasets import QM9
from torch_geometric.data import Summary

dataset = QM9("data/QM9")
s = Summary(dataset)
print(s)
```
outputs:
```bash
100%|██████████| 130831/130831 [00:16<00:00, 7841.34it/s]
Summary QM9(130831)
         nodes     edges
----  --------  --------
mean  18.0325   37.3269
std    2.94371   6.29847
min    3         4
25%   16        34
50%   18        38
75%   20        42
max   29        56
```

Can also query properties on the `Summary` object:
```python
s.max_num_nodes, s.max_num_edges
```
outputs
```
(29, 56)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.2.0] - 2022-MM-DD
 ### Added
+- Added `Summary` method for `torch_geometric.data.Dataset` interface ([#5438](https://github.com/pyg-team/pytorch_geometric/pull/5438))
 - Added `FeaturePropagation` transform ([#5387](https://github.com/pyg-team/pytorch_geometric/pull/5387))
 - Added `PositionalEncoding` ([#5381](https://github.com/pyg-team/pytorch_geometric/pull/5381))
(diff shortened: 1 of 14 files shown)


hatemhelal:
Thanks @EdisonLeeeee I updated the patch with your suggestion and updated some of the tests.  I found that the type annotation in `Summary` resulted in a circular module dependencies so I removed the type annotation for now.

I think I'd also like to add a minimum dataset size for showing the progress bar since it looks a bit odd in the unittests.

Happy to hear any other thoughts you have 🙏

EdisonLeeeee:
Hi @hatemhelal 

You've done an amazing job! Left some minor suggestions:
+ Would it support more statistics beyond `nodes` and `edges` in the summarized results? For example, `density/sparsity`, which can be simply calculated by
```python
df['density'] = df['edges'] / df['nodes']**2
```
+ How about merging these properties `*_num_nodes` (e.g., `min_num_nodes`, `max_num_nodes`, etc) as one single method? In this way, you can get rid of implementing each of them whenever new statistics are added. 

+ Do you think it is necessary to include some global statistics such as `num_classes` and `num_features`  in the summarized results?

rusty1s:
@EdisonLeeeee These are all great suggestions to include into `Summary`. Happy to include them in a follow-up PR.

hatemhelal:
All good @rusty1s, thanks @EdisonLeeeee for the kind words too!

PR #5447, opened 2022-09-15
Add `batch = None` for global pool methods

chenzhaiyu:
It seems logical to assign a default `None` value to the optional `batch` argument.


chenzhaiyu:
> Thanks for the PR, although I am a bit skeptical. On the one hand, it makes sense to support this as a default argument. On the other, I am afraid this leads to wrong usage since most of the time, a `batch` argument is desperately required to perform global pooling correctly. With `batch=None` as a default, such errors may be unnoticed. WDYT?

Solid concern. Leaving it here :)

PR #5466, opened 2022-09-17
Add `DGraphFin` dynamic graph dataset

EdisonLeeeee:
This PR added a large-scale dynamic/temporal graph dataset `DGraphFin`. The paper has been accepted by NeurIPS 2022.
I found this dataset useful for facilitating the development of `TemporalData` and some dynamic GNNs. If this PR makes sense, I can add an example for it in the follow-up commits.

+ Home page:https://dgraph.xinye.com/introduction
+ Paper: https://openreview.net/forum?id=2rQPxsmjKF&referrer=%5Bthe%20
+ Baselines: https://github.com/DGraphXinye/DGraphFin_baseline

## Usage
```python
from torch_geometric.datasets import DGraphFin
dataset = DGraphFin("data/DGraphFin")
data = dataset[0]
>>> data
Data(x=[3700550, 17], edge_index=[2, 4300999], edge_attr=[4300999], y=[3700550], edge_time=[4300999], train_mask=[3700550], val_mask=[3700550], test_mask=[3700550])
```

For now, I used `Data` instead of `TemporalData` for storing this dataset. Should I adapt it for `TemporalData`?

### DIFF
--- a/torch_geometric/datasets/dgraph.py
+++ b/torch_geometric/datasets/dgraph.py
@@ -1,0 +1,84 @@
+import os
+from typing import Callable, Optional
+
+import numpy as np
+import torch
+
+from torch_geometric.data import (
+    Data,
+    InMemoryDataset,
+    download_url,
+    extract_zip,
+)
+from torch_geometric.utils import index_to_mask
+
+
+class DGraphFin(InMemoryDataset):
+    r"""The DGraphFin networks from the
+    `"DGraph: A Large-Scale Financial Dataset for Graph Anomaly Detection"
+    <https://arxiv.org/abs/2207.03579>`_ paper.
+    It is a directed, unweighted dynamic graph consisting of millions of
+    nodes and edges, representing a realistic user-to-user social network
+    in financial industry.
+    Node represents a Finvolution user, and an edge from one
+    user to another means that the user regards the other user
+    as the emergency contact person.
+
+
+    Args:
+        root (string): Root directory where the dataset should be saved.
+        transform (callable, optional): A function/transform that takes in an
+            :obj:`torch_geometric.data.Data` object and returns a transformed
+            version. The data object will be transformed before every access.
+            (default: :obj:`None`)
+        pre_transform (callable, optional): A function/transform that takes in
+            an :obj:`torch_geometric.data.Data` object and returns a
+            transformed version. The data object will be transformed before
+            being saved to disk. (default: :obj:`None`)
+    """
+
+    url = "https://dgraph.oss-cn-shanghai.aliyuncs.com/DGraphFin.zip"
+
+    def __init__(self, root: str, transform: Optional[Callable] = None,
+                 pre_transform: Optional[Callable] = None):
+        super().__init__(root, transform, pre_transform)
+        self.data, self.slices = torch.load(self.processed_paths[0])
+
+    def download(self):
+        path = download_url(self.url, self.raw_dir)
+        extract_zip(path, self.raw_dir)
+        os.remove(path)
+
+    @property
+    def raw_file_names(self) -> str:
+        return 'DGraphFin.npz'
+
+    @property
+    def processed_file_names(self) -> str:
+        return 'data.pt'
+
+    def process(self):
+        with np.load(self.raw_paths[0]) as loader:
+            x = torch.from_numpy(loader['x']).to(torch.float)
+            y = torch.from_numpy(loader['y']).to(torch.long)
+            edge_index = torch.from_numpy(loader['edge_index']).to(torch.long)
+            edge_attr = torch.from_numpy(loader['edge_type']).to(torch.long)
+            edge_time = torch.from_numpy(loader['edge_timestamp']).to(
+                torch.long)
+            train_nodes = torch.from_numpy(loader['train_mask']).to(torch.long)
+            val_nodes = torch.from_numpy(loader['valid_mask']).to(torch.long)
+            test_nodes = torch.from_numpy(loader['test_mask']).to(torch.long)
+
+            train_mask = index_to_mask(train_nodes, size=x.size(0))
+            val_mask = index_to_mask(val_nodes, size=x.size(0))
+            test_mask = index_to_mask(test_nodes, size=x.size(0))
+            data = Data(x=x, edge_index=edge_index.t(), edge_attr=edge_attr,
+                        edge_time=edge_time, y=y, train_mask=train_mask,
+                        val_mask=val_mask, test_mask=test_mask)
+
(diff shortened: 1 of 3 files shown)


torch_geometric/datasets/dgraph.py
```diff
@@ -0,0 +1,84 @@
...
+    Data,
+    InMemoryDataset,
+    download_url,
+    extract_zip,
+)
+from torch_geometric.utils import index_to_mask
+
+
+class DGraphFin(InMemoryDataset):
+    r"""The DGraphFin networks from the
```
Padarn, line 17:
```suggestion
    r"""The DGraphFin networks dataset from the
```
Maybe more clear?

torch_geometric/datasets/dgraph.py
```diff
@@ -0,0 +1,84 @@
...
+    def processed_file_names(self) -> str:
+        return 'data.pt'
+
+    def process(self):
+        with np.load(self.raw_paths[0]) as loader:
+            x = torch.from_numpy(loader['x']).to(torch.float)
+            y = torch.from_numpy(loader['y']).to(torch.long)
+            edge_index = torch.from_numpy(loader['edge_index']).to(torch.long)
+            edge_attr = torch.from_numpy(loader['edge_type']).to(torch.long)
+            edge_time = torch.from_numpy(loader['edge_timestamp']).to(
```
Padarn, line 66: Does this mean the time that the edge was added?

Padarn:
This looks like a useful dataset! I think it would be cool to have an example dataset that directly returns `TemporalData`. Any concerns with doing this?

torch_geometric/datasets/dgraph.py
```diff
@@ -0,0 +1,84 @@
...
+            test_mask = index_to_mask(test_nodes, size=x.size(0))
+            data = Data(x=x, edge_index=edge_index.t(), edge_attr=edge_attr,
+                        edge_time=edge_time, y=y, train_mask=train_mask,
+                        val_mask=val_mask, test_mask=test_mask)
+
+        data = data if self.pre_transform is None else self.pre_transform(data)
+        data, slices = self.collate([data])
+        torch.save((data, slices), self.processed_paths[0])
+
+    def __repr__(self) -> str:
```
Padarn, line 83: No need for this - base class implementation does the same thing

torch_geometric/datasets/dgraph.py
```diff
@@ -0,0 +1,84 @@
...
+
+            train_mask = index_to_mask(train_nodes, size=x.size(0))
+            val_mask = index_to_mask(val_nodes, size=x.size(0))
+            test_mask = index_to_mask(test_nodes, size=x.size(0))
+            data = Data(x=x, edge_index=edge_index.t(), edge_attr=edge_attr,
+                        edge_time=edge_time, y=y, train_mask=train_mask,
+                        val_mask=val_mask, test_mask=test_mask)
+
+        data = data if self.pre_transform is None else self.pre_transform(data)
+        data, slices = self.collate([data])
```
Padarn, line 80: Do we need to do this if there is only one data object?

torch_geometric/datasets/dgraph.py
EdisonLeeeee: Yes. This makes the graph a dynamic one.

torch_geometric/datasets/dgraph.py
EdisonLeeeee: Got it.

torch_geometric/datasets/dgraph.py
EdisonLeeeee: I saw the implementation in other datasets did the same thing even though there is only one data object. It seems that it is a necessary step for datasets.

EdisonLeeeee:
Thanks @Padarn for your helpful feedback! The major concern I have is that `TemporalData` is under development and might lack some necessary functionality. For example, I am not sure how to define `x` and `y` for `TemporalData` as they remain unchanged over time.

Planning to add an example of using `NeighborLoader` for temporal sampling, make sense?

torch_geometric/datasets/dgraph.py
Padarn: Actually it seems to do this
```python
    @staticmethod
    def collate(
            data_list: List[Data]) -> Tuple[Data, Optional[Dict[str, Tensor]]]:
        r"""Collates a Python list of :obj:`torch_geometric.data.Data` objects
        to the internal storage format of
        :class:`~torch_geometric.data.InMemoryDataset`."""
        if len(data_list) == 1:
            return data_list[0], None
       ...
```
I guess there is no harm, but I suspect it does nothing?

Padarn:
Yep fair enough. `TemporalData` seems better suited to messages on the edges coming over time.

Good idea on the example with `NeighbourLoader`

torch_geometric/datasets/dgraph.py
Padarn: got it, thanks.

torch_geometric/datasets/dgraph.py
EdisonLeeeee: Noted. Can we just leave it?

EdisonLeeeee:
I have to close this PR now since I was told by the authors that this dataset is not ready to be included in PyG. I'll keep in touch with the authors and report back if there are any advances. Sorry for the wasted efforts here @Padarn @rusty1s

rusty1s:
Can you clarify on this? The dataset is publicly available so we should be able to use it.

EdisonLeeeee:
They said, anyone who wants to use this dataset should explicitly agree their [clause](https://dgraph.xinye.com/clause). They are afraid this cannot be guaranteed if used through PyG:
> You shall use the Datasets only for non-commercial research and educational purposes. You may provide research associates and colleagues with access to the Datasets provided that they first agree to be bound by these terms and conditions. If you are employed by a for-profit, commercial entity, your employer shall also be bound by these terms and conditions, and you hereby represents that he or she is fully authorized to enter into this agreement on behalf of such employer.

rusty1s:
In that case, we can make the download manually and just provide a `process` script (similar to the `FAUST` dataset).

EdisonLeeeee:
This sounds feasible. Let me check with them first. I'll report back ASAP : )

DGraphXinye:
@rusty1s @Padarn @rfdavid @desktable @EdisonLeeeee   Thanks for your appreciation and support of DGraph. Regarding the clause issue in the current version, we have discussed it with the submitter. We will support @EdisonLeeeee  to submit a new version like the FAUST case, and we anticipate DGraph will be able to help the graph community.

PR #5474, opened 2022-09-19
Fix broken url in `MD17` dataset

EdisonLeeeee:
This PR is to fix the broken URL in `benzene` and `benzene FHI-aims` datasets as the dataset URLs are changed now:
```python
from torch_geometric.datasets import MD17
dataset = MD17("data/MD17/", "benzene")
# HTTPError: HTTP Error 404: Not Found
```


rusty1s (approved):
Thanks!

PR #5507, opened 2022-09-22
Fix `None` case of `sampler` or `batch_sampler`

RBendias:
Passing `sampler=None` to LightningNodeData currently results in no shuffling.


rusty1s (approved):
Thanks for the great catch!

PR #5510, opened 2022-09-22
Minor fix for `pyg-lib` usage in `HeteroLinear` and `RGCNConv`

puririshi98:
pyg_lib pathway is chosen if cuda is available and pyg_lib is available. but if the input is not cuda we should not be using the pyg_lib pathway. simple fix

without this a ton of CI fails for testing rgcnconv and heterolinear w/ cpu inputs

[errors.txt](https://github.com/pyg-team/pytorch_geometric/files/9629902/errors.txt)


EdisonLeeeee:
Use `x.is_cuda` for device check?

rusty1s:
This PR will currently break TorchScript functionality since we cannot check for device within a TorchScript program (that's why we have this awkward
```python
self._WITH_PYG_LIB = torch.cuda.is_available() and _WITH_PYG_LIB
```
check as a workaround in `__init__`. I suggest that we simply integrate CPU `segment_matmul` support in `pyg-lib` via the classic for-loop instead.

rusty1s:
cc @dszwicht

dszwicht:
@rusty1s I have had a minimal implementation of `[segment | grouped]_matmul` ready for some time. Was working on optimization, but in such a case I'll create a PR with basic solution later today 😉

puririshi98:
okay id assume @dszwicht will PR that to pyg_lib. in that case then
`self._WITH_PYG_LIB = torch.cuda.is_available() and _WITH_PYG_LIB
`
should change to 
`self._WITH_PYG_LIB =  _WITH_PYG_LIB`

then once pyg_lib is updated all pathways should work

rusty1s:
Indeed. PR to `pyg-lib` just landed, so this should be resolved soon.

rusty1s:
Thanks for the update @puririshi98. I noticed that `segment_matmul` still does not support TorchScript because the Python implementation via `torch.autograd.Function` does not come with TorchScript support. Now that the dispatching of `segment_matmul` is resolved in `pyg-lib`, can we move its backward implementation from Python to C++? Let me know if you have time working on this!

rusty1s (approved):

PR #5516, opened 2022-09-23
`NeighborSampler`: Sort local neighborhoods according to `time`

rusty1s:


torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -124,16 +125,33 @@ def __init__(
...
             feature_store, graph_store = data
 
+            # Obtain all node and edge metadata:
+            node_attrs = feature_store.get_all_tensor_attrs()
+            edge_attrs = graph_store.get_all_edge_attrs()
+
             # TODO support `collect` on `FeatureStore`:
             self.node_time_dict = None
             if time_attr is not None:
+                # If the `time_attr` is present, we expect that `GraphStore`
```
rusty1s, line 135: cc @mananshah99. I currently cannot think of a more elegant way to support this. IMO, we need to extend `GraphStore` to support some form node-level metadata as well.

torch_geometric/sampler/utils.py
```diff
@@ -27,16 +28,32 @@ def to_csc(
     perm: Optional[Tensor] = None
 
     if hasattr(data, 'adj'):
+        if src_node_time is not None:
+            raise NotImplementedError("Temporal sampling via 'SparseTensor' "
+                                      "format not yet supported")
         colptr, row, _ = data.adj.csc()
 
     elif hasattr(data, 'adj_t'):
+        if src_node_time is not None:
```
rusty1s, line 37: cc @mananshah99: I don't have any good idea how we can support this. `SparseTensor` will always automatically sort by column index when not explicitly initialized via `is_sorted=True`.

ZenoTan (approved):
Thanks!

test/loader/test_neighbor_loader.py
```diff
@@ -381,37 +380,57 @@ def test_custom_neighbor_loader(FeatureStore, GraphStore):
...
+        group_name='paper',
+        attr_name='time',
+        index=None,
+    )
+    hetero_data['paper'].time = data.time
+
+    # Sort according to time in local neighborhoods:
+    row, col = data.edge_index
+    perm = ((col * (data.num_nodes + 1)) + data.time[row]).argsort()
+    edge_index = data.edge_index[:, perm]
```
mananshah99, line 416: Perhaps add a note that this is required for graph store, since we currently need to pass `is_sorted=True` in this case? Otherwise, folks may ask why it's necessary to have this logic in a test

torch_geometric/sampler/neighbor_sampler.py
mananshah99: Hm, so it seems like the requirement here is that GraphStore should be able to return `csc()`, sorted by column, and sorted by time within local neighborhoods. Why don't we just extend the layout conversion functionality here to support `GraphStore.csc(..., src_node_time: Optional[torch.Tensor] = None)`, since this all happens in-memory anyway? We can handle this in `_edge_to_layout` there.

The alternative would be to allow GraphStore to support node-level metadata as you mentioned, and one argument in favor of this would be temporal sampling. I am also open to this; happy to discuss more in-person :)

torch_geometric/sampler/utils.py
mananshah99: At worst, we can convert to COO and back to CSC, but that's not very satisfying. Maybe it's a good stop-gap?

mananshah99 (commented):
Mostly LGTM. Happy to chat more about how to modify the `GraphStore` design to support CSC sorted by node time, but at the moment I think we can get around this by passing `node_time` to `csc()`. Lmkwyt!

test/loader/test_neighbor_loader.py
```diff
@@ -280,14 +280,13 @@ def forward(self, x, edge_index, edge_weight):
...
 
-@withPackage('torch_sparse>=0.6.14')
 def test_temporal_heterogeneous_neighbor_loader_on_cora(get_dataset):
     dataset = get_dataset(name='Cora')
     data = dataset[0]
 
     hetero_data = HeteroData()
     hetero_data['paper'].x = data.x
-    hetero_data['paper'].time = torch.arange(data.num_nodes)
+    hetero_data['paper'].time = torch.arange(data.num_nodes, 0, -1)
```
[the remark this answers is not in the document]
rusty1s: We change this here so that it is no longer sorted by time by default. This lets us catch bugs more easily.

test/loader/test_neighbor_loader.py
rusty1s: Sg.

torch_geometric/sampler/neighbor_sampler.py
rusty1s: I have thought about additional an attribute to `csc` as well. It might work here indeed. What happens if we call `csc` with a time tensor though? Do we ignore it and simply trust that the data is formatted correctly?

torch_geometric/sampler/utils.py
rusty1s: 👍

torch_geometric/sampler/utils.py
```diff
@@ -27,16 +28,32 @@ def to_csc(
...
             perm = (col * data.size(0)).add_(row).argsort()
             row = row[perm]
+        if not is_sorted and src_node_time is not None:
+            # Multiplying by raw `datetime[64ns]` values may cause overflows.
+            # As such, we normalize time into range [0, 1) before sorting:
+            src_node_time = src_node_time.to(torch.double, copy=True)
+            min_time, max_time = src_node_time.min(), src_node_time.max() + 1
+            src_node_time.sub_(min_time).div_(max_time)
+
+            perm = src_node_time[row].add_(col.to(torch.double)).argsort()
```
wsad1, line 54:
```suggestion
            perm = col.to(torch.double).add_(src_node_time[row]).argsort()
```
Nitpick: makes it look similar to the computation when `src_node_time is None`.

wsad1 (approved):
Thanks for adding this.

PR #5517, opened 2022-09-23
`GraphStore`: Refactor and remove `SparseTensor` dependency

rusty1s:


torch_geometric/data/graph_store.py
```diff
@@ -58,30 +61,39 @@ class EdgeLayout(Enum):
 class EdgeAttr(CastMixin):
     r"""Defines the attributes of an :obj:`GraphStore` edge."""
 
-    # The type of the edge
-    edge_type: Optional[EdgeType]
+    # The type of the edge:
+    edge_type: EdgeType
```
Padarn, line 65: I assume this was originally optional to account for non hetero graphs. Do we have a new convention for standard graphs to define what EdgeType should be?

torch_geometric/data/graph_store.py
```diff
@@ -104,21 +116,14 @@ def _put_edge_index(self, edge_index: EdgeTensorType,
...
-        r"""Synchronously adds an edge_index tensor to the graph store.
+        r"""Synchronously adds the :obj:`edge_index` tensor to the graph store.
 
         Args:
-            tensor(EdgeTensorType): an edge_index in a format specified in
-            attr.
-            **attr(EdgeAttr): the edge attributes.
+            tensor (EdgeTensorType): The :obj:`edge_index` in the format
+                as specified in :obj:`attr`.
+            attr (EdgeAttr): The edge attributes.
```
Padarn, line 124: Just a random comment, `edge_attr` is quite  confusing here because we also use it for edge features

torch_geometric/data/graph_store.py
rusty1s: I see. Need to think a bit more about this but I assume we can simply override `EdgeAttr` for the homogeneous case.

torch_geometric/data/graph_store.py
rusty1s: Yes :( Do you know of a better name?

Padarn: perhaps `EdgeProperties` ... bit its a little verbose.

torch_geometric/data/graph_store.py
Padarn: What do you mean override for the homogenous case? We'll still need the layout info right?

rusty1s: In the homogeneous case, we can override `EdgeAttr` to not include the `edge_type` information. Let me see if that works :)

torch_geometric/data/graph_store.py
rusty1s: Sounds good to me. `EdgeProp` also works. I will likely move this to a follow-up PR though if that works for you.

Padarn: 👍

torch_geometric/data/graph_store.py
Padarn: Oh I see what you mean. That makes sense.

PR #5537, opened 2022-09-26
Add `HydroNet` dataset

hatemhelal:
This PR adds support for working with the `HydroNet` dataset described in:
[HydroNet: Benchmark Tasks for Preserving Intermolecular Interactions and Structural Motifs in Predictive and Generative Models for Molecular Data](https://arxiv.org/abs/2012.00131)

This is quite a large dataset of ~5 million graphs where each graph consists of:

* `[num_atoms, 3]` tensor containing atomic coordinates
* `y` float label containing the energy for the water cluster in kcal/mol

The raw dataset is stored as several compressed `.xyz` files where each file contains the structures for the a fixed number of clusters (e.g. the file `WN_geoms_all.zip` contains the data for a cluster of `N` water molecules).  The entire dataset is pre-processed in parallel using pandas to read the compressed files into the PyG `Data` objects.  The number of workers for this is controlled by the `max_num_workers` argument.  A subset of the entire dataset can be selected through the `clusters` argument of the constructor.

Some additional ideas (could be handled in either this PR or as a follow up):
* a complete example using the SchNet implementation in PyG to train for a few epochs on a subset of the dataset
* automated download from the official PNNL data hub (expect this will need help from @jenna1701, @jesunsahariar, and @sutanay


rusty1s (approved):
I think this looks generally good, so happy to merge this.

The nested `InMemoryDataset` is something special though. Wondering why you did go this route? This also leads to re-processing of the data in consecutive runs of `HydroNet()` which is somewhat undesired IMO.

hatemhelal:
Thanks so much for merging this!

> The nested `InMemoryDataset` is something special though. Wondering why you did go this route? This also leads to re-processing of the data in consecutive runs of `HydroNet()` which is somewhat undesired IMO.

This probably needed a comment to explain in the source: the design is that the full `HydroNet` dataset is composed of a partition for each graph/water cluster size.  This way the wrapper class manages the downloading and initialisation of the partitions.  Each partition points to a single raw/processed file pair from the whole dataset so as long as the wrapping class is used, the data should only be processed once.  Here is an example of just calling `len` twice on two datasets to show this:

```
In [3]: dataset = HydroNet('./data/HydroNet')

In [4]: len(dataset)
Out[4]: 4464740

In [5]: dataset = HydroNet('./data/HydroNet')

In [6]: len(dataset)
Out[6]: 4464740
```

I tried as much as possible to make this lazy-initialised but there might be edge cases with say selecting the `use_processed=False` option.

abojchevski:
This breaks the installation* of PytorchGeometric on Google Colab. Namely, you get the following error

```
ImportError: cannot import name 'cached_property' from 'functools' (/usr/lib/python3.7/functools.py)
```

since 'cached_property' has been introduced only in Python 3.8.


* The installation I am referring to is:
```
import os
import torch
os.environ['TORCH'] = torch.__version__
print(torch.__version__)

!pip install -q torch-scatter -f https://data.pyg.org/whl/torch-${TORCH}.html
!pip install -q torch-sparse -f https://data.pyg.org/whl/torch-${TORCH}.html
!pip install -q git+https://github.com/pyg-team/pytorch_geometric.git
```

rusty1s:
Thanks for letting us know. Is it possible for you to jump to an earlier commit while we are fixing this?

abojchevski:
Thanks!

hatemhelal:
Thanks for fixing this up @rusty1s and @EdisonLeeeee  and really sorry for the hassle I've caused 😞

PR #5540, opened 2022-09-26
Unknown label support in `to_homogeneous`

puririshi98:
```
>>> x=torch_geometric.datasets.FakeHeteroDataset().data
>>> x.to_homogeneous().y
tensor([ 7.,  0.,  8.,  ..., -1., -1., -1.])
>>> x.to_homogeneous()
Data(edge_index=[2, 71136], y=[3497], node_type=[3497], edge_type=[71136])
```

### DIFF
--- a/torch_geometric/data/hetero_data.py
+++ b/torch_geometric/data/hetero_data.py
@@ -45,62 +45,42 @@ class HeteroData(BaseData, FeatureStore, GraphStore):
     behaviour of a regular **nested** Python dictionary.
     In addition, it provides useful functionality for analyzing graph
     structures, and provides basic PyTorch tensor functionalities.
-
     .. code-block::
-
         from torch_geometric.data import HeteroData
-
         data = HeteroData()
-
         # Create two node types "paper" and "author" holding a feature matrix:
         data['paper'].x = torch.randn(num_papers, num_paper_features)
         data['author'].x = torch.randn(num_authors, num_authors_features)
-
         # Create an edge type "(author, writes, paper)" and building the
         # graph connectivity:
         data['author', 'writes', 'paper'].edge_index = ...  # [2, num_edges]
-
         data['paper'].num_nodes
         >>> 23
-
         data['author', 'writes', 'paper'].num_edges
         >>> 52
-
         # PyTorch tensor functionality:
         data = data.pin_memory()
         data = data.to('cuda:0', non_blocking=True)
-
     Note that there exists multiple ways to create a heterogeneous graph data,
     *e.g.*:
-
     * To initialize a node of type :obj:`"paper"` holding a node feature
       matrix :obj:`x_paper` named :obj:`x`:
-
       .. code-block:: python
-
         from torch_geometric.data import HeteroData
-
         data = HeteroData()
         data['paper'].x = x_paper
-
         data = HeteroData(paper={ 'x': x_paper })
-
         data = HeteroData({'paper': { 'x': x_paper }})
-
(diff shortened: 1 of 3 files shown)


torch_geometric/data/hetero_data.py
```diff
@@ -891,3 +897,17 @@ def to_homogeneous_edge_index(
         edge_index = torch.cat(edge_indices, dim=-1)
 
     return edge_index, node_slices, edge_slices
+
+
+def to_homogeneous_label(data: HeteroData, ) -> Tensor:
```
Padarn, line 902:
```suggestion
def to_homogeneous_label(data: HeteroData) -> Tensor:
```

torch_geometric/data/hetero_data.py
```diff
@@ -891,3 +897,17 @@ def to_homogeneous_edge_index(
...
+    labeled_node_types = []
+    y = []
+    for i, node_type in enumerate(data.node_types):
+        if 'y' in data[node_type].keys():
+            labeled_node_types.append(i)
+            y.append(data[node_type].y)
+        else:
+            y.append(-1 * torch.ones(data[node_type].num_nodes))
+
+    if labeled_node_types:
```
Padarn, line 912: when would this happen? not sure if failing silently like this makes sense

torch_geometric/data/hetero_data.py
```diff
@@ -704,6 +705,11 @@ def _consistent_size(stores: List[BaseStorage]) -> List[str]:
         if not data.can_infer_num_nodes:
             data.num_nodes = list(node_slices.values())[-1][1]
 
+        # Combine labels
+        node_labels = to_homogeneous_label(self)
```
rusty1s, line 709: Can we add a test for this?

rusty1s, line 709:
```suggestion
        data.y = to_homogeneous_label(self)
```

torch_geometric/data/hetero_data.py
```diff
@@ -891,3 +897,17 @@ def to_homogeneous_edge_index(
...
+
+def to_homogeneous_label(data: HeteroData, ) -> Tensor:
+    labeled_node_types = []
+    y = []
+    for i, node_type in enumerate(data.node_types):
+        if 'y' in data[node_type].keys():
+            labeled_node_types.append(i)
+            y.append(data[node_type].y)
+        else:
+            y.append(-1 * torch.ones(data[node_type].num_nodes))
```
rusty1s, line 910: Use `torch.full` for this.

torch_geometric/data/hetero_data.py
```diff
@@ -891,3 +897,17 @@ def to_homogeneous_edge_index(
...
+
+
+def to_homogeneous_label(data: HeteroData, ) -> Tensor:
+    labeled_node_types = []
+    y = []
+    for i, node_type in enumerate(data.node_types):
+        if 'y' in data[node_type].keys():
+            labeled_node_types.append(i)
+            y.append(data[node_type].y)
+        else:
```
rusty1s, line 909: This will add node labels even in case the dataset does not have a `y` tensor at all. We should gate the for-loop by `if len(data.y_dict) > 0`.

torch_geometric/data/hetero_data.py
rusty1s, line 709: Only run this in self `data.y` does not exist on the homogeneous data object.

rusty1s (approved):

Issue #5452, opened 2022-09-15
Graph data augmentation `augmentation`/`aug` (sub)package


EdisonLeeeee:
### 🚀 The feature, motivation and pitch

There is a surge in graph data augmentation research, and it is often important for self-supervised learning on graphs.  I saw there is an implementation of DropEdge (`dropout_adj`) in the `torch_geometric.utils` module. Given the prevalence of graph augmentation methods, is there any plan to integrate them into PyG?

+ https://github.com/zhao-tong/graph-data-augmentation-papers

### Additional context

As a heavy PyG user, what I want is:
```python
import torch_geometric.augmentation as A
aug1 = A.DropEdge()
aug2 = A.DropNode()
aug3 = ...

x, edge_indexx = aug1(x, edge_index)
# or x, edge_indexx = aug2(x, edge_index)

```

rusty1s:
This sounds really cool. Currently, this would interfere a bit with our `transforms` package which is designed for augmentation. The `transforms` operates on `data`, while here it looks like we want to perform augmentations as part of the model. Do you have an idea about how to align these two?

EdisonLeeeee:
Sorry for the late reply. I intended to implement these augmentation methods as part of `transforms` and accept `data` as input. However, the problem is, that methods in `transforms` are all inplaced ones. We shouldn't make the augmentations inplaced during training, right?

I assume that `transforms` package is used as data preprocessing **before training**, while `augmentations` (let's call it by that name for now) is to provide data augmentation **during training**. They are a bit different as I see.  We can follow the implementation of `dropout_adj` to implement these methods, either as one `function` or `class`. WDYT?

EdisonLeeeee:
Since augmentation is used as part of the model during training, how about implementing it as a subpackage of `nn`?

Padarn:
I like this idea a lot.

> I assume that transforms package is used as data preprocessing before training, while augmentations (let's call it by that name for now) is to provide data augmentation during training

Could we implement both using the same base functionality? One example is the `AddSelfLoops` transform, which applies the `add_self_loops` utility. We could migrate `add_self_loops` into a new `augmentations` package and then have `Transforms` just wrap them so you can chain them as preprocessing steps.

I can't think of any clean abstraction to combine these right now but I think moving some of these out of `utils` would make it easier to discover the functionality.

WDYT?

EdisonLeeeee:
[...]

PR #5548, opened 2022-09-27
Add basic augmentation methods

EdisonLeeeee:
This PR added some basic augmentation methods for graphs:
+ NodeShuffle: randomly permute node ids
+ MaskFeature: randomly mask feature columns
+ AddEdge: randomly add edges into graphs

~~TODO~~:
+ ~~add doc-string~~
+ ~~add test~~

Link to https://github.com/pyg-team/pytorch_geometric/issues/5452


torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+
+    device = edge_index.device
+    if not training or p == 0.0:
+        edge_index_to_add = torch.tensor([], device=device).view(2, 0)
+        return edge_index, edge_index_to_add
+
+    num_nodes = (num_nodes,
+                 num_nodes) if not isinstance(num_nodes, tuple) else num_nodes
+    num_src_nodes = num_nodes[0] or edge_index[0].max().item() + 1
+    num_dst_nodes = num_nodes[1] or edge_index[1].max().item() + 1
```
wsad1, line 177:
```suggestion
    num_src_nodes = maybe_num_nodes(edge_index, num_nodes[0])
    num_dst_nodes = maybe_num_nodes(edge_index, num_nodes[1])
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+
+    The method returns (1) the retained :obj:`edge_index`, (2) the added
+    edge indices.
+
+    Args:
+        edge_index (LongTensor): The edge indices.
+        p (float, optional): Ratio of added edges to the existing edges.
+            (default: :obj:`0.2`)
+        force_undirected (bool, optional): If set to :obj:`True`, will either
+            drop or keep both edges of an undirected edge.
```
wsad1, line 119:
```suggestion
        force_undirected (bool, optional): If set to  :obj:`True` added edges will be undirected. 
            (default: :obj:`False`)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+    # To avoid inplace operation
+    x = x * (mask.float())
+    return x, mask
+
+
+def add_random_edge(edge_index, p: float = 0.2, force_undirected: bool = False,
+                    num_nodes: Optional[Union[Tuple[int], int]] = None,
+                    training: bool = True) -> Tuple[Tensor, Tensor]:
+    r"""Randomly adds edges of ratio :obj:`p` into the existing edges
+    :obj:`edge_index`.
```
wsad1, line 109:
```suggestion
    r"""Randomly adds edges  to :obj:`edge_index`.
```
Nitpick: maybe more clearer.

wsad1 (commented):
Looks good mostly. Left some initial comments.

torch_geometric/utils/augmentation.py
wsad1: For bipartite graph the user would always pass a `tuple` for `num_nodes` so we won't use `max_val` of `edge_index`, right?

EdisonLeeeee: Indeed.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,191 @@
...
+
+    The method returns (1) the retained :obj:`x`, (2) the feature
+    mask indicating where features are retained.
+
+    Args:
+        x (FloatTensor): The feature matrix.
+        p (float, optional): Masking ratio. (default: :obj:`0.5`)
+        column_wise (bool, optional): If set to :obj:`True`, this operation
+            mask columns of the feature. Otherwise, the masked features
+            are sampled uniformly. (default: :obj:`True`)
```
wsad1, line 57:
```suggestion
        column_wise (bool, optional): If set to :obj:`True`, masks whole feature
             columns with probability :obj:`p`.  (default: :obj:`True`)
```

test/utils/test_augmentation.py
```diff
@@ -0,0 +1,73 @@
...
+
+    torch.manual_seed(5)
+    out = add_random_edge(edge_index, p=0.5)
+    assert out[0].tolist() == [[0, 1, 1, 2, 2, 3, 3, 2, 3],
+                               [1, 0, 2, 1, 3, 2, 1, 2, 2]]
+
+    assert out[1].tolist() == [[3, 2, 3], [1, 2, 2]]
+
+    torch.manual_seed(6)
+    out = add_random_edge(edge_index, p=0.5, force_undirected=1)
```
wsad1, line 61:
```suggestion
    out = add_random_edge(edge_index, p=0.5, force_undirected=True)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def mask_feature(x: Tensor, p: float = 0.5, column_wise: bool = True,
+                 training: bool = True) -> Tuple[Tensor, Tensor]:
+    r"""Randomly masks feature from the feature matrix
+    :obj:`x` with probability :obj:`p` using samples from
+    a Bernoulli distribution.
+
+    The method returns (1) the retained :obj:`x`, (2) the feature
+    mask indicating where features are retained.
```
wsad1, line 50:
```suggestion
    mask, the same shape as :obj:`x`, indicating where features are retained.
```

wsad1 (approved):
Thanks a lot.

EdisonLeeeee:
Thank you @wsad1 for your feedback. Really helpful!

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        tensor([[6., 7., 8.],
+                [3., 4., 5.],
+                [0., 1., 2.]])
+        >>> node_perm
+        tensor([2, 1, 0])
+    """
+    if not training:
+        perm = torch.arange(x.size(0))
+        return x, perm
+    perm = torch.randperm(x.size(0))
```
rusty1s, line 39:
```suggestion
    perm = torch.randperm(x.size(0), device=x.device)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        >>> x, node_perm = shuffle_node(x)
+        >>> x
+        tensor([[6., 7., 8.],
+                [3., 4., 5.],
+                [0., 1., 2.]])
+        >>> node_perm
+        tensor([2, 1, 0])
+    """
+    if not training:
+        perm = torch.arange(x.size(0))
```
rusty1s, line 37:
```suggestion
        perm = torch.arange(x.size(0), device=x.device)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def mask_feature(x: Tensor, p: float = 0.5, column_wise: bool = True,
+                 training: bool = True) -> Tuple[Tensor, Tensor]:
+    r"""Randomly masks feature from the feature matrix
+    :obj:`x` with probability :obj:`p` using samples from
+    a Bernoulli distribution.
+
+    The method returns (1) the retained :obj:`x`, (2) the feature
+    mask, the same shape as :obj:`x`, indicating where features are retained.
```
rusty1s, line 50:
```suggestion
    The method returns (1) the retained :obj:`x`, (2) the feature
    mask with the same shape as :obj:`x`, indicating where features are retained.
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Args:
+        x (FloatTensor): The feature matrix.
+        p (float, optional): Masking ratio. (default: :obj:`0.5`)
```
rusty1s, line 54:
```suggestion
        p (float, optional): The masking ratio. (default: :obj:`0.5`)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        column_wise (bool, optional): If set to :obj:`True`, masks whole feature
```
rusty1s, line 55: I think we can actually add 3 modes, e.g., all, row, and col.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        return x, torch.ones_like(x, dtype=torch.bool)
+
+    if column_wise:
+        mask = torch.rand(x.size(1), device=x.device) >= p
+        mask = mask.tile(x.size(0), 1)
+    else:
+        mask = torch.randn_like(x) >= p
+
+    # To avoid inplace operation
+    x = x * (mask.float())
```
rusty1s, line 102:
```python
x = x.masked_fill(mask, 0.0)
```
avoids the conversion to `float`.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    if p < 0. or p > 1.:
+        raise ValueError(f'Masking ratio has to be between 0 and 1 '
+                         f'(got {p}')
+    if not training or p == 0.0:
+        return x, torch.ones_like(x, dtype=torch.bool)
+
+    if column_wise:
+        mask = torch.rand(x.size(1), device=x.device) >= p
+        mask = mask.tile(x.size(0), 1)
```
rusty1s, line 97:
```suggestion
        mask = mask.view(1, -1)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        raise ValueError(f'Ratio of added edges has to be between 0 and 1 '
+                         f'(got {p}')
+
+    device = edge_index.device
+    if not training or p == 0.0:
+        edge_index_to_add = torch.tensor([], device=device).view(2, 0)
+        return edge_index, edge_index_to_add
+
+    num_nodes = (num_nodes,
+                 num_nodes) if not isinstance(num_nodes, tuple) else num_nodes
```
rusty1s, line 175:
```suggestion
    if not isinstance(num_nodes, (tuple, list)):
        num_nodes = (num_nodes, num_nodes)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+        tensor([[3, 4, 1],
+                [1, 3, 2]])
+    """
+    if p < 0. or p > 1.:
+        raise ValueError(f'Ratio of added edges has to be between 0 and 1 '
+                         f'(got {p}')
+
+    device = edge_index.device
+    if not training or p == 0.0:
+        edge_index_to_add = torch.tensor([], device=device).view(2, 0)
```
rusty1s, line 171:
```suggestion
        edge_index_to_add = torch.tensor([[], []], device=device)
```
would work as well :)

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,190 @@
...
+    num_nodes = (num_nodes,
+                 num_nodes) if not isinstance(num_nodes, tuple) else num_nodes
+    num_src_nodes = maybe_num_nodes(edge_index, num_nodes[0])
+    num_dst_nodes = maybe_num_nodes(edge_index, num_nodes[1])
+
+    num_edges_to_add = round(edge_index.size(1) * p)
+    row = torch.randint(0, num_src_nodes, size=(num_edges_to_add, ))
+    col = torch.randint(0, num_dst_nodes, size=(num_edges_to_add, ))
+
+    if force_undirected:
```
rusty1s, line 183: I think `force_undirected` makes only sense for homogeneous graphs/in case `num_nodes` is an integer. In that case, we also probably want to only sample half of the edges.

torch_geometric/utils/augmentation.py
EdisonLeeeee: I'm afraid this cannot broadcast `mask` to the shape the same as `x`

rusty1s: Ok, just trying to understand why not?

EdisonLeeeee: In this case
```python
x: shape (N, M)
mask: shape (M)

mask.view(1, -1): shape  (1, M)
mask.tile(x.size(0), 1): shape (N, M)
```

rusty1s: Yes, so that should be broadcastable, right?

EdisonLeeeee: but the returned mask would get the shape (1, M) that doesn't match the shape of `x` (N, M). We should return the mask with the same shape of `x`, make sense?

wsad1: They should be broadcastable according to this [pytorch](https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics) doc.
`masked_fill` works if `x` and `mask` are broadcastable.

EdisonLeeeee: My bad, they are indeed broadcastable. But I still think we should use `mask.tile(x.size(0), 1)` rather than `mask.view(1, -1)` to make `x` and `x` the same shapes. WDYT?

wsad1: For large features with many nodes it might not be the most efficient. Lets maybe update the doc to mention that `mask` is broadcastable with x (and not necessarily the same shape). WDYT?

EdisonLeeeee: Agreed. Didn't take the efficiency into consideration.

EdisonLeeeee:
Shall I update the part "GNN operators and utilities" in README correspondingly?

wsad1:
Yeah it would be nice to mention it there.

wsad1:
Sorry the failing tests are not related to this PR. Will fix them soon.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
+from typing import Optional, Tuple, Union
+
+import torch
+from torch import Tensor
+
+from .num_nodes import maybe_num_nodes
+
+
+def shuffle_node(x: Tensor, training: bool = True) -> Tuple[Tensor, Tensor]:
```
lightaime, line 9: Should we support the `shuffle_node` for batched graphs?

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+    if mode == 'row':
+        mask = torch.rand(x.size(0), device=x.device) >= p
+        mask = mask.view(-1, 1)
+    elif mode == 'col':
+        mask = torch.rand(x.size(1), device=x.device) >= p
+        mask = mask.view(1, -1)
+    else:
+        mask = torch.randn_like(x) >= p
+
+    x = x.masked_fill(~mask, 0.)
```
lightaime, line 114: Let's support `fill_value` in args. Sometimes we may want to fill it with different values such as mean.

lightaime:
```suggestion
    x = x.masked_fill(~mask, fill_value)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+    row = torch.randint(0, num_src_nodes, size=(num_edges_to_add, ))
+    col = torch.randint(0, num_dst_nodes, size=(num_edges_to_add, ))
+
+    if force_undirected:
+        mask = row < col
+        row, col = row[mask], col[mask]
+        row, col = torch.cat([row, col]), torch.cat([col, row])
+    edge_index_to_add = torch.stack([row, col], dim=0).to(device)
+    edge_index = torch.cat([edge_index, edge_index_to_add], dim=1)
+    return edge_index, edge_index_to_add
```
lightaime, line 204: Should we support sorting for the new `edge_index`?
```suggestion
    return sort_edge_index(edge_index) if sort else edge_index, edge_index_to_add
```

lightaime (commented):
Awesome. Like it! Leave some minor comments.

torch_geometric/utils/augmentation.py
rusty1s: That's a great idea. Let's add it as a TODO.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def mask_feature(x: Tensor, p: float = 0.5, mode: str = 'col',
+                 training: bool = True) -> Tuple[Tensor, Tensor]:
+    r"""Randomly masks feature from the feature matrix
+    :obj:`x` with probability :obj:`p` using samples from
+    a Bernoulli distribution.
+
+    The method returns (1) the retained :obj:`x`, (2) the feature
+    mask broadcastable with :obj:`x` (mode='row' and mode='col') or
```
rusty1s, line 50:
```suggestion
    mask broadcastable with :obj:`x` (mode=:obj:`'row'` and mode=:obj:`'col'`) or
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+    with the same shape as :obj:`x` (mode='all'), indicating where features
```
rusty1s, line 51:
```suggestion
    with the same shape as :obj:`x` (mode=:obj:`'all'`), indicating where features
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+    are retained.
+
+    Args:
+        x (FloatTensor): The feature matrix.
+        p (float, optional): The masking ratio. (default: :obj:`0.5`)
+        mode (str, optional): The masked scheme to use for feature masking.
+            (:obj:`"row"`, :obj:`"col"` or :obj:`"all"`). (default: :obj:`col`)
```
rusty1s, line 58:
```suggestion
            (:obj:`"row"`, :obj:`"col"` or :obj:`"all"`). (default: :obj:`"col"`)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+
+    The method returns (1) the retained :obj:`x`, (2) the feature
+    mask broadcastable with :obj:`x` (mode='row' and mode='col') or
+    with the same shape as :obj:`x` (mode='all'), indicating where features
+    are retained.
+
+    Args:
+        x (FloatTensor): The feature matrix.
+        p (float, optional): The masking ratio. (default: :obj:`0.5`)
+        mode (str, optional): The masked scheme to use for feature masking.
```
rusty1s, line 57: Can we add more documentation to `mode`? For example:
If mode=col, will mask entire features of all nodes from the feature matrix. If mode=row, will mask entire nodes from the feature matrix. If mode=all, will mask individual features across all nodes.

torch_geometric/utils/augmentation.py
rusty1s: That is interesting but probably not possible to compute for `mode="all"`, right? In addition, do we need to support re-scaling of features as in the `dropout` case?

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+                    num_nodes: Optional[Union[Tuple[int], int]] = None,
+                    training: bool = True) -> Tuple[Tensor, Tensor]:
+    r"""Randomly adds edges to :obj:`edge_index`.
+
+    The method returns (1) the retained :obj:`edge_index`, (2) the added
+    edge indices.
+
+    Args:
+        edge_index (LongTensor): The edge indices.
+        p (float, optional): Ratio of added edges to the existing edges.
```
rusty1s, line 128: I think we can drop the default value here. `0.2` sounds somewhat arbitrary.

torch_geometric/utils/augmentation.py
rusty1s: I don't think this is necessary. We don't do it for `add_self_loops` either.

torch_geometric/utils/augmentation.py
EdisonLeeeee: Sounds great! I think I can add it later.

torch_geometric/utils/augmentation.py
EdisonLeeeee: Got it.

torch_geometric/utils/augmentation.py
EdisonLeeeee: > That is interesting but probably not possible to compute for mode="all"

We can let `fill_value` be a user-specified value with 0 as default?

> do we need to support re-scaling of features as in the dropout case?

That's not common in literature and I don't think we need to do that, the masked features are simply treated as missing ones.

lightaime: `fill_value` with 0 as default sounds good to me.

Re-scaling also seems great. We can support re-scaling. Maybe set it as `False` by default.

IMO, filling 0 does not work well when the data is not normalized. So some way to keep the mean unchanged could be helpful.

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,204 @@
...
+    assert mode in ['row', 'col', 'all']
+
+    if mode == 'row':
+        mask = torch.rand(x.size(0), device=x.device) >= p
+        mask = mask.view(-1, 1)
+    elif mode == 'col':
+        mask = torch.rand(x.size(1), device=x.device) >= p
+        mask = mask.view(1, -1)
+    else:
+        mask = torch.randn_like(x) >= p
```
lightaime, line 143: Using `Tensor.bernoulli` to generate the mask looks more straightforward to me.

EdisonLeeeee: But it is less efficient than simply using `torch.rand`, as suggested by https://github.com/pyg-team/pytorch_geometric/pull/5481#discussion_r976128424

lightaime: Something like
```suggestion
        mask = x.bernoulli(1-p).to(bool)
```
should be efficient?

EdisonLeeeee: Not really sure about this. @rusty1s could you give a suggestion for us?

rusty1s: When we want to use `torch.bernoulli`, we will create three intermediate tensors:
*  The input (`torch.full`)
* The output floating point tensor
* The mask
 
`torch.randn` only generates two:
* The random vector
* The mask

This optimization might not be necessary, but personally I feel they both read equally well.
Happy to go with `torch.bernoulli` if this improved readability if you strongly prefer it.

lightaime: Not sure if I missed something. I believe we do not need `torch.full` for generating masks.

```
from time import perf_counter
p = 0.2
x = torch.rand(1000, 1000)

t_start = perf_counter()
for _ in range(1000):
  mask = x.bernoulli(1-p).to(torch.bool)
t_stop = perf_counter()
print("Method 1 elapsed time:", t_stop-t_start)

t_start = perf_counter()
for _ in range(1000):
  mask = torch.randn_like(x) >= p
t_stop = perf_counter()
print("Method 2 elapsed time:", t_stop-t_start)
```
```

Method 1 elapsed time: 2.7997943980000173
Method 2 elapsed time: 7.306967894999957
```

It seems using `Tensor.bernoulli` is more efficient.

EdisonLeeeee: Thank you @lightaime. Let's update the code and use `Tensor.bernoulli` instead.

rusty1s: This is not entirely true. You need to move the creation of `x` to the for-loop.

rusty1s: But seems to be still faster (which is weird). Happy to change it then :)

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,237 @@
...
+        >>> node_perm
+        tensor([1, 0, 3, 2])
+    """
+    if not training:
+        perm = torch.arange(x.size(0), device=x.device)
+        return x, perm
+    if batch is None:
+        batch = x.new_zeros(x.size(0), dtype=torch.long)
+    num_nodes = scatter_add(batch.new_ones(x.size(0)), batch, dim=0,
+                            dim_size=int(batch.max()) + 1)
```
rusty1s, line 61:
```suggestion
    num_nodes = scatter_add(batch.new_ones(x.size(0)), batch, dim=0)
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,237 @@
...
+    num_nodes = torch.cat([batch.new_zeros(1), num_nodes])
```
rusty1s, line 62: Is there a `cumsum` missing?

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,237 @@
...
+                [ 9., 10., 11.],
+                [ 6.,  7.,  8.]])
+        >>> node_perm
+        tensor([1, 0, 3, 2])
+    """
+    if not training:
+        perm = torch.arange(x.size(0), device=x.device)
+        return x, perm
+    if batch is None:
+        batch = x.new_zeros(x.size(0), dtype=torch.long)
```
rusty1s, line 59:
```suggestion
    if batch is None:
        perm = torch.randperm(x.size(0), device=x.device)
        return x[perm], perm
```

torch_geometric/utils/augmentation.py
```diff
@@ -0,0 +1,237 @@
...
+    num_nodes = scatter_add(batch.new_ones(x.size(0)), batch, dim=0,
+                            dim_size=int(batch.max()) + 1)
+    num_nodes = torch.cat([batch.new_zeros(1), num_nodes])
+    perm = torch.cat([
+        torch.randperm(n, device=x.device) + offset
```
rusty1s, line 64: Not sure I understand this. I think it needs to be `zip(num_nodes, ptr[:-1])`.

test/utils/test_augmentation.py
```diff
@@ -0,0 +1,101 @@
...
+
+    torch.manual_seed(5)
+    out = shuffle_node(x)
+    assert out[0].tolist() == [[3.0, 4.0, 5.0], [0.0, 1.0, 2.0]]
+    assert out[1].tolist() == [1, 0]
+
+    torch.manual_seed(1)
+    x = torch.tensor([[0, 1, 2], [3, 4, 5], [6, 7, 8], [9, 10, 11]],
+                     dtype=torch.float)
+    batch = torch.tensor([0, 0, 1, 1])
```
rusty1s, line 27: Add a test with 3 graphs per batch instead.

torch_geometric/utils/augmentation.py
EdisonLeeeee: My bad. Let me fix it.

torch_geometric/utils/augmentation.py
EdisonLeeeee: Sorry, what is `prt` here? I think `zip(cumsum[:-1], num_nodes)` in the updated version should work fine.

rusty1s: Yes, `[0] + cumsum == ptr` :)

PR #5563, opened 2022-09-29
Add `torch_geometric.sampler` package to docs

seemanne:
This PR adds the code in torch_geometric/sampler to the docs.  
The section should be added, as the NodeLoader class takes a BaseSampler object in its input but that object is currently not documented in the documentation. This also fixes the currently black link in the NodeLoader documentation.


rusty1s (approved):
Thank you!

PR #5571, opened 2022-09-29
[ONNX] Prevent tensors to be detached during ONNX export

thiagocrepaldi:
`torch.onnx.export` API has some limitations regarding detached tensors during tracing.

The goal of this PR checks skips tensor detaching when ONNX export is in progress. This change does not cause any behavior/performance change when export is not in progress, nor should cause any behavior change to the ONNX exported graph as well

A unit test for SAGEConv ONNX export will be added to PyTorch's repo.

Fix https://github.com/pytorch/pytorch/issues/65138

### DIFF
--- a/torch_geometric/nn/dense/linear.py
+++ b/torch_geometric/nn/dense/linear.py
@@ -145,12 +145,15 @@ class Linear(torch.nn.Module):
         delattr(self, '_hook')
 
     def _save_to_state_dict(self, destination, prefix, keep_vars):
-        if is_uninitialized_parameter(self.weight):
+        if is_uninitialized_parameter(self.weight) or torch.onnx.is_in_onnx_export():
             destination[prefix + 'weight'] = self.weight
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/dense/linear.py
```diff
@@ -145,12 +145,16 @@ def initialize_parameters(self, module, input):
         delattr(self, '_hook')
 
     def _save_to_state_dict(self, destination, prefix, keep_vars):
-        if is_uninitialized_parameter(self.weight):
+        if is_uninitialized_parameter(
+                self.weight) or torch.onnx.is_in_onnx_export():
```
rusty1s, line 149:
```suggestion
        if (is_uninitialized_parameter(
                self.weight) or torch.onnx.is_in_onnx_export()):
```

rusty1s (approved):
Thank you!

PR #5576, opened 2022-09-30
Support for sampling last `k` neighbors as temporal sampling strategy

rusty1s:


torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -101,6 +101,9 @@ class LinkNeighborLoader(LinkLoader):
             replacement. (default: :obj:`False`)
         directed (bool, optional): If set to :obj:`False`, will include all
             edges between all sampled nodes. (default: :obj:`True`)
+        temporal_strategy (string, optional): The sampling strategy when using
```
Padarn, line 104: The PR title say "last_k", is this the same thing?

Perhaps we should add some explanation of the strategies, or link to other documentation (I guess they come from pyg_lib?)

rusty1s: Good suggestion. Added more documentation.

torch_geometric/sampler/utils.py
```diff
@@ -54,8 +54,9 @@ def to_csc(
 
     elif hasattr(data, 'adj_t'):
         if src_node_time is not None:
-            raise NotImplementedError("Temporal sampling via 'SparseTensor' "
-                                      "format not yet supported")
+            pass
+            # raise NotImplementedError("Temporal sampling via 'SparseTensor' "
```
Padarn, line 58: Not totally clear why we want to keep this comment

Padarn (approved):
This looks great, I know a few people who will be excited to see it.

Issue #5172, opened 2022-08-09
Type hints for PyG Data


a-r-j:
### 🚀 The feature, motivation and pitch

Hiya, thanks for maintaining such a great project.

I've often run into scenarios where I have multiple configurations of features used within a project. I think a 'typehinting' scheme (and possibly validation) would be a great addition. I don't think we necessarily need to check for types (I think it's a reasonable assumption that everything will be an appropriate tensor but this could certainly be something to consider) but it would certainly be useful to check whether or not a `Data` object contains the relevant tensors.

This has started making it's way into various DL frameworks ((torch)[https://github.com/patrick-kidger/torchtyping], (JAX)[https://github.com/google/jaxtyping])

It could behave something like:

```python
from torch_geometric.types import Data

def forward(x: Data[edge_index, node_feats, edge_feats]):
    print(x.node_feats)

# Enforce types
from torch_geometric.types import validate

@validate
def forward(x: Data[edge_index, node_feats, edge_feats]):
    print(x.node_feats)
# Raises TypeError if x does not contain x.edge_index, x.node_feats, x.edge_feats
```

I've previously put together something similar for NetworkX: https://github.com/a-r-j/graphtype and would be quite happy to help implement this if you think it would make a good addition.

rusty1s:
Thanks for the issue. Really interesting. Besides defining the attributes of the `data`, it is also possible to define their types? There exists a lot of attributes that get added to a `data` object during batching - do we need to specify these as well?

Can you estimate the amount of work/code change that is needed to integrate?

a-r-j:
Yes, it should be quite easy to define their types. The current impl in the project I linked allows for specification of a subset or the entirety.

E.g this could look like:

```python
@validate
def forward(x: Data[edge_index: torch.Tensor]): # X can only contain the edge_index tensor attr
    pass 

@validate
def forward(x: Data[edge_index: torch.Tensor, ...]): # X has to contain edge_index tensor amongst other attributes
    pass 
```

The solution I'd suggest for PyG would include type checking/hinting attributes considering what you've just said. Within that I think integration with torch typing to specify shapes would be the biggest impact for users (and readability downstream!).

E.g. 


```python
from torchtyping import TensorType

@validate
def forward(x: Data[features: TensorType["nodes" , "n_feats"] ...]): 
   pass
```

[...]

PR #5577, opened 2022-09-30
Add Typehinting and checking for PyG `Data` and `Batch`

a-r-j:
As per #5172 

Includes functionality to use torchtyping to check dimensions of tensors.

Looking for general feedback on implementation and style. Specifically, should we add required attributes for `Batch`es (like `ptr` and `batch`)?


dev.ipynb
```diff
@@ -0,0 +1,243 @@
+{
```
Padarn, line 1: Not sure if you intended to include this, but example in `.py` format or some test would be easier to understand I think

torch_geometric/tp.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class GraphBatch(type(Data), extra=Generic, metaclass=DataMeta):
+    """Defines type Graph to serve as annotation for PyG Data."""
+    __slots__ = ()
+
+    def __new__(cls, *args, **kwds):
+        if not hasattr(cls, "_gorg") or cls._gorg is Graph:
+            raise TypeError("Type Graph cannot be instantiated.")
+        return _generic_new(GraphBatch, cls, *args, **kwds)
```
Padarn, line 172: `_generic_new` not defined?

torch_geometric/tp.py
```diff
@@ -0,0 +1,172 @@
...
+        attributes |= _get_attribute_dtypes(p)[0]
+        dtypes.update(_get_attribute_dtypes(p)[1])
+    else:
+        raise TypeError(
+            "Dataset[col1, col2, ...]: each col must be a string, list or set."
+        )
+    return attributes, dtypes
+
+
+class Graph(type(Data), extra=Generic, metaclass=DataMeta):
```
Padarn, line 155: It seems a bit unfortunate if the type name doesn't refer to `Data`. Maybe it should be `DataT` or something?

Related to this, will it work in an editor to have
```
g: Graph
g....
```
Will I get auto-completion on the functions defined on `Data` for `g`? I tried this pulling your PR and it didn't work but perhaps it was my IDE.

This works for example with numpy's generic NDArray type.

dev.ipynb
rusty1s, line 1: Can you instead of the notebook add a test or a python example?

a-r-j: Added a collection of tests

torch_geometric/tp.py
a-r-j: I'm happy to discuss alternate names. I think reference to `Data` is sensible.

Wrt IDE autocompletion, I don't think this will work in this instance as we're inheriting from `type(Data)` rather than `Data`. I tried to inherit from `Data` directly but was running into some errors that I got stuck on. I agree it's not ideal. FWIW it's possible to use this implementation to inherit from `np.ndarray` (rather than `type(np.ndarray)` so there may be something in the `Data` implementation that's making this an issue.

a-r-j: @Padarn I've renamed to `DataT` and `BatchT`.

I believe I've resolved the autocomplete issue. Let me know if it's still a problem in your IDE.

a-r-j:
@rusty1s @Padarn Looking for some further comments :)

torch_geometric/tp.py
Padarn: Hey sorry for the slow reply been a bit busy: I just checked it out and it seems to be working. Thanks!

Padarn:
Sorry for being a bit slow @a-r-j I've been a bit busy with work. I will review in detail again in the next day or so. 

At a very short glance there are quite a few warnings about cases not covered by tests. I haven't looked to see if these make sense to  test or not though.

setup.py
```diff
@@ -42,10 +42,7 @@
     'wandb',
 ]
 
-test_requires = [
-    'pytest',
-    'pytest-cov',
-]
+test_requires = ['pytest', 'pytest-cov', 'torchtyping']
```
Padarn, line 45:
```suggestion
test_requires = [
    'pytest', 
    'pytest-cov', 
    'torchtyping',
]
```
The typical formatting in PyG (the trailing comma tops it from becoming a list)

torch_geometric/tp.py
```diff
@@ -0,0 +1,220 @@
...
+        attributes |= _get_attribute_dtypes(p)[0]
+        dtypes.update(_get_attribute_dtypes(p)[1])
+    else:
+        raise TypeError("DataT[attr1, attr2, ...]: each attribute must be \
+            a string, list or set.")
+    return attributes, dtypes
+
+
+class DataT(Data, extra=Generic, metaclass=DataMeta):
+    """Defines type DataT to serve as annotation for PyG Data."""
```
Padarn, line 198: Could we maybe have a quick example in the docstring?

torch_geometric/tp.py
```diff
@@ -0,0 +1,220 @@
...
+
+    __slots__ = ()
+    check_only_specified = None
+    attributes = set()
+    dtypes = {}
+
+    def __new__(cls, *args, **kwds):
+        if not hasattr(cls, "_gorg") or cls._gorg is DataT:
+            raise TypeError("Type GraphT cannot be instantiated.")
```
Padarn, line 207: maybe a more informative error? I don't know how anyone would debug this

Padarn:
Sorry I'm so slow: I left a few final minor comments. I feel it might be nice to add this to the docs or in an example, otherwise it might fall under the radar, but its very cool.

Padarn:
(also maybe need to fix your tests)

vincrichard:
Hello, thank you for the work on these features, I think it will be a great addition for better visibility of the library. 
Any reason this PR as left as is ?
I see there are only a few comments left, related to additional details on error message / documentation. 
Maybe I am missing something, but any chance this will be continued ?

a-r-j:
Hi @vincrichard it unfortunately fell down my priority list a little. In principle, i'm open to picking it back up but I think it makes most sense to switch to jaxtyping which may require a little work

a-r-j:
Hi @rusty1s @wsad1 @mananshah99 ,

I finally got around to picking this up again. I think it's in better shape now:

* I switched out the `torchtyping` dependency for it's more modern replacements`jaxtyping` and `beartype` for tensor shape and dtype hinting and enforcing type checking with the `@torch_geometric.data.typecheck()` decorator. N.b. `jaxtyping` is not specific to `jax` and supports type hinting `torch.Tensor`.
* General clean up and improved documentation
* The failing test in the`torch==1.13.0` build seems to be unrelated.
* The failing check in `mypy` also seems to be unrelated.

Would be grateful for a (hopefully!) final round of review.

a-r-j:
For the failing test in the `torch==1.13.0` build, this seems like a plausible mitigation: https://github.com/facebookresearch/fairseq/issues/3395#issuecomment-968256371

Any concerns with me adding this?

a-r-j:
Hi @rusty1s is this feature still interesting to you?

a-r-j:
Hi @rusty1s; friendly bump :)

PR #5587, opened 2022-10-01
Add `assortativity` to `torch_geometric.utils`

EdisonLeeeee:
This PR implements the computation of degree assortativity coefficient, as described in [Mixing patterns in networks](https://arxiv.org/abs/cond-mat/0209450). Assortativity is a preference for a network's nodes to attach to others that are similar in some way.  A larger assortativity indicates nodes tend to be connected with other nodes with similar degree values, and vice versa. Link to [Slack Discussion](https://torchgeometricco.slack.com/archives/C01DN0B3B1N/p1664308511056059)

The implementation is based on `networkx.degree_assortativity_coefficient` but much faster than using `networkx`.

## Test

+ Generate a graph (without duplicated edges)
```python
import torch
from torch_geometric.utils import coalesce

torch.manual_seed(12345)
num_nodes = 10000
num_edges = 1000000
edge_index = torch.randint(0, num_nodes, size=(2, num_edges))
edge_index = coalesce(edge_index)
```
+ `networkx.degree_assortativity_coefficient` 
```python
import networkx as nx
nxg = nx.from_edgelist(edge_index.t().tolist(), create_using=nx.DiGraph)
t = time.perf_counter()
result = nx.degree_assortativity_coefficient(nxg)
print(f"result={result}, time={time.perf_counter()-t}")

>>> result=-0.0007873635931715391, time=0.48579016700000466
```

+ `torch_geometric.utils.assortativity` (On CPU)
```python
from torch_geometric.utils import assortativity
t = time.perf_counter()
result = assortativity(edge_index)
print(f"result={result}, time={time.perf_counter()-t}")

>>> result=-0.0007878749747760594, time=0.07563883299985719
```

+ `torch_geometric.utils.assortativity` (On GPU)
```python
>>> result=-0.0007857615128159523, time=0.0036659350153058767
```

The slight difference in results is caused by two different frameworks (Numpy and PyTorch).


torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
+import torch
+from torch_sparse import SparseTensor
+
+from torch_geometric.typing import Adj
+from torch_geometric.utils import coalesce, degree, to_dense_adj
+
+
+def assortativity(edge_index: Adj) -> float:
+    r"""The degree assortativity coeffeicent from the
```
rusty1s, line 9:
```suggestion
    r"""The degree assortativity coefficient from the
```

torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
...
+    `"Mixing patterns in networks"
+    <https://arxiv.org/abs/cond-mat/0209450>`_ paper
```
rusty1s, line 11:
```suggestion
    <https://arxiv.org/abs/cond-mat/0209450>`_ paper.
```

torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
...
+    device = row.device
+    out_deg = degree(row, dtype=torch.long)
+    in_deg = degree(col, dtype=torch.long)
+    degrees = torch.unique(torch.cat([out_deg, in_deg]))
+    mapping = row.new_zeros(degrees.max().item() + 1)
+    mapping[degrees] = torch.arange(degrees.size(0), device=device)
+
+    # Compute degree mixing matrix (joint probability distribution) `M`
+    num_degrees = degrees.size(0)
+    M = torch.zeros(num_degrees, num_degrees, device=device)
```
rusty1s, line 44: This matrix seems to not be used.

rusty1s (approved):

test/utils/test_assortativity.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_assortativity():
+    torch.manual_seed(12345)
+
+    num_nodes = 100
+    num_edges = 1000
+    edge_index = torch.randint(0, num_nodes, size=(2, num_edges))
+    edge_index = coalesce(edge_index)
+    out = assortativity(edge_index)
+    assert pytest.approx(out, abs=1e-5) == -0.05669079348444939
```
Padarn, line 15: Its not totally clear what this is testing to me? Is it possible for us to have a fixed example where it is clear what that `assortativity` value should be?

torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
...
+from torch_geometric.typing import Adj
+from torch_geometric.utils import coalesce, degree, to_dense_adj
+
+
+def assortativity(edge_index: Adj) -> float:
+    r"""The degree assortativity coeffeicent from the
+    `"Mixing patterns in networks"
+    <https://arxiv.org/abs/cond-mat/0209450>`_ paper
+    Assortativity measures the similarity of connections in
+    the graph with respect to the node degree. It is computed by
```
Padarn, line 13:
```suggestion
    the graph with respect to the node degree. It is computed from
```

Padarn: Does this make more sense?

torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
...
+        row, col, _ = edge_index.coo()
+    else:
+        row, col = edge_index
+
+    device = row.device
+    out_deg = degree(row, dtype=torch.long)
+    in_deg = degree(col, dtype=torch.long)
+    degrees = torch.unique(torch.cat([out_deg, in_deg]))
+    mapping = row.new_zeros(degrees.max().item() + 1)
+    mapping[degrees] = torch.arange(degrees.size(0), device=device)
```
Padarn, line 40: This is not really relevant to your change, but I'm wondering why there isn't a more elegant want to handle this device stuff - like maybe a context manager that changes the default device within a block.

torch_geometric/utils/assortativity.py
```diff
@@ -0,0 +1,65 @@
...
+    # Pearson correlation coefficient of the node degrees
+    x = y = degrees.float()
+    a, b = M.sum(0), M.sum(1)
+
+    vara = (a * x**2).sum() - ((a * x).sum())**2
+    varb = (b * x**2).sum() - ((b * x).sum())**2
+    xy = torch.outer(x, y)
+    ab = torch.outer(a, b)
+    out = (xy * (M - ab)).sum() / (vara * varb).sqrt()
+    return out.item()
```
Padarn, line 65: Sorry stupid question, should we just return a tensor?

torch_geometric/utils/assortativity.py
EdisonLeeeee: Used later to compute row/col sum
```python
a, b = M.sum(0), M.sum(1)
```

torch_geometric/utils/assortativity.py
EdisonLeeeee: Agreed. Maybe we need a context manager similar to `with tf.device(device):` in tensorflow.

test/utils/test_assortativity.py
EdisonLeeeee: Got it. Let me fix it later.

torch_geometric/utils/assortativity.py
EdisonLeeeee: Personally, I like to return a float value as it is more straightforward for users, but returning a tensor looks good to me as well.

torch_geometric/utils/assortativity.py
Padarn: +1 .. but from reading pytorch forums they don't like it much. I might have a play around to see if I can make anything useful

torch_geometric/utils/assortativity.py
Padarn: No we can leave, I just wondered if there was a reason I didn't understand

torch_geometric/utils/assortativity.py
rusty1s: Not sure I understand. You re-assign it in line 51.

rusty1s (approved):
Looks good to me. Please also add it to the `CHANGELOG.md`.

torch_geometric/utils/assortativity.py
EdisonLeeeee: Oh I missed it. Now it has been fixed (removed).

PR #5598, opened 2022-10-03
Fix C++ example compilation by linking cusparse

geekLucian:
I followed instructions in [examples/cpp/README.md](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/cpp/README.md) to build C++ example, and I encountered errors as below:

```bash
$ cmake --build .
[ 50%] Building CXX object CMakeFiles/hello-world.dir/main.cpp.o
[100%] Linking CXX executable hello-world
/usr/bin/ld: /usr/local/lib/libtorchsparse.so: undefined reference to `cusparseDcsrgemm2_bufferSizeExt'
/usr/bin/ld: /usr/local/lib/libtorchsparse.so: undefined reference to `cusparseXcsrgemm2Nnz'
/usr/bin/ld: /usr/local/lib/libtorchsparse.so: undefined reference to `cusparseScsrgemm2_bufferSizeExt'
/usr/bin/ld: /usr/local/lib/libtorchsparse.so: undefined reference to `cusparseDcsrgemm2'
/usr/bin/ld: /usr/local/lib/libtorchsparse.so: undefined reference to `cusparseScsrgemm2'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/hello-world.dir/build.make:114: hello-world] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/hello-world.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

It seems that the linker cannot find lib `cusparse`. Adding `cusparse` in target link libraries fix this for me.


rusty1s:
Thanks for this PR! I am wondering why we need this here though. Shouldn't this be moved to the `cmake` file of `pytorch_sparse`?

geekLucian:
`cusparse` is the dynamic library needed by `pytorch_sparse`. If we move this fix this to `pytorch_sparse`, I cannot come up with a solution other than statically link `cusparse` into `pytorch_sparse`, but this would cause more problems (such as larger `libtorchsparse` size and versioning issue of `cusparse`). IMO, fixing this at where we compile the target applications prevails.

rusty1s:
Makes sense. Not sure why it worked for me before.

rusty1s (approved):

PR #5602, opened 2022-10-04
Fix `LinkNeighborLoader` with sorted timestamps

rusty1s:
`pyg-lib.neighbor_sample` introduced the `seed_time` argument, which allows us to override the timestamp for the given set of seed nodes. This PR introduces the change to PyG, allowing link neighbor loader tests to pass again.

cc @wsad1

### DIFF
--- a/test/loader/test_link_neighbor_loader.py
+++ b/test/loader/test_link_neighbor_loader.py
@@ -204,7 +204,6 @@ def test_temporal_heterogeneous_link_neighbor_loader():
 
     # With edge_time:
     edge_time = torch.arange(data['paper', 'paper'].edge_index.size(1))
-    paper_time_original = data['paper'].time.clone()
     loader = LinkNeighborLoader(
         data,
         num_neighbors=[-1] * 2,
@@ -221,8 +220,6 @@ def test_temporal_heterogeneous_link_neighbor_loader():
         author_min = batch['author'].time.min()
         edge_min = batch['paper', 'paper'].edge_label_time.min()
         assert edge_min >= author_min
-        assert author_min >= 0
-        assert torch.allclose(data['paper'].time, paper_time_original)
 
 
 @pytest.mark.parametrize('FeatureStore', [MyFeatureStore, HeteroData])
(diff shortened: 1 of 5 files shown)


torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -234,7 +234,8 @@ def _sample(
                     self.row_dict,
                     seed,  # seed_dict
                     self.num_neighbors,
-                    kwargs.get('node_time_dict', self.node_time_dict),
+                    self.node_time_dict,
+                    kwargs.get('seed_time_dict', None),
```
wsad1, line 238: lets make `seed_time_dict` an argument to `_sample`.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -390,28 +379,39 @@ def sample_from_edges(
                     self.input_type[0]: query_src,
                     self.input_type[-1]: query_dst,
                 }
+                if edge_label_time is not None:
+                    seed_time_dict = {
+                        self.input_type[0]:
+                        scatter_min(edge_label_time, reverse_src)[0],
```
wsad1, line 385: Nitpick: lets call it `inverse_src` instead of `reverse_src`

test/loader/test_link_neighbor_loader.py
```diff
@@ -194,18 +194,26 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
+            data,
+            num_neighbors=[-1] * 2,
+            edge_label_index=('paper', 'paper'),
+            batch_size=32,
+            time_attr='time',
+        )
 
     # With edge_time:
     edge_time = torch.arange(data['paper', 'paper'].edge_index.size(1))
     paper_time_original = data['paper'].time.clone()
```
wsad1, line 207: don't need this variable anymore.

test/loader/test_link_neighbor_loader.py
```diff
@@ -194,18 +194,26 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
+    loader = LinkNeighborLoader(
+        data,
+        num_neighbors=[-1] * 2,
+        edge_label_index=('paper', 'paper'),
+        edge_label_time=edge_time,
+        batch_size=32,
+        time_attr='time',
+        neg_sampling_ratio=0.5,
+    )
     for batch in loader:
```
wsad1, line 217: suprised that the `author_min>=0` test is passing, as all authors have a negative timestamp.

wsad1 (approved):
Thanks for fixing this cleanly. looks good, left some comments.

test/loader/test_link_neighbor_loader.py
rusty1s: It looks like only `paper` nodes have negative timestamp.

PR #5621, opened 2022-10-07
Fix typo in `aggr/base.py`.

BuddhiWathsala:
Since it has index values with 0, 1, and 2, the output shape should be [3, 64]. Running this example code will give you [3, 64] output.


wsad1 (approved):
Thanks for fixing.

PR #5627, opened 2022-10-07
Support `in_channels` with `tuple` in `GENConv` for bipartite message passing

lightaime:
This PR supports passing `in_channels` with `tuple` to `GENConv` for bipartite message passing by projecting channels to the same dimension if channels mismatch.

### DIFF
--- a/torch_geometric/nn/conv/gen_conv.py
+++ b/torch_geometric/nn/conv/gen_conv.py
@@ -145,11 +145,11 @@ class GENConv(MessagePassing):
 
         channels = [in_channels[0]]
         for i in range(num_layers - 1):
-            channels.append(in_channels[0] * 2)
+            channels.append(out_channels * 2)
         channels.append(out_channels)
         self.mlp = MLP(channels, norm=norm)
 
-        if in_channels[0] != in_channels[1]:
+        if in_channels[0] != in_channels[1] and in_channels[0] != -1:
             self.lin_r = Linear(in_channels[1], in_channels[0], bias=True)
 
         if msg_norm:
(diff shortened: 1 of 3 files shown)


torch_geometric/nn/conv/gen_conv.py
```diff
@@ -62,8 +62,8 @@ class GENConv(MessagePassing):
         ogbn_proteins_deepgcn.py>`_.
 
     Args:
-        in_channels (int): Size of each input sample, or :obj:`-1` to derive
-            the size from the first input(s) to the forward method.
+        in_channels (int or tuple): Size of each input sample, or :obj:`-1` to
+            derive the size from the first input(s) to the forward method.
```
rusty1s, line 66:
```suggestion
        in_channels (int or tuple): Size of each input sample, or :obj:`-1` to
            derive the size from the first input(s) to the forward method.
            A tuple corresponds to the sizes of source and target
            dimensionalities.
```

torch_geometric/nn/conv/gen_conv.py
```diff
@@ -101,41 +101,65 @@ class GENConv(MessagePassing):
...
+
+        channels = [in_channels[0]]
         for i in range(num_layers - 1):
-            channels.append(in_channels * 2)
+            channels.append(in_channels[0] * 2)
         channels.append(out_channels)
         self.mlp = MLP(channels, norm=norm)
 
-        self.msg_norm = MessageNorm(learn_msg_scale) if msg_norm else None
+        if in_channels[0] != in_channels[1]:
```
rusty1s, line 150: and `in_channels=[0] != -1?

rusty1s (approved):
Can we add some corresponding tests?

torch_geometric/nn/conv/gen_conv.py
lightaime: Good point!

lightaime:
> Can we add some corresponding tests?

Sure. Will update it later.

Issue #5581, opened 2022-09-30
RandomLinkSplit gives no `neg_edge_label` on dense graphs


andreimargeloiu:
### 🐛 Describe the bug

**Context:** I'm programmatically creating undirected graphs for training a Graph Autoencoder (GAE). Some graphs are very dense, containing 90% of all possible edges. 

Training the GAE requires sampling missing edges as follows:
```T.RandomLinkSplit(num_val=0.15, num_test=0, is_undirected=True, split_labels=True, add_negative_train_samples=True, neg_sampling_ratio=1.0)```

**Issue:** Because the graphs are too dense, there aren't enough missing edges to sample negative edges. Consider a graph with 100 nodes and 9000 edges (from a maximum of 9900 edges). There are only 900 missing edges to sample as negative edges, but because I'm setting `neg_sampling_ratio=1.0`, PyG expected to find 9000 missing edges.

**The issue is that PyG doesn't check there aren't enough missing edges, and it creates training datasets without setting `neg_edge_label`.** When creating the [training data](https://github.com/pyg-team/pytorch_geometric/blob/d8800b64ce42a8d0f8ab3a1cf22b935d445a826c/torch_geometric/transforms/random_link_split.py#L217), the current code doesn't assign any `neg_edge_index` to the training data. See below the train/valid/test data.

```
(Data(x=[141, 141], edge_index=[2, 16072], pos_edge_label=[8036], pos_edge_label_index=[2, 8036]), # Train; notice no negative edges
 Data(x=[141, 141], edge_index=[2, 16072], pos_edge_label=[1417], pos_edge_label_index=[2, 1417], neg_edge_label=[834], neg_edge_label_index=[2, 834]), # Validation
 Data(x=[141, 141], edge_index=[2, 18906], pos_edge_label=[0], pos_edge_label_index=[2, 0])) # Test
```

**Error:** When creating a dataloader comes the `KeyError: 'neg_edge_label'` because the training data is missing this key.

```
gnn_dataloader = tg.loader.DataLoader(graphs_dataset, batch_size=len(graphs_dataset), shuffle=True)
gnn_batch_train, _, _ = next(iter(gnn_dataloader))
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_175984/1912375602.py in <module>
[...]

PR #5642, opened 2022-10-10
Add ratio adjustment in `RandomLinkSplit` when not enough negative edges exist

Padarn:
This change addresses https://github.com/pyg-team/pytorch_geometric/issues/5581

The problem being solved is that it is possible for there to not be enough negative edges available to satisfy the requested sampling ratio.

This change makes it so that in this case the ratio assigned to each split should reflect the desired ratio between the splits. I.e, the % assigned to each of train, test and validation is the same as requested.

### DIFF
--- a/test/transforms/test_random_link_split.py
+++ b/test/transforms/test_random_link_split.py
@@ -205,6 +205,6 @@ def test_random_link_split_insufficient_negative_edges():
                                 is_undirected=False, neg_sampling_ratio=2,
                                 split_labels=True)
     train_data, val_data, test_data = transform(data)
-    assert len(train_data.neg_edge_label) == 2
-    assert len(val_data.neg_edge_label) == 2
-    assert len(test_data.neg_edge_label) == 2
+    assert train_data.neg_edge_label_index.size() == (2, 2)
+    assert val_data.neg_edge_label_index.size() == (2, 2)
+    assert test_data.neg_edge_label_index.size() == (2, 2)
--- a/test/transforms/test_random_link_split.py
+++ b/test/transforms/test_random_link_split.py
@@ -199,12 +199,14 @@ def test_random_link_split_on_undirected_hetero_data():
 
 
 def test_random_link_split_insufficient_negative_edges():
-    data = Data()
-    data.edge_index = torch.tensor([[0, 0, 1, 1, 2, 2], [1, 3, 0, 2, 0, 1]])
+    edge_index = torch.tensor([[0, 0, 1, 1, 2, 2], [1, 3, 0, 2, 0, 1]])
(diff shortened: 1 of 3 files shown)


torch_geometric/transforms/random_link_split.py
```diff
@@ -208,6 +209,18 @@ def __call__(self, data: Union[Data, HeteroData]):
...
                                                method='sparse')
 
+            # Adjust ratio if not enough negative edges exist
+            if len(neg_edge_index) < num_neg:
+                num_neg_found = neg_edge_index.size(1)
+                ratio = num_neg_found / num_neg
+                warnings.warn(
+                    f"There are not enough negative edges to satisfy "
+                    "the provided sampling ratio. The ratio will be "
+                    f"adjusted to {ratio}")
```
rusty1s, line 219:
```suggestion
                    f"adjusted to {ratio:.3f}")
```

torch_geometric/transforms/random_link_split.py
```diff
@@ -208,6 +209,18 @@ def __call__(self, data: Union[Data, HeteroData]):
                                                num_neg_samples=num_neg,
                                                method='sparse')
 
+            # Adjust ratio if not enough negative edges exist
+            if len(neg_edge_index) < num_neg:
```
rusty1s, line 213:
```suggestion
            if neg_edge_index.size(1) < num_neg:
```

test/transforms/test_random_link_split.py
```diff
@@ -196,3 +196,16 @@ def test_random_link_split_on_undirected_hetero_data():
...
+    data = Data()
+    data.edge_index = torch.tensor([[0, 0, 0, 1, 1, 2, 2],
+                                    [1, 2, 3, 0, 2, 0, 1]])
+    transform = RandomLinkSplit(num_val=0.33, num_test=0.33,
+                                is_undirected=False, neg_sampling_ratio=2,
+                                split_labels=True)
+    train_data, val_data, test_data = transform(data)
+    assert len(train_data.neg_edge_label) != 0
+    assert len(val_data.neg_edge_label) != 0
+    assert len(test_data.neg_edge_label) != 0
```
rusty1s, line 211: Can we check here for the actual returned number of edges? I guess it may be
```python
assert len(train_data.neg_edge_label) == 1
assert len(val_data.neg_edge_label) == 1
assert len(test_data.neg_edge_label) == 0
```

rusty1s (approved):
Thank you!

test/transforms/test_random_link_split.py
Padarn: They're all non-zero, but yes we can check the exact size, I'll add that.

PR #5667, opened 2022-10-12
[TypeHints] `OneHotDegree`

sebastian-montero:


sebastian-montero:
TorchScript still doesn't support variable list of arguments with *args
```
E           torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults:
E             File "/Users/sebastianmontero/Desktop/pytorch_geometric/torch_geometric/data/data.py", line 735
E               def __call__(self, *args: List[str]) -> Iterable:
E                                  ~~~~~ <--- HERE
E                   r"""Iterates over all attributes :obj:`*args` in the data, yielding
E                   their attribute names and values.
```

torch_geometric/transforms/one_hot_degree.py
```diff
@@ -19,12 +22,14 @@ class OneHotDegree(BaseTransform):
...
-    def __init__(self, max_degree, in_degree=False, cat=True):
+    def __init__(self, max_degree: int, in_degree: bool = False,
+                 cat: bool = True):
         self.max_degree = max_degree
         self.in_degree = in_degree
         self.cat = cat
 
-    def __call__(self, data):
+    def __call__(self, data: Union[Data,
+                                   HeteroData]) -> Union[Data, HeteroData]:
```
EdisonLeeeee, line 32: `HeteroData` is not supported for this method.

rusty1s, line 32:
```suggestion
    def __call__(self, data: Data) -> Data:
```

rusty1s (approved):
Thank you!

PR #5668, opened 2022-10-12
[TypeHints] `ToDense`

sebastian-montero:


sebastian-montero:
Note that TorchScript still doesn't support variable number of arguments `*args`
```
E           torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults:
E             File "/Users/sebastianmontero/Desktop/pytorch_geometric/torch_geometric/data/data.py", line 735
E               def __call__(self, *args: List[str]) -> Iterable:
E                                  ~~~~~ <--- HERE
E                   r"""Iterates over all attributes :obj:`*args` in the data, yielding
E                   their attribute names and values.
```

sebastian-montero:
TorchScript still doesn't support variable list of arguments with `*args`
```
E           torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults:
E             File "/Users/sebastianmontero/Desktop/pytorch_geometric/torch_geometric/data/data.py", line 735
E               def __call__(self, *args: List[str]) -> Iterable:
E                                  ~~~~~ <--- HERE
E                   r"""Iterates over all attributes :obj:`*args` in the data, yielding
E                   their attribute names and values.
```

rusty1s (approved):
Thank you!

PR #5684, opened 2022-10-13
[Type Hints] `nn.InstanceNorm`

wwymak:
The test for nn.InstanceNorm already has a test for torchscript support, so I have not changed that file


torch_geometric/nn/norm/instance_norm.py
```diff
@@ -36,8 +36,9 @@ class InstanceNorm(_InstanceNorm):
             uses instance statistics in both training and eval modes.
             (default: :obj:`False`)
     """
-    def __init__(self, in_channels, eps=1e-5, momentum=0.1, affine=False,
-                 track_running_stats=False):
+    def __init__(self, in_channels: int, eps: float = 1e-5,
+                 momentum: float = 0.1, affine: bool = False,
+                 track_running_stats: bool = False):
```
rusty1s, line 41:
```suggestion
                 track_running_stats: bool = False, ):
```

rusty1s (approved):
Thank you!

PR #5697, opened 2022-10-14
Fix `pyg-lib` `hetero_neighbor_sampler` usage

mszarma:


rusty1s (approved):
Wow, missed that. Thanks for fixing :)

rusty1s:
It looks like tests will run through once the nightly version has been built. Will wait for that.

PR #5700, opened 2022-10-14
Add `node_types` argument to `transforms.Constant`

EdisonLeeeee:
~~Add `AddNodeFeature` transforms for `Data`. Currently it supports adding `torch.Tensor` or add generated tensor in different ways `(torch.ones, torch.zeros, torch.randn)`.~~

TODO:

- [x] ~~Refactor `AddLaplacianEigenvectorPE` and `AddRandomWalkPE` with `AddNodeFeature`~~
- [x] Add `HeteroData` support and refactor the examples `hetero_conv_dblp.py` and `hgt_dblp.py` with transformation`Constant `:
```diff
- data['conference'].x = torch.ones(data['conference'].num_nodes, 1)
+ data = Constant(node_types='conference')(data)
```
- [x] Add test
- [x] Update doc-string
- [x] CHANGELOG

I'm not sure that `AddNodeFeature` or `AddNodeAttr` which one is better?


torch_geometric/transforms/add_node_feature.py
```diff
@@ -0,0 +1,56 @@
...
+import torch
+from torch import Tensor
+
+from torch_geometric.data import Data, HeteroData
+from torch_geometric.data.datapipes import functional_transform
+from torch_geometric.transforms import BaseTransform
+
+
+@functional_transform('add_node_feature')
+class AddNodeFeature(BaseTransform):
```
rusty1s, line 12: `Add` is a bit confusing since we are concatenating features instead of summing them up.

torch_geometric/transforms/add_node_feature.py
```diff
@@ -0,0 +1,56 @@
...
+    r"""Adds the node feature to :obj:`attr`
+    (functional name: :obj:`add_node_feature`).
+
+    Args:
+        value (str, Tensor, optional): The value of the attribute to add.
```
rusty1s, line 17: What's the use-case of `value == Tensor`? Since transforms should be data-agnostic (i.e., independent from the number of nodes), how would this be used?

torch_geometric/transforms/add_node_feature.py
rusty1s, line 12: This feels quite related to `transforms.Constant`. Can we unify them?

torch_geometric/transforms/add_node_feature.py
EdisonLeeeee: `value == Tensor` is to refactor `AddLaplacianEigenvectorPE` and `AddRandomWalkPE` and address the TODO in `add_node_attr`. I'm also not sure this is indeed necessary.

torch_geometric/transforms/add_positional_encoding.py
```diff
@@ -13,20 +13,7 @@
     to_scipy_sparse_matrix,
 )
 
-
-def add_node_attr(data: Data, value: Any,
-                  attr_name: Optional[str] = None) -> Data:
-    # TODO Move to `BaseTransform`.
```
EdisonLeeeee, line 19: To replace this function with `AddNodeFeature`.

torch_geometric/transforms/add_node_feature.py
EdisonLeeeee: Sounds great. But `Constant` does not support adding random features, as introduced as a trick in [Random Features Strengthen Graph Neural Networks](https://arxiv.org/abs/2002.03155).

rusty1s: But that probably deserves its own transform anyway (with a proper cite to the paper). WDYT?

torch_geometric/transforms/add_node_feature.py
rusty1s: I think the idea would be more to have this as a utility function in `BaseTransform` so that every transform can make use of it. WDYT?

EdisonLeeeee: Agreed.

torch_geometric/transforms/add_node_feature.py
EdisonLeeeee: Agreed. I'm gonna unify them into `transforms.Constant`.

EdisonLeeeee: @rusty1s Will that make sense now?

torch_geometric/transforms/constant.py
```diff
@@ -14,15 +14,28 @@ class Constant(BaseTransform):
...
         self.value = value
         self.cat = cat
-
-    def __call__(self, data: Union[Data, HeteroData]):
-        for store in data.node_stores:
+        self.node_types = node_types
+
+    def __call__(self, data: Union[Data,
+                                   HeteroData]) -> Union[Data, HeteroData]:
+        if isinstance(data, HeteroData) and self.node_types is not None:
```
Padarn, line 32: should we just assert `self.node_types` is not None if using HeteroData?

EdisonLeeeee: Constants will be added to all node features if  `self.node_types=None`. This is to be consistent with the previous implementation where constants are added to all existing features by default. Or perhaps we should assert `self.node_types` is None if using Data?

Padarn: okay makes sense 👍

rusty1s (approved):
Thank you! This is cool :)

PR #5716, opened 2022-10-15
[Type Hints] `nn.GNNExplainer`

sebastian-montero:
@rusty1s Please note that I am using `torch.jit.export()` instead of `torch.jit.script()` for the tests. Can you please confirm if this is ok?


torch_geometric/nn/models/gnn_explainer.py
```diff
@@ -97,7 +112,12 @@ def _clear_masks(self):
         self.node_feat_masks = None
         self.edge_mask = None
 
-    def _loss(self, log_logits, prediction, node_idx: Optional[int] = None):
+    def _loss(
+        self,
+        log_logits: Tensor,
+        prediction: Tensor,
+        node_idx: Optional[Tensor] = None,
+    ):
```
rusty1s, line 120:
```suggestion
    ) -> Tensor:
```

rusty1s (commented):
`torch.jit.export` works here. Can we also add type hints to the function inside `GNNExplainer`?

sebastian-montero:
> 

Hi @rusty1s, I think I already did. I added type hints to all methods in the class `GNNExplainer` here `torch_geometric/nn/models/gnn_explainer.py` 

Am I missing something?

rusty1s:
Ah, sorry. Overlooked that single file, my bad.

sebastian-montero:
> Ah, sorry. Overlooked that single file, my bad.

No problem! I'll resolve the Changelog conflicts and this should be ready to merge.

rusty1s (approved):

PR #5717, opened 2022-10-15
`LinkNeighborLoader`: Disjoint sampling for links

wsad1:
With this PR each seed edge in `LinkNeigbhorLoader` will have a disjoint subgraph. Similar to how each node has a disjoint subgraph in `NeighborLoader`.


test/loader/test_link_neighbor_loader.py
```diff
@@ -206,16 +206,17 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
-        edge_label_time=edge_time,
-        batch_size=32,
-        time_attr='time',
-        neg_sampling_ratio=0.5,
-    )
+    loader = LinkNeighborLoader(data, num_neighbors=[-1] * 2,
+                                edge_label_index=('paper', 'paper'),
+                                edge_label_time=edge_time, batch_size=32,
+                                time_attr='time', neg_sampling_ratio=0.5,
+                                drop_last=True)
```
rusty1s, line 213:
```suggestion
                                drop_last=True, )
```

test/loader/test_link_neighbor_loader.py
```diff
@@ -206,16 +206,17 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
     for batch in loader:
+        # check if each seed edge has a different batch
+        assert batch['paper'].batch.max() == 47
```
rusty1s, line 216: I guess this happens due to `batch_size==32` and `neg_sampling_ratio == 0.5`, right? Can we add a comment for this and check via `batch.max() <= 32 + 16`?

test/loader/test_link_neighbor_loader.py
```diff
@@ -206,16 +206,17 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
+        # check if each seed edge has a different source
+        # and destination node
+        assert len(batch['paper'].x) > 96
```
rusty1s, line 219:
```suggestion
        assert batch['paper'].num_nodes >= 96
```

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -207,6 +208,15 @@ def _set_num_neighbors_and_num_hops(self, num_neighbors):
...
                                  f"{self.num_hops} entries (got {len(value)})")
 
+    @property
+    def disjoint_sampling(self) -> bool:
+        """True if nodes have time attribute. If `True`
+        each seed node will have a disjoint subgraph"""
+        if hasattr(self, 'node_time_dict'):  # hetero graph
+            return self.node_time_dict is not None
+        else:  # homogenous graph
+            return self.node_time is not None
```
rusty1s, line 218:
```suggestion
        return (getattr(self, 'node_time') is not None or getattr(self, 'node_time_dict') is not None)
```

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -365,57 +373,94 @@ def sample_from_edges(
...
+                    seed_dst = edge_label_index[1]
+                    seed_dst, inverse_dst = seed_dst.unique(
+                        return_inverse=True)
+                    edge_label_index = torch.stack([inverse_src, inverse_dst],
+                                                   0)
+                    seed_dict = {
+                        self.input_type[0]: seed_src,
+                        self.input_type[-1]: seed_dst,
+                    }
+                    if edge_label_time is not None:
```
rusty1s, line 408: Can be dropped.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -365,57 +373,94 @@ def sample_from_edges(
...
-                if edge_label_time is not None:
-                    tmp = torch.cat([edge_label_time, edge_label_time])
-                    seed_time_dict = {
-                        self.input_type[0]: scatter_min(tmp, inverse)[0]
-                    }
+                if self.disjoint_sampling:
+                    seed_nodes = edge_label_index.view(-1)
+                    edge_label_index = torch.arange(0,
+                                                    2 * num_seed_edges).view(
+                                                        2, -1)
```
rusty1s, line 421: Don't we need a `repeat` here as well?

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -365,57 +373,94 @@ def sample_from_edges(
...
+                    if edge_label_time is not None:
+                        tmp = torch.cat([edge_label_time, edge_label_time])
+                        seed_time_dict = {self.input_type[0]: tmp}
+                else:
+                    seed_nodes = edge_label_index.view(-1)
+                    seed_nodes, inverse = seed_nodes.unique(
+                        return_inverse=True)
+                    edge_label_index = inverse.view(2, -1)
+                    seed_dict = {self.input_type[0]: seed_nodes}
+                    if edge_label_time is not None:
```
rusty1s, line 432: Drop.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -365,57 +373,94 @@ def sample_from_edges(
...
             output.metadata = (edge_label_index, edge_label, edge_label_time)
 
         elif issubclass(self.data_cls, Data):
             assert self.node_time is None  # TODO
-            seed_nodes = edge_label_index.view(-1)
-            seed_nodes, inverse = seed_nodes.unique(return_inverse=True)
-            edge_label_index = inverse.view(2, -1)
+            if self.disjoint_sampling:
+                seed_nodes = edge_label_index.view(-1)
+                edge_label_index = torch.arange(0, 2 * num_seed_edges).view(
```
rusty1s, line 452: Here again. Missing `repeat`?

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -365,57 +373,94 @@ def sample_from_edges(
...
                 seed_time_dict=seed_time_dict,
             )
 
+            if self.disjoint_sampling:
+                output.batch = update_batch_for_sample_edge(
+                    output.batch, num_seed_edges)
             output.metadata = (edge_label_index, edge_label, edge_label_time)
 
         elif issubclass(self.data_cls, Data):
             assert self.node_time is None  # TODO
```
rusty1s, line 449: I think we could support this now easily? Let's drop the `assert` and later pass `self._sample(seed=seed_nodes, seed_time=self.node_time)`. WDYT?

torch_geometric/sampler/utils.py
```diff
@@ -149,6 +149,32 @@ def add_negative_samples(
...
+    This function modifies the batch id so that a subgraph of
+    a seed edge (`src`, `dst`) has same batch id.
+    Args:
+        batch: batch vector or dict assiging each
+            node to a seed node.
+        num_edges: Number of seed edges for sampling.
+
+    Retruns updated batch values.
+    """
+    if isinstance(batch, Dict):
```
rusty1s, line 168:
```suggestion
    if isinstance(batch, dict):
```

torch_geometric/sampler/utils.py
```diff
@@ -149,6 +149,32 @@ def add_negative_samples(
...
+        for key, value in batch.items():
+            batch[key] = torch.where(value >= num_seed_edges,
+                                     value - num_seed_edges, value)
```
rusty1s, line 171:
```suggestion
            batch[key] = value % num_seed_edges
```

torch_geometric/sampler/utils.py
```diff
@@ -149,6 +149,32 @@ def add_negative_samples(
...
+    else:
+        batch = torch.where(batch >= num_seed_edges, batch - num_seed_edges,
+                            batch)
```
rusty1s, line 174:
```suggestion
        batch = batch % num_seed_edges
```

torch_geometric/sampler/neighbor_sampler.py
wsad1: No because we use `2*num_seed_edges`. The dst nodes have id from `(0,num_seed_edges-1)` and the src nodes of the same type have ids `(num_seed_edges, 2*num_seed_edges-1)`

torch_geometric/sampler/neighbor_sampler.py
wsad1: You mean support batch dict when we don't do disjoint sampling. Do we even want to support that, because the batch dict is meaningless.

test/loader/test_link_neighbor_loader.py
wsad1: good point, updated the assert stmt.

torch_geometric/sampler/neighbor_sampler.py
rusty1s: No, I mean temporal sampling in homogeneous graphs.

rusty1s (approved):

PR #5724, opened 2022-10-15
[Type Hints] utils.grid

nihal-rao:
This is a draft PR, as the jit scripting test in test_grid.py is failing.
The failure is as follows

### DIFF
--- a/test/utils/test_grid.py
+++ b/test/utils/test_grid.py
@@ -1,3 +1,6 @@
+import torch
+
+from torch_geometric.testing import is_full_test
 from torch_geometric.utils import grid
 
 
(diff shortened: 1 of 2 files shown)


EdisonLeeeee:
Replacing
https://github.com/pyg-team/pytorch_geometric/blob/907d2280fb27f8c1680f882b35b5563d8fc45672/torch_geometric/utils/coalesce.py#L77-L80
with
```python
if edge_attr is not None:
    if isinstance(edge_attr, Tensor):
        edge_attr = edge_attr[perm]
    else:
        edge_attr = [e[perm] for e in edge_attr]
```
should address the error.

rusty1s (approved):
I fixed this by adding TorchScript support to `coalesce` via `overload` decorators :)

PR #5738, opened 2022-10-16
[Type Hints] `nn.EdgePooling`

shhs29:
This is a draft pull request.

A few points to confirm:

1. What type hint can we use for unpooling_description ? Using `NamedTuple` seems to be not working.
2. In the test for unpool, I have created a jit object for only the unpool function and not the forward function. Is this right?
3. `compute_edge_score_tanh` and `compute_edge_score_sigmoid` have parameters `edge_index` and `num_nodes` which are not used. Should these be removed ?


rusty1s:
1. This was indeed much harder than I expected. I took a look as while and could resolve it by defining a `UnpoolInfo(NamedTuple)` class.
2. Also added support for `forward`
3. I made these optional.

rusty1s (approved):

shhs29:
Thanks a lot for the help! 
However, I had a question regarding `compute_edge_score_tanh` and `compute_edge_score_sigmoid` parameters `edge_index` and `num_nodes`. These are not used at all times, so should it still be Optional or removed completely ?

rusty1s:
I made these optional so they don't need to be passed when calling this function explicitly. However, for the sake of the interface, it makes sense to keep them as we can simply call all functions using the same set of arguments.

shhs29:
That makes sense. Thanks a lot for the quick response.

PR #5757, opened 2022-10-17
[Type Hints] `nn.DiffGroupNorm`

akash-vartak:
…roup_distance_ratio


torch_geometric/nn/norm/diff_group_norm.py
```diff
@@ -40,8 +42,11 @@ class DiffGroupNorm(torch.nn.Module):
             uses batch statistics in both training and eval modes.
             (default: :obj:`True`)
     """
-    def __init__(self, in_channels, groups, lamda=0.01, eps=1e-5, momentum=0.1,
-                 affine=True, track_running_stats=True):
+    def __init__(self, in_channels: int, groups: int,
+                 lamda: Optional[float] = 0.01, eps: Optional[float] = 1e-5,
+                 momentum: Optional[float] = 0.1,
+                 affine: Optional[bool] = True,
+                 track_running_stats: Optional[bool] = True):
```
rusty1s, line 49:
```suggestion
    def __init__(self, in_channels: int, groups: int,
                 lamda: float = 0.01, eps: float = 1e-5,
                 momentum: float = 0.1,
                 affine: bool = True,
                 track_running_stats: bool = True, ):
```

torch_geometric/nn/norm/diff_group_norm.py
```diff
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import torch
```
rusty1s, line 3:
```suggestion
import torch
```

rusty1s (approved):

PR #5758, opened 2022-10-17
[Type Hints] `nn.MetaLayer`

camillepradel:


torch_geometric/nn/meta.py
```diff
@@ -104,9 +106,10 @@ def reset_parameters(self):
                 item.reset_parameters()
 
     def forward(
-            self, x: Tensor, edge_index: Tensor,
-            edge_attr: Optional[Tensor] = None, u: Optional[Tensor] = None,
-            batch: Optional[Tensor] = None) -> Tuple[Tensor, Tensor, Tensor]:
+        self, x: Tensor, edge_index: Tensor,
+        edge_attr: Optional[Tensor] = None, u: Optional[Tensor] = None,
+        batch: Optional[Tensor] = None
+    ) -> Tuple[Optional[Tensor], Optional[Tensor], Optional[Tensor]]:
```
rusty1s, line 112:
```suggestion
    ) -> Tuple[Tensor, Optional[Tensor], Optional[Tensor]]:
```

torch_geometric/nn/meta.py
```diff
@@ -104,9 +106,10 @@ def reset_parameters(self):
                 item.reset_parameters()
 
     def forward(
-            self, x: Tensor, edge_index: Tensor,
-            edge_attr: Optional[Tensor] = None, u: Optional[Tensor] = None,
-            batch: Optional[Tensor] = None) -> Tuple[Tensor, Tensor, Tensor]:
+        self, x: Tensor, edge_index: Tensor,
+        edge_attr: Optional[Tensor] = None, u: Optional[Tensor] = None,
+        batch: Optional[Tensor] = None
```
rusty1s, line 111:
```suggestion
        batch: Optional[Tensor] = None, 
```

torch_geometric/nn/meta.py
```diff
@@ -90,7 +90,9 @@ def forward(self, x, edge_index, edge_attr, u, batch):
         op = MetaLayer(EdgeModel(), NodeModel(), GlobalModel())
         x, edge_attr, u = op(x, edge_index, edge_attr, u, batch)
     """
-    def __init__(self, edge_model=None, node_model=None, global_model=None):
+    def __init__(self, edge_model: Optional[torch.nn.Module] = None,
+                 node_model: Optional[torch.nn.Module] = None,
+                 global_model: Optional[torch.nn.Module] = None):
```
rusty1s, line 95:
```suggestion
                 global_model: Optional[torch.nn.Module] = None, ):
```

torch_geometric/nn/meta.py
camillepradel: Hi! I agree it makes more sense to remove the `Optional[]`. However, current `test_meta_layer` test fails with this type hinting, as we use a dummy model for `node_model` which always return `None` (and thus returned `x` value is `None` in that case).
Maybe I could update the dummy model to return `edge_attr` for `edge_model`, `x` for `node_model` and `u` for `global_model` ?

rusty1s: Yeah, it probably makes sense to define them separately with the correct type hints.

rusty1s (approved):
Thank you so much!

PR #5766, opened 2022-10-18
[Type Hints] `nn.AttentiveFP`

TanveshT:
## Changes

- added missing type hints to `GATEConv` & `AttentiveFP`


torch_geometric/nn/models/attentive_fp.py
```diff
@@ -29,7 +34,7 @@ def __init__(self, in_channels: int, out_channels: int, edge_dim: int,
 
         self.reset_parameters()
 
-    def reset_parameters(self):
+    def reset_parameters(self) -> None:
```
rusty1s, line 37:
```suggestion
    def reset_parameters(self)
```

rusty1s (approved):

torch_geometric/nn/models/attentive_fp.py
```diff
@@ -29,7 +34,7 @@ def __init__(self, in_channels: int, out_channels: int, edge_dim: int,
 
         self.reset_parameters()
 
-    def reset_parameters(self):
+    def reset_parameters(self)
```
rusty1s, line 37:
```suggestion
    def reset_parameters(self):
```

PR #5783, opened 2022-10-20
Update `load_csv.py`

dbushpw:
load_node_csv() used  mapping = {index: i for i, index in enumerate(df.index.unique())} . But this will generate a mapping of the index of the dataframe to itself, since the index of the dataframe is a RangeIndex. To get the value from the index_col (the movie ID), enumerate(df[index_col]) should be used instead.

### DIFF
--- a/examples/hetero/load_csv.py
+++ b/examples/hetero/load_csv.py
@@ -16,7 +16,7 @@ rating_path = osp.join(root, 'ml-latest-small', 'ratings.csv')
 
 def load_node_csv(path, index_col, encoders=None, **kwargs):
     df = pd.read_csv(path, index_col=index_col, **kwargs)
-    mapping = {index: i for i, index in enumerate(df.index.unique())}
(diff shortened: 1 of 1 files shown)


examples/hetero/load_csv.py
```diff
@@ -16,7 +16,7 @@
 
 def load_node_csv(path, index_col, encoders=None, **kwargs):
     df = pd.read_csv(path, index_col=index_col, **kwargs)
-    mapping = {index: i for i, index in enumerate(df.index.unique())}
```
rusty1s, line 19: I am not sure I understand the change. `pd.read_csv(index_col=index_col)` will move the `index_col` to `df.index`, so we should not be able to access it via `df[index_col]`.

dbushpw: You are right, I loaded the dataframes outside of the function and didnt reassign the indexes.

PR #5786, opened 2022-10-20
Update Schnet RBF block's lambda coefficient

justin1228:
When I see the paper of "SchNet", the coefficient for RBF function does not same with this code.


rusty1s (approved):

rusty1s:
This looks to be inline with https://github.com/atomistic-machine-learning/schnetpack/blob/3437251e700180d29e125f4faa0faec64a66db96/src/schnetpack/nn/acsf.py#L172 though.

justin1228:
With reference from the paper "SchNet: A continuous-filter convolutional neural network for modeling quantum interactions", It said
> We avoid this by expanding the distance with radial basis functions
> ek(ri − rj) = exp(−γ∥dij − μk∥2)
> located at centers 0Å ≤ μk ≤ 30Å every 0.1Å with γ = 10Å.
When you look at the relationship between distance offset and gamma, it is just reciprocal!

justin1228:
@EdisonLeeeee I'm ready to get a review from you.

EdisonLeeeee (approved):

PR #5787, opened 2022-10-20
Added `return_semantic_attention_weights` parameter to `HANConv`

manuel-dileo:
Added return_semantic_attention_weights parameter to forward method of HANConv, in the same way GATConv returns its attention_weights. 

return_semantic_attention_weights (bool, optional): If set to :obj:`True`, will additionally return the tensor :obj:`semantic_attention_weights`, holding the computed attention weights for each edge type at semantic-level attention. (default: :obj:`None`). 

Further work could be dedicated to return also node-level attention weights as dictionary of (edge_type, attention_weights) items.

### DIFF
--- a/torch_geometric/nn/conv/han_conv.py
+++ b/torch_geometric/nn/conv/han_conv.py
@@ -12,7 +12,8 @@ from torch_geometric.utils import softmax
 
 
 def group(xs: List[Tensor], q: nn.Parameter, k_lin: nn.Module,
-          return_semantic_attention_weights=None):
+          return_semantic_attention_weights: bool = False) -> Union[OptTensor, Tuple[Tensor, Tensor]]:
+
     if len(xs) == 0:
         return None
     else:
--- a/torch_geometric/nn/conv/han_conv.py
+++ b/torch_geometric/nn/conv/han_conv.py
@@ -113,7 +113,7 @@ class HANConv(MessagePassing):
         edge_index_dict: Dict[EdgeType, Adj],
         return_semantic_attention_weights: bool = False,
     ) -> Union[Dict[NodeType, OptTensor], Tuple[Dict[NodeType, OptTensor],
-                                                Dict[NodeType, Tensor]]]:
+                                                Dict[NodeType, OptTensor]]]:
         r"""
         Args:
             x_dict (Dict[str, Tensor]): A dictionary holding input node
--- a/torch_geometric/nn/conv/han_conv.py
+++ b/torch_geometric/nn/conv/han_conv.py
@@ -109,7 +109,8 @@ class HANConv(MessagePassing):
 
     def forward(self, x_dict: Dict[NodeType, Tensor],
                 edge_index_dict: Dict[EdgeType, Adj],
-                return_semantic_attention_weights=None):
+                return_semantic_attention_weights: bool = False) -> Union[Dict[NodeType, OptTensor], TupleDict[NodeType, OptTensor], Tensor]]:
+
         r"""
         Args:
             x_dict (Dict[str, Tensor]): A dictionary holding input node
--- a/torch_geometric/nn/conv/han_conv.py
+++ b/torch_geometric/nn/conv/han_conv.py
@@ -23,7 +23,7 @@ def group(
         num_edge_types = len(xs)
         out = torch.stack(xs)
         if out.numel() == 0:
-            return out.view(0, out.size(-1))
+            return out.view(0, out.size(-1)), None
         attn_score = (q * torch.tanh(k_lin(out)).mean(1)).sum(-1)
         attn = F.softmax(attn_score, dim=0)
         out = torch.sum(attn.view(num_edge_types, 1, -1) * out, dim=0)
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/conv/han_conv.py
```diff
@@ -11,8 +11,8 @@
 from torch_geometric.utils import softmax
 
 
-def group(xs: List[Tensor], q: nn.Parameter,
-          k_lin: nn.Module) -> Optional[Tensor]:
+def group(xs: List[Tensor], q: nn.Parameter, k_lin: nn.Module,
+          return_semantic_attention_weights=None):
```
rusty1s, line 15: Shall we rename to `return_attention_weights`?

rusty1s, line 15:
```suggestion
          return_semantic_attention_weights: bool = False) -> Union[OptTensor, Tuple[Tensor, Tensor]]:

```

torch_geometric/nn/conv/han_conv.py
```diff
@@ -23,6 +23,8 @@ def group(xs: List[Tensor], q: nn.Parameter,
         attn_score = (q * torch.tanh(k_lin(out)).mean(1)).sum(-1)
         attn = F.softmax(attn_score, dim=0)
         out = torch.sum(attn.view(num_edge_types, 1, -1) * out, dim=0)
+        if isinstance(return_semantic_attention_weights, bool):
```
rusty1s, line 26:
```suggestion
        if return_semantic_attention_weights:

```

torch_geometric/nn/conv/han_conv.py
```diff
@@ -103,10 +105,9 @@ def reset_parameters(self):
         self.k_lin.reset_parameters()
         glorot(self.q)
 
-    def forward(
-        self, x_dict: Dict[NodeType, Tensor],
-        edge_index_dict: Dict[EdgeType,
-                              Adj]) -> Dict[NodeType, Optional[Tensor]]:
+    def forward(self, x_dict: Dict[NodeType, Tensor],
+                edge_index_dict: Dict[EdgeType, Adj],
+                return_semantic_attention_weights=None):
```
rusty1s, line 110:
```suggestion
                return_semantic_attention_weights: bool = False) -> Union[Dict[NodeType, OptTensor], TupleDict[NodeType, OptTensor], Tensor]]:

```

torch_geometric/nn/conv/han_conv.py
```diff
@@ -116,12 +117,20 @@ def forward(
...
-        :rtype: :obj:`Dict[str, Optional[Tensor]]` - The output node embeddings
-            for each node type.
-            In case a node type does not receive any message, its output will
-            be set to :obj:`None`.
+            return_semantic_attention_weights (bool, optional):
+                If set to :obj:`True`,
+                will additionally return the tensor
+                :obj:`semantic_attention_weights`, holding the computed
+                attention weights for each edge type
+                at semantic-level attention. (default: :obj:`None`)
```
rusty1s, line 125:
```suggestion
                at semantic-level attention. (default: :obj:`False`)

```

torch_geometric/nn/conv/han_conv.py
```diff
@@ -116,12 +117,20 @@ def forward(
...
         """
+        # NOTE: semantic_attention weights will be returned whenever
+        # `return_attention_weights` is set to a value, regardless of its
+        # actual value (might be `True` or `False`). This is a current somewhat
+        # hacky workaround to allow for TorchScript support via the
+        # `torch.jit._overload` decorator, as we can only change the output
+        # arguments conditioned on type (`None` or `bool`), not based on its
+        # actual value.
```
rusty1s, line 133: Can be dropped since we do not support TorchScript here anyway.

torch_geometric/nn/conv/han_conv.py
```diff
@@ -149,13 +158,19 @@ def forward(
 
         # iterate over node types:
         for node_type, outs in out_dict.items():
-            out = group(outs, self.q, self.k_lin)
+            if isinstance(return_semantic_attention_weights, bool):
```
rusty1s, line 161:
```suggestion
            if return_semantic_attention_weights:

```

torch_geometric/nn/conv/han_conv.py
```diff
@@ -149,13 +158,19 @@ def forward(
...
+                    return_semantic_attention_weights)
+            else:
+                out = group(outs, self.q, self.k_lin)
 
             if out is None:
                 out_dict[node_type] = None
                 continue
             out_dict[node_type] = out
-
+        if isinstance(return_semantic_attention_weights, bool):
```
rusty1s, line 172:
```suggestion
        if return_semantic_attention_weights:

```

torch_geometric/nn/conv/han_conv.py
manuel-dileo: Thank you for your review! HAN has two level attention mechanism 1) a node-level one that represents how much the features of the nodes in the neighborhood are important for a certain node 2) a semantic-level attention which can be interpreted as the contribution of a certain metapath for a specific task. I think that the renaming to simple return_attention_weights could be too generic and lead to potential usage errors.

rusty1s: Ok, let's keep the name to avoid confusion :)

rusty1s (approved):

jasmineChouLujie:
Does the return of conv.HANConv in torch_geometric contain the 2 parts tasks which are node-level attention and  semantic-level attention in the paper?

Issue #5629, opened 2022-10-07
GNN Explanation Settings


RexYing:
### 🚀 User Configuration of Explanation Settings

Allow for easily configurable setting for different use cases of GNN explanations. Various explanability methods can be adapted to tackle these different explanation problem settings.

There are many dimensions to consider when making these evaluations, including

- [x] Explanation of underlying phenomenon vs. model prediction behavior
- [x] Soft / hard masks as a form of explanation (via thresholding)
- [ ] Thresholding of masks to produce a concise explanation (constraint on the size and connectivity of explanation)
The algorithm can be as similar to a greedy algorithm:
1. Start from the seed node
2. At every step, identify all edges that are adjacent to any of the selected nodes
3. Pick the edge adjacent with the largest weight, and if the edge reaches a previously not selected node, add that into the set of selected nodes
4. Iterate 2 and 3, until a desired explanation size is achieved

More details can be found in https://arxiv.org/abs/2206.09677

The goal is to provide a class to allow users to easily set these options. Different GNN explainability algorithms can adapt to these settings 

To start with, we can consider gradient-based methods and perturbation-based methods like GNNExplainer, and adapt their objectives to different requirements of the user (the 3 aspects mentioned above and potentially more).

PR #5804, opened 2022-10-22
GNN explanation settings 

dufourc1:
This is a PR working towards #5629.

## what it does
It changes the current `Explainer`  class by splitting its responsibilities between three classes: 
- `Explanation` represent an explanation as a `Data` object with 4 basic masks (node, edge, node-features, edge-features) which can be extended. It should include any visualisation methods.
- `Explainer` class is now in charge of the “meta configuration” of the explanations (i.e. model vs phenomenon, post-processing methods, …)
- `ExplainerAlgorithm` is in charge of computing the explanations: for example, the class `GNNExplainer` will be a child. For now, there is a dummy explainer `RandomExplainer`. 

### Explainer
- [x] Add task level in `Explainer` → allows to differentiate between node-level tasks vs graph-level tasks.

- [x] Implement the thresholding methods and other post-processing methods 
     - [x]  `hard`
     - [x] `topk`
     - [ ] `connected`? (won't do for now)
     - [ ] rescaling of the explanations after thresholding/ before thresholding ? (won't do for now)
- [x] Typo & formatting docs
- [x] add target index as in captum
- [x] Remove old implementations in `torch_geometric.nn.models`
- [ ] Add examples for how to use explainer and how to combine model and interface. (wait for implementation in another PR: almost ready, just need to clean the code a bit)

Happy to take any suggestion onboard 👍🏽


torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+        # non optimal for now, but captum requires the model to instantiate
+        # the explanation algorithm
+        explainer = getattr(attr, self.method)(model)
+        input, additional_forward_args = self._arange_inputs(
+            x, edge_index, mask_type, **kwargs)
+
+        attributions = explainer.attribute(
+            **self._arange_args(input, target, additional_forward_args))
+
+        return self._create_explanation_from_masks(x, edge_index, attributions)
```
RexYing, line 48: How well does captumexplainer work for edge masks?

torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+    def explain(
+        self,
+        x: torch.Tensor,
+        edge_index: torch.Tensor,
+        target: torch.Tensor,
+        model: CaptumModel,
+        mask_type: str = "node",
+        **kwargs,
+    ) -> Explanation:
+
```
RexYing, line 36: we could probably add docstrings for various arguments (also specify what kwargs could include).

torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+        explainer = getattr(attr, self.method)(model)
+        input, additional_forward_args = self._arange_inputs(
+            x, edge_index, mask_type, **kwargs)
+
+        attributions = explainer.attribute(
+            **self._arange_args(input, target, additional_forward_args))
+
+        return self._create_explanation_from_masks(x, edge_index, attributions)
+
+    def supports(
```
RexYing, line 50: wonder if we should add some test coverage. I'm also not too sure. But perhaps use some mocks?

torch_geometric/explainability/algo/dummyexplainer.py
```diff
@@ -0,0 +1,39 @@
+import torch
+
+from torch_geometric.explainability.explanations import Explanation
+
+from .base import ExplainerAlgorithm
+
+
+class RandomExplainer(ExplainerAlgorithm):
```
RexYing, line 8: Is this for testing purpose?

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,91 @@
...
+        r"""Check if the explainer supports the user-defined settings.
+
+        Returns true if the explainer supports the settings.
+
+        Responsability of the children to exclude the settings that are not
+        supported.
+
+        Args:
+            explanation_type (str): the type of explanation to compute.
+            mask_type (str): the type of mask to use.
```
RexYing, line 89: specify the supported values?

torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+    Wraper for captum explainers.
+
+    Note:
+        This algorithm expect the model to be callable with the following
+        signature: model(x, edge_index, **kwargs).
+
+
+    Args:
+        method (str, optional): name of the captum explainer.
+        Defaults to "Saliency".
```
RexYing, line 20:
```suggestion
            Defaults to "Saliency".
```

torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+            **self._arange_args(input, target, additional_forward_args))
+
+        return self._create_explanation_from_masks(x, edge_index, attributions)
+
+    def supports(
+        self,
+        explanation_type: str,
+        mask_type: str,
+    ) -> bool:
+        return mask_type != "layers"
```
RexYing, line 55: what does layers mask_type mean? Is it compatible to node / edge mask types?

We might also consider cases where the user enters an unknown mask_type?

torch_geometric/explainability/algo/captumexplainer.py
```diff
@@ -0,0 +1,147 @@
...
+        Returns:
+            Tuple[torch.Tensor,Tuple[torch.Tensor]]: main input and additional
+                forward arguments.
+        """
+        input_mask = torch.ones((1, edge_index.shape[1]), dtype=torch.float,
+                                requires_grad=True)
+
+        if mask_type == "node":
+            input = x.clone().unsqueeze(0)
+            additional_forward_args = (edge_index, )
```
RexYing, line 77: is x not required for explaining node features?

torch_geometric/explainability/explainer.py
```diff
@@ -0,0 +1,221 @@
...
+from torch_geometric.explainability.algo.captumexplainer import CaptumExplainer
+from torch_geometric.explainability.explanations import Explanation
+from torch_geometric.explainability.utils import to_captum
+
+
+@dataclass
+class Threshold:
+    type: str
+    value: float
+    _valid_type = ["soft", "hard", "connected"]
```
RexYing, line 16: "soft" means that we don't threshold right?

should we also distinguish hard by "top K" or an absolute value?

torch_geometric/explainability/explainer.py
```diff
@@ -0,0 +1,221 @@
...
+        model: torch.nn.Module,
+        model_return_type: str,
+        threshold_value: float = 0.5,
+    ) -> None:
+
+        super().__init__()
+
+        self.model = model
+        # details of the model
+        if model_return_type.lower() not in ["logits", "probs", "regression"]:
```
RexYing, line 84: can multi-label be supported? i.e. a list of probs.

RexYing (commented):
Thanks for the overall structure. I think it's well-written.

torch_geometric/explainability/algo/dummyexplainer.py
dufourc1: For testing and benchmarking: if someone proposes a new explaination algorithms it can be compared to it (in the same spirit as a dummy model from sklearn).

torch_geometric/explainability/algo/captumexplainer.py
dufourc1: This is a try to allow explanations of neurons in the layer of a model, but maybe we should restrict it to node/edge at the beginning?

torch_geometric/explainability/algo/captumexplainer.py
dufourc1: I was thinking that one could try to get node importance in a graph with only edge features (the simplest way would be to first compute the edge attribution and then dispatch it to the nodes it connect)

torch_geometric/explainability/explainer.py
dufourc1: Yes you're totally right! `TopK` sounds good

torch_geometric/explainability/algo/captumexplainer.py
dufourc1: I would first check the shape of the explanations and the range of the values. Maybe we could try to implement tests based on `BA2-shapes` (the house vs cycle task)?

torch_geometric/explainability/explainer.py
dufourc1: We could do it like `captum`, i.e. ask the user for a `target_index`

torch_geometric/explainability/algo/captumexplainer.py
RexYing: Yeah i think that's ok. We can also add `#TODO` or NotImplementedError for the option and make it for future PRs

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
+from abc import abstractmethod
```
rusty1s, line 1: Can we re-name the package to `torch_geoemtric.explain`?

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+from torch_geometric.data import Data
+from torch_geometric.explainability.explanations import Explanation
+
+
+class ExplainerAlgorithm(torch.nn.Module):
+    r"""An abstract class for explainer algorithms."""
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.objective: Callable[
```
rusty1s, line 15: Pass as input argument?

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+        Subclasses should override this method to define their own loss.
+
+        Args:
+            y_hat (torch.Tensor): the output of the explanation algorithm.
+                (e.g. the forward pass of the model with the mask applied).
+            y (torch.Tensor): the target output.
+        """
+
+    @property
+    def accept_new_loss(self) -> bool:
```
rusty1s, line 45: Not totally sure why we need this function since the loss is defined via implementing the abstract method.

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+        """Whether the algorithm can accept a new loss function."""
+        return True
+
+    def set_objective(
```
rusty1s, line 49: I guess we don't need this one if we set the objective as input argument.

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        self.objective = objective
+
+    @torch.no_grad()
+    def get_initial_prediction(
+        self,
+        model: torch.nn.Module,
+        g: Data,
+        batch: Optional[torch.Tensor] = None,
+        **kwargs,
```
rusty1s, line 67: I would replace those with `**kwargs` to denote the input of the model. Let's not operate expect a `Data` object here.

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+            torch.Tensor: output of the underlying model.
+        """
+        out = model(
+            g,
+            **dict(batch=batch, **kwargs),
+        )
+        return out
+
+    @abstractmethod
+    def explain(
```
rusty1s, line 85: Here again.

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+            target (torch.Tensor): target of the GNN.
+            target_index (int, optional): Index of the target to explain. Used
+                in case of multi-outputs. Defaults to None.
+            batch (torch.Tensor, optional): _description_. Defaults to None.
+
+        Returns:
+            Explanation: explanation of the GNN.
+        """
+
+    def forward(
```
rusty1s, line 108: What's the purpose of `explain` when we can directly use `forward` for this?

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+        Args:
+            explanation_type (str): the type of explanation to compute.
+                Should be in ["model", "phenomenon"]
+            mask_type (str): the type of mask to use.
+                Should be in ["node", "edge", "node_and_edge", "layers"]
+        """
+
+    # TODO :this creates coupling between the explainer and the explanation,
+    #  which is not ideal. We should find a way to remove this.
+    def _create_explanation_from_masks(self, g, attributions, mask_type):
```
rusty1s, line 142: Add type hints.

torch_geometric/explainability/algo/base.py
```diff
@@ -0,0 +1,168 @@
...
+            Explanation: explanation object.
+        """
+        node_features_mask = None
+        edge_mask = None
+        if mask_type == "node":
+            node_features_mask = attributions.squeeze(0)
+        elif mask_type == "edge":
+            edge_mask = attributions.squeeze(0)
+        elif mask_type == "node_and_edge":
+            node_features_mask = attributions[0].squeeze(0)
```
rusty1s, line 160: Check that `attributions` is actually a tuple/list.

torch_geometric/explainability/algo/base.py
rusty1s, line 1: This file should also not be in `algo` package.

rusty1s (commented):
Thanks for starting this effort.

I would like to proceed a bit more structured here to simplify reviewing efforts and to remove duplicated code in PyG. As such, I suggest to split this PR into multiple ones:

1. Create `explain` package and add `Explainer` base class and utility functions. Set up documentation, remove the corresponding counterparts in `nn.models`.
2. Move the implementations from `GNNExplainer`, `to_captum` and the dummy explainer in separate PRs. Remove the counterpart code in `nn.models`.

torch_geometric/explainability/algo/base.py
dufourc1: Brain on autopilot, will update.

torch_geometric/explainability/algo/base.py
dufourc1: Is there an advantage of using `**kwargs` to pass around the input of the model vs a `Data` object ? With `**kwargs` it seemed a bit confusing, hence the `Data` object.

torch_geometric/explainability/algo/base.py
dufourc1: Where should I put it ?

rusty1s: I would put it into `torch_geometric/explain/base.py`

torch_geometric/explainability/algo/base.py
rusty1s: I personally think it is bad practice to use `def forward(data)` since this way it is not specified which arguments are actually used within the model. Prefer something like `def forward(x, edge_index)` (or similar) to make the inputs of the model well-defined.

docs/source/modules/explain.rst
```diff
@@ -0,0 +1,62 @@
...
+----------
+
+The ref::`torch_geometric.explain` module provides a set of tools to explain the
+predictions of a :class:`torch_geometric.nn.model`  or to explain the underlying
+phenomenon of a dataset (see `"GraphFramEx: Towards Systematic Evaluation of Explainability Methods for Graph Neural Networks"
+<https://arxiv.org/abs/2206.09677>`_  for more details).
+
+We represent explanations using the :class:`Explanation` class, which is a `Data` object containing masks for the nodes, edges, features and any attributes of the data.
+
+The :class:`~torch_geometric.explain.Explainer` class is designed to habdle all the explainability parameters:
```
Padarn, line 21:
```suggestion
The :class:`~torch_geometric.explain.Explainer` class is designed to handle all the explainability parameters:
```

test/explain/test_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+                                             ("connected", 1, True),
+                                             ("connected", 0, False),
+                                             ("connected", -1, False),
+                                             ("connected", 0.5, False),
+                                             ("invalid", None, False)])
+def test_classification_threshold_init(dummyexplainer, model,
+                                       model_return_type, task_level,
+                                       explanation_type, mask_type,
+                                       threshold_pairs):
+    if threshold_pairs[2]:
```
Padarn, line 141: (nit) its pretty hard to tell what this test is trying to test: maybe it would be a bit clearer if we added a line
```
threshold, threshold_value, allowed = threshold_pairs
```

torch_geometric/explain/algorithm/dummyexplainer.py
```diff
@@ -0,0 +1,53 @@
+from typing import Optional
+
+import torch
+
+from torch_geometric.data import Data
+from torch_geometric.explain.base import ExplainerAlgorithm
+from torch_geometric.explain.explanations import Explanation
+
+
+class RandomExplainer(ExplainerAlgorithm):
```
Padarn, line 10: Maybe this would be better defined in the tests? I can't see a reason users would want to use this?

torch_geometric/explain/algorithm/dummyexplainer.py
```diff
@@ -0,0 +1,53 @@
...
+                **kwargs) -> Explanation:
+        return self.forward(x=g.x, edge_index=g.edge_index,
+                            edge_attr=g.edge_attr, **kwargs)
+
+    def supports(
+        self,
+        explanation_type: str,
+        mask_type: str,
+    ) -> bool:
+        return mask_type != "layers"
```
Padarn, line 50: (nit) I'd probably prefer that this enumerated the supported types, makes it very clear what is supported since its a free string anyway

torch_geometric/explain/base.py
```diff
@@ -0,0 +1,121 @@
...
+        self,
+        x: torch.Tensor,
+        edge_index: torch.Tensor,
+        *,
+        model: torch.nn.Module,
+        target: torch.Tensor,
+        target_index: int = 0,
+        batch: Optional[torch.Tensor] = None,
+        task_level: str = "graph",
+        return_type: str = "regression",
```
Padarn, line 56: Sorry naive question: Is it likely for a single explainer to support multiple return types?

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+            message += "Check the documentation for more details."
+
+            raise ValueError(message)
+
+        # update model mode based on return type
+        if self.return_type == "regression":
+            self.mode = "regression"
+
+
+class Explainer(torch.nn.Module):
```
Padarn, line 95: Sorry high level thought: Intuitively it would make more sense to me to have something like a builder rather than having the explainer be a module which wraps another module? Do we need both this abstraction and the `ExplainerAlgorithm`?

torch_geometric/explainability/algo/base.py
Padarn: Agreed. The `Data` object also does not require those attributes to be set, so it makes it less clear to the developer what arguments are actually required.

torch_geometric/explain/algorithm/dummyexplainer.py
dufourc1: Maybe for benchmarking? if someone proposes a new explanation algorithms, it can be compared to the dummy explainers (in the same spirit as a dummy model from `sklearn`).

torch_geometric/explain/base.py
dufourc1: If you look at the gradient based methods, they would support both classification and regression tasks? But this might be added in the `supports` method of the `ExplainerAlgorithm`.

torch_geometric/explain/explainer.py
dufourc1: I think you're right. I have to think about how explanations can be used in model training to check if the inheritance is needed  (like in _Schramowski, P., Stammer, W., Teso, S. et al. Making deep neural networks right for the right scientific reasons by interacting with their explanations. Nat Mach Intell 2, 476–486 (2020). https://doi.org/10.1038/s42256-020-0212-3_).

dufourc1: For the post-processing of the explanation? (I know it's quite simple for now, but it may be upgraded).
This would ensure that the only responsibility of the `ExplainerAlgorithm` are to return masks and then the rest is always taken care of in the same way.

torch_geometric/explain/base.py
Padarn: Makes sense. I haven't worked with many different approaches to this so I wasn't sure. Thanks

torch_geometric/explain/explainer.py
Padarn: I was thinking of `Explainer` as being a base class and the various `ExplainerAlgorithm` just implement the functions needed to provide the masks.

Similar to how our `Aggregation` package works.

dufourc1: I'll have a look at the 'Aggregation' package then 👍

dufourc1: Ok, I see what you meant. You're probably right since the `ExplainerAlgorithm` is only called once in `_compute_explanation`.  Maybe @RexYing has some more insights ?

torch_geometric/explain/algorithm/dummyexplainer.py
```diff
@@ -0,0 +1,53 @@
...
+                Defaults to None.
+
+        Returns:
+            Explanation: random explanation based on the shape of the inputs.
+        """
+        node_features_mask = torch.rand(x.shape)
+        node_mask = torch.rand(x.shape[0])
+        edge_mask = torch.rand(edge_index.shape[1])
+        masks = (node_mask, edge_mask, node_features_mask)
+        mask_keys = ("node_mask", "edge_mask", "node_features_mask")
```
rusty1s, line 30: Why not save in a dictionary?
```python
mask_dict = {
    'node_mask': node_mask,
    ...
}
```
This way, we don't need to track both `masks` and `mask_keys`.

torch_geometric/explain/algorithm/dummyexplainer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def explain(self, g: Data, model: torch.nn.Module,
+                **kwargs) -> Explanation:
+        return self.forward(x=g.x, edge_index=g.edge_index,
+                            edge_attr=g.edge_attr, **kwargs)
+
+    def supports(
+        self,
+        explanation_type: str,
+        mask_type: str,
```
rusty1s, line 48: Can we add `Enum` classes for these?

torch_geometric/explain/algorithm/dummyexplainer.py
```diff
@@ -0,0 +1,53 @@
...
+            x (torch.Tensor): node features.
+            edge_index (torch.Tensor): edge indices.
+            edge_attr (torch.Tensor, optional): edge attributes.
+                Defaults to None.
+
+        Returns:
+            Explanation: random explanation based on the shape of the inputs.
+        """
+        node_features_mask = torch.rand(x.shape)
+        node_mask = torch.rand(x.shape[0])
```
rusty1s, line 27: Ensure correct devices.

torch_geometric/explain/base.py
```diff
@@ -0,0 +1,121 @@
...
+from torch_geometric.data import Data
+from torch_geometric.explain.explanations import Explanation
+
+
+class ExplainerAlgorithm(torch.nn.Module):
+    """Abstract class for explanation algorithms."""
+    @abstractmethod
+    def loss(self, y_hat: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        r"""
+        This method compute the loss to be used for the explanation algorithm.
```
rusty1s, line 15:
```suggestion
        This method computes the loss to be used for the explanation algorithm.
```

torch_geometric/explain/base.py
```diff
@@ -0,0 +1,121 @@
...
+        Subclasses should override this method to define their own loss.
```
rusty1s, line 16: Drop. Since it is an abstractmethod, we have to override it.

torch_geometric/explain/base.py
```diff
@@ -0,0 +1,121 @@
...
+            y_hat (torch.Tensor): the output of the explanation algorithm.
+                (e.g. the forward pass of the model with the mask applied).
+            y (torch.Tensor): the reference output.
+        """
+
+    @torch.no_grad()
+    def get_initial_prediction(
+        self,
+        model: torch.nn.Module,
+        g: Data,
```
rusty1s, line 28: Still prefer simply `*args` and `**kwargs` which is used as input to the model.

torch_geometric/explain/base.py
```diff
@@ -0,0 +1,121 @@
...
+            edge_index (torch.Tensor): edge indices.
+            model (torch.nn.Module): the model to explain.
+            target (torch.Tensor): the target of the model.
+            target_index (int): the index of the target to explain.
+                By default suppose the target is a single value.
+            batch (torch.Tensor, optional): batch indicator.
+            **kwargs: additional arguments to pass to the GNN.
+        """
+
+    @abstractmethod
```
rusty1s, line 72: Drop IMO.

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+
+from torch_geometric.data import Data
+from torch_geometric.explain.base import ExplainerAlgorithm
+from torch_geometric.explain.explanations import Explanation
+
+
+@dataclass
+class Threshold:
+    """Class to store and validate threshold parameters."""
+    type: str
```
rusty1s, line 15: make a `Enum` and override `init` to support both str and `Enum`:
```python
def __init__(self, type: Union[ExplanationType, str], value):
    self.type = ExplanationType(type)

File: torch_geometric/explain/explainer.py [hunk 34]
```diff
@@ -0,0 +1,342 @@
...
+                raise ValueError(f'Invalid threshold value {self.value}. '
+                                 f'Valid values are in [0, 1].')
+        if self.type in ["topk", "connected", "topk_hard"]:
+            if self.value <= 0 or not isinstance(self.value, int):
+                raise ValueError(f'Invalid threshold value {self.value}. '
+                                 f'Valid values are positif integers.')
+
+
+@dataclass
+class ExplainerConfig:
```
rusty1s, line 38: Can you add some doc-strings about what each of these means?

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+    """Class to store and validate high level explanation parameters."""
+    explanation_type: str
+    mask_type: str
+    _valid_explanation_type = ["model", "phenomenon"]
+    _valid_mask_type = [
+        "node", "edge", "node_and_edge", "layers", "node_feat", "edge_feat",
+        "node_feat_and_edge_feat", "node_feat_and_edge", "node_and_edge_feat"
```
rusty1s, line 45: Wonder if we can separate the `mask_type` into `node_mask_type` and `edge_mask_type` for a cleaner interface.

torch_geometric/explain/explainer.py
rusty1s: I don't have a strong opinion on this one, but I personally like the separation between algorithm and an `Explainer` that takes care of the rest. Makes it easy to see the concrete interface one has to implement for a new `ExplainerAlgorithm`.

rusty1s, line 95: Not sure why this needs to be a `Module`?

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+
+        self.model_config = ModelConfig(model=model,
+                                        return_type=model_return_type.lower(),
+                                        task_level=task_level.lower())
+
+        self.explanation_config = ExplainerConfig(
+            explanation_type=explanation_type.lower(),
+            mask_type=mask_type.lower())
+
+        # details for post-processing the ouput of the explanation algorithm
```
rusty1s, line 205:
```suggestion
        # details for post-processing the output of the explanation algorithm
```

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+                f'Invalid explanation type {self.explanation_type}. '
+                f'Valid types are {self._valid_explanation_type}.')
+        if self.mask_type not in self._valid_mask_type:
+            raise ValueError(f'Invalid mask type {self.mask_type}. '
+                             f'Valid types are {self._valid_mask_type}.')
+
+
+@dataclass
+class ModelConfig:
+    model: torch.nn.Module
```
rusty1s, line 60: A `ModelConfig` shouldn't hold the model IMO.

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+                       **kwargs) -> torch.Tensor:
+        r"""Returns the prediction of the model on the input graph.
+
+        Args:
+            g (torch_geometric.data.Data): the input graph.
+            batch (torch.Tensor, optional): the batch vector.
+                (default: :obj:`None`)
+        """
+        with torch.no_grad():
+            return self.model_config.model(g, **dict({
```
rusty1s, line 231:
```suggestion
            return self.model(g, **dict({
```

torch_geometric/explain/explainer.py
```diff
@@ -0,0 +1,342 @@
...
+                "batch": batch,
+                **kwargs
+            }))
+
+    def forward(self, g: Data, target: Optional[torch.Tensor] = None,
```
rusty1s, line 236:
```suggestion
    def __call__(self, g: Data, target: Optional[torch.Tensor] = None,
```

torch_geometric/explain/explanations.py
```diff
@@ -0,0 +1,124 @@
+"""Class representing explanations."""
+from typing import Optional, Tuple, Union
+
+from torch import Tensor
+
+from torch_geometric.data import Data
+
+
+class Explanation(Data):
```
rusty1s, line 9: Not totally convinced yet that this needs to be a `Data` object. Any reason for that? What is the functionality form `Data` we need here? Do we need it to create subgraphs from the original graph?

dufourc1: For creating visualization or downstream tasks based on explanations, we might need methods to create subgraphs. Also, the explanation is a graph with specific attributes, I thought it would be good to reuse the methods that can be already in `pyg` for handling this kind of data.

torch_geometric/explain/explainer.py
dufourc1: I think I will stay with the split then, I removed the useless fact that `Explainer` was a module (thanks @rusty1s  for the review btw 👍🏽 ), making it more clear.

torch_geometric/explain/base.py
dufourc1: So I switched `Data` to  `**kwargs` but did not include `*args`, because when using `*args` you have to define an ordering of the input you expect, which may vary between two instances of `ExplainerAlgorithm` and make tests like this which I find messy: 
https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/explainer.py#L81-L105.

Is this ok or are the `*args` necessary somehow ?

RexYing (approved):
Thanks Matthias and Padarn for the detailed suggestions on code structure! 

I believe that the concerns are addressed but feel free to discuss further. Thanks Charles for incorporating all suggestions.

wsad1 (changes requested):
I see we've removed `models.gnn_explainer` and `models.to_captum` in this PR, but we haven't added any equivalent modules in `explain.algorithm`, I don't think we should remove them until we add equivalent modules. WDYT?
Further implementing `gnn_explainer` and `to_captum` behind this interface would validate it, again we can do that in follow up PRs.

dufourc1:
> I see we've removed `models.gnn_explainer` and `models.to_captum` in this PR, but we haven't added any equivalent modules in `explain.algorithm`, I don't think we should remove them until we add equivalent modules. WDYT?
> Further implementing `gnn_explainer` and `to_captum` behind this interface would validate it, again we can do that in follow up PRs.

Hi! Thanks for having a look 👍
So for letting the `gnn_explainer` in, we would have to let the old  `Explainer` class as well, which could have been confusing? Or we could add some kind of future depreciation warnings (I'm fine with both). 

Either way, the `CaptumAlgorithms` and `GnnExplainer` classes will arrive shortly, I'll do two PRs this week.

wsad1 (commented):
Yeah lets leave `gnn_explianer` and `to_captum`, and add a depreciation warning when we add them behind the `ExplainerAlgorithm` interface.

rusty1s (approved):
Thanks! I made a pass and made some adjustments. The overall interface has stayed the same, while I ensured that structure and documentation now act as a great basis for further improvements.

The one adjustment I made was to drop the `index` argument in `explainer.__call__` since I wasn't really sure what it does :( Please feel free to bring it back in later PRs where we can see its usage.

PR #5816, opened 2022-10-24
fixed issue when importing from networkx graph in node ordering. Afte…

DanielFPerez:
Found bug when importing from_networkx after performing a data augmentation procedure on the networkx graph. The data augmentation was changing the node-ids of the original graph.
Using node-IDs strings as positional encodings, the data augmentation step re-shuffled the node-ID strings of the graph. 
E.g.:
- Before data augmentation: G1.nodes(): ['A0', 'A1', 'A2' 'A3', 'A4']
- A possible data augmentation: G1_augm.nodes(): ['A2', 'A0', 'A4', 'A1', 'A3']

However, when converting from_networkx, if the "sorted"-odering option is not passed in nx.convert_node_labels_to_integers (Line 133 in torch_geometric.utils.convert.py), the function would ignore the new node-id strings in G1_augm and assing integer values (e.g., over-writing node-ID 'A2' with '0').

Same behavior is expected even if the node-ids are integers.


rusty1s (approved):

rusty1s:
Thanks! I don't think the current implementation is correct (see failing tests), as edge indices are shuffled but not node features. Can you take a look?

DanielFPerez:
Hi! Yes, I have a work-related deadline on Nov 1st, so I can take a look at it after that if that's ok

rusty1s:
Sure, thanks!

PR #5826, opened 2022-10-25
refactor(sampler): separate `node_sample` and `edge_sample`

mananshah99:
As titled.


torch_geometric/sampler/neighbor_sampler_utils.py
```diff
@@ -0,0 +1,150 @@
...
+
+# Node-based sampling #########################################################
+
+
+def node_sample(
+    index: NodeSamplerInput,
+    sample_fn: Callable,
+    input_type: Optional[str] = None,
+    **kwargs,
+) -> Union[SamplerOutput, HeteroSamplerOutput]:
```
wsad1, line 21: Add a short doc explaining the arguments.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -359,103 +340,9 @@ def sample_from_edges(
...
-
-            output.metadata = (index, edge_label_index, edge_label,
-                               edge_label_time)
-
-        else:
-            raise TypeError(f"'{self.__class__.__name__}'' found invalid "
-                            f"type: '{self.data_cls}'")
-
-        return output
+        return edge_sample(index, self._sample, self.num_src_nodes,
```
wsad1, line 343: Do we need to move the implementation to `edge_sample` and `node_sample`. Can't we keep them here and have classes inherit from `NeighborSampler`?

torch_geometric/sampler/neighbor_sampler_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        output = sample_fn(seed=input_nodes, seed_time=input_time)
+        output.metadata = index
+
+    return output
+
+
+# Edge-based sampling #########################################################
+
+
+def edge_sample(
```
wsad1, line 44: add docs.

wsad1 (approved):
Thanks this looks good. 
One thing we could do in a follow up PR is separate out hetero and homo neighbour samplers into separate classes. With this we could avoid all these if conditions and just have single if condition in the loader when it chooses the appropriate sampler. WDYT?

torch_geometric/sampler/neighbor_sampler.py
mananshah99: I don't think it's easy for classes to inherit from NeighborSampler, since NeighborSampler makes a lot of assumptions about in-memory sampling that folks may want to avoid when writing a custom neighbor sampler. If we want to make NeighborSampler a more well-defined base class, I am open to it, but for now I think it makes more sense to just separate things out.

rusty1s (approved):
This looks all good to me, sorry for the delayed review. Since the functions of `node_sample` and `edge_sample` are tightly coupled to `NeighborSampler`, can we keep them in the same file? I don't think there is an urgent need for `neighbor_sampler_utils.py` right now.

PR #5833, opened 2022-10-27
Improve `data.validate()` to check for `edge_index.shape`

rusty1s:


torch_geometric/data/data.py
```diff
@@ -526,6 +526,14 @@ def validate(self, raise_on_error: bool = True) -> bool:
...
 
+        if 'edge_index' in self:
+            if self.edge_index.dim() != 2 or self.edge_index.size(0) != 2:
+                status = False
+                warn_or_raise(
+                    f"'edge_index' needs to be of shape [2, num_edges] in "
+                    f"'{cls_name}' (found {self.edge_index.size()})",
+                    raise_on_error)
+
         if 'edge_index' in self and self.edge_index.numel() > 0:
```
wsad1, line 537: combine the above `if` with this?

wsad1 (approved):

torch_geometric/data/data.py
rusty1s: Not possible since I also want to check in case of `numel() == 0`.

PR #5834, opened 2022-10-27
Add Bipartite GraphSage example

husimplicity:
This PR adds an example implementation of a recommendation model bipartite Graphsage.
The current implementation is a supervised version on MovieLens. The i2i co-occurrence matrix is applied to construct i2i-graph. When applying Graphsage, the model follows i-i-i pattern to encode item embedding and i-i-u for user-side respectively. Since the task to predict u2i rates is supervised,  we apply MSE loss with labels.
For next step, we will apply this model to an unsupervised task on another dataset with negative sampling and triplet loss.

### DIFF
--- a/examples/hetero/bipartite_sage.py
+++ b/examples/hetero/bipartite_sage.py
@@ -1,0 +1,268 @@
+import argparse
+import os.path as osp
+
+import torch
+import numpy as np
+import torch.nn.functional as F
+from torch.nn import Linear
+
+import torch_geometric.transforms as T
+from torch_geometric.datasets import MovieLens
+from torch_geometric.nn import SAGEConv
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--use_weighted_loss', action='store_true',
+                    help='Whether to use weighted MSE loss.')
+args = parser.parse_args()
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '../../data/MovieLens')
+user_sim_dict_path = 'swing_user_sim_dict.json'
+dataset = MovieLens(path, model_name='all-MiniLM-L6-v2')
+data = dataset[0].to(device)
+
+
+data['user'].x = torch.LongTensor(torch.arange(0, data['user'].num_nodes)).to(device)
+del data['user'].num_nodes
+
+# Add a reverse ('movie', 'rev_rates', 'user') relation for message passing:
+data = T.ToUndirected()(data)
+del data['movie', 'rev_rates', 'user'].edge_label  # Remove "reverse" label.
+
+# Perform a link-level split into training, validation, and test edges:
+train_data, val_data, test_data = T.RandomLinkSplit(
+    num_val=0.1,
+    num_test=0.1,
+    neg_sampling_ratio=0.0,
+    edge_types=[('user', 'rates', 'movie')],
+    rev_edge_types=[('movie', 'rev_rates', 'user')],
+)(data)
+
+
+def to_u2i_mat(edge_index, i_num, u_num):
+    # Convert bipartite edge_index format to matrix format
+    u2imat = torch.zeros((u_num, i_num))
+    for i in range(edge_index.shape[1]):
+        u2imat[edge_index[0][i], edge_index[1][i]] = 1
+
+    return u2imat
+
+
+def get_coocur_mat(train_mat, num_neighbors):
+    # Generate the co-occurrence matrix with normalisation and top-k filtering
+    A = train_mat.T @ train_mat
+    n_items = A.shape[0]
+    res_mat = torch.zeros((n_items, num_neighbors))
+    res_sim_mat = torch.zeros((n_items, num_neighbors))
+
+    items_D = torch.sum(A, axis=0).reshape(-1)
+    users_D = torch.sum(A, axis=1).reshape(-1)
+    beta_uD = (1 / torch.sqrt(users_D + 1)).reshape(-1, 1)
+    beta_iD = (1 / torch.sqrt(items_D + 1)).reshape(1, -1)
+    all_ii_constraint_mat = beta_uD @ beta_iD
+
+    for i in range(n_items):
+        row = all_ii_constraint_mat[i] * A[i]
+        row_sims, row_idxs = torch.topk(row, num_neighbors)
+        res_mat[i] = row_idxs
+        res_sim_mat[i] = row_sims
+
+    return res_mat.long(), res_sim_mat.float()
+
+
+def get_i2i_sim_graph(i2imat, i2isimmat):
+    # Generate the i2i graph from the co-occurrence matrix
+    i2i, i2i_sim = [[], []], []
+    for i in range(len(i2imat)):
+        # Filter nodes with no neighbors in co-occurrence matrix
+        if i2imat[i][0] == i and i2isimmat[i][0] > 1e-5:
+            i2i_sim.append(i2isimmat[i][1:])
+            for j in i2imat[i]:
+                i2i[0].append(i)
+                i2i[1].append(j)
+
+    return torch.tensor(i2i), torch.stack(i2i_sim)
+
+
+u2imat = to_u2i_mat(train_data.edge_index_dict[('user', 'rates', 'movie')],
+                    train_data['movie'].x.shape[0],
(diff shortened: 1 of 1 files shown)


examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '../../data/MovieLens')
+user_sim_dict_path = 'swing_user_sim_dict.json'
+dataset = MovieLens(path, model_name='all-MiniLM-L6-v2')
+data = dataset[0].to(device)
+
+data['user'].x = torch.LongTensor(torch.arange(
+    0, data['user'].num_nodes)).to(device)
```
rusty1s, line 26:
```suggestion
data = dataset[0]
data['user'].x = torch.arange(data['user'].num_nodes)
data = data.to(device)
```

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+train_data, val_data, test_data = T.RandomLinkSplit(
+    num_val=0.1,
+    num_test=0.1,
+    neg_sampling_ratio=0.0,
+    edge_types=[('user', 'rates', 'movie')],
+    rev_edge_types=[('movie', 'rev_rates', 'user')],
+)(data)
+
+
+def to_u2i_mat(edge_index, i_num, u_num):
```
rusty1s, line 43: You can actually use `utils.to_dense_adj` for this.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+    res_mat = torch.zeros((n_items, num_neighbors))
+    res_sim_mat = torch.zeros((n_items, num_neighbors))
+
+    items_D = torch.sum(A, axis=0).reshape(-1)
+    users_D = torch.sum(A, axis=1).reshape(-1)
+    beta_uD = (1 / torch.sqrt(users_D + 1)).reshape(-1, 1)
+    beta_iD = (1 / torch.sqrt(items_D + 1)).reshape(1, -1)
+    all_ii_constraint_mat = beta_uD @ beta_iD
+
+    for i in range(n_items):
```
rusty1s, line 65: Is there a need for this for-loop here? I assume you can apply `topk` in `dim=1` for this as well.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+        row = all_ii_constraint_mat[i] * A[i]
+        row_sims, row_idxs = torch.topk(row, num_neighbors)
+        res_mat[i] = row_idxs
+        res_sim_mat[i] = row_sims
+
+    return res_mat.long(), res_sim_mat.float()
+
+
+def get_i2i_sim_graph(i2imat, i2isimmat):
```
rusty1s, line 74: `nonzero()` + masking should work as well, right?

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+    return torch.tensor(i2i), torch.stack(i2i_sim)
+
+
+u2i_mat = to_u2i_mat(train_data.edge_index_dict[('user', 'rates', 'movie')],
+                     train_data['movie'].x.shape[0],
+                     train_data['user'].x.shape[0])
+i2i_mat, i2i_sim_mat = get_coocur_mat(u2i_mat, 9)
+i2i, _ = get_i2i_sim_graph(i2i_mat, i2i_sim_mat)
+
+# Add the generated i2i graph for high-order information
```
rusty1s, line 94: I assume you should use the `edge_index` from `train_data`, `val_data` and `test_data` for this. Currently, you are using it only from `train_data`.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+        self.conv1 = SAGEConv(-1, hidden_channels)
+        self.conv2 = SAGEConv(-1, hidden_channels)
+        self.lin1 = Linear(hidden_channels, hidden_channels)
+        self.lin2 = Linear(hidden_channels, out_channels)
+
+    def reset_parameters(self):
+        self.conv1.reset_parameters()
+        self.conv2.reset_parameters()
+        self.lin1.reset_parameters()
+        self.lin2.reset_parameters()
```
rusty1s, line 126: Can be dropped IMO.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def forward(self, x_dict, edge_index_dict):
```
rusty1s, line 128:
```suggestion
    def forward(self, x, edge_index):
```
No need to work with dicts here.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+        super().__init__()
+
+        self.conv1 = SAGEConv((-1, -1), hidden_channels)
+        self.conv2 = SAGEConv((-1, -1), hidden_channels)
+        self.conv3 = SAGEConv((-1, -1), hidden_channels)
+        self.lin1 = Linear(hidden_channels, hidden_channels)
+        self.lin2 = Linear(hidden_channels, hidden_channels)
+        self.lin3 = Linear(hidden_channels, out_channels)
+
+    def reset_parameters(self):
```
rusty1s, line 148: Can be dropped.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+
+class Model(torch.nn.Module):
+    def __init__(self, input_size, hidden_channels, out_channels):
+        super().__init__()
+        self.item_encoder = ItemGNNEncoder(hidden_channels, out_channels)
+        self.user_encoder = UserGNNEncoder(hidden_channels, out_channels)
+        self.decoder = EdgeDecoder(hidden_channels)
+        self.user_emb = Embedding(input_size, hidden_channels, device=device)
+
+    def reset_parameters(self):
```
rusty1s, line 200: Can be dropped.

examples/hetero/bipartite_sage.py
```diff
@@ -0,0 +1,262 @@
...
+        return z.view(-1)
+
+
+class Model(torch.nn.Module):
+    def __init__(self, input_size, hidden_channels, out_channels):
+        super().__init__()
+        self.item_encoder = ItemGNNEncoder(hidden_channels, out_channels)
+        self.user_encoder = UserGNNEncoder(hidden_channels, out_channels)
+        self.decoder = EdgeDecoder(hidden_channels)
+        self.user_emb = Embedding(input_size, hidden_channels, device=device)
```
rusty1s, line 198: Move to the top.

rusty1s (commented):
This is cool, thanks for adding! Left a few comments.

examples/hetero/bipartite_sage.py
husimplicity: Fixed.

examples/hetero/bipartite_sage.py
husimplicity: Sure, good advice!

examples/hetero/bipartite_sage.py
husimplicity: Good suggestion, thx.

examples/hetero/bipartite_sage.py
husimplicity: Our concern here is the information leak from val_data and test_data to train_data if we use the whole data to construct graph.

examples/hetero/bipartite_sage.py
husimplicity: dropped.

examples/hetero/bipartite_sage.py
husimplicity: fixed.

examples/hetero/bipartite_sage.py
husimplicity: dropped.

examples/hetero/bipartite_sage.py
husimplicity: dropped.

examples/hetero/bipartite_sage.py
husimplicity: moved.

Issue #5727, opened 2022-10-16
Model summary functionality for PyG models


nowyouseemejoe:
### 🚀 The feature, motivation and pitch

A model summary functionality to print a string summary of the network/model, like the `tensorflow.Keras.Model.summary()` method. A model summary would also help us debug the model and figure out the shapes of each layer's output. To be specific, `summary` can be implemented as a functional method, such that:
```python
from torch_geometric.nn import summary
model = GCN(...)
summary(model, ...)

# Outputs (just an example):
_________________________________________________________________
Layer (type)                 Output Shape               #Param   
=================================================================
GCNConv1                   (-1, 16)              4253864   
_________________________________________________________________
GCNConv2                   (-1, 7)                  244       
=================================================================
Total params: 4,254,108
Trainable params: 4,232,220
Non-trainable params: 21,888
_________________________________________________________________
```

I saw there is a summary method for datasets, would there be a summary method for models? Thanks!

### Alternatives

FYI: 
+ https://github.com/TylerYep/torchinfo
+ https://github.com/sksq96/pytorch-summary
+ https://github.com/ceykmc/pytorch_model_summary

rusty1s:
Does there exist some external package that already supports this? Otherwise, I am happy to include this one into PyG.

nowyouseemejoe:
Found an external package [torchinfo](https://github.com/TylerYep/torchinfo) that is not directly applicable to PyG-like inputs though. It would be amazing if PyG support this feature. Thanks!

rusty1s:
I see. I guess this one fails since shapes are not static in PyG. In that case, I am supportive of the idea on integrating it directly into PyG. Let me know if you want to work on this together.

nowyouseemejoe:
That would be great. Thank you @rusty1s @EdisonLeeeee

rusty1s:
This is now available thanks to @EdisonLeeeee

PR #5859, opened 2022-10-31
Add `summary` function for PyG/PyTorch models

EdisonLeeeee:
As discussed in https://github.com/pyg-team/pytorch_geometric/issues/5727, this PR is to add a summary function for all PyTorch particularly PyG defined models. Current implementation summarizes the information includes:
+ Layer names, 
+ Input/output shapes,
+ Number of parameters, 
+ Excutation time.

## Usage
```python
import torch
from torch_geometric.nn import summary
from torch_geometric.nn.models import GCN

model = GCN(128, 64, 2, out_channels=32)
x = torch.randn(100, 128)
edge_index = torch.randint(100, size=(2, 20))
print(summary(model, x, edge_index))

+-----------------------------------------+---------------------+----------------+----------+--------+
| Layer                                   | Input Shape         | Output Shape   | #Param   | Time   |
|-----------------------------------------+---------------------+----------------+----------+--------|
| GCN                                     | [100, 128], [2, 20] | [100, 32]      | 10,336   | 0.0008 |
| ├─(act)ReLU                             | [100, 64]           | [100, 64]      | --       | 0.0000 |
| ├─(convs)ModuleList                     | --                  | --             | 10,336   | --     |
| │    └─(0)GCNConv                       | [100, 128], [2, 20] | [100, 64]      | 8,256    | 0.0006 |
| │    │    └─(aggr_module)SumAggregation | [119, 64], [119]    | [100, 64]      | --       | 0.0000 |
| │    │    └─(lin)Linear                 | [100, 128]          | [100, 64]      | 8,192    | 0.0001 |
| │    └─(1)GCNConv                       | [100, 64], [2, 20]  | [100, 32]      | 2,080    | 0.0002 |
| │    │    └─(aggr_module)SumAggregation | [119, 32], [119]    | [100, 32]      | --       | 0.0000 |
| │    │    └─(lin)Linear                 | [100, 64]           | [100, 32]      | 2,048    | 0.0000 |
+-----------------------------------------+---------------------+----------------+----------+--------+
```

Set `max_depth=1` for a better view
```python
print(summary(model, x, edge_index, max_depth=1))

+---------------------+---------------------+----------------+----------+--------+
| Layer               | Input Shape         | Output Shape   | #Param   | Time   |
|---------------------+---------------------+----------------+----------+--------|
| GCN                 | [100, 128], [2, 20] | [100, 32]      | 10,336   | 0.0021 |
| ├─(act)ReLU         | [100, 64]           | [100, 64]      | --       | 0.0000 |
| ├─(convs)ModuleList | --                  | --             | 10,336   | --     |
+---------------------+---------------------+----------------+----------+--------+
```

## Update
```python
import torch
from torch_geometric.nn import summary
from torch_geometric.nn.models import GCN

model = GCN(128, 64, 2, out_channels=32)
x = torch.randn(100, 128)
edge_index = torch.randint(100, size=(2, 20))
print(summary(model, x, edge_index))

+---------------------+---------------------+----------------+----------+
| Layer               | Input Shape         | Output Shape   | #Param   |
|---------------------+---------------------+----------------+----------|
| GCN                 | [100, 128], [2, 20] | [100, 32]      | 10,336   |
| ├─(act)ReLU         | [100, 64]           | [100, 64]      | --       |
| ├─(convs)ModuleList | --                  | --             | 10,336   |
| │    └─(0)GCNConv   | [100, 128], [2, 20] | [100, 64]      | 8,256    |
| │    └─(1)GCNConv   | [100, 64], [2, 20]  | [100, 32]      | 2,080    |
+---------------------+---------------------+----------------+----------+
```

Any suggestion would be highly appreciated :)


torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
+from collections import defaultdict
```
rusty1s, line 1: A small test would be cool :)

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        x = torch.randn(100, 128)
+        edge_index = torch.randint(100, size=(2, 20))
+        print(summary(model, x, edge_index))
+
+    Args:
+        model (nn.Module): PyTorch model to summarize.
+        inputs (Sequence[Any]): Arguments for the model's
+            forward pass.
+        device (torch.device): Device for model and input_data.
+            (default: :obj:`"cpu"`)
```
rusty1s, line 35: Do we desperately need this argument? If the model is on GPU, inputs should be on GPU as well.

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        max_depth (int): Depth of nested layers to display.
+            Nested layers below this depth will not be displayed
+            in the summary. (default: :obj:`"3"`)
+    """
+    def register_pre_hook(info):
+        def hook(module, input):
+            info['input_time'].append(perf_counter())
```
rusty1s, line 42: I think time information is somewhat cool but gets tricky to implement on CUDA. We would either need CUDA events to track or synchronize the model in order to get correct timings.

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        hooks[module_id] = (
+            module.register_forward_pre_hook(register_pre_hook(info)),
+            module.register_forward_hook(register_hook(info)),
+        )
+        module_items = reversed(module._modules.items())
+        stack += [(f"({name}){get_name(mod)}", mod, depth + 1)
+                  for name, mod in module_items if mod is not None]
+
+    # set `eval` mode
+    model.eval()
```
rusty1s, line 90: Need to reset after execution?

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+    # make a forward pass
+    device = torch.device(device)
+    with torch.no_grad():
```
rusty1s, line 93: Let's drop this here :)

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        def hook(module, input):
+            info['input_time'].append(perf_counter())
+
+        return hook
+
+    def register_hook(info):
+        def hook(module, input, output):
+            info['input_shape'].append(get_shape(input))
+            info['output_shape'].append(get_shape(output))
+            info['output_time'].append(perf_counter())
```
rusty1s, line 50: First save `output_time` and then remaining stats. We don't want to benchmark `get_shape()` here :)

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+
+        return hook
+
+    hooks = {}
+    depth = 0
+    stack = [(get_name(model), model, depth)]
+
+    info_list = []
+    input_shape = defaultdict(list)
```
rusty1s, line 59: Any reason these need to be lists?

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        var_name, module, depth = stack.pop()
+        module_id = id(module)
+        if module_id in hooks:
+            for hook in hooks[module_id]:
+                hook.remove()
+
+        info = {}
+        info['name'] = var_name
+        info['layer'] = module
+        info['input_shape'] = input_shape[id(module)]
```
rusty1s, line 73: We can also set these in the forward hook, such that this loop only registers the hook.

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+            info['#param'], info['duration']
+        ]
+        content.append(row)
+
+    body = tabulate(content, headers='firstrow', tablefmt='psql')
+
+    return body
+
+
+def get_shape(input):
```
rusty1s, line 144: Type hints.

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+        input = (input, )
+    out = []
+    for x in input:
+        if hasattr(x, 'size'):
+            out.append(str(list(x.size())))
+    out = ', '.join(out)
+    return out
+
+
+def get_name(module):
```
rusty1s, line 155: Type hints.

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,156 @@
...
+
+import torch
+import torch.nn as nn
+
+
+def summary(
+    model: nn.Module,
+    *inputs: Sequence[Any],
+    device: torch.device = 'cpu',
+    max_depth: int = 3,
```
rusty1s, line 13: In addition to `max_depth`, we could think about adding a function about which layers should be treated as "leaf" modules. For example, it probably makes sense to let any instance of `MessagePassing` be treated as a leaf module. WDYT?

torch_geometric/nn/summary.py
EdisonLeeeee: Agreed.

torch_geometric/nn/summary.py
EdisonLeeeee: Makes sense. Let's drop the time information from the summarized results.

torch_geometric/nn/summary.py
EdisonLeeeee: That seems weird but it makes sense for cases with some modules reused.  For example:
```python
relu = torch.nn.ReLU()
model = torch.nn.Sequential(torch.Linear(10, 10), relu, torch.nn.Linear(10, 10), relu)
```
where `relu` module would be called twice during forward but we only registered one hook for this module. Therefore, a list is used for sharing the input/output information for `info_list[0] (first relu)` and `info_list[3] (second relu)`. That also gets tricky but currently I don't have a better idea to tackle this case. WDYT?

torch_geometric/nn/summary.py
EdisonLeeeee: That was my first attempt, but the problem is that some of the modules (e.g. `torch.nn.ModuleList`) don't have a forward method implemented and would be ignored then.

torch_geometric/nn/summary.py
EdisonLeeeee: Got it. That's indeed necessary.

torch_geometric/nn/summary.py
EdisonLeeeee: Sorry, did you mean to drop this line `with torch.no_grad()`? Any reason?

torch_geometric/nn/summary.py
EdisonLeeeee: Not sure I understand you, but I'll have a try.

torch_geometric/nn/summary.py
rusty1s: No I meant the `device` :D

test/nn/test_model_summary.py
```diff
@@ -0,0 +1,112 @@
...
+
+@withPackage('tabulate')
+def test_basic_summary() -> None:
+
+    model = GCN(32, 16, num_layers=2, out_channels=32)
+    x = torch.randn(100, 32)
+    edge_index = torch.randint(100, size=(2, 20))
+    adj_t = SparseTensor.from_edge_index(edge_index,
+                                         sparse_sizes=(100, 100)).t()
+    print(summary(model, x, edge_index))
```
Padarn, line 45: Can we actually test the output here?

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,146 @@
...
+from torch_sparse import SparseTensor
+
+from torch_geometric.nn.conv import MessagePassing
+
+
+def summary(
+    model: nn.Module,
+    *inputs: Sequence[Any],
+    max_depth: int = 3,
+    leaf_module: Union[nn.Module, Tuple[nn.Module],
```
Padarn, line 16:
```suggestion
    leaf_modules: List[Union[nn.Module, Tuple[nn.Module],
```
Should  this be a list? Maybe multiple modules you want to treat this way?

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,146 @@
...
+            info['name'] = prefix + info['name']
+
+        if info['input_shape']:
+            info['input_shape'] = info['input_shape'].pop(0)
+            info['output_shape'] = info['output_shape'].pop(0)
+        else:
+            info['input_shape'] = "--"
+            info['output_shape'] = "--"
+
+    from tabulate import tabulate
```
Padarn, line 113: [nit] can we move this to the top of the function definition? feel its easier to debug

test/nn/test_model_summary.py
EdisonLeeeee: Currently I have no idea to test the output string as it's too long and complicated :(

torch_geometric/nn/summary.py
EdisonLeeeee: No, it is used to check if the input model `isinstance(model, A)` or `isinstance(model, (A, B))`. So it's better to be a tuple if multiple modules are provided.

torch_geometric/nn/summary.py
EdisonLeeeee: Sounds good!

torch_geometric/nn/summary.py
```diff
@@ -0,0 +1,152 @@
...
+        if isinstance(x, SparseTensor):
+            out.append(str(list(x.sparse_sizes())))
+        elif hasattr(x, 'size'):
+            out.append(str(list(x.size())))
+    out = ', '.join(out)
+    return out
+
+
+def get_name(module: nn.Module) -> str:
+    return str(module.__class__).split(".")[-1].split("'")[0]
```
rfdavid, line 115: You can use:
`return module.__class__.__name__`

test/nn/test_model_summary.py
rfdavid: For the tests, I know the output is too long, so I was thinking of having some method that outputs the list (content) from `make_table` method before tabulate, so that you can check the values. However, I still believe the best way is to test the table as the library tabulate do in [test_output.py](https://github.com/zackdever/python-tabulate/blob/master/test/test_output.py).

I also suggest splitting test_summary_basic using fixtures:

```  
@pytest.fixture
def gcn(): 
    model = GCN(32, 16, num_layers=2, out_channels=32)
    x = torch.randn(100, 32)
    edge_index = torch.randint(100, size=(2, 20))
    adj_t = SparseTensor.from_edge_index(edge_index,
                                         sparse_sizes=(100, 100)).t()
    return { 'model': model,
             'x': x,
             'edge_index': edge_index,}
             'adj_t': adj_t }

    
def test_summary_edge_index(gcn) -> None:
    expected = '\n'.join(['+---------------------+--------------------+----------------+----------+',
                          '| Layer               | Input Shape        | Output Shape   | #Param   |',
                          '|---------------------+--------------------+----------------+----------|',
                          '| GCN                 | [100, 32], [2, 20] | [100, 32]      | 1,072    |',
                          '| ├─(act)ReLU         | [100, 16]          | [100, 16]      | --       |',
                          '| ├─(convs)ModuleList | --                 | --             | 1,072    |',
                          '| │    └─(0)GCNConv   | [100, 32], [2, 20] | [100, 16]      | 528      |',
                          '| │    └─(1)GCNConv   | [100, 16], [2, 20] | [100, 32]      | 544      |',
                          '+---------------------+--------------------+----------------+----------+'])

    assert summary(gcn['model'], gcn['x'], gcn['edge_index']) == expected

def test_summary_adj_t(gcn) -> None:
    expected = '\n'.join(['+---------------------+-----------------------+----------------+----------+',
                         '| Layer               | Input Shape           | Output Shape   | #Param   |',
                         '|---------------------+-----------------------+----------------+----------|',
                         '| GCN                 | [100, 32], [100, 100] | [100, 32]      | 1,072    |',
                         '| ├─(act)ReLU         | [100, 16]             | [100, 16]      | --       |',
                         '| ├─(convs)ModuleList | --                    | --             | 1,072    |',
                         '| │    └─(0)GCNConv   | [100, 32], [100, 100] | [100, 16]      | 528      |',
                         '| │    └─(1)GCNConv   | [100, 16], [100, 100] | [100, 32]      | 544      |',
                         '+---------------------+-----------------------+----------------+----------+'])

    assert summary(gcn['model'], gcn['x'], gcn['adj_t']) == expected

```

I hope that helps.

torch_geometric/nn/summary.py
EdisonLeeeee: Thanks, much cleaner!

test/nn/test_model_summary.py
EdisonLeeeee: Super! I'll update tests taking your suggestion into account, thanks!

torch_geometric/nn/summary.py
Padarn: Oh I see your point 👍

test/nn/test_model_summary.py
```diff
@@ -0,0 +1,196 @@
...
+        '| │    │    └─(aggr_module)SumAggregation | [120, 16], [120]   | [100, 16]      | --       |',  # noqa
+        '| │    │    └─(lin)Linear                 | [100, 32]          | [100, 16]      | 512      |',  # noqa
+        '| │    └─(1)GCNConv                       | [100, 16], [2, 20] | [100, 32]      | 544      |',  # noqa
+        '| │    │    └─(aggr_module)SumAggregation | [120, 32], [120]   | [100, 32]      | --       |',  # noqa
+        '| │    │    └─(lin)Linear                 | [100, 16]          | [100, 32]      | 512      |',  # noqa
+        '+-----------------------------------------+--------------------+----------------+----------+',  # noqa
+    ])
+
+    assert summary(model, x, edge_index, leaf_module=None) == expected
+
```
rfdavid, line 101: Neat! I would recommend having only one assert per test (split test_basic_summary into multiple functions and use that fixture to avoid re-computation). It is easier to figure out what is wrong when a test fails, and if there is more than one failed test, we can see all of them at once.

EdisonLeeeee: Sounds great! Will update tests soon. Thanks for your helpful suggestions!

test/nn/test_model_summary.py
```diff
@@ -0,0 +1,217 @@
...
+        '| Layer                     | Input Shape   | Output Shape   | #Param   |',  # noqa
+        '|---------------------------+---------------+----------------+----------|',  # noqa
+        '| ModuleDictModel           | [100, 32]     | [100, 32]      | 1        |',  # noqa
+        '| ├─(activations)ModuleDict | --            | --             | 1        |',  # noqa
+        '| │    └─(lrelu)LeakyReLU   | --            | --             | --       |',  # noqa
+        '| │    └─(prelu)PReLU       | [100, 32]     | [100, 32]      | 1        |',  # noqa
+        '+---------------------------+---------------+----------------+----------+',  # noqa
+    ])
+
+    summary(model, x, 'prelu') == expected
```
rfdavid, line 197: missing assert here

EdisonLeeeee: Fixed. Thanks for catching.

rusty1s (approved):
This is amazing, thank you! Sorry for taking me so long to have a final look.

PR #5866, opened 2022-11-01
Add `GNNFF` model

ken2403:
This PR adds an implementation of a graph convolutional neural network model called [`GNNFF`.](https://www.nature.com/articles/s41524-021-00543-3)

- It uses the same inputs as the already implemented [`DimeNet`.](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/dimenet.py)
- It is fast because it can find the forces between atoms in a material without automatic differentiation.
- It constructs a force field that is translationally-invariant and rotationally-covariant with respect to the atomic coordinates.


test/nn/models/test_gnnff.py
```diff
@@ -0,0 +1,38 @@
+import torch
+import torch.nn.functional as F
+
+from torch_geometric.data import Data
+from torch_geometric.nn import GNNFF
+from torch_geometric.testing import is_full_test, onlyFullTest
+
+
+@onlyFullTest
```
rusty1s, line 9:
```suggestion
from torch_geometric.testing import is_full_test


```

ken2403:
Thank you for your kind review.
I'm proud to have contributed in any small way to such a great projec!

I am one of the users who was fascinated by the tremendous potential of PyG!
I will continue to actively contribute!

Issue #5867, opened 2022-11-01
[Roadmap] PyTorch SparseTensor Support 🚀


EdisonLeeeee:
### 🚀 The feature, motivation and pitch

# 🚀 The feature, motivation and pitch
PyG currently accepts `torch.LongTensor: edge_index` and `torch_sparse.SparseTensor: adj` in `MessagePassing`, which limits its flexibility for users who use native PyTorch SparseTensor. As such, an additional step is required for them to convert one to another.

The goal of this roadmap is to track the integration of native PyTorch SparseTensor support into PyG. After integration, the `edge_index` in `MessagePassing` can also accept PyTorch SparseTensor while allowing backpropagation like `torch_sparse.SparseTensor`.

## General Roadmap

- [ ] Add PyTorch SparseTensor support for `torch_geometric.nn.MessagePassing`
- [ ] Add PyTorch SparseTensor support for `torch_geometric.transforms`
- [ ] Add PyTorch SparseTensor support for `torch_geometric.loader`
- [ ] Replace `torch_sparse.SparseTensor` based implementations with PyTorch functionality
- [ ] ...

EdisonLeeeee:
So far I can think of a way to support torch SparseTensor with the least effort in `MessagePassing` is:
```python
def hook(model, input):
    edge_index, size, kwargs = input
    if is_torch_coo_tensor(adj):
        edge_index = adj._indices()
        kwargs['edge_weight'] = adj._values()
    return edge_index, size, kwargs

layer.register_propagate_forward_pre_hook(hook)
``` 
we can register such a hook in `__init__` of `MessagePassing`, WDYT? @rusty1s

rusty1s:
I think this would only be a temporary solution. Ultimately, we want direct support for it similar to `SparseTensor`. Wondering what are the main challenges to support this?

EdisonLeeeee:
IMO, there are some challenges to support this:

1. It only supports `sum` aggregation when using PyTorch SparseTensor. For other cases including advanced aggregations, we still need to convert it to the form of `(edge_index, edge_weight)`.
2. PyTorch SparseTensor does not support `spspmm` or `mul` with broadcasting, which makes it challenging to implement some operations such as `gcn_norm` (e.g., $A\cdot D^{-\frac{1}{2}}$ ) efficiently.
[...]

PR #5881, opened 2022-11-02
Enable `SparseTensor` in training benchmarks

dszwicht:
Enable `SparseTensor` in training benchmarks.


dszwicht:
Hi @rusty1s, while verifying this I noticed that `rgcn + ogbn-mag` case is failing with `Floating point exception` coming from `torch_sparse.matmul`. Could you please help to identify the root cause?

rusty1s:
Will check.

dszwicht:
> Mh, it runs through for me both on CPU and CUDA, but I needed to remove the `data = dataset.to(device)` call and replace it with `data = dataset` in order to let the script succeed (we don't support sampling on GPU right now).

@rusty1s, I tried to run `rgcn + ogbn-mag` on a fresh build - got the same result. Let me just confirm, did you pass `--use-sparse-tensor` option to the benchmark script?

rusty1s:
Yes, I was using `--use-sparse-tensor` option, and verified it is working correctly. Mmh. Is that for CPU or GPU execution?

We are also working on #5867 in order to deprecate `torch_sparse.SparseTensor` soon, so let's not waste too much time on it.

dszwicht:
> Yes, I was using `--use-sparse-tensor` option, and verified it is working correctly. Mmh. Is that for CPU or GPU execution?
> 
> We are also working on #5867 in order to deprecate `torch_sparse.SparseTensor` soon, so let's not waste too much time on it.

@rusty1s , it failed in both cases, but I just found out that in a script that runs a bunch of examples, `--num-layers` was set to 1 <- this is the case where failure occurs.

I guess we can proceed with this PR.

rusty1s (approved):

PR #5888, opened 2022-11-03
Add `SparseTensor` support for `SuperGATConv`

EdisonLeeeee:
To address https://github.com/pyg-team/pytorch_geometric/discussions/5885

### DIFF
--- a/test/nn/conv/test_supergat_conv.py
+++ b/test/nn/conv/test_supergat_conv.py
@@ -18,8 +18,7 @@ def test_supergat_conv(att_type):
     assert out.size() == (4, 64)
 
     adj_t = SparseTensor.from_edge_index(edge_index, sparse_sizes=(4, 4)).t()
-    out2 = conv(x, adj_t)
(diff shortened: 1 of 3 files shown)


test/nn/conv/test_supergat_conv.py
```diff
@@ -16,6 +17,10 @@ def test_supergat_conv(att_type):
     out = conv(x, edge_index)
     assert out.size() == (4, 64)
 
+    adj_t = SparseTensor.from_edge_index(edge_index, sparse_sizes=(4, 4)).t()
+    out2 = conv(x, adj_t)
+    assert out2.size() == (4, 64)
+
```
wsad1, line 23:
```suggestion
assert torch.allclose(out, out2)
```
Can we also compare the output of `get_attention_loss` ?

wsad1 (approved):
Thanks for fixing this.

rusty1s (approved):
Thanks!

PR #5898, opened 2022-11-04
Enable to set custom number of steps in benchmarks

kgajdamo:
- Added --num-steps argument, that allows to set number of the steps (batches two iterate over) in inference and training benchmarks.
- Added wrapper for NeighborLoader, that handles custom number of steps.


yanbing-j:
Hi @kgajdamo , thanks for your hard work! For benchmark purpose, the function of setting `num_steps` is very useful, which decrases much time that wasted in running the entire epoch. LGTM!

torch_geometric/loader/steps_loader.py
```diff
@@ -0,0 +1,44 @@
+class StepsLoader():
```
rusty1s, line 1: I personally don't think we need this. This should be automatically supported by adding a custom sampler from PyTorch to `NeighborLoader`, e.g.:
```python
sampler = torch.utils.data.RandomSampler(range(len(train_nodes)), num_samples=num_steps * batch_size)
```

kgajdamo: This is a very good idea, but it doesn't work with shuffle. So it will work for the inference but not for the training. Of course in training benchmark steps handling can be done just in a train_homo/train_hetero function so if You prefer such option I can switch to the one You proposed.

mszarma: @rusty1s @kgajdamo what about adding it as bench supporting/specific to https://github.com/pyg-team/pytorch_geometric/blob/master/benchmark/utils/utils.py ?

kgajdamo: Ok, so I checked this RandomSampler and it already do shuffle. So I think we can use it, but then the shuffle argument we pass to NeighborLoader must be set to false. And everything should work.

kgajdamo: I have made corrections. Please take another look :)

benchmark/training/training_benchmark.py
```diff
@@ -96,13 +103,9 @@ def run(args: argparse.ArgumentParser) -> None:
...
-                        data,
-                        num_neighbors=num_neighbors,
-                        input_nodes=mask,
-                        batch_size=batch_size,
-                        shuffle=True,
-                        num_workers=args.num_workers,
-                    )
+                        data, num_neighbors=num_neighbors, input_nodes=mask,
+                        batch_size=batch_size, shuffle=shuffle,
+                        num_workers=args.num_workers, sampler=sampler)
```
rusty1s, line 108:
```suggestion
                        num_workers=args.num_workers, sampler=sampler, )
```

benchmark/inference/inference_benchmark.py
```diff
@@ -29,6 +29,12 @@ def run(args: argparse.ArgumentParser) -> None:
         hetero = True if dataset_name == 'ogbn-mag' else False
         mask = ('paper', None) if dataset_name == 'ogbn-mag' else None
         degree = None
+
+        assert args.num_steps == -1 or args.num_layers == [
+            1
+        ] or hetero, 'In case of the custom number of steps \
+        and layer-wise inference, the number of layers must be equal to 1.'
```
rusty1s, line 36:
```suggestion
        if args.num_layers != [1] and not hetero:
            assert args.num_steps == -1, 'Layer-wise inference requires full `steps=-1`'
```

rusty1s (approved):

benchmark/inference/inference_benchmark.py
```diff
@@ -29,6 +29,10 @@ def run(args: argparse.ArgumentParser) -> None:
         hetero = True if dataset_name == 'ogbn-mag' else False
         mask = ('paper', None) if dataset_name == 'ogbn-mag' else None
         degree = None
+
+        if args.num_layers != [1] and not hetero:
+            assert args.num_steps == -1, 'Layer-wise inference requires full `steps=-1`'
```
rusty1s, line 34:
```suggestion
        if args.num_layers != [1] and not hetero and args.num_steps != -1:
            raise ValueError("Layer-wise inference requires `steps=-1`")
```

rusty1s (approved):

kgajdamo:
@rusty1s please do not merge it yet. I noticed that for the same number of steps the version with RandomSampler is x4 faster. While the time should be the same. I haven't had time to look into it yet.

kgajdamo:
@rusty1s please do not merge it yet. I noticed that for the same number of steps the version with RandomSampler is x4 faster. While the time should be the same. I haven't had time to look into it yet.

rusty1s:
Ok :)

kgajdamo:
@rusty1s I checked it and found out what was the reason off such timing. I provided wrong argument to the RandomSampler:
```
sampler = torch.utils.data.RandomSampler(
                     **range(len(data))**, num_samples=args.num_steps *
                     batch_size)
```
But when I put instead of `range(len(data))` - `range(len(data.train_mask))` I am receiving an index out of range error. What do You think about it?

rusty1s:
I think you need to use `range(int(data.train_mask.sum()))`.

kgajdamo:
> I think you need to use `range(int(data.train_mask.sum()))`.

@rusty1s Thank You! Now it's working and You can merge.

PR #5902, opened 2022-11-04
Fix `cached_property` import on Python 3.7

rusty1s:


EdisonLeeeee:
I think we can use `lru_cache` + `property` to implement `cached_property`,  which could also benefit the data loading, e.g., 
```python
    def _dataset(self) -> ConcatDataset:
        dataset = ConcatDataset(self._partitions)

        if self.name == "small":
            dataset = self._load_small_split(dataset)

        return dataset

    if cached_property is not None:
        _dataset = cached_property(_dataset)
    else:
        _dataset = property(fget=lru_cache(maxsize=1)(_dataset))
```

rusty1s:
Yes, that works as well. I guess it isn't super important since most users are on > Python 3.7, so I wanted to unblock ASAP. Feel free to fix this :)

EdisonLeeeee:
Actually, `cached_property` has been introduced since 3.8 :D

I can make a PR for it though.

FYI: https://stackoverflow.com/questions/49436833/difference-between-cached-property-and-lru-cache-decorator

EdisonLeeeee:
One more thing, the following code cannot make a function a property:
```python
    def cached_property(func):
        return func
```

Just a gentle reminder :)

rusty1s:
Yes, you are absolutely correct. Looks like I was in weekend mode already :(

PR #5906, opened 2022-11-05
Add `spmm` and `is_torch_sparse_tensor`

EdisonLeeeee:
This PR implements:
+ `spmm` function for both PyTorch and torch_sparse SparseTensor
+ `is_torch_sparse_tensor` function to check if a tensor is a PyTorch SparseTensor

TODO:
+ Add test and docstring for `spmm`

Any comments would be appreciated :)


torch_geometric/nn/functional/__init__.py
```diff
@@ -1,6 +1,7 @@
 from .bro import bro
 from .gini import gini
+from .spmm import spmm
 
-__all__ = ['bro', 'gini']
+__all__ = ['bro', 'gini', 'spmm']
```
rusty1s, line 5:
```suggestion
__all__ = ['bro', 'gini', 'spmm', ]
```

torch_geometric/nn/functional/spmm.py
```diff
@@ -0,0 +1,22 @@
+import torch
+from torch import Tensor
+from torch_sparse import SparseTensor
+
+from torch_geometric.utils import is_torch_sparse_tensor
+
+
+def spmm(src: SparseTensor, other: Tensor, reduce: str = "sum") -> Tensor:
```
rusty1s, line 8: I personally think we should move this to `utils` as well since I see it potentially used in other places as well, e.g., transforms.

rusty1s, line 8: Can we make this TorchScript compatible via the `overload` decorator?

torch_geometric/utils/torch_sparse_tensor.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def is_torch_sparse_tensor(x: Any) -> bool:
+    """Returns :obj:`True` if the input :obj:`x` is
+    PyTorch SparseTensor (COO or CSR format).
+
+    Args:
+        x (Any): The input object to be checked.
+    """
+    return getattr(x, 'is_sparse', False)
```
rusty1s, line 11:
```suggestion
    return isinstance(x, Tensor) and x.is_sparse
```

torch_geometric/nn/functional/spmm.py
EdisonLeeeee: Ok, got it.

EdisonLeeeee: Sure :)

EdisonLeeeee:
Thanks @rusty1s for the updates and making it merged :)

PR #5919, opened 2022-11-07
Add support for customizable interaction graph in `SchNet` model 

hatemhelal:
This introduces a new `interaction_graph` arg that defaults to `radius_graph` to allow callers to customize how interatomic edges are defined in the `SchNet` architecture.

This removes the support that was added for pre-computed edges in #5401 so also updated the example to reflect this change.


torch_geometric/nn/models/schnet.py
```diff
@@ -289,7 +296,11 @@ def forward(self, z: Tensor, pos: Tensor, batch: OptTensor = None,
         if not self.dipole and self.atomref is not None:
             h = h + self.atomref(z)
 
-        out = scatter(h, batch, dim=0, reduce=self.readout)
+        # zero out embeddings for any padding atoms
+        h.masked_fill_((z == 0).view(-1, 1), 0.)
```
hatemhelal, line 300: this can probably be removed

torch_geometric/nn/models/schnet.py
```diff
@@ -253,21 +267,13 @@ def forward(self, z: Tensor, pos: Tensor, batch: OptTensor = None,
             batch (LongTensor, optional): Batch indices assigning each atom to
                 a separate molecule with shape :obj:`[num_atoms]`.
                 (default: :obj:`None`)
-            edge_index (LongTensor, optional): Indices of interacting pairs of
-                atoms with shape :obj:`[2, num_edges]`. Will be computed in
-                every forward pass if not provided. (default: :obj:`None`)
         """
-        assert z.dim() == 1 and z.dtype == torch.long
+        assert z.dim() == 1 and z.dtype in (torch.uint8, torch.int8,
+                                            torch.int32, torch.int64)
```
rusty1s, line 272:
```suggestion
        assert z.dim() == 1 and not z.dtype.is_floating_point()
```

torch_geometric/nn/models/schnet.py
```diff
@@ -90,12 +96,14 @@ def __init__(
...
         cutoff: float = 10.0,
+        interaction_graph: Optional[Callable] = None,
         max_num_neighbors: int = 32,
         readout: str = 'add',
         dipole: bool = False,
         mean: Optional[float] = None,
         std: Optional[float] = None,
-        atomref: OptTensor = None,
+        atomref: Optional[torch.Tensor] = None,
+        batch_size: Optional[int] = None,
```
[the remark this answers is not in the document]
hatemhelal: I think we could drop it at least for now: the background is we need to know sizes through out the computational graph when we compile a static graph.  I think we might be able to do this using a `torch.fx` pass but I'm still figuring out the details.

torch_geometric/nn/models/schnet.py
```diff
@@ -90,12 +96,14 @@ def __init__(
         num_interactions: int = 6,
         num_gaussians: int = 50,
         cutoff: float = 10.0,
+        interaction_graph: Optional[Callable] = None,
```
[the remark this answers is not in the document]
hatemhelal: Sure I'll add docs to `RadiusInteractionGraph` in this file and can figure out how to link it to this module.  Let me know if you are aware of some examples for how to accomplish this with sphinx

rusty1s (approved):

torch_geometric/nn/models/schnet.py
```diff
@@ -268,7 +268,7 @@ def forward(self, z: Tensor, pos: Tensor,
                 a separate molecule with shape :obj:`[num_atoms]`.
                 (default: :obj:`None`)
         """
-        assert z.dim() == 1 and not z.dtype.is_floating_point
+        assert z.dim() == 1 and not z.dtype.is_floating_point()
```
hatemhelal, line 271: @rusty1s Did something change in torch?  I get this error when I try to invoke `is_floating_point` 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [20], line 2
      1 x = torch.arange(10)
----> 2 x.dtype.is_floating_point()

TypeError: 'bool' object is not callable
```

I'm on torch 1.13 in case that makes a difference?

rusty1s: I replaced it with `torch.is_floating_point(z)`. Hopefully, this will work across all versions.

hatemhelal: sounds good to me

PR #5930, opened 2022-11-08
Huggingface hub integration 5545

wwymak:
This is an initial draft of what the wrapper could look like. 

Some questions I have:
- is this a sensible approach? (I have overridden the from_pretrained and save_pretrained methods so we can have defaults for the model cards, and a slightly different way of initialising the PyG model as some of them require the data at init time)
- should the wrapper live somewhere else in the codebase? 
- do we want to have a nicer template for PyG models or just use the default one? Are there fields we want to have prepopulated by default
- should the wrapper also contain methods to log the training/evaluation metrics by default?
@rusty1s would be great if you comment :)

The screenshot is an example of what it currently looks like if you use the wrapper to upload to huggingface model hub
Screenshot from 2022-11-08 22-08-56


torch_geometric/nn/models/huggingface_hub_wrapper.py
```diff
@@ -0,0 +1,124 @@
+import json
```
rusty1s, line 1: Just name this `model_hub.py`?

torch_geometric/nn/models/huggingface_hub_wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+import logging
+import os
+from pathlib import Path
+from typing import Dict, Optional
+
+import requests
+from huggingface_hub import (
```
rusty1s, line 8: I guess you would need to wrap this around `try` and `except ImportError` to treat `huggingface_hub` correctly as an optional dependency.

torch_geometric/nn/models/huggingface_hub_wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+    hf_hub_download,
+)
+
+CONFIG_NAME = 'config.json'
+MODEL_HUB_ORGANIZATION = "pytorch_geometric"
+
+
+def hfhub_model(PyGModel, dataset=None):
+    # @wraps(PyGModel)
+    class HFModel(PyGModel, PyTorchModelHubMixin):
```
rusty1s, line 21:
```suggestion
    class ModelHubMixin(PyGModel, PyTorchModelHubMixin):
```

torch_geometric/nn/models/huggingface_hub_wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+    ModelCardData,
+    PyTorchModelHubMixin,
+    hf_hub_download,
+)
+
+CONFIG_NAME = 'config.json'
+MODEL_HUB_ORGANIZATION = "pytorch_geometric"
+
+
+def hfhub_model(PyGModel, dataset=None):
```
rusty1s, line 19: Can you add some basic doc-string on how to use this?

rusty1s, line 19: I originally thought we should treat this as a `Mixin` as well. For example, `SchNet(ModelHubMixin)` now provides a function `from_pretrained(..., dataset='QM9')`.

rusty1s:
This looks great. Happy to fine-tune it in later PRs as well.

> is this a sensible approach? (I have overridden the from_pretrained and save_pretrained methods so we can have defaults for the model cards, and a slightly different way of initialising the PyG model as some of them require the data at init time)

I think so. Currently, we are creating a new model while I would prefer something like `SchNet(ModelHubMixin)`.

> should the wrapper live somewhere else in the codebase?

I think it should live in `torch_geometric.nn` rather than `torch_geometric.nn.models`.

> do we want to have a nicer template for PyG models or just use the default one? Are there fields we want to have prepopulated by default

I think the default one works for now. Happy to fine-tune it in later PRs if that is possible. Please feel free to get creative :)

> should the wrapper also contain methods to log the training/evaluation metrics by default?

I am not sure. Let's go with a basic version first.

wwymak:
Hi @rusty1s I have converted the integration to a mixin, and added an example -- is this more along the lines of what you are thinking of?

rusty1s:
Thanks, this is super cool. Wondering how we can integrate a basic test for this. Any idea?

wwymak:
> Wondering how we can integrate a basic test for this. Any idea?

I am thinking of a simple version of https://github.com/huggingface/huggingface_hub/blob/main/tests/test_hubmixin.py ? Since ours is based on what they've done for the pytorch mixin?

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,164 @@
...
+except ImportError:
+    print('please install huggingface hub to use this mixin')
+
+CONFIG_NAME = 'config.json'
+MODEL_HUB_ORGANIZATION = "pytorch_geometric"
+MODEL_WEIGHTS_NAME = 'model.pth'
+
+
+class PyGModelHubMixin(ModelHubMixin):
+    """
```
lightaime, line 23:
```suggestion
    r"""
```

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,164 @@
...
+        model_name='node2vec', dataset_name='Cora',
+        edge_index=data.edge_index)
+
+    Note: at the moment the model card is fairly basic--
+    override the `construct_model_card` method if you want to have
+    a more detailed model card
+
+    """
+    def __init__(self, model_name: str, dataset_name: str, model_kwargs: Dict):
+        """
```
lightaime, line 65:
```suggestion
        r"""
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,117 @@
...
+
+@withPackage('huggingface_hub')
+def test_save_pretrained(model, tmp_path):
+    save_directory = f"{str(tmp_path / REPO_NAME)}"
+    # model._save_pretrained = Mock()
+    model.save_pretrained(save_directory)
+    files = os.listdir(save_directory)
+    assert "model.pth" in files
+    assert len(files) >= 1
+    #
```
lightaime, line 57: Remove this comment?

lightaime (approved):
Awesome work! left some small comments.

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Args:
+            model_name (str): Name of the model as shown on the model card
+            dataset_name (str): Name of dataset you trained model against
+            model_kwargs (Dict): arguments that will be passed to the PyG model
+        """
+        ModelHubMixin.__init__(self)
+        self.model_config = {
+            k: v
+            for k, v in model_kwargs.items() if type(v) in [str, int, float]
```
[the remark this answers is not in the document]
wwymak: yup-- have added a comment explaining this

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,164 @@
...
+        path = os.path.join(save_directory, MODEL_WEIGHTS_NAME)
+        model_to_save = self.module if hasattr(self, "module") else self
+        torch.save(model_to_save.state_dict(), path)
+
+    def save_pretrained(self, save_directory: str, push_to_hub: bool = False,
+                        **kwargs):
+
+        config = self.model_config
+        if 'config' in kwargs.keys():
+            kwargs.pop('config')
```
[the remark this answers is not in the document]
wwymak: no-- I have added a comment why I did it that way. Let me know if it is unclear :)

torch_geometric/nn/model_hub.py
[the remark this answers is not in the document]
lightaime: cool

torch_geometric/nn/model_hub.py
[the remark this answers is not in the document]
lightaime: I see. You can also do `kwargs.pop('config', None)` without the `if` statement.

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,171 @@
...
+
+try:
+    from huggingface_hub import (
+        ModelCard,
+        ModelCardData,
+        ModelHubMixin,
+        hf_hub_download,
+    )
+except ImportError:
+    print('please install huggingface hub to use this mixin')
```
wsad1, line 15:
```suggestion
    raise ImportError("'PyGModelHubMixin' requires 'huggingface_hub'")
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+        def load_state_dict(self, state_dict, strict):
+            return None
+
+    return DummyModel
+
+
+@withPackage('huggingface_hub')
+@pytest.fixture
+def model(dummy_model_class):
+    return dummy_model_class(MODEL_NAME, DATASET_NAME, CONFIG)
```
wsad1, line 37:
```suggestion
  class DummyModel(torch.nn.Module, PyGModelHubMixin):
      def __init__(self, model_name, dataset_name, model_kwargs):
          torch.nn.Module.__init__(self)
          PyGModelHubMixin.__init__(self, model_name, dataset_name,
                                    model_kwargs)

      def load_state_dict(self, state_dict, strict):
          return None

@withPackage('huggingface_hub')
@pytest.fixture
def model():
    return DummyModel
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+REPO_NAME = "test"
+MODEL_NAME = 'test_model'
+DATASET_NAME = 'some_dataset'
+CONFIG = {"hello": "world"}
+
+
+@withPackage('huggingface_hub')
+@pytest.fixture
+def dummy_model_class():
+    class DummyModel(torch.nn.Module, PyGModelHubMixin):
```
wsad1, line 22: Can we use a simple `GNN` here instead of `torhc.nn.Module`.

wsad1 (commented):
Thanks @wwymak for building this. I think this is almost ready to merge.
Left a few comments.

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+    return DummyModel
+
+
+@withPackage('huggingface_hub')
+@pytest.fixture
+def model(dummy_model_class):
+    return dummy_model_class(MODEL_NAME, DATASET_NAME, CONFIG)
+
+
+@withPackage('huggingface_hub')
```
wsad1, line 40:
```suggestion
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+def test_model_init(dummy_model_class):
+    model = dummy_model_class(
+        MODEL_NAME, DATASET_NAME, model_kwargs={
+            **CONFIG, "tensor": torch.Tensor([1, 2, 3])
+        })
+    assert model.model_config == CONFIG
+
+
+@withPackage('huggingface_hub')
```
wsad1, line 49:
```suggestion
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+def test_save_pretrained(model, tmp_path):
+    save_directory = f"{str(tmp_path / REPO_NAME)}"
+    # model._save_pretrained = Mock()
+    model.save_pretrained(save_directory)
+    files = os.listdir(save_directory)
+    assert "model.pth" in files
+    assert len(files) >= 1
+
+
+@withPackage('huggingface_hub')
```
wsad1, line 59:
```suggestion
```

test/nn/test_model_hub.py
```diff
@@ -0,0 +1,116 @@
...
+    model.save_pretrained(save_directory, push_to_hub=True, repo_id="CustomID",
+                          config=CONFIG)
+    model.push_to_hub.assert_called_with(repo_id="CustomID", config=CONFIG)
+
+    # Push to hub with default repo_id (based on dir name)
+    model.save_pretrained(save_directory, push_to_hub=True, config=CONFIG)
+    model.push_to_hub.assert_called_with(repo_id=REPO_NAME, config=CONFIG)
+
+
+@withPackage('huggingface_hub')
```
wsad1, line 90:
```suggestion
```

wsad1 (approved):
Thanks again for the great work.

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,275 @@
...
+                The directory where weights are saved, to a file
+                called :obj:`"model.pth"`.
+            push_to_hub: If :obj:`True`, push the model to the model hub.
+                (default: :obj:`False`)
+            repo_id (str, optional): The repository name in the hub.
+                If not provided will default to the name of
+                :obj:`save_directory` in your namespace.
+                (default: :obj:`None`)
+            **kwargs: Additional keyword arguments passed along to
+                :obj:`huggingface_hub.ModelHubMixin.save_pretrained`.
```
wsad1, line 132:
```suggestion
            save_directory (str, Path):
                The directory where weights are saved, to a file
                called :obj:`"model.pth"`.
            push_to_hub(bool): If :obj:`True`, push the model to the
                model hub. (default: :obj:`False`)
            repo_id (str, optional): The repository name in the hub.
                If not provided will default to the name of
                :obj:`save_directory` in your namespace.
                (default: :obj:`None`)
            **kwargs: Additional keyword arguments passed along to
                :obj:`huggingface_hub.ModelHubMixin.save_pretrained`.
```

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,275 @@
...
+            cache_dir (`Union[str, os.PathLike]`, *optional*):
+                Path to a directory in which a downloaded pretrained
+                model configuration should be cached if the standard
+                cache should not be used.
+            local_files_only(`bool`, *optional*, defaults to `False`):
+                Whether to only look at local files (i.e., do not try to
+                download the model).
+            model_kwargs (`Dict`, *optional*):
+                model_kwargs will be passed to the model during
+                initialization
```
wsad1, line 264:
```suggestion
            pretrained_model_name_or_path (str, Path):
                Can be either:
                - A string, the `model id` of a pretrained model
                hosted inside a model repo on huggingface.co.
                Valid model ids can be located at the root-level,
                like `bert-base-uncased`, or namespaced under a
                user or organization name, like
                `dbmdz/bert-base-german-cased`.
                - You can add `revision` by appending `@` at the end
                of model_id simply like this:
                `dbmdz/bert-base-german-cased@main` Revision is
                the specific model version to use. It can be a
                branch name, a tag name, or a commit id, since we
                use a git-based system for storing models and
                other artifacts on huggingface.co, so `revision`
                can be any identifier allowed by git.
                - A path to a `directory` containing model weights
                saved using
                [`~transformers.PreTrainedModel.save_pretrained`],
                e.g., `./my_model_directory/`.
                - `None` if you are both providing the configuration
                and state dictionary (resp. with keyword arguments
                :obj:`config` and :obj:`state_dict`).
            force_download (bool): Whether to force the (re-)download of the model
                weights and configuration files, overriding the cached versions
                if they exist. (default: :obj:`False`)
            resume_download (bool): Whether to delete incompletely received files.
                Will attempt to resume the download if such a file exists. (default: :obj:`False`)
            proxies (Dict[str, str], optional): A dictionary of proxy servers to use by protocol or
                endpoint, e.g., `{'http': 'foo.bar:3128', 'http://hostname': 'foo.bar:4012'}`.
                The proxies are used on each request. (default: :obj:`None`)
            token (str, bool, optional): The token to use as HTTP bearer authorization
                for remote files. If `True`, will use the token generated when
                running `transformers-cli login` (stored in
                `~/.huggingface`). It is **required** if you
                want to use a private model. (default: :obj:`None`)
            cache_dir (`Union[str, os.PathLike]`, *optional*):
                Path to a directory in which a downloaded pretrained
                model configuration should be cached if the standard
                cache should not be used.
            local_files_only(`bool`, *optional*, defaults to `False`):
                Whether to only look at local files (i.e., do not try to
                download the model).
            model_kwargs (`Dict`, *optional*):
                model_kwargs will be passed to the model during
                initialization
```

torch_geometric/nn/model_hub.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+    ..note::
+        At the moment the model card is fairly basic. Override the
+        :obj:`construct_model_card` method if you want a more detailed
+        model card
+
+    Args:
+        model_name (str): Name of the model shown on the model card
+            hugging face hub.
```
wsad1, line 76:
```suggestion
        model_name (str): Name of the model shown on the model card
            on hugging face hub.
```

PR #5938, opened 2022-11-09
Support symbolic tracing of `SchNet`

hatemhelal:
This PR removes/modifies a couple control flow patterns that were used in the `SchNet` implementation that are not traceable with `torch.fx`:

* removing the assertion in the forward pass
* Replaced calling `scatter` directly with an `Aggregation` module

Without these changes, attempting symbolic tracing:
```
from torch_geometric.nn.fx import Transformer
from torch_geometric.nn import SchNet

model = SchNet()
tx = Transformer(model, debug=True)
tx.transform()
```

fails with:
```
TraceError: symbolically traced variables cannot be used as inputs to control flow
```


torch_geometric/nn/models/schnet.py
```diff
@@ -112,9 +112,9 @@ def __init__(
         self.num_interactions = num_interactions
         self.num_gaussians = num_gaussians
         self.cutoff = cutoff
-        self.readout = readout
+        self.aggr = aggr_resolver(readout)
+        self.sum_aggr = SumAggregation()
         self.dipole = dipole
-        self.readout = 'add' if self.dipole else self.readout
```
lightaime, line 117: Could you clarify this change? Shouldn't we resolve the aggr of `readout` after this?

lightaime (commented):
Thanks! LGTM.

torch_geometric/nn/models/schnet.py
hatemhelal: yes, good spot! I'll fix this soon

lightaime (approved):
Cool. Thanks!

rusty1s (approved):

PR #5939, opened 2022-11-09
Add `to_fixed_size` graph transformer

hatemhelal:
This PR adds a graph transformer for applying a fixed `dim_size` argument to any global pooling layers in a network.  This will be useful when exporting to ahead-of-time compilation environments since in this case we cannot rely on inferring the size from the `max` index used in the pooling.


rusty1s (approved):
Very clean.

PR #5940, opened 2022-11-09
Add `Pad` transform

bwroblew:
This PR adds a transform that allows node and edge features to be padded to a specified size. It enforces consistent shapes of tensors in the dataset.


torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+
+import torch
+import torch.nn.functional as F
+
+from torch_geometric.data import Data, HeteroData
+from torch_geometric.data.datapipes import functional_transform
+from torch_geometric.transforms import BaseTransform
+
+
+class _AttributeTypeCache:
```
rusty1s, line 14: Is that such an expensive op? Can we consider moving this to the `Data` class directly?

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+class Pad(BaseTransform):
+    r"""Applies padding to enforce consistent tensor shapes
+    (functional name: :obj:`pad`).
+
+    This transform will pad node and edge features up to a maximum allowed
+    size in the node or edge feature dimension. By default :obj:`0.0` is used
+    as the padding value and can be configured by setting :obj:`node_pad_value`
+    and :obj:`edge_pad_value`.
+
+    Note that in case of applying Pad to
```
rusty1s, line 65:
```suggestion
    Note that in case of applying :class:`Pad` to
```

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+    :class:`~HeteroData` input, the :obj:`node_pad_value`
```
rusty1s, line 66:
```suggestion
    a :class:`~torch_geometric.data.HeteroData` object, the :obj:`node_pad_value`
```

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+
+    Args:
+        max_num_nodes (int or Dict[str, int]): The maximum number of nodes.
+            In heterogeneous graphs, may also take in a dictionary denoting the
+            maximum number of nodes for specific node types.
+        max_num_edges (int or Dict[Tuple[str, str, str], int], optional):
+            The maximum number of edges.
+            In heterogeneous graphs, may also take in a dictionary denoting the
+            maximum number of edges for specific edge types.
+        node_pad_value (float or Dict[str, float], optional): The fill value to
```
rusty1s, line 91: Is there a use-case for passing a dictionary here? Would prefer to keep this simple and just expect a `float`.

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+            In heterogeneous graphs, may also take in a dictionary denoting the
+            fill value for specific edge types.
+            Note that in case of :obj:`edge_index` feature the tensors are
+            padded with the first padded node index (which represents a set of
+            self loops on the padded node).
+        train_mask_pad_value (bool, optional): The fill value to use for
+            :obj:`train_mask` (default: :obj:`False`).
+        test_mask_pad_value (bool, optional): The fill value to use for
+            :obj:`test_mask` (default: :obj:`False`).
+        include_keys (List[str] or Tuple[str], optional): Keys to include from
```
rusty1s, line 106: Shall we just make this `mask_pad_value`?

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+                 max_num_edges: Optional[Union[int, Dict[Tuple[str, str, str],
+                                                         int]]] = None,
+                 node_pad_value: Optional[Union[float, Dict[str,
+                                                            float]]] = None,
+                 edge_pad_value: Optional[Union[float, Dict[Tuple[str, str,
+                                                                  str],
+                                                            float]]] = None,
+                 train_mask_pad_value: Optional[bool] = None,
+                 test_mask_pad_value: Optional[bool] = None,
+                 include_keys: Optional[Union[List[str], Tuple[str]]] = None):
```
rusty1s, line 119:
```suggestion
                 include_keys: Optional[Union[List[str], Tuple[str]]] = None, ):
```

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+            'train_mask':
+            0.0 if train_mask_pad_value is None else train_mask_pad_value,
+            'test_mask':
+            0.0 if test_mask_pad_value is None else test_mask_pad_value
+        }
+
+        self.include_keys = include_keys
+        self.attribute_cacher = _AttributeTypeCache()
+
+    class ParamWithDesc():
```
rusty1s, line 143:
```suggestion
    class ParamWithDesc:
```

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+            param (object): The parameter to be wrapped.
+        """
+        def __init__(self, param):
+            self.value = param
+            self.is_number = isinstance(param, numbers.Number)
+
+    def _set_max_num_edges(self, max_num_nodes: Union[int, Dict[str, int]],
+                           max_num_edges: Union[int, Dict[Tuple[str, str, str],
+                                                          int]],
+                           include_keys: Union[List[str], Tuple[str]]):
```
rusty1s, line 157:
```suggestion
                           include_keys: Union[List[str], Tuple[str]], ):
```

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+        train_mask_pad_value (bool, optional): The fill value to use for
+            :obj:`train_mask` (default: :obj:`False`).
+        test_mask_pad_value (bool, optional): The fill value to use for
+            :obj:`test_mask` (default: :obj:`False`).
+        include_keys (List[str] or Tuple[str], optional): Keys to include from
+            the input data object to apply the padding transformation to.
+    """
+    def __init__(self, max_num_nodes: Union[int, Dict[str, int]],
+                 max_num_edges: Optional[Union[int, Dict[Tuple[str, str, str],
+                                                         int]]] = None,
```
rusty1s, line 111: Not sure I understand why this can be optional.

torch_geometric/transforms/pad.py
```diff
@@ -0,0 +1,548 @@
...
+                    'The max_num_edges option must be at least as large '\
+                    'as\n\n' \
+                    '    max(data.num_edges)\n\n' \
+                    'for all examples in the dataset.'
+
+    @singledispatchmethod
+    def __call__(self, data):
+        raise ValueError(f'Unsupported data type: {type(data)}')
+
+    @__call__.register
```
rusty1s, line 376: I personally think the implementation gets cleaner if you merge both `Data` and `HeteroData` handling into a single function and just iterate over its stores. There is a lot of functionality to be shared here.

rusty1s (commented):
Thanks for the PR. I personally didn't expected such a big one :) 

Overall, this looks mostly good:
* Can we move caching of `is_node_attr` and `is_edge_attr` to `Storage` (why do we need it in the first-place)?
* is there any chance we can merge the impls for `Data` and `HeterData`?
* I am not totally sure I understand the motivation behind `ParamWithDesc`. Can you explain?

torch_geometric/transforms/pad.py
bwroblew: I believe the request to handle it per node type in case of `HeteroData` came came from our research team. @hatemhelal, could you please help providing a use-case for that?

torch_geometric/transforms/pad.py
bwroblew: @hatemhelal, do you know if we have any use cases for having separate masks?

bwroblew:
@rusty1s, thanks for the review! Answering your questions:
1. I moved that to the GlobalStorage class, created a separate commit for that. It was added just to improve performance a bit. We have a microbenchmark that iterates over 10k graphs from QM9 dataset applying the Pad transform and it runs around 40% faster with that caching optimization.
2. Will try to unify that tomorrow.
3. It was initially created to not have to check the type of the argument (whether it’s a number or a dict) everytime that information is needed, just cache that as a field in a class. I’m not the author tbh but it also makes sense to me. Maybe the naming is not the best, I can rename it to `ParamMaybeNumeric` or sth like that?

torch_geometric/transforms/pad.py
bwroblew: Moved to the `GlobalStorage` class, see [the comment](https://github.com/pyg-team/pytorch_geometric/pull/5940#issuecomment-1326719344) about the performance reasons.

torch_geometric/transforms/pad.py
bwroblew: Will try to unify that tomorrow

torch_geometric/transforms/pad.py
hatemhelal: Not sure I have a use case for separate masks at the moment.

torch_geometric/transforms/pad.py
hatemhelal: The per-node padding value comes up in cases where I wanted to pad two different node features that have different meanings.  For small molecules:

* `z`: vector of atomic numbers -> use zero as a padding value to represent fake atoms 
* `pos`: 3d atomic position data -> use a large value to exclude padding atoms from any pairwise interactions. this is necessary to avoid wrong answers with padding.

torch_geometric/transforms/pad.py
hatemhelal: I think this can be omitted in the case where the dataset does not have any edge-features.

bwroblew:
@rusty1s, I just updated the PR, sorry for such a late update.
I redesigned it all a lot based on your comments (to unify the implementetion for `Data` and `HeteroData` objects and just iterate over the stores) and our internal discussions.

I would really appreciate if you could take another look at the PR.

One question from me:
As we need separate padding values for different node/edge types and different attributes, I created the `Padding` classes, so it's more explicit than just passing a dictionary of dictionaries. I added those classes in the global scope of `pad` module. The problem is that they're not rendered in the documentation, I didn't include them in the `__all__` list. How should I include such classes in the documentation? Where is the best place to put them?
(just adding them to transforms' `__all__` doesn't solve the problem as they're rendered on the list of transforms, I'm afraid it's not the best place, and some checks in `fill_config_store` fail then).

rusty1s:
I think it would be sufficient to just show a small example in the `Pad` doc-string on how to use - then we can likely avoid to add this class to the documentation altogether. Would that work for you?

bwroblew:
Sure, I extended the documentation of `Pad` with an example showing basically all kinds of padding values and a description.

bwroblew:
@rusty1s, could you please take another look at this PR?

rusty1s:
Definitely, sorry for the delay on my end.

test/transforms/test_pad.py
```diff
@@ -0,0 +1,536 @@
...
+    _check_data(original, padded, num_nodes, num_edges)
+
+
+@pytest.mark.parametrize('num_nodes', [32, {'v0': 64, 'v1': 36}])
+@pytest.mark.parametrize('num_edges', [300, {('v0', 'e0', 'v1'): 397}])
+def test_pad_heterodata_explicit_edges(num_nodes, num_edges):
+    data = fake_hetero_data()
+    original = deepcopy(data)
+    transform = Pad(max_num_nodes=num_nodes, max_num_edges=num_edges)
+
```
zhiyil-graphcore, line 291: from this test and the test below `test_pad_data_pad_values`, we pass `max_num_nodes` to the `Pad` transform. Does this mean that all types of nodes in the heterogeneous graph are padded to the same size?
For example, if we have `author, affiliated_with, institution` relation. In one batch, we have 1750 author nodes, 54 institution nodes. If it possible to pad author node to be 1800 for all batches and institution nodes to 60 for all batches? The same might apply on the edges. In this case, we don't waste space on institution nodes by padding it to 1800. Let me know if this makes sense or if I  understand this wrong.

bwroblew: It makes sense and it is possible in the current implementation. See that in one test case of `test_pad_heterodata_explicit_edges`, `max_num_nodes` is `32` but in the other we pass a dictionary `{'v0': 64, 'v1': 36}` meaning that `v0` node is going to be padded to be `64` and `v1` -- to `36`. See also the description of the `max_num_nodes` argument in the `Pad` class:

> The number of nodes after padding.
In heterogeneous graphs, may also take in a dictionary denoting the
number of nodes for specific node types. The dictionary must
specify values for all the possible node types.

So in your case you could pass sth like `{'author': 1800, 'institution': 60}`. The same applies to edges, `max_num_edges` can also be a dictionary with values for specific edge types.

zhiyil-graphcore: That's excellent! Sounds good to me. Thanks for the explanation.

bwroblew:
Thanks for all the changes, @rusty1s! Do you need anything from my side? Do you have any comments or doubts?

rusty1s (approved):

PR #5941, opened 2022-11-09
Add `lr_scheduler` helper functions

lightaime:
Add a helper to get any learning rate scheduler implemented in pyg and pytorch from its name or type.


lightaime:
Drop this in favor of https://github.com/pyg-team/pytorch_geometric/pull/5942.

PR #5942, opened 2022-11-09
Add `lr_scheduler_solver` and  customized `lr_scheduler` classes

lightaime:
Add `lr_scheduler_solver` and customized `lr_scheduler` classes to support getting any learning rate scheduler implemented in pyg and pytorch from its name. Maybe a better location than `utils` like `torch_geometric.optim` in case we develop specific optimization methods for GNNs?


rusty1s (approved):
Amazing!

PR #5944, opened 2022-11-10
Add PyTorch SparseTensor support for `MessagePassing`

EdisonLeeeee:
This PR adds the PyTorch SparseTensor support for the base layer `MessagePassing`. There are some points to be confirmed (as marked with TODO):
+ ~~In `__collect__`: Since `adj._values()` returns a detached tensor, should we use a coalesced matrix instead (e.g., `adj.coalesce().values()`)? This is for the case of computing sparse gradients of `adj`.~~ (Solved)
+ In `__collect__`: Should we store the `ptr` for PyTorch SparseTensor when fused aggregation is not available?
+ In `__lift__`: Should we use `gather_csr` for PyTorch SparseTensor?

Also, `torch.jit.script` is not available for PyTorch SparseTensor. Will figure it out soon.


EdisonLeeeee:
Testing failed due to numerical instability. It seems we should use `torch.allclose` instead.

rusty1s:
This looks pretty good. Is it ready to review?

EdisonLeeeee:
Yes. Please take a look at your convenience :)

rusty1s (approved):
This is super cool. I think we can also think about adding a `is_sparse_tensor` utility to PyG that checks for either `torch.sparse.Tensor` or `SparseTensor` to clean-up some duplicated code in a follow-up.

EdisonLeeeee:
Sure :) Will also add it to the roadmap.

PR #5948, opened 2022-11-10
Support for `int32` graphs in `NeighborLoader`

rusty1s:


rusty1s:
@ZenoTan Please take a look :)

torch_geometric/loader/utils.py
```diff
@@ -91,7 +95,7 @@ def filter_edge_store_(store: EdgeStorage, out_store: EdgeStorage, row: Tensor,
             edge_attr = value.storage.value()
             if edge_attr is not None:
                 index = index.to(edge_attr.device)
-                edge_attr = edge_attr[index]
+                edge_attr = index_select(edge_attr, index, dim=0)
```
ZenoTan, line 98: If we already guaranteed int64, do we still need to change here?

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -228,12 +228,17 @@ def _sample(
         if self.data_cls == 'custom' or issubclass(self.data_cls, HeteroData):
             if _WITH_PYG_LIB:
                 # TODO (matthias) `return_edge_id` if edge features present
+                # TODO (matthias) Ideally, `seed` should inherit the type of
+                # `colptr_dict` and `row_dict`.
+                colptrs = list(self.colptr_dict.values())
+                dtype = colptrs[0].dtype if len(colptrs) > 0 else torch.int64
```
ZenoTan, line 234: nice :)

torch_geometric/sampler/utils.py
```diff
@@ -68,7 +68,12 @@ def to_csc(
         row, col = data.edge_index
         if not is_sorted:
             row, col, perm = sort_csc(row, col, src_node_time)
-        colptr = torch.ops.torch_sparse.ind2ptr(col, data.size(1))
+
+        if hasattr(torch, '_convert_indices_from_coo_to_csr'):
+            colptr = torch._convert_indices_from_coo_to_csr(
```
ZenoTan, line 73: Curious is it supported in torch with later versions so we have that if-else?

ZenoTan (approved):

ZenoTan:
Also I guess this diff would also be good for every other integer types?

torch_geometric/sampler/utils.py
rusty1s: Yes, this is needed for PyTorch 1.10 support :D

torch_geometric/loader/utils.py
rusty1s: I don't think the `index` is guaranteed to be `int64` here.

PR #5951, opened 2022-11-10
Fix `collate_fn` type hints

avgupta456:
I just cloned the repository and my VSCode extensions picked up on a Union of a single type.


rusty1s (approved):
Thanks!

Issue #5924, opened 2022-11-08
Example of Using the New Explainability Framework


RexYing:
### 🚀 The feature, motivation and pitch

Create an example of using the new GNN explainability interface #5804 .
The example could include:
- Loading of an existing dataset and a basic GNN training
- Demo of creating a configuration setting for the desired explainability setting #5629
- Explanation of the basic GNN using the `Explainer` class (using any explainability method)
- Basic output of an `Explanation` object that corresponds to an explanation of a particular instance (node) in the dataset

Issue #5925, opened 2022-11-08
Explainability for Link Prediction


RexYing:
### 🚀 The feature, motivation and pitch

From #5804 , we already have explanation setting for node classification and graph classification, through the `task_level` argument of the `Explainer` class. 

We will create an additional option to enable link prediction explanation setting, and provide a basic implementation of explaining link prediction through e.g. GNNExplainer, SubgraphX or other methods. 

The idea is that through the general interface of `explain.base.ExplainerAlgorithm` class, we can plug in different explainability algorithms to perform link prediction explanation. 

A basic form of explanation output can be specified through the `Explanation` class.

avgupta456:
In #5967, ModelTaskLevel is an Enum accepting `node`, `graph`, and `edge`. Can we understand link prediction as a subset of edge regression, or should we create a separate option for `link_pred`? In the `edge` task level, the current implementation doesn't allow classifying/regressing an edge not in the graph, but negative link prediction explanations might be desired. Thoughts?

Additionally, should a separate issue be created for edge classification/regression explanation?

rusty1s:
I think this should be supported if you pass in an `edge_label_index` as additional argument to the model. Would this resolve your concern?

avgupta456:
That's true, I can incorporate an edge_label_index. That solves most of the problem I think.

Additionally, I was wondering if it is okay to combine "Link Prediction" and "Edge classification/regression" under the same ModelTaskLevel.edge category, by considering link prediction as edge regression. I was unsure if link prediction is more common and would be better having a separate label.

avgupta456:
Can we close this now?

rusty1s:
Yes :)

PR #5967, opened 2022-11-13
`GNNExplainer` migration

dufourc1:
- Implements the `GNNExplainer` in the new `explain` module.
- Add the argument `node_index` and `target_index` to the `explain.Explainer` class to allow for explanations of model with multi-outputs and specify the node.
- Update the examples of the `GNNExplainer` (#5924)


dufourc1:
Not so sure how to deal with multi-outputs for classification (node/graph). Does it even make sens ?

torch_geometric/explain/algorithm/base.py
```diff
@@ -37,6 +38,11 @@ def forward(
             target (torch.Tensor): the target of the model.
             target_index (int or torch.Tensor, optional): The target indices to
```
wsad1, line 39: Is my understanding correct that we need this only for multi-task learning?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+            node_feat_mask = None
+        self.clear(model)
+
+        # build explanation
+        return Explanation(x=x, edge_index=edge_index, edge_mask=edge_mask,
+                           node_mask=node_mask, node_feat_mask=node_feat_mask)
+
+    def explain_graph(self, model: torch.nn.Module, x: Tensor,
+                      edge_index: Tensor, explainer_config: ExplainerConfig,
+                      model_config: ModelConfig, target: Tensor,
```
wsad1, line 122: `target` Can we compute this internally an not have the user pass it in , like we did in `gnn_explainer` perviously?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+        node_index: Optional[int] = None,
+    ):
+        if target_idx is not None:
+            y_hat = y_hat[target_idx]
+            y = y[target_idx]
+
+        if return_type == ModelReturnType.probs:
+            y_hat = y_hat.log()
+        if return_type == ModelReturnType.raw:
+            y_hat = y_hat.log_softmax(dim=-1)
```
wsad1, line 262: Use `if` , `elif` here.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+        self.coeffs.update(kwargs)
+        self.shared_mask = shared_mask
+        self.node_mask = None
+        self.edge_mask = None
+
+    def forward(self, model: torch.nn.Module, x: Tensor, edge_index: Tensor,
+                explainer_config: ExplainerConfig, model_config: ModelConfig,
+                target: Tensor, target_index: Optional[Union[int,
+                                                             Tensor]] = None,
+                index: Optional[Union[int, Tensor]] = None,
```
wsad1, line 79:
```suggestion
                index: Optional[int] = None,
```
At the moment the `loss` function only supports training for one node.

wsad1 (commented):
Thanks for adding this. Left some initial comments. Will take a look again tomorrow.

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: if the type of explanation required is `phenomenon`, the target should be the groundtruth, while if the explanation required is `model` then it can be computed as the `gnn_explainer` did previously. The  `Explainer` class takes care of this switch if needed, so the `ExplainerAlgorithm` does not have to do it.

torch_geometric/explain/algorithm/base.py
dufourc1: yes, I'm just unsure if this is a good way to do it. I'm a bit lost with the multi-task classification setting

wsad1: Ah I see. Lets mention in the doc that this is only needed for multi-task learning. 
Also i don't think we have an actual example of multi-task learning on graphs, so i wouldn't worry about it till someone comes up with an example an then uses `GNNExpaliner` on it.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,317 @@
...
+        node_index: Optional[int] = None,
+    ):
+        if target_idx is not None:
+            y_hat = y_hat[target_idx]
+            y = y[target_idx]
+
+        if return_type == ModelReturnType.probs:
+            y_hat = y_hat.log()
+        if return_type == ModelReturnType.raw:
+            y_hat = y_hat.log_softmax(dim=-1)
```
wsad1, line 261: Lets move this chunk of code to a method in `base.py`. Lots of explainers will have to do similar processing before computing loss, wdyt?
Another think you could do is as was done in `gnn_explainer` have `loss` in `base.py` do this processing and then call `_loss` which is implemented in the child class.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,317 @@
...
+        shared_mask: bool = True,
+        **kwargs,
+    ):
+        super().__init__()
+        self.epochs = epochs
+        self.lr = lr
+        self.coeffs.update(kwargs)
+        self.shared_mask = shared_mask
+        self.node_mask = None
+        self.edge_mask = None
```
wsad1, line 73: Can we move this to `base.py``__init__`? I am guessing most explainers will share this.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,317 @@
...
+            node_mask = None
+        else:
+            node_feat_mask = None
+        self.clear(model)
+
+        # build explanation
+        return Explanation(x=x, edge_index=edge_index, edge_mask=edge_mask,
+                           node_mask=node_mask, node_feat_mask=node_feat_mask)
+
+    def explain_graph(self, model: torch.nn.Module, x: Tensor,
```
wsad1, line 119:
```suggestion
    def _explain_graph(self, model: torch.nn.Module, x: Tensor,
```
Nitpick: lets add `_` to methods that are not used outside this class.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,317 @@
...
+                          node_mask_type: MaskType, edge_mask_type: MaskType):
+        (N, F), E = x.size(), edge_index.size(1)
+        std = 0.1
+        if node_mask_type == MaskType.object:
+            self.node_mask = torch.nn.Parameter(torch.randn(N, 1) * std)
+        elif node_mask_type == MaskType.attributes:
+            if self.shared_mask:
+                self.node_mask = torch.nn.Parameter(torch.randn(1, F) * std)
+            else:
+                self.node_mask = torch.nn.Parameter(torch.randn(N, F) * std)
```
wsad1, line 222: Lets test this line by setting `shared_mask=False`

wsad1 (commented):
The `GNNExplainer` part looks good. Will review the test and examples next.

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: Yes we can have a `get_prediction` method. Actually the `Explainer` class also has one, maybe we should have it as a utility function ?

torch_geometric/explain/algorithm/base.py
dufourc1: Well now that I think about it, it is also usefull if suppose you have $GNN: (A,X) \mapsto \mathbb{R}^K$ and you are interested in what is important for the $k ^{th}$ coordinate of your output, you'd set _target_index_ to $k$.

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: Although it can be a bit confusing: here, `node_mask` is used interchangeably for the `node_mask` and the `node_feature_mask`.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+        ent = -m * torch.log(m + EPS) - (1 - m) * torch.log(1 - m + EPS)
+        loss = loss + self.coeffs['node_feat_ent'] * ent.mean()
+
+        return loss
+
+    def supports(self, explainer_config: ExplainerConfig,
+                 model_config: ModelConfig) -> bool:
+        if model_config.task_level not in [
+                ModelTaskLevel.node, ModelTaskLevel.graph
+        ]:
```
rusty1s, line 300: Can we print/log something here?

torch_geometric/explain/algorithm/utils.py
```diff
@@ -0,0 +1,58 @@
+from typing import Dict
```
rusty1s, line 1: Is this a copy-paste? Wondering why it doesn't pick up the diff :(

torch_geometric/nn/models/__init__.py
```diff
@@ -10,7 +10,6 @@
 from .schnet import SchNet
 from .dimenet import DimeNet, DimeNetPlusPlus
 from .explainer import Explainer, to_captum
-from .gnn_explainer import GNNExplainer
```
rusty1s, line 13: Can we maintain some form of backward compatibility for the old interface. I don't want existing scripts from users to break here :(

torch_geometric/explain/algorithm/base.py
```diff
@@ -37,6 +38,11 @@ def forward(
...
                 explain. (default: :obj:`None`)
+            index (Union[int, Tensor], optional): the node/edge index
```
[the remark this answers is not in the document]
dufourc1: In my mind it would be used like this: if you have a model that outputs a $y \in \mathbb{R}^K$ with $K>1$ for all your nodes, and you are interested in what influences $y[k]$ for a specific node $i$, you would set `target_index=k` and `index=i`. The target index influences the index of the output (regardless of the task level of the GNN), while the `index` helps selecting the specific node (and edge in the future).

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+                explainer_config: ExplainerConfig, model_config: ModelConfig,
+                target: Tensor, target_index: Optional[Union[int,
+                                                             Tensor]] = None,
+                index: Optional[int] = None, **kwargs) -> Explanation:
+
+        if isinstance(target_index, Tensor):
+            raise ValueError(
+                'GNNExplainer only supports single target index for now')
+
+        # can it be an issue for some models ?
```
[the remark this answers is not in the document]
dufourc1: I was just wondering if a model could have masks somehow that would be deleted by this method and not restored.

torch_geometric/explain/algorithm/base.py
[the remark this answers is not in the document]
rusty1s: Oh, I see. `target_index` references a second dimension of potentially multiple targets. I thought it would index in the first dimension (and therefore didn't see the difference to `index`). Wonder if it might be better to allow tuple of indices then?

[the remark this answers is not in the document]
dufourc1: Yes, exactly (I might still this comment to put it in the docs) ! 
Perhaps it would be confusing ? I find it a bit clearer if they are separated (maybe renaming  `index` to `node_index` might help.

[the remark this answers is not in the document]
rusty1s: Sounds good.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,318 @@
...
+
+    Args:
+        epochs (int, optional): The number of epochs to train.
+            (default: :obj:`100`)
+        lr (float, optional): The learning rate to apply.
+            (default: :obj:`0.01`)
+        shared_mask (bool, optional): Whether to share the mask for node
+            features. When set to :obj:`True`, the node feature mask will be
+            shared across all nodes. (default: :obj:`True`). Only used when the
+            node mask type is :obj:`MaskType.attributes`.
```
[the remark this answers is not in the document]
rusty1s: Let's do it here.

torch_geometric/explain/algorithm/utils.py
dufourc1: yes it's just moving utilities from the old `Explainer` class

torch_geometric/nn/models/__init__.py
dufourc1: I added a `GNNExplainer_` that should do the trick in [22a3376](https://github.com/pyg-team/pytorch_geometric/pull/5967/commits/22a33764a16b4327390428d8ab8377def1fd4ea8). No idea how to actually format it and where to put it.

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: I'm not sure about forcing the preprocessing by the base class though, will all the explainer algorithms use log_probs for computing their classification loss ?

wsad1: Yeah thats a good point, lets have utility method in the base class instead that converts the probabilities to log_prob.

dufourc1: It's done 👍🏽

torch_geometric/explain/algorithm/base.py
wsad1: Ah I see. In that case can me mention this in the doc. Pretty much let the user know that they need to use this if their model outputs a tensor of values, and can ignore it otherwise.

torch_geometric/nn/models/__init__.py
wsad1: For full backward compatibility we should put `GNNExplaienr_` in `nn/models/gnn_explainer` and call it `GNNExplainer` with a depreciation warning, then we can remove it in a while.

test/explain/algoritms/test_gnn_explainer.py
```diff
@@ -0,0 +1,672 @@
...
+        return_type=return_type,
+        mode="regression",
+    )
+
+    model = model()
+
+    x = torch.randn(8, 3)
+    edge_index = torch.tensor([[0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7],
+                               [1, 0, 2, 1, 3, 2, 4, 3, 5, 4, 6, 5, 7, 6]])
+    out = model(x, edge_index, None, None)
```
wsad1, line 495: Can we define these at the top of the file are reuse for all tests?

test/explain/algoritms/test_gnn_explainer.py
```diff
@@ -0,0 +1,672 @@
...
+
+
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize('model', [
+    GNN_regression_singleoutput,
+    GNN_regression_multioutput,
+])
+@pytest.mark.parametrize('return_type', return_types_regression)
+def test_gnn_explainer_graph_regression(
```
wsad1, line 423: Correct me if I am wrong but doesn't `test_gnn_explainer_with_meta_explainer_regression_graph` make this test redundant?

wsad1 (commented):
Tests look good too mostly.

test/explain/algoritms/test_gnn_explainer.py
dufourc1: yes good point 👍🏽 , will do

test/explain/algoritms/test_gnn_explainer.py
dufourc1: yes, they were just useful when rewriting the algorithm, now I will delete them

torch_geometric/nn/models/__init__.py
dufourc1: thanks, I'll do that

torch_geometric/explain/algorithm/base.py
```diff
@@ -37,6 +43,11 @@ def forward(
...
             target_index (int or torch.Tensor, optional): The target indices to
                 explain. (default: :obj:`None`)
+            node_index (Union[int, Tensor], optional): the node/edge index
+                to explain. Can be a single index if no batch is provided, or a
+                tuple of indices if a batch is provided. only used if the model
+                task level is :obj:`"node"` or :obj:`"edge"`.
+                (default: :obj:`None`)
             **kwargs (optional): Additional keyword arguments passed to
                 :obj:`model`.
         """
```
avgupta456, line 53: Regarding the loss function right below this, is there a reason this is in the base explainer instead of just GNNExplainer? Do other explanation methods commonly use a loss function?

torch_geometric/explain/algorithm/base.py
```diff
@@ -37,6 +43,11 @@ def forward(
             target (torch.Tensor): the target of the model.
             target_index (int or torch.Tensor, optional): The target indices to
                 explain. (default: :obj:`None`)
+            node_index (Union[int, Tensor], optional): the node/edge index
+                to explain. Can be a single index if no batch is provided, or a
+                tuple of indices if a batch is provided. only used if the model
+                task level is :obj:`"node"` or :obj:`"edge"`.
+                (default: :obj:`None`)
```
avgupta456, line 50: What do you mean by "tuple of indices"? I see that a Tensor type is accepted but not a tuple. How exactly does this work for `edge` task level? It would seem to be a bit of a misnomer...

torch_geometric/explain/config.py
```diff
@@ -75,6 +76,9 @@ class ExplainerConfig(CastMixin):
 
                 - :obj:`"both"`: Will mask both the node and its attributes..
 
+                - :obj:`"common_attributes"`: Will share the attribute mask
+                    across all nodes.
+
```
avgupta456, line 81: What is the difference between common_attributes and either attributes or both? My understanding is common_attributes is (N, F) like both but is just attributes repeated N times. If that is the case, can it be extracted out of the individual explanation methods (treat independent of GNNExplainer logic)?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,546 @@
...
```
avgupta456, line 44: Outdated? Shared mask never used afterwards

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,546 @@
...
+            node_mask, edge_mask = self._explain_graph(model, x, edge_index,
+                                                       explainer_config,
+                                                       model_config, target,
+                                                       target_index, **kwargs)
+
+        if explainer_config.node_mask_type == MaskType.object:
+            node_feat_mask = None
+        else:
+            node_feat_mask = node_mask
+            node_mask = None
```
avgupta456, line 114: More of a broader point, but what is the purpose of separating node_mask vs node_feat_mask in the explanation class?

node_mask is of shape [N], while node_feat_mask is of shape [N, F] or [1, F]. Given the dimension is already variable, perhaps we could just combine them? Same for edge_mask/edge_feat_mask

Also, is the docstring in explanations.py outdated?
`node_feat_mask (Tensor, optional): Node-level feature mask with shape [num_nodes, num_node_features]`

torch_geometric/explain/config.py
```diff
@@ -75,6 +76,9 @@ class ExplainerConfig(CastMixin):
 
                 - :obj:`"both"`: Will mask both the node and its attributes..
```
avgupta456, line 78: Is both outdated? Doesn't seem to be referenced anymore.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,546 @@
...
+                               edge_mask_type=explainer_config.edge_mask_type)
+
+        if explainer_config.edge_mask_type == MaskType.object:
+            set_masks(model, self.edge_mask, edge_index, apply_sigmoid=True)
+            parameters = [self.node_mask, self.edge_mask]
+        else:
+            parameters = [self.node_mask]
+        optimizer = torch.optim.Adam(parameters, lr=self.lr)
+
+        for _ in range(1, self.epochs + 1):
```
avgupta456, line 220: Could just be `range(self.epochs)`

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,546 @@
...
+
+        if explainer_config.node_mask_type is None:
+            logging.error("Node mask type not supported.")
+            return False
+
+        if model_config.task_level not in [
+                ModelTaskLevel.node, ModelTaskLevel.graph
+        ]:
+            logging.error("Model task level not supported.")
+            return False
```
avgupta456, line 347: Redundant from L330-334?

avgupta456 (commented):
Hi! I'm looking to get started with link explanation soon and wanted to familiarize myself with current progress. I left some comments mostly to help myself understand, looking forward to reading your response.

avgupta456:
Echoing a previous comment, I also think it might be helpful to combine target_index and node/edge_index into a tuple of tensors used to index the output of the model relevant for the GNNExplainer optimization. This would allow for greater flexibility and better transfer to the link explanation setting.

dufourc1:
> Hi! I'm looking to get started with link explanation soon and wanted to familiarize myself with current progress. I left some comments mostly to help myself understand, looking forward to reading your response.

Thanks for the review, I'll address your comments asap. Lots of good catch 👍🏽 


> Echoing a previous comment, I also think it might be helpful to combine target_index and node/edge_index into a tuple of tensors used to index the output of the model relevant for the GNNExplainer optimization. This would allow for greater flexibility and better transfer to the link explanation setting.

So if two people suggest this, I might be missing something. Currently, if your model has a multioutput and you want to explain the $k^{th}$ element of this output you would just call `explainer(..., target_index=k, ...)`. If in addition you wanted to explain a specific node (or edge), let's say the $i^{th}$ you would just need to add `explainer(..., target_index=k, node_index=i ...)`. On the other hand if you had a node/edge level task and wanted the explanation for the $i^{th}$ node but did not care about the output index, you would use `explainer(..., node_index=i, ...)`. (Of course if available you can replace the indices by tensors). Could you show how we would do this with the tuple of tensors? What I have in mind would just be exactly the same (e.g., `index= (target_index,node_index)`).

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: 👍🏽  yep thanks !

torch_geometric/explain/config.py
dufourc1: The idea is to have users go through the `Explainer` class, and not the `ExplainerAlgorithm` class. This choice might also be useful for other explanation methods. 

I have reshaped the output to be a node_feat_mask so that any downstream task does not have to worry about how the feature mask of the nodes was computed, but just take it as a node_feat_mask. Maybe adding a new type of mask might make sens, but it would be additional to the node_feat_mask ?

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: So for now the node_feat_mask is always [N,F] (see comment above).

torch_geometric/explain/algorithm/base.py
dufourc1: This should be replaced by tensor in the docs. For now, the implementation does not work with `edge_level` task, but we could either add an `edge_index` which would be a tuple of node indices in the arguments, or just reuse this (like ` node_index (Union[int, Tensor, Tuple[int,int], List[Tuple[int,int]]]`, where the first two are for selecting nodes, while the second two are for selecting edges. 

And you always know which one you expect with the `model_config.task_level` attributes. What do you think ?

wsad1 (approved):
Thanks for the great work.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,536 @@
...
+        for _ in range(self.epochs):
+            optimizer.zero_grad()
+            h = x * self.node_mask.sigmoid()
+            out = model(x=h, edge_index=edge_index, **kwargs)
+            loss_value = self.loss(
+                out, target, return_type=model_config.return_type,
+                target_idx=target_index, node_index=index,
+                edge_mask_type=explainer_config.edge_mask_type,
+                model_mode=model_config.mode)
+            loss_value.backward(retain_graph=True)
```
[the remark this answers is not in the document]
dufourc1: I had an issue computing the gradient otherwise, the error indicated I should add that. Could not figure out why, but I checked in the examples the explainer is learning sensible masks.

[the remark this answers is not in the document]
wsad1: Could you paste the error here, lets try to resolve it. 
Setting `retain_graph` to `True` is pretty inefficient.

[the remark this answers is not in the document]
dufourc1: If you set `retrain_graph=True` and run the tests, you will get (for example) 

```
_____________________ test_gnn_explainer_with_meta_explainer_regression_node[phenomenon-raw-GCN_multioutput_regression-common_attributes-None] _____________________

edge_mask_type = None, node_mask_type = 'common_attributes', model = GCN_multioutput_regression(
  (conv1): GCNConv(3, 16)
  (conv2): GCNConv(16, 7)
)
return_type = 'raw', explanation_type = 'phenomenon'

    @pytest.mark.parametrize('edge_mask_type', edge_mask_types)
    @pytest.mark.parametrize('node_mask_type', node_mask_types)
    @pytest.mark.parametrize(
        'model', [GCN_single_output_regression, GCN_multioutput_regression])
    @pytest.mark.parametrize('return_type', return_types_regression)
    @pytest.mark.parametrize("explanation_type", ["model", "phenomenon"])
    def test_gnn_explainer_with_meta_explainer_regression_node(
        edge_mask_type,
        node_mask_type,
        model,
        return_type,
        explanation_type,
    ):
        explainer_config = ExplainerConfig(
            explanation_type=explanation_type,
            node_mask_type=node_mask_type,
            edge_mask_type=edge_mask_type,
        )
        model_config = ModelConfig(
            task_level='node',
            return_type=return_type,
            mode="regression",
        )
    
        model = model()
    
        out = model(x, edge_index)
    
        if isinstance(model, GCN_multioutput_regression):
            target_index = 0
        else:
            target_index = None
    
        explainer = Explainer(model=model, algorithm=GNNExplainer(),
                              explainer_config=explainer_config,
                              model_config=model_config)
    
        # try to explain prediction for node 0
>       explanation = explainer(x=x, edge_index=edge_index, target=out,
                                target_index=target_index, node_index=2)

test/explain/algoritms/test_gnn_explainer.py:378: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
torch_geometric/explain/explainer.py:115: in __call__
    explanation = self.algorithm(
env/lib/python3.9/site-packages/torch/nn/modules/module.py:1130: in _call_impl
    return forward_call(*input, **kwargs)
torch_geometric/explain/algorithm/gnn_explainer.py:98: in forward
    node_mask, edge_mask = self._explain_node(model, x, edge_index,
torch_geometric/explain/algorithm/gnn_explainer.py:172: in _explain_node
    self._train_node_edge_mask(model, x, edge_index, explainer_config,
torch_geometric/explain/algorithm/gnn_explainer.py:228: in _train_node_edge_mask
    loss_value.backward(retain_graph=False)
env/lib/python3.9/site-packages/torch/_tensor.py:396: in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tensors = (tensor([[0.6374]], grad_fn=<AddBackward0>),), grad_tensors = None, retain_graph = False, create_graph = False, grad_variables = None, inputs = ()

    def backward(
        tensors: _TensorOrTensors,
        grad_tensors: Optional[_TensorOrTensors] = None,
        retain_graph: Optional[bool] = None,
        create_graph: bool = False,
        grad_variables: Optional[_TensorOrTensors] = None,
        inputs: Optional[_TensorOrTensors] = None,
    ) -> None:
        r"""Computes the sum of gradients of given tensors with respect to graph
        leaves.
    
        The graph is differentiated using the chain rule. If any of ``tensors``
        are non-scalar (i.e. their data has more than one element) and require
        gradient, then the Jacobian-vector product would be computed, in this
        case the function additionally requires specifying ``grad_tensors``.
        It should be a sequence of matching length, that contains the "vector"
        in the Jacobian-vector product, usually the gradient of the differentiated
        function w.r.t. corresponding tensors (``None`` is an acceptable value for
        all tensors that don't need gradient tensors).
    
        This function accumulates gradients in the leaves - you might need to zero
        ``.grad`` attributes or set them to ``None`` before calling it.
        See :ref:`Default gradient layouts<default-grad-layouts>`
        for details on the memory layout of accumulated gradients.
    
        .. note::
            Using this method with ``create_graph=True`` will create a reference cycle
            between the parameter and its gradient which can cause a memory leak.
            We recommend using ``autograd.grad`` when creating the graph to avoid this.
            If you have to use this function, make sure to reset the ``.grad`` fields of your
            parameters to ``None`` after use to break the cycle and avoid the leak.
    
        .. note::
    
            If you run any forward ops, create ``grad_tensors``, and/or call ``backward``
            in a user-specified CUDA stream context, see
            :ref:`Stream semantics of backward passes<bwd-cuda-stream-semantics>`.
    
        .. note::
    
            When ``inputs`` are provided and a given input is not a leaf,
            the current implementation will call its grad_fn (even though it is not strictly needed to get this gradients).
            It is an implementation detail on which the user should not rely.
            See https://github.com/pytorch/pytorch/pull/60521#issuecomment-867061780 for more details.
    
        Args:
            tensors (Sequence[Tensor] or Tensor): Tensors of which the derivative will be
                computed.
            grad_tensors (Sequence[Tensor or None] or Tensor, optional): The "vector" in
                the Jacobian-vector product, usually gradients w.r.t. each element of
                corresponding tensors. None values can be specified for scalar Tensors or
                ones that don't require grad. If a None value would be acceptable for all
                grad_tensors, then this argument is optional.
            retain_graph (bool, optional): If ``False``, the graph used to compute the grad
                will be freed. Note that in nearly all cases setting this option to ``True``
                is not needed and often can be worked around in a much more efficient
                way. Defaults to the value of ``create_graph``.
            create_graph (bool, optional): If ``True``, graph of the derivative will
                be constructed, allowing to compute higher order derivative products.
                Defaults to ``False``.
            inputs (Sequence[Tensor] or Tensor, optional): Inputs w.r.t. which the gradient
                be will accumulated into ``.grad``. All other Tensors will be ignored. If
                not provided, the gradient is accumulated into all the leaf Tensors that
                were used to compute the attr::tensors.
        """
        if grad_variables is not None:
            warnings.warn("'grad_variables' is deprecated. Use 'grad_tensors' instead.")
            if grad_tensors is None:
                grad_tensors = grad_variables
            else:
                raise RuntimeError("'grad_tensors' and 'grad_variables' (deprecated) "
                                   "arguments both passed to backward(). Please only "
                                   "use 'grad_tensors'.")
        if inputs is not None and len(inputs) == 0:
            raise RuntimeError("'inputs' argument to backward() cannot be empty.")
    
        tensors = (tensors,) if isinstance(tensors, torch.Tensor) else tuple(tensors)
        inputs = (inputs,) if isinstance(inputs, torch.Tensor) else \
            tuple(inputs) if inputs is not None else tuple()
    
        grad_tensors_ = _tensor_or_tensors_to_tuple(grad_tensors, len(tensors))
        grad_tensors_ = _make_grads(tensors, grad_tensors_, is_grads_batched=False)
        if retain_graph is None:
            retain_graph = create_graph
    
        # The reason we repeat same the comment below is that
        # some Python versions print out the first line of a multi-line function
        # calls in the traceback and some print out the last line
>       Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
            tensors, grad_tensors_, retain_graph, create_graph, inputs,
            allow_unreachable=True, accumulate_grad=True)  # Calls into the C++ engine to run the backward pass
E       RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.

env/lib/python3.9/site-packages/torch/autograd/__init__.py:173: RuntimeError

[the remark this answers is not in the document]
Comment (wsad1) [reply torch_geometric/explain/algorithm/gnn_explainer.py:225]: Do all tests give you this error, also do the examples give you this error?  Will try running this on my end too.
This error usually means something isn't working as expected.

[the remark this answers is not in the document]
Comment (dufourc1) [reply torch_geometric/explain/algorithm/gnn_explainer.py:225]: Found the issue: in the test I was calling the model to compute the ground_truth without a `torch.no_grad()`. For the classification case this went unnoticed as I had the `argmax` operator was breaking the gradients anyway. 
This has no impact on the implementation, as the Explainer class correctly sets a context manager `with torch.no_grad()`

File: torch_geometric/explain/algorithm/base.py (hunk 18 above)
Comment (avgupta456) [reply torch_geometric/explain/algorithm/base.py]: I was thinking in the edge explanation setting, we would pass in indices to the edge_index tensor (shape [2, num_edges]). I think it would be confusing to pass in edge index into node_index, so maybe we have separate arguments or rename to a broader term.

File: torch_geometric/explain/config.py (hunk 19 above)
Comment (avgupta456) [reply torch_geometric/explain/config.py:81]:
```python
if explainer_config.node_mask_type == MaskType.common_attributes:
            node_mask = self.node_mask.detach().sigmoid().squeeze(0)
            node_mask = self._reshape_common_attributes(node_mask, num_nodes)
```

In this snippet from GNNExplainer, the logic seems completely separable from the specific method. Could this maybe be moved outo f GNNExplainer and into ExplainerAlgorithm?

torch_geometric/explain/algorithm/gnn_explainer.py
avgupta456: Ah makes sense now.

torch_geometric/explain/config.py
avgupta456: I understand. Maybe change the order to be `object`, `attribute`, `common_attribute`, `both` in the Enum and ExplainerConfig docstring. Then describe both as enabling `object` and `attribute`. Is there a way to enable `object` and `common_attribute`?

avgupta456:
> > Hi! I'm looking to get started with link explanation soon and wanted to familiarize myself with current progress. I left some comments mostly to help myself understand, looking forward to reading your response.
> 
> Thanks for the review, I'll address your comments asap. Lots of good catch 👍🏽
> 
> > Echoing a previous comment, I also think it might be helpful to combine target_index and node/edge_index into a tuple of tensors used to index the output of the model relevant for the GNNExplainer optimization. This would allow for greater flexibility and better transfer to the link explanation setting.
> 
> So if two people suggest this, I might be missing something. Currently, if your model has a multioutput and you want to explain the kth element of this output you would just call `explainer(..., target_index=k, ...)`. If in addition you wanted to explain a specific node (or edge), let's say the ith you would just need to add `explainer(..., target_index=k, node_index=i ...)`. On the other hand if you had a node/edge level task and wanted the explanation for the ith node but did not care about the output index, you would use `explainer(..., node_index=i, ...)`. (Of course if available you can replace the indices by tensors). Could you show how we would do this with the tuple of tensors? What I have in mind would just be exactly the same (e.g., `index= (target_index,node_index)`).

My initial thought was combining target_index and node_index would simplify the user interface and add more functionality where a mask is required that filters different target_index for different node_index --> ex. simultaneously explain why node i was target k_1 and node j was target k_2.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,539 @@
...
+        **kwargs,
+    ) -> Tuple[Tensor, Optional[Tensor]]:
+        self._train_node_edge_mask(model, x, edge_index, explainer_config,
+                                   model_config, target, target_index, None,
+                                   **kwargs)
+
+        node_mask = self.node_mask.detach().sigmoid().squeeze(-1)
+        if explainer_config.edge_mask_type == MaskType.object:
+            edge_mask = self.edge_mask.detach().sigmoid()
+        else:
```
avgupta456, line 136: In a couple other areas, you do
```python
if mask_type == A:
    ...
elif mask_type == B:
    ...
```

The benefit of that is seeing clearly what triggers the second clause, but the downside is an invalid input might cause no action. Would if/elif/else raise Exception() be a better framework? Regardless I think it would be good to standardize.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,539 @@
...
+        if model_config.task_level not in [
+                ModelTaskLevel.node, ModelTaskLevel.graph
+        ]:
+            logging.error("Model task level not supported.")
+            return False
+        if explainer_config.edge_mask_type not in [MaskType.object, None]:
+            logging.error("Edge mask type not supported.")
+            return False
+
+        if explainer_config.node_mask_type in [MaskType.both, None]:
```
avgupta456, line 335: I would replace this with not in [...] to specifically clarify what is allowed. This would also make it more robust in case additional `node_mask_types` are added later.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,539 @@
...
+        model_config: ModelConfig,
+        target: Tensor,
+        target_index: Optional[Union[int, Tensor]] = None,
+        **kwargs,
+    ) -> Tuple[Tensor, Optional[Tensor]]:
+        self._train_node_edge_mask(model, x, edge_index, explainer_config,
+                                   model_config, target, target_index, None,
+                                   **kwargs)
+
+        node_mask = self.node_mask.detach().sigmoid().squeeze(-1)
```
avgupta456, line 133: Does this work for common_attributes? I also noticed in `check_explanation` in `test_gnn_explainer`, `MaskType.common_attributes` is not considered...

Additionally, can this and the mask transforms in `_explain_node()` be consolidated to prevent future errors.

test/explain/algoritms/test_gnn_explainer.py
```diff
@@ -0,0 +1,383 @@
...
+    explanation,
+):
+    if node_mask_type == MaskType.attributes:
+        assert explanation.node_features_mask.shape == x.shape
+        assert explanation.node_features_mask.min() >= 0
+        assert explanation.node_features_mask.max() <= 1
+    elif node_mask_type == MaskType.object:
+        assert explanation.node_mask.shape == x.shape[0]
+        assert explanation.node_mask.min() >= 0
+        assert explanation.node_mask.max() <= 1
```
avgupta456, line 178: Maybe also check expected behavior when the mask is None, in addition to `common_attributes` and `both`.

avgupta456 (commented):
Left some more comments.

torch_geometric/explain/config.py
dufourc1: The `_reshape_common_attributes` function is already in the base class, I'm not sure I see what you mean.

torch_geometric/explain/algorithm/base.py
dufourc1: Oh yes that could make sens. It's probably worth to be in another PR since this is closely related to #5925

torch_geometric/explain/config.py
dufourc1: yes you're right, I'll update the docs, it can be make clearer

>  Is there a way to enable object and common_attribute?
Not yet, not sure how to do it properly. Another mask type ?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,539 @@
...
+        gnn_explainer.py>`_.
+
+    Args:
+        epochs (int, optional): The number of epochs to train.
+            (default: :obj:`100`)
+        lr (float, optional): The learning rate to apply.
+            (default: :obj:`0.01`)
+        **kwargs (optional): Additional hyper-parameters to override default
+            settings in
+            :attr:`~torch_geometric.explain.algorithm.GNNExplainer.coeffs`.
```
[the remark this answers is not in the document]
dufourc1: good point, will add.

test/explain/algoritms/test_gnn_explainer.py
dufourc1: The initial  method does not support the `None` attributes for the node mask, I'll update it in the docs.

https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/gnn_explainer.py#L49-L53

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: yes I could uniformize that, it's a bit shaky at the moment (this also relates to your previous comment on the assert being handled in the `supports` function)

torch_geometric/explain/algorithm/gnn_explainer.py
avgupta456: I was mainly wondering if _explain_graph() handles explainer_config.node_mask_type == MaskType.common_attributes. In _explain_node(), _reshape_common_attributes is called on the node_mask, but that check isn't present in _explain_graph. If that is correct, then I have no additional feedback.

If we do need to _reshape_common_attributes() in _explain_graph(), then there is a section of code in both _explain_graph and _explain_node that handles creating the correct node_mask and edge_mask, and this can be abstracted out to either forward() or another function called in forward().

Does that make sense? I can clarify further if needed.

dufourc1: Oh yeah it make sens. It could but I'm not sure it's worth it, it might make some mechanism in the explain_node a bit less obvious (the part were the computational subgraph is selected, which requires some update on the node mask).

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: Sorry, could you point me to the precise loop you're referring ? I can't seem to find it.

torch_geometric/explain/algorithm/gnn_explainer.py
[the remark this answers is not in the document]
wsad1: Good catch.

dufourc1:
> My initial thought was combining target_index and node_index would simplify the user interface and add more functionality where a mask is required that filters different target_index for different node_index --> ex. simultaneously explain why node i was target k_1 and node j was target k_2.

What do you think about using `explainer(..., target = [k_1,k_2], node_index = [i,j], ...)` ?

rusty1s (commented):
I think this looks pretty good, but we should align on the notation. I think `index` is a more fitting term to refer to the model output indices. This also lets us merge all the different functions (`explain_node`, `explain_graph`, ...).

BTW: Thanks for maintaining the old interface as well, this is awesome!

torch_geometric/explain/config.py
```diff
@@ -16,6 +16,7 @@ class MaskType(Enum):
     object = 'object'
     attributes = 'attributes'
     both = 'both'
+    common_attributes = 'common_attributes'
```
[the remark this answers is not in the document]
dufourc1: a method masking both nodes and all the features at the same time, but you could encode that as a mask on the features multiplied by the mask of each node. Maybe let's delete until someone comes up with a method like this.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,558 @@
...
+                           node_mask=node_mask, node_feat_mask=node_feat_mask)
+
+    def _explain_graph(
+        self,
+        model: torch.nn.Module,
+        x: Tensor,
+        edge_index: Tensor,
+        explainer_config: ExplainerConfig,
+        model_config: ModelConfig,
+        target: Tensor,
```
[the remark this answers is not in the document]
dufourc1: that's a good point. Not sure how to go about the concrete implementation yet.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,558 @@
...
+        target_index: Optional[Union[int, Tensor]] = None,
+        **kwargs,
+    ) -> Explanation:
+
+        if isinstance(index, Tensor) and index.numel() > 1:
+            raise NotImplementedError(
+                "GNNExplainer only supports single node index for now")
+        if isinstance(target_index, Tensor) and target_index.numel() > 1:
+            raise NotImplementedError(
+                "GNNExplainer only supports single target index for now")
```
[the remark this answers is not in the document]
dufourc1: We would just need to change the code, but maybe we can do another PR so that this one get merged and does not block the explainability sprint ?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,558 @@
...
+        if target_idx is not None:
+            y_hat = y_hat[target_idx]
+            y = y[target_idx]
+
+        y_hat = self._to_log_prob(y_hat, return_type)
+
+        if node_index is not None:
+            loss = -y_hat[node_index, y[node_index]]
+        else:
+            loss = -y_hat[0, y[0]]
```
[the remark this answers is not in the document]
dufourc1: Not sure I understand, this is supposed to compute 

$$-\sum_{c=1}^C \mathbb{1}[y=c] \log P_{\Phi}\left(Y=y \mid G=A_c \odot \sigma(M), X=X_c\right),$$

and here `y_hat[0,k]` = $\log P_{\Phi}\left(Y=k \mid G=A_c \odot \sigma(M), X=X_c\right)$ and `y[0]` = $c$

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,558 @@
...
+                                   **kwargs)
+
+        node_mask = self.node_mask.detach().sigmoid().squeeze(-1)
+        if explainer_config.edge_mask_type == MaskType.object:
+            edge_mask = self.edge_mask.detach().sigmoid()
+        else:
+            edge_mask = None
+        return node_mask, edge_mask
+
+    def _explain_node(
```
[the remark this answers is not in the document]
dufourc1: And how you put back the explanations after computing them. Not sure how "nice" I can make it, but I'll try

torch_geometric/explain/algorithm/gnn_explainer.py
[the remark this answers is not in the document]
dufourc1: I guess the only change will happen in the loss function

avgupta456 (commented):
Some small testing bugs

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,389 @@
...
+    model = model()
+
+    out = model(x, edge_index, None, None).argmax(dim=-1)
+
+    if isinstance(model, GNN_classification_multioutput):
+        target_index = 0
+    else:
+        target_index = None
+
+    explainer = Explainer(model=model, algorithm=GNNExplainer(),
```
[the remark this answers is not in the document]
dufourc1: I even reduced it to `epochs=2`, we don't need the training part

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,389 @@
...
+        return_type=return_type,
+        mode="classification",
+    )
+
+    model = model()
+
+    out = model(x, edge_index, None, None).argmax(dim=-1)
+
+    if isinstance(model, GNN_classification_multioutput):
+        target_index = 0
```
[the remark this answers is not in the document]
dufourc1: Oh I think I know: for now we only support single index but I forgot to unpack if the tensor is only containing one element: good catch 👍🏽

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,389 @@
...
+    if edge_mask_type == MaskType.object:
+        assert explanation.edge_mask.shape == (edge_index.shape[1], )
+        assert explanation.edge_mask.min() >= 0
+        assert explanation.edge_mask.max() <= 1
+
+
+# type of masks allowed for GNNExplainer
+node_mask_types = ["attributes", "object", "common_attributes"]
+edge_mask_types = ["object", None]
+return_types_classification = ["log_probs", "raw", "probs"]
```
avgupta456, line 195: I think you forgot to use this

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,389 @@
...
+                      explanation)
+
+
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize(
+    'model', [GNN_classification_singleoutput, GNN_classification_multioutput])
+@pytest.mark.parametrize("explanation_type", ["model", "phenomenon"])
+@pytest.mark.parametrize('return_type', return_types_regression)
+def test_gnn_explainer_with_meta_explainer_classification_graph(
```
avgupta456, line 261: Small, optional request: Could you put the node tests before the graph tests since the node models are defined first. Also pick an ordering between classification/regression and single/multi output and stay consistent across model definitions and tests. 

Not sure if this is truly necessary but helps with readability for me.

avgupta456 (commented):
Mostly stylistic requests for tests

test/explain/algorithm/test_gnn_explainer.py
avgupta456: Sorry, this refers to `return_types_classification`

dufourc1: good catch, forgot to switch for classification 👍🏽

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,390 @@
...
+        x = self.conv2(x, edge_index)
+        return x.log_softmax(dim=1)
+
+
+# --------------------------------------------------------------------
+# model for graph level tasks
+# --------------------------------------------------------------------
+
+
+class GNN_regression_multi_output(torch.nn.Module):
```
avgupta456, line 89: I would rename to `GCN_graph_regression_multi_output`, and the node task version to `GCN_node_regression_multi_output`. Similarly for the other models.

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -0,0 +1,390 @@
...
+
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize("explanation_type", ["model", "phenomenon"])
+@pytest.mark.parametrize('model', [
+    GNN_regression_single_output,
+    GNN_regression_multi_output,
+])
+@pytest.mark.parametrize('return_type', return_types_regression)
+def test_gnn_explainer_with_meta_explainer_regression_graph(
```
avgupta456, line 302: Put test_classification_graph function before test_regression_graph to be consistent with node test order.

avgupta456 (approved):

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -0,0 +1,550 @@
...
+                                                  node_mask.size(-1),
+                                                  dtype=torch.float)
+                    node_feat_mask_[subset] = node_mask
+                elif task_level == ModelTaskLevel.graph:
+                    node_feat_mask_ = node_mask
+                else:
+                    raise NotImplementedError
+            elif explainer_config.node_mask_type == MaskType.common_attributes:
+                node_feat_mask_ = self._reshape_common_attributes(
+                    node_mask, num_nodes)
```
wsad1, line 184: if `task_level==node` then wouldn't we want to zero out the mask of nodes outside the subgraph?

wsad1 (approved):
Looks good. Thanks again for building this. 
Lets address `batch` support in another PR because i don't think its too trivial.

torch_geometric/explain/algorithm/gnn_explainer.py
dufourc1: Yes very good point, I'll update that ASAP.

rusty1s (approved):
Thank you! Overall, this looks great. I made some small adjustments (e.g., `check_explanation` was never called), and there was a bug with masks not being put on the correct device. Fixed now :)

ZeynepP:
Hi, 
a small question : why _get_masks does not return edge_feat_mask ?

rusty1s:
`GNNExplainer` does not support `edge_feat_mask` :(

PR #5972, opened 2022-11-14
Projected and Greedy Randomized Block Coordinate Descent (NeurIPS 2021)

sigeisler:
Hi PyG-team,

Thank you for the tremendous effort in putting this code base together and maintaining it to this high level of quality.

Similar to the `Explainer` and `GNN-Explainer` models, I hereby submit the code for our adversarial attacks based on [Randomized Block Coordinate Descent (RBCD) that appeared (NeurIPS 2021)](https://www.cs.cit.tum.de/daml/robustness-of-gnns-at-scale/). Our efficient (i.e sparse) gradient-based attacks apply to most models in PyG and cover the most popular threat models (see below for more details).

Adversarial robustness is quite related to explainability (here considering w.r.t. graph structure). Specifically, in an adversarial attack, we are asking for a limited number of edges that harm the prediction the most, and (GNN-) Explainer seeks for the set of friendly edges (i.e. minimal set without changing prediction). Due to the similarities, I have integrated attacks similar to how GNNExplainer was integrated.

Adversarial robustness of GNNs is an emerging field (the two seminal works were cited more than 1000 times) and it would be of even greater practical relevance if the "entry barrier" would be lowered. Moreover, having practical adversarial attacks in PyGs code base comes with opportunities for its users. These include: (1) seamless adversarial training for PyG's large collection of models; (2) adversarial attacks can be understood as a building block towards counterfactual explainability (and explainability is already part of PyG's codebase); (3) Adversarial attacks can be used for unsupervised representation learning (similar to Deep Graph Infomax). 

Importantly, having adversarial attacks generically implemented that can be directly applied to models (i.e. basic adaptive attacks) would counteract a common antipattern in the graph learning community. In our work
[Are Defenses for Graph Neural Networks Robust? (NeurIPS 2022)](https://www.cs.cit.tum.de/daml/are-gnn-defenses-robust/), we show that such adaptive attacks would greatly help to obtain more realistic robustness estimates.

Our Projected- and Greedy Randomized Block Coordinate Descent attacks are (afaik) the only practical general-purpose gradient-based adversarial attacks on GNNs (i.e. work with sparse matrices and come with a linear overhead for test time attacks). Our gradient-based approach relaxes the weights of the adjacency matrix from {0,1} -> [0,1] _during the attack_ and, therefore, seamlessly integrates with all models that support weighted adjacency matrices and do not apply non-differentiable operations to the edge weights. Although, non-differentiable models can often be made differentiable (e.g. using the gumble softmax trick).

Also with respect to tasks and threat models, this implementation covers the most popular cases. The defaults are set for node and graph classification (see unit tests for the latter) and other tasks (e.g. regression) can be handled by providing an appropriate model loss pair. The threat models covered are: (a) "global attacks": the adversary targets a large fraction of nodes at once (e.g. the test set); (b) "local attacks": the adversary targets the prediction of a single node in node-classification; any combination with (I) "evasion": attack at test time (most relevant for inductive learning); (II) "poisoning": attack at training time (most relevant for transductive learning, see poisoning example).

I am happy to include further suggestions and answer any questions. Please let me know if you have any feedback.


torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+        # The learning rate is refined heuristically, s.t. (1) it is
+        # independent of the number of perturbations (assuming an undirected
+        # adjacency matrix) and (2) to decay learning rate during fine-tuning
+        # (i.e. fixed search space).
+        lr = (budget / self.n * self.lr /
+              np.sqrt(max(0, epoch - self.epochs_resampling) + 1))
+        self.block_edge_weight.data.add_(lr * gradient)
+
+    @staticmethod
+    def _num_possible_edges(n: int, is_undirected_graph: bool) -> int:
```
sigeisler, line 617: I guess below here I could also move the methods into a "utils" file as DimeNet does, but I have no strong opinion.

sigeisler:
Thank you @wsad1

I am a bit confused, though, that you assigned me. Is there any action required from my side at the moment?

rusty1s:
@sigeisler No action required until reviews lands. This just means you are the author of this PR :)

torch_geometric/nn/models/__init__.py
```diff
@@ -22,43 +22,16 @@
...
-    'LightGCN',
-    'MaskLabel',
-    'GroupAddRev',
+    'MLP', 'GCN', 'GraphSAGE', 'GIN', 'GAT', 'PNA', 'EdgeCNN',
+    'JumpingKnowledge', 'Node2Vec', 'DeepGraphInfomax', 'InnerProductDecoder',
+    'GAE', 'VGAE', 'ARGA', 'ARGVA', 'SignedGCN', 'RENet', 'GraphUNet',
+    'SchNet', 'DimeNet', 'DimeNetPlusPlus', 'Explainer', 'to_captum',
+    'GNNExplainer', 'MetaPath2Vec', 'DeepGCNLayer', 'TGNMemory',
+    'LabelPropagation', 'CorrectAndSmooth', 'AttentiveFP', 'RECT_L', 'LINKX',
+    'LightGCN', 'MaskLabel', 'GroupAddRev', 'RBCDAttack'
```
EdisonLeeeee, line 34: Adding a comma after `RBCDAttack` could avoid autoformatting.

examples/rbcd_attack.py
```diff
@@ -0,0 +1,189 @@
...
+from torch_geometric.datasets import Planetoid
+from torch_geometric.nn import GATConv, GCNConv, RBCDAttack
+from torch_geometric.utils import softmax
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+# IMPORTANT: Edge weights are being ignored later and most adjacency
+# preprocessings should be part of the model (part of backpropagation)
+transform = T.Compose([T.NormalizeFeatures()])
+dataset = Planetoid(path, dataset, transform=transform)
```
EdisonLeeeee, line 20:
```suggestion
dataset = Planetoid(path, dataset, transform=T.NormalizeFeatures())
```

examples/rbcd_attack_poisoning.py
```diff
@@ -0,0 +1,197 @@
...
+import torch_geometric.transforms as T
+from torch_geometric.datasets import Planetoid
+from torch_geometric.nn import GCNConv, RBCDAttack
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+# IMPORTANT: Edge weights are being ignored later and most adjacency
+# preprocessings should be part of the model (part of backpropagation)
+transform = T.Compose([T.NormalizeFeatures()])
+dataset = Planetoid(path, dataset, transform=transform)
```
EdisonLeeeee, line 27:
```suggestion
dataset = Planetoid(path, dataset, transform=T.NormalizeFeatures())
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+        :rtype: (:class:`Tensor`, :class:`Tensor`)
+        """
+
+        raise NotImplementedError('Abstractmethod needs to be implemented')
+
+    def __call__(self, *args, **kwargs):
+        return self.attack(*args, **kwargs)
+
+
+class RBCDAttack(Attack):
```
EdisonLeeeee, line 53: Just a suggestion: could we implement PRBCG and GRBCD separately and inherit from RBCDAttack? E.g.:
```python
class RBCDAttack(nn.Module):
    ...
class PRBCDAttack(RBCD):
    ...
class GRBCDAttack(RBCD):
  ....
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+        return edge_index, edge_weight
+
+    def _filter_self_loops(self, with_weight: bool):
+        is_not_sl = self.block_edge_index[0] != self.block_edge_index[1]
+        self.current_block = self.current_block[is_not_sl]
+        self.block_edge_index = self.block_edge_index[:, is_not_sl]
+        if with_weight:
+            self.block_edge_weight = self.block_edge_weight[is_not_sl]
+
+    def _sample_random_block(self, budget: int = 0) -> None:
```
EdisonLeeeee, line 486:
```suggestion
    def _sample_random_block(self, budget: int = 0):
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+
+            self.block_edge_weight = torch.full_like(self.current_block,
+                                                     self.coeffs['eps'],
+                                                     dtype=torch.float32)
+            if self.current_block.size(0) >= budget:
+                return
+        raise RuntimeError('Sampling random block was not successful. '
+                           'Please decrease `budget`.')
+
+    def _resample_random_block(self, budget: int) -> None:
```
EdisonLeeeee, line 509:
```suggestion
    def _resample_random_block(self, budget: int):
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+
+    @staticmethod
+    def _linear_to_full_idx(n: int, lin_idx: Tensor) -> Tensor:
+        """Convert a linear index to index of matrix."""
+        row_idx = torch.div(lin_idx, n, rounding_mode='floor')
+        col_idx = lin_idx % n
+        return torch.stack((row_idx, col_idx))
+
+    @staticmethod
+    def _project(budget: int, values: Tensor, eps: float = 1e-7):
```
EdisonLeeeee, line 642: typehints

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+        r"""Project `values`: $budget \ge \sum \Pi_{[0, 1]}(\text{values})$."""
+        if torch.clamp(values, 0, 1).sum() > budget:
+            left = (values - 1).min()
+            right = values.max()
+            miu = RBCDAttack._bisection(values, left, right, budget)
+            values = values - miu
+        return torch.clamp(values, min=eps, max=1 - eps)
+
+    @staticmethod
+    def _bisection(edge_weights, a, b, n_pert, eps=1e-5, max_iter=1e3):
```
EdisonLeeeee, line 652: typehints

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,770 @@
...
+
+        # Prepare attack and define `self.iterable` to iterate over
+        self._prepare(x, edge_index, labels, budget, idx_attack, **kwargs)
+
+        if self.log:  # pragma: no cover
+            pbar = tqdm(total=len(self.step_sequence))
+            pbar.set_description('Attack model')
+
+        # Loop over the epochs (Algorithm 1, line 5)
+        for step in self.step_sequence:
```
EdisonLeeeee, line 240: [nit] we can use `for step in tqdm(self.step_sequence, disable=not self.log):`

EdisonLeeeee (commented):
Left a few initial comments. Will take a look again soon :)

torch_geometric/nn/models/rbcd_attack.py
sigeisler: Yes, I first had it like this but felt like it was (1) unnecessarily inflating the documentation, (2) changing the attack through configuration is convenient, and (3) subclassing it is somehow nicer this way (e.g. see poisoning example).

Of course, the cons are (a) the RBCD class is more complicated this way and (b) there are some attributes that are only used by `PRBCD`.

As I was unsure, I also asked some lab mates about what interface they preferred. The results were roughly 50/50.

So long story short, I am happy to go with the subclass solution, if you think it is a better fit. I do not have a very strong opinion on this tbh.

EdisonLeeeee (commented):
Sorry for the late. Left some minor points. Will review again these days.

torch_geometric/nn/models/rbcd_attack.py
EdisonLeeeee: I would vote for the subclass solution as it makes the interfaces clear and understandable for users. 
Alternatively, we can just implement two classes for them:
```python
class GRBCD(nn.Module or Attack):
    ...
class PRBCD(GRBCD):
   ...
```

But this requires a lot of effort for the changes to this version. @rusty1s could you give us some advice?

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,763 @@
...
+
+    @staticmethod
+    def _bisection(edge_weights: Tensor, a: float, b: float, n_pert: int,
+                   eps=1e-5, max_iter=1e3) -> Tensor:
+        """Bisection search for projection."""
+        def shift(offset: float):
+            return (torch.clamp(edge_weights - offset, 0, 1).sum() - n_pert)
+
+        miu = a
+        for _ in range(int(max_iter)):
```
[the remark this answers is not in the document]
sigeisler: Of course, but then I need to increment a variable and do the break after `max_iter` with an if statement etc. The issue is that (due to numerical issues) the `bisection` is not guaranteed to finish. Thus, I added an upper limit on iterations in contrast to the standard L0-PGD.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,763 @@
...
+        edge_index = torch.cat(
+            (self.edge_index.to(self.device), block_edge_index), dim=-1)
+        edge_weight = torch.cat(
+            (self.edge_weight.to(self.device), block_edge_weight))
+
+        edge_index, edge_weight = torch_sparse.coalesce(
+            edge_index, edge_weight, m=self.n, n=self.n, op='sum')
+
+        # Allow (soft) removal of edges
+        edge_weight[edge_weight > 1] = 2 - edge_weight[edge_weight > 1]
```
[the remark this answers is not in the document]
sigeisler: In the lines above I just stacked and then coalesced the `block_edge_index` and `block_edge_weight` with sum aggregation. Hence, if in the clean edges we already had an edge, the `edge_weight` is in [1 + eps, 2]. If the edge already existed in the clean graph, though, we actually want to *softly* drop it. Thus, `2 - 2 - edge_weight[edge_weight > 1]` will be in `[0, 1]` again. Moreover, the closer `block_edge_weight` is to 1, the closer is the `edge_weight` then to 0 (i.e. the more we are convinced that the edge should be dropped).

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,763 @@
...
+    @staticmethod
+    def _num_possible_edges(n: int, is_undirected_graph: bool) -> int:
+        """Determine number of possible edges for graph."""
+        if is_undirected_graph:
+            return n * (n - 1) // 2
+        else:
+            return int(n**2)  # We filter self-loops later
+
+    @staticmethod
+    def _linear_to_triu_idx(n: int, lin_idx: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
sigeisler: The closest references I know (they include the main diagonal though):
- https://dongkwan-kim.github.io/blogs/indices-for-the-upper-triangle-matrix/
- https://stackoverflow.com/questions/53233695/numpy-efficient-way-to-convert-indices-of-a-square-matrix-to-its-upper-triangul/53234021#53234021
- https://stackoverflow.com/questions/242711/algorithm-for-index-numbers-of-triangular-matrix-coefficients/244550#244550

torch_geometric/nn/models/rbcd_attack.py
sigeisler: I am happy to implement whatever is preferred. It is also not that much work as the interfaces etc. are all already clear.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,763 @@
...
+                RBCDAttack._num_possible_edges(self.n,
+                                               self.is_undirected_graph),
+                (n_edges_resample, ), device=self.device)
+
+            current_block = torch.cat((self.current_block, lin_index))
+            self.current_block, unique_idx = torch.unique(
+                current_block, sorted=True, return_inverse=True)
+
+            if self.is_undirected_graph:
+                self.block_edge_index = RBCDAttack._linear_to_triu_idx(
```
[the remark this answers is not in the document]
sigeisler: I also left a comment about how to handle these "utility" methods. So would you keep them but reference them with `self.`? If so, I would adapt the code accordingly but consistently.

torch_geometric/nn/models/rbcd_attack.py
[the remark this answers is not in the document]
sigeisler: Do you think it makes sense to add e.g. the equation to the docstring or rather some of the references?

torch_geometric/nn/models/rbcd_attack.py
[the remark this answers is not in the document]
EdisonLeeeee: ok, thanks for your explanation.

torch_geometric/nn/models/rbcd_attack.py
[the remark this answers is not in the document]
EdisonLeeeee: I think adding one single representative reference with some proper explanations is fine.

torch_geometric/nn/models/rbcd_attack.py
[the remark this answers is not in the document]
sigeisler: Fixed for all occasions where possible (i.e. all references but from static methods themselves).

torch_geometric/nn/models/rbcd_attack.py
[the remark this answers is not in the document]
sigeisler: Tried to make this more explicit

torch_geometric/nn/models/rbcd_attack.py
sigeisler: @EdisonLeeeee I have now switched to the suggested subclass solution. Please let me know if this is the version to continue. I think we should come to a decision on this before I continue any future feedback.

EdisonLeeeee: LGTM. Will review there again. Thanks for your effort!

test/nn/models/test_rbcd_attack.py
```diff
@@ -0,0 +1,119 @@
...
+        x = global_add_pool(x, batch)
+        x = self.lin(x)
+        return x
+
+
+@pytest.mark.parametrize('attack_type', [PRBCDAttack, GRBCDAttack])
+@pytest.mark.parametrize('model', [GCN, GNN])
+@pytest.mark.parametrize('budget', [1])
+@pytest.mark.parametrize('is_undirected_graph', [False, True])
+def test_attack(attack_type, model, budget, is_undirected_graph):
```
EdisonLeeeee, line 51: Also add the test for `early_stopping` if possible.

examples/rbcd_attack_poisoning.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        self.conv1 = gcn_conv(dataset.num_features, hidden_dim)
+        self.conv2 = gcn_conv(hidden_dim, dataset.num_classes)
+        self.gcn_norm = torch_geometric.nn.conv.gcn_conv.gcn_norm
+
+    def forward(self, x, edge_index, edge_weight=None, skip_norm=False):
+        # Normalizing once lowers memory footprint and caching is not possible
+        # during attack (for training it would be possible)
+        if not skip_norm:
+            edge_index, edge_weight = self.gcn_norm(edge_index, edge_weight,
```
EdisonLeeeee, line 48: [nit] we can import gcn_norm out of the class and call it directly.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,758 @@
...
+    out of the box (sampling needs to be adapted).
+
+    .. note::
+        For examples of using the GRBCD Attack, see
+        `examples/rbcd_attack.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        rbcd_attack.py>`_
+        for a test time attack (evasion) or `examples/rbcd_attack_poisoning.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        rbcd_attack_poisoning.py>`_
```
EdisonLeeeee, line 48: It seems we don't have the poisoning example for GRBCD.

CHANGELOG.md
```diff
@@ -101,11 +104,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 - Allow any integer tensors when checking edge_index input to message passing ([5281](https://github.com/pyg-team/pytorch_geometric/pull/5281))
+
 ### Removed
+
 - Removed `scatter_reduce` option from experimental mode ([#5399](https://github.com/pyg-team/pytorch_geometric/pull/5399))
 
 ## [2.1.0] - 2022-08-17
+
 ### Added
+
```
EdisonLeeeee, line 115: Let's remove these blanks.

EdisonLeeeee, line 115: Let's remove these blanks.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,758 @@
...
+        'with_early_stopping': True,
+        'eps': 1e-7
+    }
+
+    def __init__(self, model: torch.nn.Module, block_size: int = 250_000,
+                 epochs: int = 125, epochs_resampling: int = 100,
+                 loss: Optional[Union[str, LOSS_TYPE]] = 'prob_margin',
+                 metric: Optional[Union[str, LOSS_TYPE]] = None,
+                 lr: float = 1_000, is_undirected_graph: bool = True,
+                 log: bool = True, **kwargs) -> None:
```
EdisonLeeeee, line 527:
```suggestion
                 log: bool = True, **kwargs):
```

EdisonLeeeee (commented):
Thanks for addressing all the comments till now. This looks good to me mostly, left some last comments 😄 .

sigeisler:
@EdisonLeeeee thanks for the feedback! I have addressed all comments.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,758 @@
...
+        is_one_mask = torch.isclose(edge_weight, torch.tensor(1.))
+        self.edge_index = edge_index[:, is_one_mask]
+        self.edge_weight = edge_weight[is_one_mask]
+        # self.edge_weight = torch.ones_like(self.edge_weight)
+        assert self.edge_index.size(1) == self.edge_weight.size(0)
+
+        # Sample initial search space (Algorithm 2, line 3-4)
+        self._sample_random_block(step_size)
+
+        return {}
```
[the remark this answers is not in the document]
sigeisler: Good catch, I indeed wanted to add sth here.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}()'
+
+
+class GRBCDAttack(PRBCDAttack):
+    r"""Greedy Randomized Block Coordinate Descent (GRBCD) adversarial attack
+    from the `Robustness of Graph Neural Networks at Scale
+    <https://www.cs.cit.tum.de/daml/robustness-of-gnns-at-scale>`_ paper.
+
+    GRBCD shares most of the properties and requirements with PRBCD (see doc).
```
EdisonLeeeee, line 642:
```suggestion
    GRBCD shares most of the properties and requirements with :class:`PRBCDAttack`
```
Link to PRBCD

test/nn/models/test_rbcd_attack.py
```diff
@@ -0,0 +1,153 @@
...
+    x = torch.randn(8, 3)
+    edge_index = torch.tensor([[0, 1, 2, 3, 4, 5, 6], [1, 2, 3, 4, 5, 6, 7]])
+    if is_undirected_graph:
+        edge_index = to_undirected(edge_index)
+
+    if model == GNN:
+        y = torch.tensor([0])
+        # all nodes belong to same graph
+        kwargs = dict(
+            batch=torch.zeros(x.shape[0], dtype=int, device=x.device))
```
EdisonLeeeee, line 123:
```suggestion
            batch=edge_index.new_zeros(x.size(0))
```

test/nn/models/test_rbcd_attack.py
```diff
@@ -0,0 +1,153 @@
...
```
EdisonLeeeee, line 67:
```suggestion
            batch=edge_index.new_zeros(x.size(0))
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        rbcd_attack_poisoning.py>`_
+        for a training time (poisoning) attack.
+
+    Args:
+        model (torch.nn.Module): The GNN module to assess.
+        block_size (int, optional): Number of randomly selected elements in the
+            adjacency matrix to consider. (default: :obj:`250_000`)
+        epochs (int, optional): Number of epochs (aborts early if
+            :obj:`mode='greedy'` and budget is satisfied) (default: :obj:`125`)
```
EdisonLeeeee, line 58: The argument `mode` could be safely removed from the doc of PRBCD and GRBCD.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+        # independent of the number of perturbations (assuming an undirected
+        # adjacency matrix) and (2) to decay learning rate during fine-tuning
+        # (i.e. fixed search space).
+        lr = (budget / self.num_nodes * self.lr /
+              np.sqrt(max(0, epoch - self.epochs_resampling) + 1))
+        self.block_edge_weight.data.add_(lr * gradient)
+
+    @staticmethod
+    def _project(budget: int, values: Tensor, eps: float = 1e-7) -> Tensor:
+        r"""Project `values`: $budget \ge \sum \Pi_{[0, 1]}(\text{values})$."""
```
EdisonLeeeee, line 472:
```suggestion
        r"""Project :obj:`values`: :math: `budget \ge \sum \Pi_{[0, 1]}(\text{values})`."""
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+        # Sample initial search space (Algorithm 2, line 3-4)
+        self._sample_random_block(step_size)
+
+        # Return debug information
+        scalars = {
+            'number_positive_entries_in_gradient': (gradient > 0).sum().item()
+        }
+        return scalars
+
+    @torch.no_grad()
```
EdisonLeeeee, line 752:
```suggestion

```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+
+        flip_edge_index = self.block_edge_index[:, topk_edge_index]
+        flip_edge_weight = torch.ones_like(flip_edge_index[0],
+                                           dtype=torch.float32)
+
+        self.flipped_edges = torch.cat((self.flipped_edges, flip_edge_index),
+                                       axis=-1)
+
+        if self.is_undirected_graph:
+            flip_edge_index, flip_edge_weight = to_undirected(
```
EdisonLeeeee, line 725: [nit] The following code look better to me:
```python
flip_edge_index, flip_edge_weight = to_undirected(flip_edge_index,
                                                  flip_edge_weight
                                                  num_nodes=self.num_nodes,
                                                  reduce='mean')


```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+            if metric > best_metric:
+                best_metric = metric
+                best_edge_weight = self.block_edge_weight.clone().cpu()
+
+        # Recover best sample
+        self.block_edge_weight = best_edge_weight.to(self.device)
+        flipped_edges = self.block_edge_index[:,
+                                              torch.where(best_edge_weight)[0]]
+
+        edge_index, edge_weight = self._get_modified_adj(
```
EdisonLeeeee, line 453: [nit] The following code look better to me:
```python
edge_index, edge_weight = self._get_modified_adj(self.edge_index,
                                                 self.edge_weight,
                                                 self.block_edge_index,
                                                 self.block_edge_weight)
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+        if keep_above < sorted_idx.size(0) // 2:
+            keep_above = sorted_idx.size(0) // 2
+        sorted_idx = sorted_idx[keep_above:]
+
+        self.current_block = self.current_block[sorted_idx]
+
+        # Sample until enough edges were drawn
+        for _ in range(self.coeffs['max_trials_sampling']):
+            n_edges_resample = self.block_size - self.current_block.size(0)
+            lin_index = torch.randint(
```
EdisonLeeeee, line 384: [nit] Split this line of code into two lines:
```python
num_possible_edges = self._num_possible_edges(...)
lin_index = torch.randint(...)
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+                                         budget).indices] = 1
+            else:
+                sampled_edges = torch.bernoulli(block_edge_weight).float()
+
+            if sampled_edges.sum() > budget:
+                # Allowed budget is exceeded
+                continue
+            self.block_edge_weight = sampled_edges
+
+            edge_index, edge_weight = self._get_modified_adj(
```
EdisonLeeeee, line 437: [nit] this looks better to me:
```python
edge_index, edge_weight = self._get_modified_adj(self.edge_index,
                                                 self.edge_weight,
                                                 self.block_edge_index,
                                                 self.block_edge_weight)
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+        is_not_sl = self.block_edge_index[0] != self.block_edge_index[1]
+        self.current_block = self.current_block[is_not_sl]
+        self.block_edge_index = self.block_edge_index[:, is_not_sl]
+        if with_weight:
+            self.block_edge_weight = self.block_edge_weight[is_not_sl]
+
+    def _sample_random_block(self, budget: int = 0):
+        for _ in range(self.coeffs['max_trials_sampling']):
+            self.current_block = torch.randint(
+                self._num_possible_edges(self.num_nodes,
```
EdisonLeeeee, line 350: [nit] Split this line of code into two lines:
```python
num_possible_edges = self._num_possible_edges(...)
self.current_block = torch.randint(...)
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,755 @@
...
+            block_edge_index, block_edge_weight = to_undirected(
+                block_edge_index, block_edge_weight, num_nodes=self.num_nodes,
+                reduce='mean')
+
+        modified_edge_index = torch.cat(
+            (edge_index.to(self.device), block_edge_index), dim=-1)
+        modified_edge_weight = torch.cat(
+            (edge_weight.to(self.device), block_edge_weight))
+
+        modified_edge_index, modified_edge_weight = coalesce(
```
EdisonLeeeee, line 329: [nit] This looks better to me:
```python
modified_edge_index, modified_edge_weight = coalesce(modified_edge_index,
                                                     modified_edge_weight,
                                                     num_nodes=self.num_nodes,
                                                     reduce='sum')
```

EdisonLeeeee (approved):
Left a few final minor comments :)

torch_geometric/nn/models/rbcd_attack.py
sigeisler: Does hit the max number of characters

torch_geometric/nn/models/rbcd_attack.py
sigeisler: Does hit the max number of characters

torch_geometric/nn/models/rbcd_attack.py
sigeisler: Does hit the max number of characters

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,758 @@
...
+
+        self.model = model
+        self.block_size = block_size
+        self.epochs = epochs
+
+        if isinstance(loss, str):
+            if loss == 'masked':
+                self.loss = self._masked_cross_entropy
+            elif loss == 'margin':
+                self.loss = self._margin
```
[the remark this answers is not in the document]
EdisonLeeeee: There seems still an error during backward propagation as `_margin_loss` returns a Tensor instead of a scalar.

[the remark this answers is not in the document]
sigeisler: Oh damn... :) Fixed it now and added it to the unit tests to verify

sigeisler:
Hi @rusty1s I think we are done at the moment and would appreciate input from your side.

rusty1s:
Sounds good, thanks for the effort.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+class PRBCDAttack(torch.nn.Module):
+    r"""Projected Randomized Block Coordinate Descent (PRBCD) adversarial
+    attack from the `Robustness of Graph Neural Networks at Scale
+    <https://www.cs.cit.tum.de/daml/robustness-of-gnns-at-scale>`_ paper.
+
+    This attack uses an efficient gradient based approach that (during the
+    attack) relaxes the discrete entries in the adjacency matrix
+    :math:`\{0, 1\}` to :math:`[0, 1]` and solely perturbs the adjacency matrix
+    (no feature perturbations). Thus, this attack supports all models that can
+    handle weighted graphs that are differentiable w.r.t. these edge weights.
```
rusty1s, line 26:
```suggestion
    handle weighted graphs that are differentiable w.r.t. these edge weights, *e.g.*, :class:`~torch_geometric.nn.conv.GCNConv` or :class:`~torch_geometric.nn.conv.GraphConv`.
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+    For non-differentiable models you might be able to e.g. use the gumble
```
rusty1s, line 27:
```suggestion
    For non-differentiable models you might be able to, *e.g.*, use the gumble
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+        rbcd_attack.py>`_
+        for a test time attack (evasion) or `examples/rbcd_attack_poisoning.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        rbcd_attack_poisoning.py>`_
+        for a training time (poisoning) attack.
+
+    Args:
+        model (torch.nn.Module): The GNN module to assess.
+        block_size (int, optional): Number of randomly selected elements in the
+            adjacency matrix to consider. (default: :obj:`250_000`)
```
rusty1s, line 57: Is it reasonable to set a default here? I assume this highly depends on the chosen dataset. I think we should drop it.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+        epochs (int, optional): Number of epochs (aborts early if
+            :obj:`mode='greedy'` and budget is satisfied) (default: :obj:`125`)
+        epochs_resampling (int, optional): Number of epochs to resample the
+            random block. (default: obj:`100`)
+        loss (str or Callable, optional): A loss to quantify the "strength" of
+            an attack. Note that this function must match the output format of
+            :attr:`model`. By default, it is assumed that the task is
+            classification, that the model returns raw predictions (i.e. no
```
rusty1s, line 65:
```suggestion
            classification, and that the model returns raw predictions (*i.e.* no
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+            Either pass Callable or one of: :obj:`'masked'`, :obj:`'margin'`,
+            :obj:`'prob_margin'`, :obj:`'tanh_margin'`.
+            (default: :obj:`'prob_margin'`)
+        metric (Callable, optional): Second (potentially
+            non-differentiable) loss for monitoring or early stopping (if
+            :obj:`mode='greedy'`). (default: same as :attr:`loss`)
+        lr (float, optional): Learning rate for updating edge weights.
+            Additionally, it is heuristically corrected for :attr:`block_size`,
+            budget (see :attr:`attack`) and graph size. (default: :obj:`1_000`)
+        is_undirected_graph (bool, optional): If :obj:`True` the graph is
```
rusty1s, line 77:
```suggestion
        is_undirected (bool, optional): If :obj:`True` the graph is
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+        else:
+            self.loss = loss
+
+        self.is_undirected_graph = is_undirected_graph
+        self.log = log
+
+        if metric is None:
+            self.metric = self.loss
+        else:
+            self.metric = metric
```
rusty1s, line 129:
```suggestion
        if metric is None:
            self.metric = metric or self.loss
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+
+        self.epochs_resampling = epochs_resampling
+        self.lr = lr
+
+        self.coeffs.update(kwargs)
+
+    def attack(self, x: Tensor, edge_index: Tensor, labels: Tensor,
+               budget: int, idx_attack: Optional[Tensor] = None,
+               **kwargs) -> Tuple[Tensor, Tensor]:
```
rusty1s, line 138:
```suggestion
               **kwargs, ) -> Tuple[Tensor, Tensor]:
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+        """Attack the predictions for the provided model and graph.
+
+        A subset of predictions may be specified with :attr:`idx_attack`. The
+        attack is allowed to flip (i.e. add or delete) :attr:`budget` edges and
+        will return the strongest perturbation it can find. It returns both the
+        resulting perturbed :attr:`edge_index` as well as the perturbations.
+
+        Args:
+            x (Tensor): The node feature matrix. We assume `x` to be located
+                on target device.
```
rusty1s, line 148:
```suggestion
            x (Tensor): The node feature matrix.
```

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+            budget (int): The number of allowed perturbations (i.e.
+                number of edges that are flipped at most).
+            idx_attack (Tensor, optional): Filter for predictions/labels.
+                Shape and type must match that it can index :attr:`labels`
+                and the model's predictions.
+            **kwargs (optional): Additional arguments passed to the GNN module.
+
+        :rtype: (:class:`Tensor`, :class:`Tensor`)
+        """
+        assert self.block_size > budget, (
```
rusty1s, line 160: Can we change the `assert` to `raise ValueError`?

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+
+        self.model.eval()
+
+        self.device = x.device
+        assert kwargs.get('edge_weight', None) is None, \
+            '`edge_weight` is not supported'
+        edge_weight = torch.ones(edge_index.size(1), device=self.device)
+        self.edge_index = edge_index.cpu().clone()
+        self.edge_weight = edge_weight.cpu().clone()
+        self.num_nodes = x.size(0)
```
rusty1s, line 172: What's the reason we need to store them inside the model? Can we avoid this?

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+            f'greater than the number of permutations ({budget})')
+
+        self.model.eval()
+
+        self.device = x.device
+        assert kwargs.get('edge_weight', None) is None, \
+            '`edge_weight` is not supported'
+        edge_weight = torch.ones(edge_index.size(1), device=self.device)
+        self.edge_index = edge_index.cpu().clone()
+        self.edge_weight = edge_weight.cpu().clone()
```
rusty1s, line 171: Any reason these need to be on CPU? Why do we create `edge_weight` on the GPU and convert it to CPU afterwards?

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+        """Merges adjacency matrix with current block (incl. weights)"""
+        if self.is_undirected_graph:
+            block_edge_index, block_edge_weight = to_undirected(
+                block_edge_index, block_edge_weight, num_nodes=self.num_nodes,
+                reduce='mean')
+
+        modified_edge_index = torch.cat(
+            (edge_index.to(self.device), block_edge_index), dim=-1)
+        modified_edge_weight = torch.cat(
+            (edge_weight.to(self.device), block_edge_weight))
```
rusty1s, line 327: As mentioned above, would be great to get rid of these device transfers.

torch_geometric/nn/models/rbcd_attack.py
```diff
@@ -0,0 +1,760 @@
...
+            adjacency matrix to consider. (default: :obj:`250_000`)
+        epochs (int, optional): Number of epochs (aborts early if
+            :obj:`mode='greedy'` and budget is satisfied) (default: :obj:`125`)
+        epochs_resampling (int, optional): Number of epochs to resample the
+            random block. (default: obj:`100`)
+        loss (str or Callable, optional): A loss to quantify the "strength" of
+            an attack. Note that this function must match the output format of
+            :attr:`model`. By default, it is assumed that the task is
+            classification, that the model returns raw predictions (i.e. no
+            output activation) or uses :obj:`logsoftmax`, and that the number
```
rusty1s, line 66: Does it not matter whether we return raw or log-probs?

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+class GCN(torch.nn.Module):
+    """GCN that normalizes adjacency matrix once for all layers (no cache)."""
+    def __init__(self, hidden_dim: int = 16):
+        super().__init__()
+        gcn_conv = functools.partial(
+            GCNConv,
+            cached=False,  # Important to backpropagate to the adj. matrix
+            add_self_loops=False,  # We add them in `self.gcn_norm`
+            normalize=False  # It is better to normalize once
+        )
```
rusty1s, line 32: This is interesting. Would prefer to just instantiate via `GCNConv(..., normalize=False)` since I feel this is easier to understand.

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+        self.conv1 = gcn_conv(dataset.num_features, hidden_dim)
+        self.conv2 = gcn_conv(hidden_dim, dataset.num_classes)
+
+    def forward(self, x, edge_index, edge_weight=None):
+        # Normalizing once lowers memory footprint and caching is not possible
+        # during attack (for training it would be possible)
```
rusty1s, line 38:
```suggestion
        # Normalizing once lowers memory footprint:
```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+        edge_index, edge_weight = gcn_norm(edge_index, edge_weight, x.size(0),
+                                           add_self_loops=True)
+        x = F.relu(self.conv1(x, edge_index, edge_weight))
```
rusty1s, line 41:
```suggestion
        x = self.conv1(x, edge_index, edge_weight).relu()
```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+                    ptr: Optional[Tensor], size_i: Optional[int]) -> Tensor:
+        # Given edge-level attention coefficients for source and target nodes,
+        # we simply need to sum them up to "emulate" concatenation:
+        alpha = alpha_j if alpha_i is None else alpha_j + alpha_i
+        alpha = F.leaky_relu(alpha, self.negative_slope)
+
+        if edge_attr is not None:
+            assert edge_attr.dim() == 1, 'Only scalar edge weights supported'
+            edge_attr = edge_attr.view(-1, 1)
+            # `alpha`` unchanged if edge_attr == 1 and -Inf if edge_attr == 0;
```
rusty1s, line 60:
```suggestion
            # `alpha` unchanged if edge_attr == 1 and -Inf if edge_attr == 0;
```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+        gat_conv = functools.partial(
+            WeightedGATConv,
+            add_self_loops=True,  # In case the dataset has none
+            fill_value=1.,  # Default edge weight (for self-loops)
+        )
+        self.conv1 = gat_conv(dataset.num_features, hidden_dim)
+        self.conv2 = gat_conv(hidden_dim, dataset.num_classes)
+
+    def forward(self, x, edge_index, edge_weight=None, **kwargs):
+        x = F.relu(self.conv1(x, edge_index, edge_weight))
```
rusty1s, line 82:
```suggestion
        x = self.conv1(x, edge_index, edge_weight).relu()
```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+gcn = GCN().to(device)
+gat = GAT().to(device)
+data = data.to(device)
+
+x, edge_index, y = data.x, data.edge_index, data.y
+
+
+def train(model, x, edge_index, edge_weight=None, epochs=200, lr=0.01,
+          weight_decay=5e-4):
```
rusty1s, line 97:
```suggestion
def train(model, data, epochs=200, lr=0.01,
          weight_decay=5e-4):
```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+        x = self.conv2(x, edge_index, edge_weight)
+        return x
+
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+gcn = GCN().to(device)
+gat = GAT().to(device)
+data = data.to(device)
+
+x, edge_index, y = data.x, data.edge_index, data.y
```
rusty1s, line 93:
```suggestion

```

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+                               data.y[data.train_mask])
+        loss.backward()
+        optimizer.step()
+
+
+def accuracy(pred, lab, mask):
+    return (pred.argmax(-1)[mask] == lab[mask]).float().mean()
+
+
+gcn.train()
```
rusty1s, line 113: move `model.train()` to `train(...)` function.

examples/rbcd_attack.py
```diff
@@ -0,0 +1,187 @@
...
+        optimizer.zero_grad()
+        prediction = model(x, edge_index, edge_weight)
+        loss = F.cross_entropy(prediction[data.train_mask],
+                               data.y[data.train_mask])
+        loss.backward()
+        optimizer.step()
+
+
+def accuracy(pred, lab, mask):
+    return (pred.argmax(-1)[mask] == lab[mask]).float().mean()
```
rusty1s, line 110: Would prefer a `test(model, data)` function instead, which also calls `model.eval()`.

examples/rbcd_attack_poisoning.py
```diff
@@ -0,0 +1,197 @@
...
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+# IMPORTANT: Edge weights are being ignored later and most adjacency
+# preprocessings should be part of the model (part of backpropagation)
+dataset = Planetoid(path, dataset, transform=T.NormalizeFeatures())
+data = dataset[0]
+
+
+class GCN(torch.nn.Module):
```
rusty1s, line 30: Can we import this model from the previous example (similar to the PointNet examples).

examples/rbcd_attack_poisoning.py
```diff
@@ -0,0 +1,197 @@
...
+train(gcn, x, edge_index)
+gcn.eval()
+
+print('\n------------- GCN: Global Poisoning -------------\n')
+
+clean_accuracy = accuracy(gcn(x, edge_index), y, data.test_mask).item()
+print(f'Clean accuracy: {clean_accuracy:.3f}')
+
+
+class PoisoningPRBCDAttack(PRBCDAttack):
```
rusty1s, line 95: Any reason this is implemented in the example rather than as a module in `nn/models`?

examples/rbcd_attack_poisoning.py
```diff
@@ -0,0 +1,197 @@
...
+
+        return loss, gradient
+
+
+# The metric in PRBCD is assumed to be best if lower (i.e. like a loss)
+def metric(*args, **kwargs):
+    return -accuracy(*args, **kwargs)
+
+
+# for i in range(10):
```
rusty1s, line 162:
```suggestion

```

rusty1s (commented):
Thank you! Overall, this looks great. There are some questions for me on the need of adding model arguments as attributes of the module, and the need for the device transfers. Other than that, most comments are personal nits :)

sigeisler:
@rusty1s thank you for your time and the comments!

Here are my responses to the high-level questions:

> ... the need of adding model arguments as attributes of the module...

This was in analogy to the (old) implementation of GNNExplainer (have not checked how it is now). But I agree that we could add it to the `attack` method. I think the sole advantage of having this as a class attribute is, that one can move the attack (including) the model to a specific device. Although, the user anyhow has to handle anyways the device of the inputs of the `attack` method.

> ... the need for the device transfers.

The idea is to save (valuable) memory of the GPU/... Note that here we maintain a *clean* graph and construct its *perturbed* counterpart. Thus, without moving the clean *clean* graph to `cpu`, the hardware accelerator needs to store two graphs of similar size.

rusty1s:
Thanks for the clarification. This makes sense to me.

torch_geometric/nn/models/rbcd_attack.py
sigeisler: Applying `log_softmax` multiple consecutive times does not make a difference (except for a slightly higher memory footprint).

examples/rbcd_attack_poisoning.py
sigeisler: The main reason is that the train loop can have so many different variations. Of course, we could pass a method that defines the training procedure instead of tons of configurations. Moreover, the implementation here requires an additional dependency (i.e., `higher`).

If you prefer to have the "poisoning" as a dedicated model or configurable option, I am more than happy to implement it.

torch_geometric/nn/models/rbcd_attack.py
sigeisler: See last week's high-level comment

torch_geometric/nn/models/rbcd_attack.py
sigeisler: See last week's high-level comment

torch_geometric/nn/models/rbcd_attack.py
sigeisler: See last week's high-level comment

sigeisler:
@rusty1s I worked through your comments and adapted the code accordingly. I left some comments to explain etc. Please let me know if there are further action items.

Thanks again for your time and the helpful feedback!

sigeisler:
Hi @rusty1s I know you must be quite busy, however, I am wondering how to proceed. Thanks.

rusty1s:
This should be ready to merge. Will take a last look and merge it soon.

rusty1s (approved):

sigeisler:
Hi @rusty1s thanks for seeing this through :-) 

I have seen that you now have moved this to the `contrib` package, out of curiosity, what are the intended differences? Is there a lower level of quality or is this due to maintenance (e.g you do not plan to support all features in the long haul etc.)?

Btw. if you need any input whatsoever for promoting this as one of the models in the new `contrib` package, please let me know.

rusty1s:
Hi @sigeisler. Yes, I indeed moved this to `contrib` since we plan to move any big new model their first going forward. This does not have anything to do with code quality (that one is awesome), more so with maintenance and API stability. Hope that's okay :)

sigeisler:
@rusty1s makes a lot of sense! Thanks for clarifying.

PR #5992, opened 2022-11-16
pyg-lib `ToHeteroModule`

puririshi98:
[benchmark of new HeteroModule](https://github.com/puririshi98/rgcn_pyg_lib_forward_bench/blob/main/heteromodule_bench.py):
image
sufficient acceleration achieved
Remaining TODO for the next PR:
- integrate into transformer setup


rusty1s:
@puririshi98 I moved `Linear` and `MessagePassing` to its own classes instead of merging them two one. Hope that works for you :)

rusty1s (approved):

PR #6004, opened 2022-11-18
Triplet sampling in `NeighborLoader`

rusty1s:
This PR implements triplet sampling in link-level tasks, where we want only want to sample a negative destination node.
This PR got quite large, especially because I stumbled upon some weird things in the current link-level pipeline on temporal graphs, e.g., we don't want to sample nodes that do not yet exist for a given edge label timestamp.

When `neg_sampling="triplet"`, then the returned `Data` object holds:
* a `src_index` which denotes the index from left-hand-side embeddings
* a `dst_pos_index` which denotes the index from positive right-hand-side embeddings
* a `dst_neg_index` which denotes the index from negative right-hand-side embeddings

Overall, it introduces
* `NegativeSamplingConfig` with `strategy=binary/triplet`
* Deprecates `neg_sampling_ratio` in favor of `neg_sampling`
* Introduces a new method called `neg_sampling` that respects the timestamp of nodes while sampling
* Refactors and extends `sample_from_edges` => Overall, this function got quite large, and I will consider separating them into multiple functions in the future.


torch_geometric/loader/link_loader.py
```diff
@@ -54,9 +59,36 @@ class LinkLoader(torch.utils.data.DataLoader):
...
+            label from :obj:`0` to :obj:`num_classes - 1`.
+            After negative sampling, label :obj:`0` represents negative edges,
+            and labels :obj:`1` to :obj:`num_classes` represent the labels of
+            positive edges.
+            Note that returned labels are of type :obj:`torch.float` for binary
+            classification (to facilitate the ease-of-use of
+            :meth:`F.binary_cross_entropy`) and of type
+            :obj:`torch.long` for multi-class classification (to facilitate the
+            ease-of-use of :meth:`F.cross_entropy`).
+            If set to :obj:`"triplet"`, will randomly sample negative
```
wsad1, line 81: Can we mention that `edge_label` cannot be passed for `triplet`.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -396,106 +389,237 @@ def edge_sample(
...
+            else:
+                dst_node_time = node_time
+
+            dst_neg = neg_sample(dst, neg_sampling.amount, num_dst_nodes,
+                                 dst_time, dst_node_time)
+            dst = torch.cat([dst, dst_neg], dim=0)
+
+            if edge_label is None:
+                edge_label = torch.ones(num_pos)
+            elif edge_label.min() == 0:
```
wsad1, line 437: Edge case:
`edge_label.min()==0` is `True` for the whole data, but for a given minibatch this isn't True

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -396,106 +389,237 @@ def edge_sample(
...
+
+
+def neg_sample(seed: Tensor, num_samples: int, num_nodes: int,
+               seed_time: Optional[Tensor],
+               node_time: Optional[Tensor]) -> Tensor:
+    if node_time is None:
+        return torch.randint(num_nodes, (seed.numel() * num_samples, ))
+
+    # Otherwise, we can only sample negatives that exist in the graph:
+    # TODO See if this greedy algorithm here can be improved.
```
wsad1, line 604: Add a comment on why we can only sample negatives that exist in the graph.

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -396,106 +389,237 @@ def edge_sample(
...
+    seed_time = seed_time.view(1, -1).expand(num_samples, -1)
+    out = torch.randint(num_nodes, (num_samples, seed.numel()))
+    mask = node_time[out] >= seed_time
+    neg_sampling_complete = False
+    for i in range(5):  # pragma: no cover
+        if not mask.any():
+            neg_sampling_complete = True
+            break
+
+        # Greedily search for alternative negatives.
```
wsad1, line 615: is there some way to avoid sampling false negatives, given we have `edge_label_index`

torch_geometric/loader/link_loader.py
```diff
@@ -106,20 +140,24 @@ def __init__(
...
         )
-        self.neg_sampling_ratio = neg_sampling_ratio
+        self.neg_sampling = NegativeSamplingConfig.cast(neg_sampling)
         self.transform = transform
         self.filter_per_worker = filter_per_worker
 
+        if (self.neg_sampling is not None and self.neg_sampling.is_triplet()
+                and edge_label is not None):
+            raise ValueError("'edge_label' needs to be undefined for "
+                             "'triplet' negative sampling")
```
wsad1, line 150: Explain in the error or in comments why `triplet` cant' support `edge_label`.

examples/graph_sage_unsup_ppi.py
```diff
@@ -20,7 +20,7 @@
 # Group all training graphs into a single graph to perform sampling:
 train_data = Batch.from_data_list(train_dataset)
 loader = LinkNeighborLoader(train_data, batch_size=2048, shuffle=True,
-                            neg_sampling_ratio=0.5, num_neighbors=[10, 10],
+                            neg_sampling_ratio=1.0, num_neighbors=[10, 10],
```
wsad1, line 23: Update example to use `neg_sampling` and point to this example in `link_loader` docs.

test/data/test_lightning_datamodule.py
```diff
@@ -329,9 +329,7 @@ def test_lightning_hetero_link_data():
     )
     assert isinstance(datamodule.neighbor_sampler, NeighborSampler)
     for batch in datamodule.train_dataloader():
-        assert 'edge_label' in batch['author', 'paper']
```
wsad1, line 332: Why remove this check? isn't the default `binary` which adds `edge_label`.

wsad1 (commented):
Left some initial comments.

torch_geometric/sampler/base.py
```diff
@@ -66,6 +70,47 @@ class HeteroSamplerOutput:
...
+    binary = 'binary'
+    # 'triplet': Randomly sample negative destination nodes for each positive
+    # source node.
+    triplet = 'triplet'
+
+
+@dataclass
+class NegativeSamplingConfig(CastMixin):
+    strategy: NegativeSamplingStrategy
+    amount: int = 1
```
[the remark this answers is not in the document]
rusty1s: We can, but only for binary and non-temporal cases, because the current negative sampling behavior assumes fixed-sized neighbors.

torch_geometric/sampler/neighbor_sampler.py
rusty1s: Will add a TODO for this. I think this is currently out-of-scope for this PR.

examples/graph_sage_unsup_ppi.py
rusty1s: Will update examples once PyG 2.2 lands to avoid breaking examples for all users with PyG 2.1

test/data/test_lightning_datamodule.py
rusty1s: We no longer add `edge_label` in case `neg_sampling = None` and no `edge_label` was passed.

rusty1s:
Incorporated your review. Thank you!

torch_geometric/loader/link_neighbor_loader.py
```diff
@@ -114,22 +115,38 @@ class LinkNeighborLoader(LinkLoader):
...
-            If :obj:`neg_sampling_ratio > 0` and in case :obj:`edge_label`
-            does not exist, it will be automatically created and represents a
-            binary classification task (:obj:`1` = edge, :obj:`0` = no edge).
-            If :obj:`neg_sampling_ratio > 0` and in case :obj:`edge_label`
-            exists, it has to be a categorical label from :obj:`0` to
-            :obj:`num_classes - 1`.
+        neg_sampling (NegativeSamplingConfig, optional): The negative sampling
+            strategy. Can be either :obj:`"binary"` or :obj:`"triplet"`, and
+            can be further customized by an additional :obj:`amount` argument
+            to control the number of negatives to sample.
```
wsad1, line 121:
```suggestion
            to control the ratio of negative to positive samples.
```

test/loader/test_link_neighbor_loader.py
```diff
@@ -213,14 +212,12 @@ def test_temporal_heterogeneous_link_neighbor_loader():
         edge_label_time=edge_time,
         batch_size=32,
         time_attr='time',
-        neg_sampling_ratio=0.5,
+        neg_sampling_ratio=1.0,
```
wsad1, line 215:
```suggestion
        neg_sampling_ratio=0.5
```
Lets test the case of `amount<1`.

test/loader/test_link_neighbor_loader.py
```diff
@@ -213,14 +212,12 @@ def test_temporal_heterogeneous_link_neighbor_loader():
...
         drop_last=True,
     )
     for batch in loader:
         # Check if each seed edge has a different batch:
-        assert int(batch['paper'].batch.max()) + 1 == 32 + 16
-        # Check if each seed edge has a different source and dstination node:
-        assert batch['paper'].num_nodes >= 2 * (32 + 16)
+        assert int(batch['paper'].batch.max()) + 1 == 32
```
wsad1, line 220: Why isn't this 64? Won't the negative samples also have a disjoint batch?

wsad1 (commented):
Left some final comments.

test/loader/test_link_neighbor_loader.py
rusty1s: I personally think it makes more sense to group the negatives to its own batch, in particular because they now share time information.

rusty1s:
@wsad1 Updated based on your comments.

wsad1 (approved):
Looks great. Thanks.

PR #5746, opened 2022-10-17
Added `cpu_worker_affinity_cores` attribute to `node_loader` init

JakubPietrakIntel:
This PR enables CPU affinitization for `node_loader` using a function from `psutil` package (`psutil.Process().cpu_affinity()`).
Affinitization is switched on by passing a correct input to `NeihborLoader` instantiation, namely:

- `use_cpu_worker_affinity = True`
- `cpu_worker_affinity_cores = <list of workers' CPU IDs>` , i.e. list(range(num_workers))

CPU affinitization allows for maximizing CPU performance for Pytorch as described in [Maximize Performance of Intel® Optimization for PyTorch* on CPU](https://www.intel.com/content/www/us/en/developer/articles/technical/how-to-get-better-performance-on-pytorchcaffe2-with-intel-acceleration.html).  For PyG it can be used to optimize execution of the NeighborLoader threads. The gain is observed for dual socket machines, especially at large batch-sizes, so when the performance bottleneck is on the side of data-loading, rather than compute threads. 

Primary HW to run benchmark is:
Dual Socket IceLake CPU (Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz) 
512 GB RAM

Affinization may be used with additional Intel OpenMP arguments that bind threads to physical CPUs, in example:
- OMP_NUM_THREADS = <nr_physical_cores> - <nr_workers>
- GOMP_CPU_AFFINITY = "< list of CPU IDs not in cpu_worker_affinity_cores > " 
i.e. for 2x CPU with 40 cores each and nr_workers=2;  cpu_worker_affinity_cores=[0,1]; OMP_NUM_THREADS=78; GOMP_CPU_AFFINITY="2-79"
**Important:** above settings need to be changed before the script execution had started, therefore can't be changed 2from within a python script.

**[WIP]**
The complete proof is still under construction, as we're running benchmarks on different machines to gather more data. Combination of CPU affinitization and hyperthreading is under investigation.
Results analysis is in progress.

PR #6005, opened 2022-11-18
DataLoader threads affinitization

JakubPietrakIntel:
**Feature description:**
DataLoader threads affintization can be enabled with `NodeLoader.enable_cpu_affinity()` as a context manager function.
It can affinitize to a user-provided list of CPU IDs or by default to numa0 node cores starting at core1, where core0 is left out for running any other background threads: system-level, OMP, etc. The affinitization is implemented with psutil method and applied to each worker_id at the time of initialization 
 https://github.com/JakubPietrakIntel/pytorch_geometric/blob/28a4694a85914196962cb9363e53bbd33b5eebba/torch_geometric/loader/node_loader.py#L173

Additional helper function `get_numa_nodes()` has been added to utils. It is responsible for reading CPU core ids from `/sys/devices/system/node/` for Linux machines. For other operating systems standard `psutil.cpu_count(logical=False)` will be applied.

https://github.com/JakubPietrakIntel/pytorch_geometric/blob/28a4694a85914196962cb9363e53bbd33b5eebba/torch_geometric/loader/utils.py#L210-L253

**Benchmarking:**
To test the feature, extensive benchmarks were ran on a dual socket system with Intel(R) Xeon(R) Gold 6348 CPU @ 2.60GHz (ICX) with 28 cores each and 512GB of RAM DDR4. The script to measure inference time was `inference_benchmark.py` with a GCN model and ogbn-products dataset over a range of hyperparameters: 

- hidden feature size [16, 128]
- number of layers [2,3]
- hyperthreading (HT) on/off
- sparse or dense tensor matrix multiplication (`spmm`/`scatter_add`)
- range of DataLoader workers [1, 2, 4, 8, 16]
- batch_size [512, 1024, 2048, 4096, 8192]

In the plots presented at the bottom of this page the results are presented for CPU setup without hyperthreading, which yielded overall better results than with HT on. In this scenario 3 example affinitization configurations are presented in comparison with the baseline config.
- DL - DataLoader affinitzation only
- DL + CPU All -  compute threads affinitzation for all cores not used by DataLoader
- DL + CPU1 - decreased number of compute threads running on the same NUMA node/CPU socket as DataLoader

**Results discussion:**
In majority of cases DL threads affinitization yields a measurable improvement up to 2x for small hidden feature size and large batch size. The inference time can be optimized further using compute threads affinitzation that, as of now, can only be done by setting additional OMP parameters, prior to launching python script: limiting number of threads with OMP_NUM_THREADS and pinning them to specific CPUs with GOMP_CPU_AFFINITY (see CPU All and CPU1 cases). 
There's no single one-fit-all solution for CPU config that can be applied to all models and architectures. Complexity of the problem arises from mutual dependency between performance of data-loading and compute threads and HW architecture of a CPU model. Therefore, further research will be performed to develop more user-friendly interfaces for thread management. 

**Further work:**
- Reserach on possibility of compute threads affinitization from within pytorch multiprocessing library and without external OMP commands.
- Increasing dual-socket CPU utilization and performance with multiple DataLoader workers acting in parallel.

**Plots:**
SPMM
spmm-feat16HT0
spmm-feat128HT0
SCATTER
scatter-feat16HT0
scatter-feat128HT0


yanbing-j:
Hi @JakubPietrakIntel ,Thanks for your hard work!
I have some questions about this PR.

1. Is this PR the replacement of https://github.com/pyg-team/pytorch_geometric/pull/5746 What's the difference between them?
2. This PR requires at least 1 DL workers. https://github.com/pyg-team/pytorch_geometric/pull/6005/files#diff-7f627486416900d3cbaa74871470419e5a15c044e1902acd4e0921fbcc684755R200-R203 Is this reasonable? And what about the performance of num_workers = 0?
3. Do you test on other smaller datasets, and also get the same improvement as ogbn-products?
4. From the plots, CPU1 can get the smallest inf time relatively. So running in single socket can perform better than running in two sockets, right?

JakubPietrakIntel:
Hi Yanbing, let me answer your questions.
> 1. Is this PR the replacement of [Added `cpu_worker_affinity_cores` attribute to `node_loader` init #5746](https://github.com/pyg-team/pytorch_geometric/pull/5746) What's the difference between them?

This PR is a replacement for #5746 and implements a more complete solution that uses contextmanager to facilitate affinitzation function. It also automatically retrieves information about HT setting and numa core ids.
> 2. This PR requires at least 1 DL workers. https://github.com/pyg-team/pytorch_geometric/pull/6005/files#diff-7f627486416900d3cbaa74871470419e5a15c044e1902acd4e0921fbcc684755R200-R203 Is this reasonable? And what about the performance of num_workers = 0?

In most cases included in my benchmarks DL operating in main process (nr_workers=0) is worse than using separate processes for the DL. However, for some corner cases, i.e  large hidden feature size, using sparse data repr. and with HT off I've noticed that main process DL performs better. This indicates a perf bottleneck that needs to be eliminated.
> 3. Do you test on other smaller datasets, and also get the same improvement as ogbn-products?

Initially, I performed some manual runs that indicated that there will be a perf. boost also while running with a smaller dataset. I will provide benchmark results for a smaller dataset, i.e. Reddit soon. 
> 4. From the plots, CPU1 can get the smallest inf time relatively. So running in single socket can perform better than running in two sockets, right?

Correct, in majority of cases running on a single socket leads to better performance, because it eliminates the remote memory access which causes significant memory bound. I will be investigating how to utilize dual socket CPU better. It will most likely require compute threads affinitzation & binding. This is also an ongoing research.

benchmark/inference/inference_benchmark.py
```diff
@@ -108,26 +109,28 @@ def run(args: argparse.ArgumentParser) -> None:
...
                         model.eval()
-
-                        with amp:
-                            for _ in range(args.warmup):
-                                model.inference(subgraph_loader, device,
-                                                progress_bar=True)
-                            with timeit():
-                                # becomes a no-op if vtune_profile == False
-                                with emit_itt(args.vtune_profile):
+                        with subgraph_loader.enable_cpu_affinity(
```
rusty1s, line 112: There is a lot of nesting going on here, but it looks like Python supports chaining of context managers. For example, we could simplify this to:
```python
if args.cpu_affinity:
    cpu_affinity = subgraph_loader.enable_cpu_affinity(args.loader_cores)
else:
    cpu_affinity = nullcontext()
profile = torch_profile() if args.profile else nullcontext()
with cpu_affinity, amp, timeit(), emit_itt(args.vtune_profile):
    pass
```

torch_geometric/loader/node_loader.py
```diff
@@ -89,9 +92,17 @@ def __init__(
         self.input_data = InputData(input_nodes, input_time)
         self.transform = transform
         self.filter_per_worker = filter_per_worker
+        # CPU Affinitization for loader and compute cores
```
rusty1s, line 95: Can we add a TODO here to unify this behind a `BaseDataLoader` class?

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
         return DataLoaderIterator(super()._get_iterator(), self.filter_fn)
 
+    def __enter__(self):
+        return self
+
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}()'
+
+    @contextmanager
+    def enable_cpu_affinity(self, loader_cores: List[int] = None):
```
rusty1s, line 172: Is it possible to add a basic test for this?

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+        """ Helper method for enabling cpu affinity for compute threads and dataloader workers
```
rusty1s, line 173:
```suggestion
        r"""A context manager to enable cpu affinity for compute threads and dataloader workers
```

torch_geometric/loader/node_loader.py
```diff
@@ -89,9 +92,17 @@ def __init__(
         self.input_data = InputData(input_nodes, input_time)
         self.transform = transform
         self.filter_per_worker = filter_per_worker
+        # CPU Affinitization for loader and compute cores
+        self.num_workers = kwargs.get('num_workers', 0)
+        self.device = torch.device(
+            'cuda' if torch.cuda.is_available() else 'cpu')
```
rusty1s, line 98: Can we rename that to something like
```python
self.is_cuda_available = torch.cuda.is_available()
```

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
         return f'{self.__class__.__name__}()'
+
+    @contextmanager
+    def enable_cpu_affinity(self, loader_cores: List[int] = None):
+        """ Helper method for enabling cpu affinity for compute threads and dataloader workers
+        Only for CPU devices
+        Uses NUMA node 0 by default for multi-node systems
+
+        Args:
+            loader_cores : [int] (optional)
```
rusty1s, line 178:
```suggestion
            loader_cores ([int], optional):
```

torch_geometric/loader/node_loader.py
rusty1s, line 173: A few sentences about explaining CPU affinity would be great here :)

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+                List of cpu cores to which dataloader workers should affinitize to.
+                By default cpu0 is reserved for all auxiliary threads & ops.
+                DataLoader wil affinitize to cores starting at cpu1.
+                default: node0_cores[1:num_workers]
+        Usage:
+        with dataloader.enable_cpu_affinity(losfrt):
+        <training loop>
```
rusty1s, line 185: Define as `.. code-block:: python`.

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+        """
+        if self.device.type == 'cpu':
+            if not self.num_workers > 0:
+                raise Exception(
```
rusty1s, line 189:
```suggestion
                raise ValueError(
```

torch_geometric/loader/node_loader.py
rusty1s, line 172:
```suggestion
    def enable_cpu_affinity(self, loader_cores: Optional[List[int]] = None):
```

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+                    'ERROR: affinity should be used with at least one DL worker'
+                )
+            if loader_cores and len(loader_cores) != self.num_workers:
```
rusty1s, line 192:
```suggestion
            if loader_cores is not None and len(loader_cores) != self.num_workers:
```

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+                try:
+                    psutil.Process().cpu_affinity([loader_cores[worker_id]])
+                except:
+                    raise Exception(
+                        f'ERROR: cannot use affinity id={worker_id} cpu={loader_cores}'
+                    )
+
+                worker_init_fn_old(worker_id)
+
+            if not loader_cores:
```
rusty1s, line 213:
```suggestion
            if loader_cores is None:
```

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+            if loader_cores and len(loader_cores) != self.num_workers:
+                raise Exception(
+                    'ERROR: cpu_affinity incorrect '
+                    f'number of loader_cores={loader_cores} for num_workers={self.num_workers}'
+                )
+
+            worker_init_fn_old = self.worker_init_fn
+            affinity_old = psutil.Process().cpu_affinity()
+            nthreads_old = torch.get_num_threads()
+            loader_cores = loader_cores[:] if loader_cores else []
```
rusty1s, line 201:
```suggestion
            loader_cores = loader_cores[:] if loader_cores else None
```

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+
+                # set default loader core ids
+                loader_cores = loader_cores or node0_cores[1:self.num_workers +
+                                                           1]
+
+            try:
+                # set cpu affinity for dataloader
+                self.worker_init_fn = init_fn
+                self.cpu_affinity_enabled = True
+                print(
```
rusty1s, line 232: Shall we use `logging` module for this?

torch_geometric/loader/node_loader.py
```diff
@@ -143,8 +154,103 @@ def _get_iterator(self) -> Iterator:
...
+                # restore omp_num_threads and cpu affinity
+                psutil.Process().cpu_affinity(affinity_old)
+                torch.set_num_threads(nthreads_old)
+                self.worker_init_fn = worker_init_fn_old
+                self.cpu_affinity_enabled = False
+        else:
+            yield
+
+
+class WorkerInitWrapper(object):
```
rusty1s, line 246:
```suggestion
class WorkerInitWrapper:
```

rusty1s, line 246: Can we move this to `loader/base.py`?

rusty1s (commented):
This looks great, thanks for all the hard effort. Most commons are only super nit :)

yanbing-j:
Hi @JakubPietrakIntel , thanks for your explanation. Still have some questions.

For the corner cases, does `num-hidden_channels` = 64 belongs to large hidden feature sizes? And if use `numactl -C 0-xx -m 0` to fix physical cores in socket 0, is this equal to CPU1 without HT?

I run with the above options, and cannot get the improvement as expected. Can you share your configs? Do you use iomp or jemalloc? Single socket config is enough for me.

JakubPietrakIntel:
Hi! Let's keep the discussion ongoing :) I'd like everything to be clear, please feel free to ask.

> For the corner cases, does `num-hidden_channels` = 64 belongs to large hidden feature sizes? And if use `numactl -C 0-xx -m 0` to fix physical cores in socket 0, is this equal to CPU1 without HT?
> I run with the above options, and cannot get the improvement as expected. Can you share your configs? Do you use iomp or jemalloc? Single socket config is enough for me.

**TLDR:** The important check would be to see if you're able to get an improvement with simple DataLoader affinitization default setting and without any additional thread control.

**Long explanation:**
Firstly, to turn off Hyperthreading you need to set it up in BIOS or on a Linux machine you may do simply set 
`echo off > /sys/devices/system/cpu/smt/control`, first making sure that you have the access rights to modify the file.  You can check if the HT is off by running `cat /sys/devices/system/cpu/smt/active` which should result in `0`, else it would give `1` if HT on. You might also check off-line CPUs in lscpu output.
To limit the amount of threads I am using OMP_NUM_THREADS and to set threads affinity GOMP_CPU_AFFINITY. To the best of my knowledge using GOMP commands doesn't allow you to control if there will be thread switching between logical cores, this could be done using low-level API - KMP_AFFINITY. Therefore if HT is on and you affinitize to logical cores, you could still observe threads being migrated between logical cores of each logical cpu.   

By default the DataLoader workers will take cores [1.. nr_workers] therefore it is important to **exclude these core ids** from affinitization settings for compute processes to avoid core over-subscribing and unnecessary conflicts between threads. This will lead to immediate drop in performance.  Then, we alco exclude cpu0 from compute to leave one thread for all other ops that by default will run on cpu0. To give an example:
Say you're running on a dual socket CPU with 20 Cores (CPU0 has core ids 0-19, CPU1 has core ids 20-39) each and you want to use 3 workers and "single socket mode". If you don't provide a list with core ids to `NodeLoader.enable_cpu_affinity(loader_cores=[])` the workers take [core1, core2, core3]. 
For compute you need to specify OMP_NUM_THREADS=20-nr_workers-1=16 
GOMP takes a literal argument that defines range of core ids.
GOMP_CPU_AFFINITY=3-19 (here '-' is a dash)

In effect, if HT is off, while running htop you should see that the workload is being executed only on CPU0 cores 0-19.

JakubPietrakIntel:
Inference results for Reddit dataset with HT ON and denser tensor multiplication.
reddit-feat16HT1ST0
reddit-feat128HT1ST0
DL affinitization results with single CPU are always equal or better to baseline.

JakubPietrakIntel:
@yanbing-j I am planning to develop this feature further to include compute threads affinitization. When I have all the results ready I will write a user-friendly guide for "CPU best practices for PyG". Please stay tuned.

torch_geometric/loader/node_loader.py
JakubPietrakIntel: added

torch_geometric/loader/node_loader.py
JakubPietrakIntel: done

torch_geometric/loader/node_loader.py
JakubPietrakIntel: done

torch_geometric/loader/node_loader.py
JakubPietrakIntel: Please let me know what you think about such description:
"A context manager to enable CPU affinity for DataLoader workers.
        Only for CPU devices!
        As of now, be default it uses NUMA node 0 cores, for a multi-node system.
        This will be gradually extended to increase performance on dual socket CPUs.
        
        Affinitization places DataLoader workers threads on specific CPU cores. In effect, 
        it allows for more efficient local memory allocation and reduces remote memory calls.
        Every time a process or thread moves from one core to another, registers and caches 
        need to be flushed and reloaded. This can become very costly if it happens often, 
        and our threads may also no longer be close to their data, or be able to share data in a cache.
        
        Important:
        If you want to further affinitize compute threads (i.e. with OMP), please make sure that you exclude 
        loader_cores from the list of cores available for compute. This will cause core oversubsription
        and exacerbate performance.
        Args:
            loader_cores : [int] (optional)
                List of cpu cores to which dataloader workers should affinitize to.
                By default cpu0 is reserved for all auxiliary threads & ops.
                DataLoader wil affinitize to cores starting at cpu1.
                default: node0_cores[1:num_workers]
        Usage:
        with dataloader.enable_cpu_affinity(losfrt):
        <training loop>"

torch_geometric/loader/node_loader.py
JakubPietrakIntel: check

torch_geometric/loader/node_loader.py
JakubPietrakIntel: check

torch_geometric/loader/node_loader.py
JakubPietrakIntel: check

torch_geometric/loader/node_loader.py
JakubPietrakIntel: check

torch_geometric/loader/node_loader.py
JakubPietrakIntel: moved

benchmark/inference/inference_benchmark.py
JakubPietrakIntel: Changed for cleaner structure as suggested. I also added `reset()` function to the `timeit()` module, because I found it useful to measure times for warmup and inference loops. With reset all can be defined in the same line.

torch_geometric/loader/node_loader.py
JakubPietrakIntel: changed print for logging.debug()

torch_geometric/loader/node_loader.py
JakubPietrakIntel: Added test `test_cpu_affinity_neighbor_loader()` in `test_neighbor_loader.py` 
https://github.com/JakubPietrakIntel/pytorch_geometric/blob/2ce828aa451bd2a4c9ad7c0deb8a7e0cee99437b/test/loader/test_neighbor_loader.py#L578-L599

JakubPietrakIntel:
The test results for loader affinitization are unstable and I will need to investigate this case further before merging the PR. More info to follow.

JakubPietrakIntel:
> The test results for loader affinitization are unstable and I will need to investigate this case further before merging the PR. More info to follow.

The problem has been fixed by adding `sleep(1)` in https://github.com/JakubPietrakIntel/pytorch_geometric/blob/a705d39d5f48b156c53be71bf5bedf74cec3b3de/test/loader/test_neighbor_loader.py#L594

The issue originates from the fact that DataLoader worker needs time to initialize completely in a parallel subprocess (with worker PID). When `subprocess.Popen(taskset...)` takes precedence over worker subprocess it reads out wrong value for affinitization. This PR is ready to be merged.

torch_geometric/loader/base.py
```diff
@@ -34,3 +34,15 @@ def __len__(self) -> int:
 
     def __next__(self) -> Any:
         return self.transform_fn(next(self.iterator))
+
+
+class WorkerInitWrapper:
+    r"""Wraps the :attr:`worker_init_fn` argument of the DataLoader
+    to set the number of OMP threads to 1 for PyTorch DataLoader workers.
```
mszarma, line 41: is the class description is adjusted to newest code?

mszarma (approved):
LGTM.

JakubPietrakIntel:
@yanbing-j @mszarma @rusty1s Thank you all for the review!
This feature is now merged.

PR #6033, opened 2022-11-22
Add PyTorch SparseTensor support for `GCNConv` and `gcn_norm`

EdisonLeeeee:
For `gcn_norm`, a PyTorch SparseTensor is converted to `torch_sparse.SparseTensor` and then converted back. Not really like this but have no better solution.


rusty1s (commented):
Thanks for this PR. I think this gets really tricky to implement with TorchScript support. Sadly, they do not pass in this PR but I also failed to make them work :(

EdisonLeeeee:
Yeah, I have been blocked by this problem and also failed to find a solution these days. It seems we should drop the TorchScript support for PyTorch SparseTensor for now.

rusty1s:
Yeah, we don't have support for this anyway yet. The challenge is to ensure that original TorchScript support is maintained :(

EdisonLeeeee:
Noticed now. The problem seems related to `is_torch_sparse_tensor` - unable to distinguish between the `Tensor`(dense) and `torch.sparse.Tensor` when converting to torchscript :(

EdisonLeeeee:
@rusty1s I've made the tests passed with some changes on `gcn_norm` when dealing with `torch.sparse.Tensor`, it is somehow tricky though.

rusty1s (approved):
This is quite clever. Thank you!

PR #6036, opened 2022-11-22
`FusedAggregation` of simple scatter reductions

rusty1s:
+------------------------------+---------+---------+
    | Aggregators                  | Vanilla | Fusion  |
    +==============================+=========+=========+
    | :obj:`[sum, mean]`           | 0.4019s | 0.1666s |
    +------------------------------+---------+---------+
    | :obj:`[sum, mean, min, max]` | 0.7841s | 0.4223s |
    +------------------------------+---------+---------+
    | :obj:`[sum, mean, var]`      | 0.9711s | 0.3614s |
    +------------------------------+---------+---------+
    | :obj:`[sum, mean, var, std]` | 1.5994s | 0.3722s |
    +------------------------------+---------+---------+

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.2.0] - 2022-MM-DD
 ### Added
+- Added `FusedAggregation` of simple scatter reductions ([#6036](https://github.com/pyg-team/pytorch_geometric/pull/6036))
 - Added `HeteroData` support for `to_captum_model` and added `to_captum_input` ([#5934](https://github.com/pyg-team/pytorch_geometric/pull/5934))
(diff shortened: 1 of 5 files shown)


lightaime (commented):
Amazing. I really like it!

An open discussion: I wonder if it is possible to move these caching logistics into these aggregations (`MeanAggregation`, `VarAggregation`, `StdAggregation`) that may utilize caches. We can first sort the executions of aggregations based on caching dependence and maintain caching results for the `degree`, `sum`, `mean`, and `var` into a buffer that can be read by the later aggregations. That may make the control flow a bit clear. Just a random thought. Essentially they are the same but moving the caching logistics to the Aggregation may be easier to read.

rusty1s:
@lightaime Yeah, I would like to avoid using `torch.nn.Module` as a cache. If you have better ideas how to implement this, I am all ears. I am not super happy with the current implementation, but I failed to come up with a better approach.

PR #6038, opened 2022-11-22
Add `to_smiles` function

hichiaty:
This PR:

- Adds a PyG Graph `to_smiles` function based on the `from_smiles` function.
- Adds further options to `x_map` and `e_map` to match RDkit's map indices
- Fixes kekulize in the `from_smiles` function
- Adds test to check `from_smiles` and `to_smiles` are consistent


rusty1s (approved):
Thank you!

PR #6043, opened 2022-11-23
Minor fix for `mask_feature`

EdisonLeeeee:
A careless error - it should be `rand_like` instead of `randn_like`.


rusty1s (approved):
Good catch :)

PR #6054, opened 2022-11-24
Some usability updates to `Explanation`

BlazStojanovic:
### Motivation
When evaluating and visualizing explanations it will be very convenient to obtain the explanation and explanation complement subgraphs directly from the `Explanation` objects. Given that the class holds information of the original graph being explained this shouldn’t be difficult to achieve. I propose we add two methods to the `Explanation` class:

1. `get_explanation_subgraph` returns the explanation subgraph $G_S$,
2. `get_complement_subgraph` returns the complement of the explanation subgraph $G_{C \backslash S}$

I’ve submitted this draft PR to get opinions on this, both implementation-wise and if we should do this in the first place. 

### Alternatives
1. Delegate this to other parts of code, I.e. whenewer this is needed (e.g. when evaluation explanations or visualizing them)
2. Implement the logic in `util.py` for general Data object, e.g. a `mask_graph` method, which can then be reused here or elsewhere.


BlazStojanovic:
@rusty1s and @dufourc1, while working on this PR, I’ve encountered the following behaviour. If you create a partially empty `Explanation`, e.g.
```
node_mask = torch.tensor([False, True])
explanation = Explanation(node_mask=node_mask)
```
then trying to acces either of the other masks results in the following `AttributeError `:
```
explanation.edge_mask (or node_feat_mask, edge_feat_mask)
...
AttributeError: 'GlobalStorage' object has no attribute 'edge_mask'
```
Is this intended behaviour?

rusty1s:
I guess it is. You can access them via `data.get('edge_feat_mask')` if you want to access potentially non-existing attributes. Would that work for you?

dufourc1:
I would have expected that to return `None`, even if it does not change a lot given the workaround rusty1s just have 👍.

rusty1s (approved):

PR #6056, opened 2022-11-24
`GNNExplainer` Edge Task Level

avgupta456:
This PR extends the recent `GNNExplainer` work (#5967) to edge classification and edge regression (including link prediction). Draft until the main `GNNExplainer` PR is merged. Progress towards #5925.

**Done**
- Adds tests for the `subgraph` function in `base.py`. Uses `subgraph([node_idx_1, node_idx_2])` to get the edge k-hop subgraph.
- Implements the edge classification and edge regression tasks. Currently only supports a single index into the edge_index.
- Adds tests for the edge predictions, similar to the node/graph task tests.

**Todo**
- Support multiple indexes into the edge_index and/or passing a separate edge_label_index to explain edges not in the graph (ex. negative link prediction).
- Create examples to demonstrate edge level explanations. Are there specific datasets/tasks that would be interesting to look at.

I would like some preliminary feedback before attempting the above two tasks. I'm also unsure if these additions should be in this PR or a separate one.


wsad1:
Thanks for the work @avgupta456. It might be good to build on top of #6065, or wait for it to get merged to master as it cleans up some internal implementations and interfaces.

avgupta456:
@camillepradel After syncing with #6065, my PR actually makes no changes to the GNNExplainer class, and only contributes additional test cases. Do you think you could try moving your example onto this PR and seeing if everything works nicely? It also looks like the subgraph filtering has been removed now, which should simplify things. Let me know what you think.

rusty1s:
I agree, functionality-wise there is not much left. Adding tests for this sounds good.

One option question would be how to handle binary classification (e.g., in link prediction tasks) since the current design can only handle multi-class classification. We can also do this in a separate PR.

camillepradel:
@avgupta456 I will add my example from the other PR. But I won't be able to port it as is since this example trains a model on a training dataset and then generates an explanation for the output concerning an edge from an evaluation dataset (which contains new edges); in the link prediction task, we want to perform prediction on new edges.. With the current version of edge classification, the `Explainer` can only explain a output concerning an edge which appear in `edge_index`.

This observation is probably encompassed in @rusty1s 's one.

I will adapt my example to use an existing edge for explanation, but I think we should address this problem.

rusty1s:
@camillepradel I haven‘t had a look at your PR yet, but I don‘t see why we are not able to explain a prediction for an unseen edge. Can you explain? IMO, this should work:
```python
explainer(x=…, edge_index=…, edgel_label_index=…, index=0)

rusty1s (approved):
Thank you! Cleaned up the tests a bit further, hope that is okay :)

camillepradel:
> @camillepradel I haven‘t had a look at your PR yet, but I don‘t see why we are not able to explain a prediction for an unseen edge. Can you explain? IMO, this should work:
> 
> ```python
> explainer(x=…, edge_index=…, edgel_label_index=…, index=0)
> ```

My understanding is that by setting `index` to `0`, we will get the explanation of the first edge in `edge_index`.
 - [l. 66 of gnn_explainer.py](https://github.com/avgupta456/pytorch_geometric/blob/76c88ad5d4a85259d0535728b575d0ba699c42f3/torch_geometric/explain/algorithm/gnn_explainer.py#L166), the model outputs its predictions for all edges of `edge_index`
 - [l. 171 of gnn_explainer.py](https://github.com/avgupta456/pytorch_geometric/blob/76c88ad5d4a85259d0535728b575d0ba699c42f3/torch_geometric/explain/algorithm/gnn_explainer.py#L171), we select the element with index `index` from this output to process the loss.
So the masks are learned to explain the edge with index `index` from `edge_index`.

I don't see how I can get an explanation for an edge between two arbitrary nodes (e.g. the edge from node with index `0` to node with index `1`).

rusty1s:
The `index` just refers to the output index of what the model returns. So if your model returns predictions for unseen edges, e.g., via `edge_label_index`, you can as well index them via `index`. We now do this in the `test_gnn_explainer.py` tests.

camillepradel:
I see. The thing I didn't catch is that you can use `kwargs` to indicate the model for which edges you want the prediction and then use `edge_label_index` and `index` to get the wanted explanation.

I will update my example.

avgupta456:
@rusty1s @wsad1 I realized I forgot to perform the k_hop_subgraph hard mask for the edge task level, as this was changed by #6065 after I started this PR. Should I create a new PR or issue for this?

https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/explain/algorithm/gnn_explainer.py#L116

rusty1s:
You are right. Yes, let‘s fix it in a nee PR.

rusty1s:
Although it might not be trivial to get the subgraph in the first place. In the worst case, the user needs to apply the hard masking on its own.

rusty1s:
Another solution is to track whether masks are actually receiving gradients. Those entries who don‘t/who are never updated are save to set to zero.

avgupta456:
I think it is relatively simple to get the subgraph by passing in a tensor of nodes to k_hop_subgraph(). I will try and get a PR up soon.

rusty1s:
Yes, that‘s what I was referring, too. I don‘t think we necessarily want to mask based on `edge_index` since the resulting subgraph will nit remove any nodes or edges this way. Let me take a look as well.

PR #6065, opened 2022-11-25
`GNNExplainer` improvements

rusty1s:
* Drops `subgraph` computation. `GNNExplainer` is now always applied on the full set of edges. This fixes several bugs, e.g., that `index` now references invalid entries. Instead, we compute the hard mask to get all edges involved in message passing.
* `GNNExplainer` now correctly works with a tensor of `index`
* `target_index` clean-up.
* Moves `model.eval()` to the base class and resets it state afterwards.


torch_geometric/explain/explainer.py
```diff
@@ -128,6 +133,8 @@ def __call__(
             **kwargs,
         )
 
+        self.model.train(training)
```
wsad1, line 136: Add a test to check if model state is reset. This will prevent someone from modifying this by mistake in the future.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -224,29 +165,29 @@ def _train_node_edge_mask(
...
+            y = target
+
+            if target_index is not None:
+                y_hat, y = y_hat[target_index], y[target_index]
+            if index is not None:
+                y_hat, y = y_hat[index], y[index]
+
+            loss = self._loss(y_hat, y, explainer_config, model_config)
+
+            loss.backward(retain_graph=False)
```
wsad1, line 190:
```suggestion
            loss.backward()
```

wsad1 (commented):
We might get incorrect results if the nodes in `index` have overlapping subgraphs. Because some edges/nodes might get more importance as they are part of multiple sugbraphs.

torch_geometric/explain/algorithm/base.py
```diff
@@ -104,47 +95,68 @@ def get_initial_prediction(
...
-            elif torch.is_tensor(value) and value.size(0) == num_edges:
-                kwargs[key] = value[edge_mask]
+        node_mask = edge_index.new_zeros(num_nodes, dtype=torch.bool)
+        node_mask[mapping] = True
 
-        return x, edge_index, mapping, node_idx, edge_mask, kwargs
+        return node_mask, edge_mask
 
-    # Helper functions ########################################################
+    def _num_hops(self, model: torch.nn.Module) -> int:
```
wsad1, line 151: Make this a utility method.

torch_geometric/explain/algorithm/base.py
```diff
@@ -104,47 +95,68 @@ def get_initial_prediction(
...
-            num_hops=get_num_hops(model),
+        if hard_mask is not None:
+            mask[~hard_mask] = 0.
+
+        return mask
+
+    def _get_hard_masks(
+        self,
+        model: torch.nn.Module,
+        index: Optional[Union[int, Tensor]],
```
wsad1, line 128: Add a test for `index` of type `Tensor`.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -76,6 +77,34 @@ def __init__(self, epochs: int = 100, lr: float = 0.01, **kwargs):
 
         self.node_mask = self.edge_mask = None
 
+    def supports(
+        self,
+        explainer_config: ExplainerConfig,
+        model_config: ModelConfig,
```
wsad1, line 83: We can address this in another PR. 
What if `explainer_config` and `model_config` were inputs to `ExplainerAlgorithm.__init__()`, and we would call `self.support` while initializing the class, feels like thats more cleaner.

RBendias (approved):
LGTM! :) As Jinu suggested, I also think it's cleaner if the configs (explainer and model) are e.g. class variables and not passed through multiple methods, but we can do that in a future PR.

examples/gnn_explainer_ba_shapes.py
```diff
@@ -83,12 +82,11 @@ def test():
         explanation = explainer(data.x, data.edge_index, index=node_index,
                                 target=data.y, edge_weight=data.edge_weight)
```
RBendias, line 83: The description in `Explainer` for the target is:
>   If the explanation type is :obj:`"phenomenon"`, the target has to be provided. If the explanation type is 
     :obj:`"model"`, the target should be set to :obj:`None` and will get automatically inferred.

However, we set the target in both cases here.

examples/gnn_explainer_ba_shapes.py
```diff
@@ -83,12 +82,11 @@ def test():
...
 
-        subgraph = k_hop_subgraph(node_index, num_hops=3,
-                                  edge_index=data.edge_index)
-        expl_edge_mask = explanation.edge_mask[non_loop_mask]
-        subgraph_edge_mask = subgraph[3][non_loop_mask]
-        targets.append(data.edge_label[subgraph_edge_mask].cpu())
-        preds.append(expl_edge_mask[subgraph_edge_mask].cpu())
+        _, _, _, hard_edge_mask = k_hop_subgraph(node_index, num_hops=3,
```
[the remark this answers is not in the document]
rusty1s: 4-hop nodes do not have any importance as the layer is only 3 layer deep.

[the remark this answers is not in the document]
rusty1s: yes, but the 4 hop nodes do not have any importance score as these edge masks are never trained.

examples/gnn_explainer_ba_shapes.py
rusty1s: Good point. Let me make a TODO to clarify this.

PR #6070, opened 2022-11-25
Fix training benchmark `device` issue

rusty1s:


LukeLIN-web (commented):
Why remove the `inputs_channels`?  It will cause the    "pytorch_geometric/benchmark/utils/utils.py", line 92, in get_model
    return Model(params['inputs_channels'], params['hidden_channels'],
KeyError: 'inputs_channels'

rusty1s:
Yes, I was dump. It is fixed now.

PR #6083, opened 2022-11-28
Link prediction explanation

camillepradel:
Closes https://github.com/pyg-team/pytorch_geometric/issues/5925

_This is still a draft, but I would be happy to get early feedback from reviewers._ 

**Done:**
 - evolution of `Explainer` to adapt link prediction model output to match format of node prediction output
 - minor changes in `GNNExplainer` to support link `"edge"` as `task_level`
 - example script `gnn_explainer_link_pred.py` training a link prediction model and getting the explanation for one output

The implemented solution has the **advantage** of requiring very minor changes to the explainer for it to support link prediction tasks.
The main **drawback** is that it kind of "tricks" the explainer into believing it is explaining a node classification task (the model output and index are transformed before being sent to the explainer).

**TODO:**
 - support `batch` explanations (I will probably wait for it to be supported for other tasks: https://github.com/pyg-team/pytorch_geometric/pull/5967#pullrequestreview-1191390904)
 - restrict the computation to the two nodes of interest and their computation graph 
   https://github.com/pyg-team/pytorch_geometric/blob/b3d4b83a615bdd10618f10ba90951a8917e1c241/torch_geometric/explain/algorithm/gnn_explainer.py#L131
 - add tests

### DIFF
--- a/examples/gnn_explainer_link_pred.py
+++ b/examples/gnn_explainer_link_pred.py
@@ -1,0 +1,111 @@
+import os.path as osp
+
+import torch
+from sklearn.metrics import roc_auc_score
+
+import torch_geometric.transforms as T
+from torch_geometric.datasets import Planetoid
+from torch_geometric.explain import Explainer, ExplainerConfig, ModelConfig
+from torch_geometric.explain.algorithm import GNNExplainer
+from torch_geometric.nn import GCNConv
+from torch_geometric.utils import negative_sampling
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+transform = T.Compose([
+    T.NormalizeFeatures(),
+    T.ToDevice(device),
+    T.RandomLinkSplit(num_val=0.05, num_test=0.1, is_undirected=True,
+                      add_negative_train_samples=False),
+])
+dataset = Planetoid(path, dataset, transform=transform)
+# After applying the `RandomLinkSplit` transform, the data is transformed from
+# a data object to a list of tuples (train_data, val_data, test_data), with
+# each element representing the corresponding split.
+train_data, val_data, test_data = dataset[0]
+
+
+class Net(torch.nn.Module):
+    def __init__(self, in_channels, hidden_channels, out_channels):
+        super().__init__()
+        self.conv1 = GCNConv(in_channels, hidden_channels)
+        self.conv2 = GCNConv(hidden_channels, out_channels)
+
+    def encode(self, x, edge_index):
+        x = self.conv1(x, edge_index).relu()
+        return self.conv2(x, edge_index)
+
+    def decode(self, z, edge_label_index):
+        return (z[edge_label_index[0]] * z[edge_label_index[1]]).sum(dim=-1)
+
+    def decode_all(self, z):
+        prob_adj = z @ z.t()
+        return (prob_adj > 0).nonzero(as_tuple=False).t()
+
+    def forward(self, x, edge_index, edge_label_index):
+        z = model.encode(x, edge_index)
+        return model.decode(z, edge_label_index).view(-1)
+
+
+model = Net(dataset.num_features, 128, 64).to(device)
+optimizer = torch.optim.Adam(params=model.parameters(), lr=0.01)
+criterion = torch.nn.BCEWithLogitsLoss()
+
+
+def train():
+    model.train()
+    optimizer.zero_grad()
+
+    # We perform a new round of negative sampling for every training epoch:
+    neg_edge_index = negative_sampling(
+        edge_index=train_data.edge_index, num_nodes=train_data.num_nodes,
+        num_neg_samples=train_data.edge_label_index.size(1), method='sparse')
+
+    edge_label_index = torch.cat(
+        [train_data.edge_label_index, neg_edge_index],
+        dim=-1,
+    )
+    edge_label = torch.cat([
+        train_data.edge_label,
+        train_data.edge_label.new_zeros(neg_edge_index.size(1))
+    ], dim=0)
+
+    out = model(train_data.x, train_data.edge_index, edge_label_index)
+    loss = criterion(out, edge_label)
+    loss.backward()
+    optimizer.step()
+    return loss
+
+
+@torch.no_grad()
+def test(data):
+    model.eval()
+    out = model(data.x, data.edge_index, data.edge_label_index).sigmoid()
+    return roc_auc_score(data.edge_label.cpu().numpy(), out.cpu().numpy())
+
+
+best_val_auc = final_test_auc = 0
+for epoch in range(1, 11):
+    loss = train()
+    val_auc = test(val_data)
+    test_auc = test(test_data)
+    if val_auc > best_val_auc:
+        best_val_auc = val_auc
+        final_test_auc = test_auc
+    print(f'Epoch: {epoch:03d}, Loss: {loss:.4f}, Val: {val_auc:.4f}, '
+          f'Test: {test_auc:.4f}')
+
+print(f'Final Test: {final_test_auc:.4f}')
+
+explainer = Explainer(
+    model=model, algorithm=GNNExplainer(epochs=200),
+    explainer_config=ExplainerConfig(explanation_type="model",
+                                     node_mask_type="attributes",
+                                     edge_mask_type="object"),
+    model_config=ModelConfig(mode="classification", task_level="edge",
+                             return_type="raw"))
+edge_idx = val_data.edge_label_index[:, 0]
+explanation = explainer(x=train_data.x, edge_index=train_data.edge_index,
+                        index=edge_idx, target_index=None)
+print(explanation.available_explanations)
--- a/torch_geometric/explain/algorithm/gnn_explainer.py
+++ b/torch_geometric/explain/algorithm/gnn_explainer.py
@@ -31,7 +31,7 @@ class GNNExplainer(ExplainerAlgorithm):
 
     - :class:`torch_geometric.explain.config.ModelConfig`
 
-        - :attr:`task_level`: :obj:`"node"` or :obj:`"graph"`
+        - :attr:`task_level`: :obj:`"node"`, :obj:`"edge"` or :obj:`"graph"`
 
     - :class:`torch_geometric.explain.config.ExplainerConfig`
 
@@ -124,7 +124,13 @@ class GNNExplainer(ExplainerAlgorithm):
                 target = torch.index_select(target, 1, subset)
             else:
                 target = target[subset]
-
+        elif model_config.task_level == ModelTaskLevel.edge:
+            # if we are dealing with a node level task, we can restrict the
+            # computation to the two nodes of interest and their computation
+            # graph
+            # TODO
+            hard_edge_mask = None
+            subset = None
         elif model_config.task_level == ModelTaskLevel.graph:
             subset = None
             hard_edge_mask = None
@@ -164,7 +170,7 @@ class GNNExplainer(ExplainerAlgorithm):
         if self.node_mask is not None:
             node_mask = self.node_mask.detach().sigmoid().squeeze(-1)
             if explainer_config.node_mask_type == MaskType.object:
-                if task_level == ModelTaskLevel.node:
+                if task_level in [ModelTaskLevel.node, ModelTaskLevel.edge]:
                     node_mask_ = node_mask.new_zeros(num_nodes)
                     node_mask_[subset] = node_mask
                 elif task_level == ModelTaskLevel.graph:
@@ -172,7 +178,7 @@ class GNNExplainer(ExplainerAlgorithm):
                 else:
                     raise NotImplementedError
             elif explainer_config.node_mask_type == MaskType.attributes:
-                if task_level == ModelTaskLevel.node:
+                if task_level in [ModelTaskLevel.node, ModelTaskLevel.edge]:
                     node_feat_mask_ = node_mask.new_zeros(
                         num_nodes, node_mask.size(-1))
                     node_feat_mask_[subset] = node_mask
@@ -181,7 +187,7 @@ class GNNExplainer(ExplainerAlgorithm):
                 else:
                     raise NotImplementedError
             elif explainer_config.node_mask_type == MaskType.common_attributes:
-                if task_level == ModelTaskLevel.node:
+                if task_level in [ModelTaskLevel.node, ModelTaskLevel.edge]:
                     node_feat_mask_ = node_mask.new_zeros(
                         num_nodes, node_mask.numel())
                     node_feat_mask_[subset] = node_mask
@@ -195,7 +201,7 @@ class GNNExplainer(ExplainerAlgorithm):
 
         if self.edge_mask is not None:
             edge_mask = self.edge_mask.detach().sigmoid()
-            if task_level == ModelTaskLevel.node:
+            if task_level in [ModelTaskLevel.node, ModelTaskLevel.edge]:
                 edge_mask_ = edge_mask.new_zeros(num_edges)
                 edge_mask_[hard_edge_mask] = edge_mask
             elif task_level == ModelTaskLevel.graph:
@@ -353,7 +359,7 @@ class GNNExplainer(ExplainerAlgorithm):
         model_config: ModelConfig,
     ) -> bool:
         if model_config.task_level not in [
-                ModelTaskLevel.node, ModelTaskLevel.graph
+                ModelTaskLevel.node, ModelTaskLevel.edge, ModelTaskLevel.graph
         ]:
             logging.error("Model task level not supported.")
             return False
--- a/torch_geometric/explain/explainer.py
+++ b/torch_geometric/explain/explainer.py
@@ -1,4 +1,5 @@
 import copy
+import functools
 from typing import Optional, Union
 
 import torch
@@ -10,9 +11,42 @@ from torch_geometric.explain.config import (
     ExplanationType,
     ModelConfig,
     ModelMode,
+    ModelReturnType,
+    ModelTaskLevel,
     ThresholdConfig,
     ThresholdType,
 )
+
+
+def decorate_link_prediction_model(model, index, model_config: ModelConfig):
+    if model_config.return_type == ModelReturnType.raw:
+
+        def get_complement_and_out(out):
+            out = out.sigmoid()
+            complement = 1 - out
+            return torch.stack((complement, out), dim=-1)
+
+        model_config.return_type = ModelReturnType.probs
+    elif model_config.return_type == ModelReturnType.log_probs:
+
+        def get_complement_and_out(out):
+            complement = torch.log(1 + torch.exp((torch.log(-out))))
+            return torch.stack((complement, out), dim=-1)
+    elif model_config.return_type == ModelReturnType.probs:
+
+        def get_complement_and_out(out):
+            complement = 1 - out
+            return torch.stack((complement, out), dim=-1)
+    else:
+        raise NotImplementedError
+
+    @functools.wraps(model)
+    def wrapper(*args, **kwargs):
+        out = model(*args, edge_label_index=index, **kwargs)
+        out = get_complement_and_out(out)
+        return out
+
+    return wrapper
 
 
 class Explainer:
@@ -106,6 +140,17 @@ class Explainer:
                 (default: :obj:`None`)
             **kwargs: additional arguments to pass to the GNN.
         """
+        if self.model_config.task_level == ModelTaskLevel.edge:
+            # decorate the model forward function to make its output match the
+            # node classification format
+            self.model.forward = decorate_link_prediction_model(
+                self.model.forward, index, self.model_config)
+            assert isinstance(index, Tensor) and len(index.shape) == 1 or (
+                len(index.shape) == 2 and index.shape[0] == 1
+            ), "Explainer only supports single edge index expressed as a "
+            "tensor for now"
+            index = 0
+
         # Choose the `target` depending on the explanation type:
         if (self.explainer_config.explanation_type ==
                 ExplanationType.phenomenon):
@@ -127,6 +172,10 @@ class Explainer:
             target_index=target_index,
             **kwargs,
         )
+
+        if self.model_config.task_level == ModelTaskLevel.edge:
+            # restore original forward function in the model
+            self.model.forward = self.model.forward.__wrapped__
 
         return self._post_process(explanation)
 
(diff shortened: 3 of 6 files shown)


avgupta456:
I wonder if we could combine this PR and #6056. This PR contributes an example script while #6056 handles the k_hop_subgraph and adds edge-level tests for GNNExplainer.

camillepradel:
> I wonder if we could combine this PR and #6056. This PR contributes an example script while #6056 handles the k_hop_subgraph and adds edge-level tests for GNNExplainer.

oops, I didn't see your PR! I will have a look, thanks!

rusty1s:
@camillepradel Any reason to close this? We would still like to integrate an example of this :)

camillepradel:
you are right. I wanted to open a new PR but there is actually no good reason to not keep using this one.

camillepradel:
So I updated the example, but since it was a link prediction task, I needed the explainability framework to support binary classification. I ended up adding a new `binary_classification` mode (and renaming the original `classification` to `multiclass_classification`).
If somebody can think of a simpler way to support binary classification, I am curious to know about it.

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
+import os.path as osp
+
+import torch
+from sklearn.metrics import roc_auc_score
+
+import torch_geometric.transforms as T
+from torch_geometric.datasets import Planetoid
+from torch_geometric.explain import Explainer, ExplainerConfig, ModelConfig
+from torch_geometric.explain.algorithm import GNNExplainer
```
RBendias, line 9:
```suggestion
from torch_geometric.explain import Explainer, ExplainerConfig, ModelConfig, GNNExplainer
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+    def encode(self, x, edge_index):
+        x = self.conv1(x, edge_index).relu()
+        return self.conv2(x, edge_index)
+
+    def decode(self, z, edge_label_index):
+        src, dst = edge_label_index
+        return (z[src] * z[dst]).sum(dim=-1)
+
+    def forward(self, x, edge_index, edge_label_index,
+                cast_to_multiclass=False):
```
RBendias, line 44: Is `cast_to_multiclass` used somewhere?

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+        out = model.decode(z, edge_label_index).view(-1)
+        if cast_to_multiclass:
+            out = out.sigmoid()
+            out = torch.stack((1 - out, out), dim=-1)
+        return out
+
+
+model = Net(dataset.num_features, 128, 64).to(device)
+optimizer = torch.optim.Adam(params=model.parameters(), lr=0.01)
+criterion = torch.nn.BCEWithLogitsLoss()
```
RBendias, line 55: In previous examples, we use more `torch.nn.functional`. Therefore I would suggest using:

```
loss = F.binary_cross_entropy_with_logits(out, edge_label)
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+
+best_val_auc = final_test_auc = 0
+for epoch in range(1, 11):
+    loss = train()
+    val_auc = test(val_data)
+    test_auc = test(test_data)
+    if val_auc > best_val_auc:
+        best_val_auc = val_auc
+        final_test_auc = test_auc
+    print(f'Epoch: {epoch:03d}, Loss: {loss:.4f}, Val: {val_auc:.4f}, '
```
RBendias, line 98:
```suggestion
    print(f'Epoch: {epoch:02d}, Loss: {loss:.4f}, Val: {val_auc:.4f}, '
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+transform = T.Compose([
+    T.NormalizeFeatures(),
+    T.ToDevice(device),
+    T.RandomLinkSplit(num_val=0.05, num_test=0.1, is_undirected=True,
+                      add_negative_train_samples=False),
+])
+dataset = Planetoid(path, dataset, transform=transform)
+# After applying the `RandomLinkSplit` transform, the data is transformed from
+# a data object to a list of tuples (train_data, val_data, test_data), with
+# each element representing the corresponding split.
```
RBendias, line 25:
```suggestion
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+    test_auc = test(test_data)
+    if val_auc > best_val_auc:
+        best_val_auc = val_auc
+        final_test_auc = test_auc
+    print(f'Epoch: {epoch:03d}, Loss: {loss:.4f}, Val: {val_auc:.4f}, '
+          f'Test: {test_auc:.4f}')
+
+print(f'Final Test: {final_test_auc:.4f}')
+
+# explain the model output on an edge from the validation dataset
```
RBendias, line 103:
```suggestion
# Explain model output for an edge
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,130 @@
...
+                                     node_mask_type="attributes",
+                                     edge_mask_type="object"),
+    model_config=model_config)
+explanation = explainer(x=train_data.x, edge_index=train_data.edge_index,
+                        index=0, target_index=None,
+                        edge_label_index=edge_label_index)
+print(explanation.available_explanations)
+
+# explain a phenomenon that the model should predict on an edge from the
+# validation dataset
```
RBendias, line 119:
```suggestion
# Explain a selected target (phenomenon) for an edge
```

RBendias (commented):
Hi, thanks a lot for the PR! I left some comments in the example file. Do you think it is really necessary to introduce a new classification type, or is it possible to internally choose the specific loss based on the model output?

examples/gnn_explainer_link_pred.py
camillepradel: non, I forgot to remove it after implementing `binary_classification`. It is done now.

examples/gnn_explainer_link_pred.py
camillepradel: done

RBendias:
I checked the code again. The main reason I see for the additional type is to check if the `return_type` is set correctly in the `ModelConfig`. The `return_type` needs to be `probs`, as we use `binary_cross_entropy`. However, we could also measure the MSE between the raw/probs/log_probs values (also using the raw model output in `get_prediction` instead of setting a threshold of 0.5). @camillepradel What do you think?

camillepradel:
> I checked the code again. The main reason I see for the additional type is to check if the `return_type` is set correctly in the `ModelConfig`. The `return_type` needs to be `probs`, as we use `binary_cross_entropy`. However, we could also measure the MSE between the raw/probs/log_probs values (also using the raw model output in `get_prediction` instead of setting a threshold of 0.5). @camillepradel What do you think?

The distinction between `multiclass_classification` and `binary_classification` modes is indeed used in `ModelConfig`, but also in processing prediction in `Explainer` and in processing loss in `GNNExplainer`. In the two later cases, it allows to handle differently the output of the model according to the mode. I initially though we could also make the difference by looking at the shape of the output, but it looked tricky to me (depending on the number of classes and the optional batching, we might not be able to know), which is why I went to define explicitly two distinct modes.

You are right, we don't have to enforce `return_type` to be `probs` for bynary classification (I have never seen a `log_probs` output in that setup but it seems legit), I will change that.

camillepradel:
I updated the code to allow `raw` and `log_probs` modes, but I am still a bit confused about `log_probs`.

More specifically, I don't know how to apply MSE loss directly on `log_probs` since the range of values for log probabilities is (−∞,0]. In current version, I applied `binary_cross_entropy` to the `exp()` of `y_hat`.

torch_geometric/explain/explainer.py
```diff
@@ -64,8 +66,15 @@ def get_prediction(self, *args, **kwargs) -> torch.Tensor:
...
-        if self.model_config.mode == ModelMode.classification:
+        if self.model_config.mode == ModelMode.multiclass_classification:
             out = out.argmax(dim=-1)
+        elif self.model_config.mode == ModelMode.binary_classification:
+            if self.model_config.return_type.value == 'raw':
+                out = (out > 0).long()
+            if self.model_config.return_type.value == 'probs':
+                out = (out > 0.5).long()
+            elif self.model_config.return_type.value == 'log_probs':
+                out = (out > math.log(0.5)).long()
```
rusty1s, line 77: I don't think the combination of `binary_cross_entropy` and `log_probs` is valid - can we disable this combination during `ModelConfig` initialization?

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -201,19 +201,30 @@ def _initialize_masks(self, x: Tensor, edge_index: Tensor):
...
 
         return (-y_hat).gather(1, y.view(-1, 1)).mean()
 
+    def _loss_binary_classification(self, y_hat: Tensor, y: Tensor) -> Tensor:
+        if self.model_config.return_type.value == 'raw':
+            return F.binary_cross_entropy_with_logits(y_hat, y.float())
+        if self.model_config.return_type.value == 'probs':
+            return F.binary_cross_entropy(y_hat, y.float())
+        elif self.model_config.return_type.value == 'log_probs':
+            return F.binary_cross_entropy(y_hat.exp(), y.float())
```
rusty1s, line 219:
```suggestion
```

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -201,19 +201,30 @@ def _initialize_masks(self, x: Tensor, edge_index: Tensor):
...
 
         y_hat = self._to_log_prob(y_hat)
 
         return (-y_hat).gather(1, y.view(-1, 1)).mean()
 
+    def _loss_binary_classification(self, y_hat: Tensor, y: Tensor) -> Tensor:
+        if self.model_config.return_type.value == 'raw':
+            return F.binary_cross_entropy_with_logits(y_hat, y.float())
+        if self.model_config.return_type.value == 'probs':
+            return F.binary_cross_entropy(y_hat, y.float())
```
rusty1s, line 217:
```suggestion
            return F.binary_cross_entropy_with_logits(y_hat.view_as(y), y.float())
        if self.model_config.return_type.value == 'probs':
            return F.binary_cross_entropy(y_hat.view_as(y), y.float())
```

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -201,19 +201,30 @@ def _initialize_masks(self, x: Tensor, edge_index: Tensor):
...
         if y.dim() == 0:  # `index` was given as an integer.
             y_hat, y = y_hat.unsqueeze(0), y.unsqueeze(0)
 
         y_hat = self._to_log_prob(y_hat)
 
         return (-y_hat).gather(1, y.view(-1, 1)).mean()
 
+    def _loss_binary_classification(self, y_hat: Tensor, y: Tensor) -> Tensor:
+        if self.model_config.return_type.value == 'raw':
+            return F.binary_cross_entropy_with_logits(y_hat, y.float())
```
rusty1s, line 215: I like the reuse here of `F.binary_cross_entropy_with_logits`. We should probably do the same for `_loss_multiclass_classification` as well and use `F.cross_entropy_with_logits`.

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
+import os.path as osp
```
rusty1s, line 1: Can we add a link to this example [here](https://github.com/pyg-team/pytorch_geometric/blob/master/README.md?plain=1#L300)?

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.conv2(x, edge_index)
+
+    def decode(self, z, edge_label_index):
+        src, dst = edge_label_index
+        return (z[src] * z[dst]).sum(dim=-1)
+
+    def forward(self, x, edge_index, edge_label_index):
+        z = model.encode(x, edge_index)
+        out = model.decode(z, edge_label_index).view(-1)
+        return out
```
rusty1s, line 48:
```suggestion
        return model.decode(z, edge_label_index).view(-1)
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+from torch_geometric.utils import negative_sampling
+
+dataset = 'Cora'
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data', 'Planetoid')
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+transform = T.Compose([
+    T.NormalizeFeatures(),
+    T.ToDevice(device),
+    T.RandomLinkSplit(num_val=0.05, num_test=0.1, is_undirected=True,
+                      add_negative_train_samples=False),
```
rusty1s, line 25:
```suggestion
    T.RandomLinkSplit(num_val=0.05, num_test=0.1, is_undirected=True),
```
Can we make this example as minimal as possible? I don't think we desperately need the logic of on-the-fly negative sampling in this example.

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+        num_neg_samples=train_data.edge_label_index.size(1), method='sparse')
+
+    edge_label_index = torch.cat(
+        [train_data.edge_label_index, neg_edge_index],
+        dim=-1,
+    )
+    edge_label = torch.cat([
+        train_data.edge_label,
+        train_data.edge_label.new_zeros(neg_edge_index.size(1))
+    ], dim=0)
```
rusty1s, line 71:
```suggestion
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+
+    out = model(train_data.x, train_data.edge_index, edge_label_index)
+    loss = F.binary_cross_entropy_with_logits(out, edge_label)
```
rusty1s, line 74:
```suggestion
    out = model(train_data.x, train_data.edge_index, train_data.edge_label_index)
    loss = F.binary_cross_entropy_with_logits(out, train_data.edge_label)
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+    loss.backward()
+    optimizer.step()
+    return loss
```
rusty1s, line 77:
```suggestion
    return float(loss)
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+best_val_auc = final_test_auc = 0
+for epoch in range(1, 11):
+    loss = train()
+    val_auc = test(val_data)
+    test_auc = test(test_data)
+    if val_auc > best_val_auc:
+        best_val_auc = val_auc
+        final_test_auc = test_auc
```
rusty1s, line 94:
```suggestion
```
Would like to drop this.

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+    print(f'Epoch: {epoch:02d}, Loss: {loss:.4f}, Val: {val_auc:.4f}, '
+          f'Test: {test_auc:.4f}')
+
+print(f'Final Test: {final_test_auc:.4f}')
+
+# Explain model output for an edge
+model_config = ModelConfig(mode="binary_classification", task_level="edge",
+                           return_type="probs")
```
rusty1s, line 102:
```suggestion
                           return_type="raw")
```
We don't apply a `sigmoid` as part of the model, so it needs to be `"raw"`.

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+edge_label_index = val_data.edge_label_index[:, 0]
+explainer = Explainer(
+    model=model, algorithm=GNNExplainer(epochs=200),
+    explainer_config=ExplainerConfig(explanation_type="model",
+                                     node_mask_type="attributes",
+                                     edge_mask_type="object"),
```
rusty1s, line 108:
```suggestion
                                     edge_mask_type="object", ),
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+    model_config=model_config)
+explanation = explainer(x=train_data.x, edge_index=train_data.edge_index,
+                        index=0, target_index=None,
```
rusty1s, line 111: Since you are slicing the `edge_label_index` to only predict a single edge, there is no need to pass`index=0` here.

rusty1s, line 111:
```suggestion
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+# Explain a selected target (phenomenon) for an edge
+explainer = Explainer(
+    model=model, algorithm=GNNExplainer(epochs=200),
+    explainer_config=ExplainerConfig(explanation_type="phenomenon",
+                                     node_mask_type="attributes",
+                                     edge_mask_type="object"),
+    model_config=model_config)
+target = val_data.edge_label[0].unsqueeze(dim=0).long()
+explanation = explainer(x=train_data.x, edge_index=train_data.edge_index,
+                        target=target, index=0, target_index=None,
```
rusty1s, line 124:
```suggestion
                        target=target,
```

examples/gnn_explainer_link_pred.py
```diff
@@ -0,0 +1,126 @@
...
+                        index=0, target_index=None,
+                        edge_label_index=edge_label_index)
+print(explanation.available_explanations)
+
+# Explain a selected target (phenomenon) for an edge
+explainer = Explainer(
+    model=model, algorithm=GNNExplainer(epochs=200),
+    explainer_config=ExplainerConfig(explanation_type="phenomenon",
+                                     node_mask_type="attributes",
+                                     edge_mask_type="object"),
```
rusty1s, line 120:
```suggestion
                                     edge_mask_type="object", ),
```

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -16,7 +16,8 @@ def __init__(self, model_config: ModelConfig, multi_output: bool = False):
         self.model_config = model_config
         self.multi_output = multi_output
 
-        out_channels = 7 if model_config.mode.value == 'classification' else 1
+        out_channels = \
+            7 if model_config.mode.value == 'multiclass_classification' else 1
```
rusty1s, line 20:
```suggestion
        if model_config.mode.value == 'multiclass_classification':
            out_channels = 7
        else:
            out_channels = 1
```

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -31,10 +32,17 @@ def forward(self, x, edge_index, batch=None, edge_label_index=None):
             assert edge_label_index is not None
             x = x[edge_label_index[0]] * x[edge_label_index[1]]
 
-        if self.model_config.return_type.value == 'probs':
-            x = x.softmax(dim=-1)
-        elif self.model_config.return_type.value == 'log_probs':
-            x = x.log_softmax(dim=-1)
+        if self.model_config.mode.value == 'binary_classification':
+            x = x[:, 0]
```
rusty1s, line 36:
```suggestion
```
Let's drop this logic here to test unequal shapes of `y` and `y_hat` in the `GNNExplainer` (see comment above about `view_as`).

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -31,10 +32,17 @@ def forward(self, x, edge_index, batch=None, edge_label_index=None):
...
+            if self.model_config.return_type.value == 'probs':
+                x = x.sigmoid()
+            elif self.model_config.return_type.value == 'log_probs':
+                x = x.sigmoid().log()
```
rusty1s, line 40:
```suggestion
```

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -139,6 +147,62 @@ def test_gnn_explainer_classification(
     check_explanation(edge_mask_type, node_mask_type, explanation)
 
 
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize('explanation_type', ['model', 'phenomenon'])
+@pytest.mark.parametrize('task_level', ['node', 'edge', 'graph'])
+@pytest.mark.parametrize('return_type', ['log_probs', 'probs', 'raw'])
```
rusty1s, line 154:
```suggestion
@pytest.mark.parametrize('return_type', ['log_probs', 'probs'])
```

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -139,6 +147,62 @@ def test_gnn_explainer_classification(
...
+        edge_mask_type=edge_mask_type,
+    )
+
+    model = GCN(model_config, multi_output)
+
+    target = None
+    if explanation_type == 'phenomenon':
+        with torch.no_grad():
+            out = model(x, edge_index, batch, edge_label_index)
+            target = (out > 0.5).long()
```
rusty1s, line 184: This needs to be dependent on the `return_value`.

rusty1s (commented):
LGTM, a few nit-picky comments. I think we need to drop `log_probs` support in `binary_classification`.

torch_geometric/explain/algorithm/gnn_explainer.py
camillepradel: do you mean something like
```python
    def _loss_multiclass_classification(self, y_hat: Tensor,
                                        y: Tensor) -> Tensor:
        if y.dim() == 0:  # `index` was given as an integer.
            y_hat, y = y_hat.unsqueeze(0), y.unsqueeze(0)

        if self.model_config.return_type.value == 'raw':
            return F.cross_entropy_with_logits(y_hat.view_as(y), y.float())
        if self.model_config.return_type.value == 'probs':
            return F.cross_entropy(y_hat.view_as(y), y.float())
        if self.model_config.return_type.value == 'log_probs':
            return (-y_hat).gather(1, y.view(-1, 1)).mean()
```
if yes, the `raw` and `probs` cases are clearer but we don't get rid of the hard to decode `log_probs` case.
Maybe you have something simpler in mind?

test/explain/algorithm/test_gnn_explainer.py
camillepradel: I guess you meant `['probs', 'raw']`.

test/explain/algorithm/test_gnn_explainer.py
camillepradel: indeed!

rusty1s (approved):
Thanks for the updates. Looks great!

Issue #5961, opened 2022-11-12
[Explainability Evaluation] - GNN model (Un)Faithfulness


BlazStojanovic:
### 🚀 The feature, motivation and pitch

Implement **faithfulness** [1],  an explainability metric which quantifies how faithful the explanations are to an underlying GNN predictor. This feature request is a part of broader explainability evaluation metrics (parent issue https://github.com/pyg-team/pytorch_geometric/issues/5628) which will be crucial for the new revamp of explainability (see roadmap https://github.com/pyg-team/pytorch_geometric/issues/5520).

### Overview of the metric
**Un-Faithfulness** of a GNN predictor is given as 
$$\operatorname{GEF}\left(f, \mathcal{S}_u, \mathcal{S}_u^{\prime}\right)=1-\exp ^{-\mathrm{KL}\left(f\left(\mathcal{S}_u\right) \| f\left(\mathcal{S}_u^{\prime}\right)\right)}, $$
where:
* $\hat y_u$ is the prediction probability vector obtained by using the GNN $\hat{y}_u=f\left(\mathcal{S}_u\right)$
* $\hat y_{u^\prime}$ is also obtained with the GNN as $\hat{y}_{u^{\prime}}=f \left(\mathcal S_u ^ \prime \right)$, where $\mathcal S_u^\prime$ is a masked subgraph generated by only keeping the original values of the top-k features identified by an explanation, and
* $KL(\cdot || \cdot)$ is the [Kullback–Leibler divergence](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence)

### Implementation checklist
* Implement a method that computes the unfaithfulness of a GNN predictor

### References
[1] - [Evaluating Explainability for Graph Neural Networks](https://arxiv.org/abs/2208.09339)

PR #6090, opened 2022-11-29
[Explainability Evaluation] - GNN model (Un)Faithfulness

ZeynepP:
Here is the implementation of metric faithfulness with its examples for node and graph classification. I made some implementation decisions, we can discuss. I did not implement edge_mask, again to discuss. 
#5961


torch_geometric/explain/metrics/faithfulness.py
```diff
@@ -0,0 +1,169 @@
...
+        top_k_nodes = explanation.node_mask.topk(topk)[1]
+
+        pert_x = data.x.clone()
+        # removing not top-k
+        rem_nodes = [
+            node for node in range(data.x.shape[0]) if node not in top_k_nodes
+        ]
+        pert_x[rem_nodes] = torch.zeros_like(pert_x[rem_nodes]).to(device)
+        # getting predictions of explainer
+        pert_vec = model(pert_x, data.edge_index, **kwargs)[node_idx]
```
wsad1, line 82: Can we reuse `Expalnation.get_explanation_subgraph`. Also here you only seem to be using `node_mask` but we should be using the edge_mask too right?

torch_geometric/explain/metrics/faithfulness.py
```diff
@@ -0,0 +1,169 @@
...
+    This function should be called for node level tasks to calculate the node's faitfulness score
+
+    Args:
+        node_idx (int): node_idx for node to explain, should be the same used for explainer (it can be get from explainer)
+        data (Data) : dataset
+        explanation (Explanation): Explanation output generated by an explainer
+        model (torch.nn.Module): model used in explanation (ait can be get from explanation)
+        topk (int): k most important feature/nodes, it does not calculate percentage and
+                    the user should take into account the threshold config used in explanation
+        device (str): cpu / cuda
```
wsad1, line 52: You don't need device. Pur the new tensors in `x.device`.

torch_geometric/explain/metrics/faithfulness.py
```diff
@@ -0,0 +1,169 @@
...
+
+we obtain the prediction probability vector ŷ_u using the GNN, i.e. ŷ_u = f (S_u),
+and using the explanation, i.e. ŷ_u′ = f (S_u′), where we generate a masked subgraph S_u′ by only keeping the original
+values of the top-k features identified by an explanation, and get their respective predictions ŷ_u′
+
+Un-Faithfulness of a GNN predictor is given as
+GEF(f,S_u,S_u') = 1 - exp^-KL(ŷ_u || ŷ_u')
+
+Note that higher values indicate a higher degree of unfaithfulness
+"""
```
wsad1, line 29: Move the documentation inside the function. Also you can refer to other methods in pyg to see how docs are styled like https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/gcn_conv.py. let me know if you need help with this.

wsad1 (commented):
Thanks for adding this. Lets add some tests.

torch_geometric/explain/metrics/faithfulness.py
```diff
@@ -0,0 +1,169 @@
...
+
+import torch
+import torch.nn.functional as F
+from utils import perturb_node_features
+
+from torch_geometric.data import Data
+from torch_geometric.explain import Explanation
+
+
+def get_node_faithfulness(node_idx: int, data: Data, explanation: Explanation,
```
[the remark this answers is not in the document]
ZeynepP: I thought of it they look like sharing lots of code but not exactly. If we do that we will have so many if  and else inside that one function.

torch_geometric/explain/metrics/faithfulness.py
```diff
@@ -0,0 +1,169 @@
...
+                          model: torch.nn.Module, topk: int, device: str,
+                          **kwargs):
```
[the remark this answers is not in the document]
ZeynepP: opps ok

torch_geometric/explain/metrics/faithfulness.py
ZeynepP: I tired to use it but I did not work I do not remember why :) I will check it again tell the reason. According to definition of faithfulness, it is calculated by masking nodes or by perturbation of node features. I am not an expert on explanaibility so maybe we should discuss this in detail with the issue owners.

ZeynepP: Issue refers to [Evaluating Explainability for Graph Neural Networks](https://arxiv.org/abs/2208.09339) but it is based on the paper you linked.  I have also read this paper. They mask edges or rewire edges with some probability pr. I will try to add this.

ZeynepP: For node level tasks if I use get_explanation_subgraph, I loose node_idx as there is no mapping. Maybe I am wrong!

ZeynepP: I added a function with subgraphs but not really sure. It was just to discuss. It corresponds well to explanation on the paper : 

1. get model’s prediction for node u ŷ_u = f(S_u)
2. generate a masked subgraph S_u′ by only keeping the original values of the top-k features identified by an explanation
3. get model's prediction for S_u' 
4. calculate metric 

Explanation gets subgraph with different masks for me but I think I need to use node_idx to get the prediction for node u and there is no mapping returned. I am a kind of lost here :)

wsad1: Oh i see the problem you are facing `node_idx` gets renumbered when you call `Explanation.get_explanation_subgraph`. In that case i think its best not to use `Explanation.get_explanation_subgraph`. 
Another thing you could try is to get the new node id
```
# The node being explained should not be masked out
assert explanation.node_maks[node_idx]>0 
node_idx = (explanation.node_mask>0).cumsum(dim=0)[node_idx]
```

ZeynepP: Yepp that was my problem since the beginning. I did not know we could get it this way, this is the expert touch :) Thanks

`sub_node_idx = (explanation.node_mask>0).cumsum(dim=0)[node_idx]`
In that case I believe , `explanation.x[node_idx]` should be equal to `explanation.get_explanation_subgraph().x[sub_node_idx]`
but it is not. When I check all the nodes in subgraph to find corresponding `explanation.x[node_idx]`, i see that corresponding sub_node_idx is always 0.

ZeynepP: `sub_node_idx = (explanation.node_mask>0).cumsum(dim=0)[node_idx]`
In that case I believe , `explanation.x[node_idx]` should be equal to `explanation.get_explanation_subgraph().x[sub_node_idx]`
but it is not. When I check all the nodes in subgraph to find corresponding `explanation.x[node_idx]`, i see that corresponding sub graph node_idx is always sub_node_idx-1. What am I missing?

BlazStojanovic (commented):
Thank you very much @ZeynepP for your work, and sorry for late PR review!

I've left some comments on how to make the functionality of `faithfulness.py` more general and make the usability more consistent between all the metrics. 

### High-level comments
- First, I don't think we need two examples for this metric. The behaviour between node and graph faithfulness should be similar enough that only one example is needed to understand both. 
- Secondly, we want the metrics to be as similar as possible among each other, so let's try and define faithfulness as: `graph_faithfulness(explainer: Explainer, explanation: Explanation, index: Union[int, Tensor] = None, topk: int, **kwargs):`. Alongside more unified interface, this has the additional benefit of having the Explainer (along with all of its configs) available for various checks and branches you might want to do. 
- Let's try to leverage as much of the existing functionality of `Explainer` and `Explanation` as possible, this will make implementation of `faithfulness` much easier and clearer (see my comments for details). Also on this note, I'll try to update `Explainer` and `Explanation` so I make your job easier, this will include updating `get_explanation_subgraph` and `get_complement_subgraph` so that they include all updated attributes needed for prediction and different utils for properly (and consistently across the codebase) obtaining thresholded masks. Will follow up with PR, so you can see what functionality is available. 

### Notes about faithfulness and perturbations
You are correct that there are really **two types of faithfulness** both apply to *instance* explanations, 
1. there is `node_faithfulness` as defined in https://arxiv.org/pdf/2106.09078.pdf. Here we do the perturbations. Lets leave this one for later, and first focus on `graph_faithfulness`. 
2. `graph_faithfulness` is the simpler of the two (imo.) it works like this: We have an explanation for an instance (e.g. a node), we use the explanation to generate a subgraph by only keeping the original values of the top-$k$ features identified by an explanation. Here we can go beyond the original paper as we can apply this metric to any `Explanation` meaning any combination of masks (node, edge, node-edge-feat), for any task level (node, graph edge) should work within our implementation. 

Here are my implementation suggestions:
- start with `graph_faithfulness()` i.e. what

torch_geometric/explain/metrics/faithfulness.py
wsad1: Yeah that makes sense it should be `sub_node_idx-1`, since the tensors are 0 indexed.

BlazStojanovic: @ZeynepP This will also be included in the revamp of the `Explanation` class, so don't worry about this for now 😃

rusty1s:
I did go with graph (un)faithfulness for now. @BlazStojanovic Let me know if you think node (un)faithfulness is also necessary.

rusty1s (approved):

Issue #6014, opened 2022-11-19
[Explainability]: `explain` support for heterogenous graphs.


wsad1:
### 🚀 The feature, motivation and pitch

Add a new class `HeteroExplainer` that supports explaining heterogenous graphs. This will involve the following changes to 
`explain` module
1. Create a `BaseExplainer` which will have common functions both `HeteroExplainer` and `Explainer` needs.
2. `Explainer` class will inherit from `BaseExplainer`. Note: The `Explainer` class will have the same interface the only change is that it inherits form `BaseExplainer`
3. Add `HeteroExplanation(HeteroData)` to hold heterogenous explanations.
4. `HeteroExplainer` inherits from `BaseExplainer`.  Both `HeteroExplainer` and `Explainer` can use the same config files as input.
```
class HeteroExplainer(BaseExplainer):
    def __call__(x_dict, edge_index_dict, *, target, target_idx)-> HeteroExplanation:
     ...
    def _threshold(self, explanation: HeteroExplanation)->HeteroExplanation:
    ...
```
cc: @dufourc1 @RexYing  for thoughts

### Alternatives

An alternative will be for `Explainer` to handle both hetero and homo graphs. But I feel this makes the interface more cluttered because it has to accept both dictionary or tensors and the implementation more complicated.

### Additional context

This is part of the explainability roadmap: https://github.com/pyg-team/pytorch_geometric/issues/5520, #5630

rusty1s:
Do you think we can just share the `Explainer` class for both cases? I would like to avoid creating replicas for each algorithm and the base class. Since we usually just forward pass input arguments to the model, I believe the interface isn't necessarily cluttered at all.

wsad1:
Yeah having `Explainer` support both seems reasonable now. But we'll have some `if` conditions checking if input is of type dictionary or tensor
```
class Explainer():
    def __call__(x: Union[Tensor, Dict[NodeType, Tensor]], edge_index: Union[Tensor,Dict[EdgeType, Tensor]],   *, target, 
                   target_idx)-> Union[HeteroExplanation, Explanation]:
          ....

    def _threshold(self, explanation: Union[Explanation, HeteroExplanation])->Union[Explanation, HeteroExplanation]:
        if isinstanc(expalnation, Explanation):
            return _threshold_explanation(explanation)
       else:
             return _threshold_hetero_explantion(explanation)
```

[...]

PR #6091, opened 2022-11-29
Heterogeneous Explanation

avgupta456:
Starts heterogeneous explanations using the new `Explainer` framework. Creates `HeterogeneousExplanation` and extends `Explainer`, `ExplainerAlgorithm`, and `DummyExplainer` to handle heterogeneous graphs. Tests and functionalities are partially complete, looking to get some initial validation before continuing.

Todo
- Heterogeneous explanation tests
- Add and test thresholding for heterogeneous explanations
- Finish some heterogeneous explanation methods
- Extend GNNExplainer (maybe in separate PR)

Closes https://github.com/pyg-team/pytorch_geometric/issues/6014


torch_geometric/explain/algorithm/utils.py
```diff
@@ -56,3 +56,12 @@ def clear_masks(model: torch.nn.Module):
             module._loop_mask = None
             module._apply_sigmoid = True
     return module
+
+
+def clear_hetero_masks(model: torch.nn.Module):
```
wsad1, line 61: `clear_masks` should work for heterogenous models too. so we don't need this function.

torch_geometric/explain/explainer.py
```diff
@@ -39,11 +44,19 @@ def __init__(
         self.model = model
         self.algorithm = algorithm
 
+        # check if model is heterogeneous
+        self.is_hetero = False
+        for _, module in self.model.named_modules():
+            if isinstance(module, torch.nn.ModuleDict):
```
wsad1, line 50: This might be a flaky way of detecting wether a model is heterogenous. Because a user might use a `ModuleDict` for other reasons too. 
Lets set `self.is_hetero` in `__call__` and then call `self.algorihm.support` again if `is_hetero` is `True`.

torch_geometric/explain/explainer.py
```diff
@@ -140,22 +153,22 @@ def __call__(
...
 
-    def _threshold(self, explanation: Explanation) -> Explanation:
-        """Threshold the explanation mask according to the thresholding method.
-
-        Args:
-            explanation (Explanation): The explanation to threshold.
-        """
+    def _threshold_explanation(
+        self, explanation: Union[Explanation, HeteroExplanation]
+    ) -> Union[Explanation, HeteroExplanation]:
```
wsad1, line 171:
```suggestion
        self, explanation: Explanation
    ) -> Explanation:
```

wsad1 (commented):
Thanks. Left some initial comments.

torch_geometric/explain/explainer.py
avgupta456: I think I understand what you're saying and pushed a commit, but not 100% sure on if the implementation is what you envisioned.

torch_geometric/explain/algorithm/base.py
```diff
@@ -20,12 +21,13 @@ class ExplainerAlgorithm(torch.nn.Module):
     def forward(
         self,
         model: torch.nn.Module,
-        x: Tensor,
-        edge_index: Tensor,
+        x: Union[Tensor, Dict[NodeType, Tensor]],
+        edge_index: Union[Tensor, Dict[EdgeType, Tensor]],
```
wsad1, line 25: Update the docstring, to mention when `x` and `edge_index` need to be dicts.

torch_geometric/explain/algorithm/base.py
```diff
@@ -43,14 +45,17 @@ def forward(
                 in a multi-task learning scenario. Should be kept to
                 :obj:`None` in case the model only returns a single output
                 tensor. (default: :obj:`None`)
+            is_hetero (bool, optional): Whether the input graph is
+                heterogeneous. (default: :obj:`False`)
```
wsad1, line 49: We can detect if its Heterogenous if `x` is a dict we don't need the `is_hetero` arg.

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +131,136 @@ def _apply_masks(
...
+class HeteroExplanation(HeteroData):
+    r"""Holds all the obtained explanations of a heterogeneous graph.
+
+    The explanation object is a :obj:`~torch_geometric.hetero_data.HeteroData`
+    object and can hold node-attribution, edge-attribution,
+    feature-attribution. It can also hold the original graph if needed.
+
+    Args:
+        node_mask (Dict[NodeType, Tensor], optional): Dictionary of Node-level
+            mask with shape :obj:`[num_nodes]`. (default: :obj:`None`)
```
wsad1, line 145:
```suggestion
        node_mask (Dict[NodeType, Tensor], optional): Dictionary of Node-level
            mask with  key `NodeType` and value `Tensor` of 
            shape :obj:`[num_nodes]`. (default: :obj:`None`)
```
Nitpick: just to be clear that the shape of the dict isn't `num_nodes`.

avgupta456:
I added logic and testing for thresholding, `validate()` function in heterogeneous explanation, and fixed some bugs. The last item left for this PR is creating heterogeneous explanation `subgraph`/`complement_subgraph` using `_apply_masks`. 

I am unsure if my implementation is good, overall having trouble extracting the `x_dict`, `edge_index_dict`, and `edge_attrs` from a `HeterogeneousExplanation`

Currently I am doing it like this

```python
x_dict = None
edge_index_dict = None
edge_attr_dict = None
for key, value in self.node_items():
    if key == "x_dict":
        x_dict = value
    elif key == "edge_index_dict":
        edge_index_dict = value
    elif key == "edge_attr_dict":
        edge_attr_dict = value
        continue
```

And setting the updated dictionaries like this:

```python
out.edge_index_dict.update(edge_index_dict)
out.edge_attr_dict.update(edge_attr_dict)
```

Is there a better way to access the `HeterogeneousExplanation` underlying graph?

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +131,188 @@ def _apply_masks(
...
+        for key, value in out.node_items():
+            if key == "edge_index_dict":
+                edge_index_dict = value
+            elif key == "edge_attr_dict":
+                edge_attr_dict = value
+
+        if edge_mask is not None:
+            for key, value in self.edge_mask.items():
+                edge_index_dict[key] = edge_index_dict[key][:, value]
+                edge_attr_dict[key] = edge_attr_dict[key][value]
```
wsad1, line 310:
```suggestion
            for edge_type in self.edge_types:
               out[edge_type].edge_index = out[edge_type].edge_index[:, value]
```
This might be cleaner. Also we need to check if `edge_attr` is in `out[edge_type]` before masking it.

torch_geometric/explain/algorithm/dummy_explainer.py
```diff
@@ -46,5 +47,75 @@ def forward(
...
+        r"""Returns random explanations based on the shape of the inputs.
+
+        Args:
+            model (torch.nn.Module): The model to explain.
+            x (Union[torch.Tensor, Dict[NodeType, torch.Tensor]]): The node
+                features.
+            edge_index (Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]):
+                The edge indices.
+            edge_attr (Union[torch.Tensor, Dict[EdgeType, torch.Tensor]],
+                optional): The edge attributes. (default: :obj:`None`)
```
wsad1, line 109: Update doc string to indicate when `dict` is passed.

test/explain/algorithm/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+    explainer = Explainer(
+        model=model, algorithm=HeteroExplainerAlgorithm(),
+        explainer_config=ExplainerConfig(
+            explanation_type="model",
+            node_mask_type="object",
+            edge_mask_type=None,
+        ), model_config=ModelConfig(
+            mode="regression",
+            task_level="node",
+        ))
```
wsad1, line 106:
```suggestion
     explainer.algorithm = HeteroExplainerAlgorithm()
```

test/explain/algorithm/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+
+    explainer(data.x_dict, data.edge_index_dict)
```
wsad1, line 81: Lets drop this test. it doesn't cover a different part of the code than the below two tests.

test/explain/test_explainer_hetero.py
```diff
@@ -0,0 +1,204 @@
...
+        return self.conv2(x, edge_index)
+
+
+class HeteroSAGE(torch.nn.Module):
+    def __init__(self, metadata):
+        super().__init__()
+        self.graph_sage = to_hetero(GraphSAGE(), metadata, debug=False)
+
+    def forward(self, x_dict, edge_index_dict, *args) -> torch.Tensor:
+        return self.graph_sage(x_dict, edge_index_dict)['paper']
```
wsad1, line 48: Lets move this to a `conftest.py` and share across all tests (https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session). Happy to discuss more about this or even address it in another PR.

wsad1 (commented):
Thanks for the updates @avgupta456 . Will take a look again once you are done with TODOs in test.

torch_geometric/explain/explainer.py
```diff
@@ -198,6 +200,72 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+                        k=self.threshold_config.value,
+                    )
+
+                    out = torch.zeros_like(v.flatten())
+                    if self.threshold_config.type == ThresholdType.topk:
+                        out[index] = value
+                    else:
+                        out[index] = 1.0
+                    mask_dict[key][k] = out.reshape(v.size())
+        else:
```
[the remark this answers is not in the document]
avgupta456: Yeah that's much simpler :)

avgupta456:
I'm currently running into an issue with HeteroExplanation that is blocking my further progress.

Unlike Explanation (inherits from Data), HeteroExplanation (inherits from HeteroData) does not correctly store the x_dict, edge_index_dict, and edge_attrs_dict when initialized with these arguments. It does work if the dictionaries are set as attributes afterwards.

```python
HeteroExplanation(x_dict=x_dict, edge_index_dict=edge_index_dict, edge_attrs_dict=edge_attrs_dict)  # does not work

explanation = HeteroExplanation()  # does work
explanation.x_dict = x_dict
explanation.edge_index_dict = edge_index_dict
explanation.edge_attrs_dict = edge_attrs_dict
```

One workaround is that when initialized with the dictionaries, they are accessible through self.node_items(). Perhaps this is a bug with HeteroData? Or maybe I am doing something else incorrectly.

As a result of this bug, the subgraph method isn't working as intended. Open to any advice.

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +131,172 @@ def _apply_masks(
...
+        self,
+        node_mask: Optional[Dict[NodeType, Tensor]] = None,
+        edge_mask: Optional[Dict[EdgeType, Tensor]] = None,
+        node_feat_mask: Optional[Dict[NodeType, Tensor]] = None,
+        edge_feat_mask: Optional[Dict[EdgeType, Tensor]] = None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        # HeteroData doesn't set all kwargs as attributes, so we need to do it
```
wsad1, line 168:
```suggestion
    def __init__(
        x_dict, edge_index_dict, edge_attr_dict = None,
        node_mask: Optional[Dict[NodeType, Tensor]] = None,
        edge_mask: Optional[Dict[EdgeType, Tensor]] = None,
        node_feat_mask: Optional[Dict[NodeType, Tensor]] = None,
        edge_feat_mask: Optional[Dict[EdgeType, Tensor]] = None,
    ):
        super().__init__()
        for key, val in x_dict.items():
            self[key].x = val
        for key, val in edge_index_dict.items():
            self[key].edge_index = val
            if edge_attr_dict is not None:
                self.[key].edge_attr = edge_attr_dict[key]
        # HeteroData doesn't set all kwargs as attributes, so we need to do it
```
Assigning it this way should help you create a valid `HeteData` object on which `subgraph` will work. Note, I made the assumption that `x_dict` and `edge_index_dict` are passed as arguments, which is reasonable given that we create `HeteroExplanation` internally.

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +131,172 @@ def _apply_masks(
...
+        self,
+        node_mask: Optional[Tensor] = None,
+        edge_mask: Optional[Tensor] = None,
+    ) -> 'HeteroExplanation':
+        out = copy.copy(self)
+
+        if edge_mask is not None:
+            for edge_type in self.edge_types:
+                if edge_type not in edge_mask:
+                    continue
```
wsad1, line 293: Can we drop this? Because `validate` would fail if a `edge_type` didn't have a mask.

wsad1 (commented):
This looks clean now. Will approve once the `subgraph` issue is resolved.

dufourc1 (commented):
Thanks for the work ! LGTM, these are nitpick comments

test/explain/test_explainer_hetero.py
avgupta456: Maybe best to move to a separate PR, I agree there is a lot of repeated code across the explanation tests that can be consolidated. Should I create an issue for this?

torch_geometric/explain/explanation.py
avgupta456: Thanks @wsad1! Resolving this comment as @dufourc1's comment is also below.

test/explain/test_explainer_hetero.py
wsad1: Yeah feel free to add a TODO and also raise an issue.

torch_geometric/explain/explainer.py
```diff
@@ -198,6 +200,40 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+            mask_dict[key] = self._threshold_homogeneous_explanation(mask)
+        return mask_dict
+
+    def _threshold(
+        self, explanation: Union[Explanation, HeteroExplanation]
+    ) -> Union[Explanation, HeteroExplanation]:
+        """Threshold the explanation mask according to the thresholding method.
+
+        Args:
+            explanation (Explanation): The explanation to threshold.
```
wsad1, line 218:
```suggestion
            explanation (Explanation or HeteroExplanation): The explanation to threshold.
```

torch_geometric/explain/algorithm/dummy_explainer.py
```diff
@@ -46,5 +47,74 @@ def forward(
...
+        **kwargs,
+    ) -> HeteroExplanation:
+        r"""Returns random explanations based on the shape of the inputs.
+
+        Args:
+            model (torch.nn.Module): The model to explain.
+            x (Dict[NodeType, torch.Tensor]): The node features.
+            edge_index (Dict[EdgeType, torch.Tensor]): The edge indices.
+            edge_attr (Dict[EdgeType, torch.Tensor], optional): The edge
+                attributes. (default: :obj:`None`)
```
wsad1, line 65:
```suggestion
            x_dict (Dict[NodeType, torch.Tensor]): The node features for each node type.
            edge_index_dict (Dict[EdgeType, torch.Tensor]): The edge indices for each edge type.
            edge_attr_dict (Dict[EdgeType, torch.Tensor], optional): The edge
                attributes for each edge type. (default: :obj:`None`)
```

torch_geometric/data/hetero_data.py
```diff
@@ -580,7 +580,7 @@ def subgraph(self, subset_dict: Dict[NodeType, Tensor]) -> 'HeteroData':
             subset_dict (Dict[str, LongTensor or BoolTensor]): A dictonary
                 holding the nodes to keep for each node type.
         """
-        data = self.__class__(self._global_store)
+        data = self.__class__(**self._global_store)
```
wsad1, line 583: Why do we need this change?

test/explain/test_explainer_hetero.py
```diff
@@ -0,0 +1,204 @@
...
+
+    for key in explanation.available_explanations:
+        masks = explanation[key]
+        for mask in masks.values():
+            assert set(mask.unique().tolist()).issubset({0, 1})
+
+
+@pytest.mark.parametrize('threshold_value', list(range(1, 10)))
+@pytest.mark.parametrize('node_mask_type', ['object', 'attributes'])
+def test_topk_threshold(data, threshold_value, node_mask_type):
```
wsad1, line 139: Can we combine topk and topk_hard into one test, using parametrization?

test/explain/algorithm/test_base.py
```diff
@@ -0,0 +1,81 @@
...
+            explainer_config=ExplainerConfig(
+                explanation_type="model",
+                node_mask_type="object",
+                edge_mask_type=None,
+            ), model_config=ModelConfig(
+                mode="regression",
+                task_level="node",
+            ))
+
+        explainer(data.x_dict, data.edge_index_dict)
```
wsad1, line 77:
```suggestion

      explainer = Explainer(
          model=model, algorithm=HomoExplainerAlgorithm(),
          explainer_config=ExplainerConfig(
              explanation_type="model",
              node_mask_type="object",
              edge_mask_type=None,
          ), model_config=ModelConfig(
              mode="regression",
              task_level="node",
          ))
    with pytest.raises(ValueError):
        explainer(data.x_dict, data.edge_index_dict)
```

wsad1 (approved):
Thanks @avgupta456, this looks good.

torch_geometric/data/hetero_data.py
avgupta456: Since self.__class__ is HeteroExplanation, the dictionary self._global_store is interpreted as x_dict, and it complains for the other arguments (not to mention all the arguments are stored incorrectly in x_dict).

Since HeteroData() treats _mapping and kwargs together, this operation doesn't change the existing behavior and allows the arguments to be correctly mapped for HeteroExplanation.

test/explain/test_explainer_hetero.py
avgupta456: Created #6186

Issue #5631, opened 2022-10-07
Visualization of GNN Explanations


RexYing:
### 🚀 The feature, motivation and pitch

Improve visualization utility functions of the explanation. It can be in the form of feature importance maps, or visualization of the explanation subgraph. Support parameters such as the size of explanation and show weighted importance for edges to control the visualization output.

Plan:
- [x] Explanation subgraph visualization for thresholded explanation. The output would be a visualization of the explanation subgraph after filtering out edges according to their importance / mask values (by a threshold). 
- [x] Explanation subgraph visualization for non-threshold explanation. Potentially plot the whole graph or the neighborhood graph, where edge width / color depends on the importance value of each edge.
- [x] Visualize feature importance. Draw a bar plot that corresponds to the importance of each feature dimension (and sort them).

rjurney:
This would be great!

davodogster:
Yes this would be great!

sebastian-montero:
Hello all! I will be working on the third point for this PR. 

@RexYing I am looking at the `GNNExplainer` module in this [example](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/gnn_explainer.py). If I understand this correctly, the `explain_node` method already provides a `node_feat_mask` for a specific node. What is not clear is if this mask the actual feature importance? If so, I can wrap this in a matplotlib bar plot. 

I am planning to add the options for feature names and top N feature to plot. Is there anything else we should add for this specific point?

Thank you!

rusty1s:
This sounds good. I think we can support this option for both`MaskType.common_attributes` and `MaskType.attributes` (see #5967). For `MaskType.attributes`, we first need to reduce the feature mask from shape `(N,F)` to shape `(F,)`.

snknitin:
Does this include visualization of Heterogenous Graphs? How would one go about using the explainer module right now for HeteroData and for edge based prediction tasks?

rusty1s:
It does not include visualizations of heterogeneous graphs. For heterogeneous graphs, we can later just leverage `data.to_homogeneous()` for visualization purposes.

wsad1:
For hetero graphs it might be useful to include the node and edge types in the visualization. So we  could consider building visualizations directly on heterogenous graphs.

rusty1s:
Yes, these are available through `data.to_homogeneous()`.

akash-vartak:
[...]

PR #6094, opened 2022-11-30
Feature importance visualization

sebastian-montero:
Work in progress for feat importance visualization for issue [5631](https://github.com/pyg-team/pytorch_geometric/issues/5631)


torch_geometric/nn/models/explainer.py
```diff
@@ -388,3 +388,52 @@ def visualize_subgraph(self, node_idx: Optional[int], edge_index: Tensor,
         nx.draw_networkx_labels(G, pos, **label_kwargs)
 
         return ax, G
+
+    def visualize_feature_importance(
```
wsad1, line 392: Lets add this to the new `Explainer` class in `torch_geometric.explain`, since we are going to depreciate this in the future.

torch_geometric/nn/models/explainer.py
```diff
@@ -388,3 +388,52 @@ def visualize_subgraph(self, node_idx: Optional[int], edge_index: Tensor,
...
+        self,
+        node_feat_mask: Tensor,
```
wsad1, line 394: Pass in `Explanation` object instead, and look for `node_feat_mask` in that.

wsad1 (commented):
Thanks for adding this. Lets implement this in the new `Explainer` interface.

torch_geometric/nn/models/explainer.py
sebastian-montero: Sounds good to me! I'll update and push!

torch_geometric/nn/models/explainer.py
sebastian-montero: Done! I had to pass in the `node_index` too to filter the `node_feat_mask`

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+        n_features: int = 20,
+    ):
+        r"""Creates a bar plot of the node features importance given
+        a :attr:`node_feat_mask`
+
+        Args:
+            node_feat_mask (Tensor): The node feature mask to plot.
+            feat_labels (List[str], optional): Optional labels for features.
+            n_features (int, optional): Top N features to plot.
+                (default: :int:`20`)
```
EdisonLeeeee, line 221:
```suggestion
                (default: :obj:`20`)
```

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+        :rtype: :class:`matplotlib.axes.Axes`
+        """
+        import matplotlib.pyplot as plt
+        import pandas as pd
+
+        node_feature_explanation = explanation.node_feat_mask[index]
+        node_feature_explanation = node_feature_explanation.numpy()
```
EdisonLeeeee, line 228:
```suggestion
        node_feature_explanation = node_feature_explanation.cpu().numpy()
```
In case `node_feature_explanation` is on GPU.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+
+        if feat_labels is None:
+            feat_labels = [
```
EdisonLeeeee, line 231: `feat_labels` seems not necessary to be a string. You can just pass integers instead.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+            node_feat_mask (Tensor): The node feature mask to plot.
+            feat_labels (List[str], optional): Optional labels for features.
+            n_features (int, optional): Top N features to plot.
+                (default: :obj:`20`)
+        :rtype: :class:`matplotlib.axes.Axes`
+        """
+        import matplotlib.pyplot as plt
+        import pandas as pd
+
+        node_feature_explanation = explanation.node_feat_mask[index]
```
wsad1, line 227: It looks like we are only looking at the feature importance of a single node. What we want to do instead is reduce the `node_feat_mask` of shape `N*F` to `F` by summing across all nodes. Also we don't need `index` input as `node_feat_mask` will be zero for nodes that are not in the subgraph of the node that was explained.

wsad1, line 227: `node_feat_mask` might not exist in `Explanation`, lets that it exists, and throw a descriptive error if it doesn't.

wsad1 (commented):
Thanks for the update.

sebastian-montero:
@EdisonLeeeee @wsad1, thank you for your comments! I have updated method.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+        n_features: int = 20,
+    ):
+        r"""Creates a bar plot of the node features importance given
+        a :attr:`node_feat_mask`
+
+        Args:
+            node_feat_mask (Tensor): The node feature mask to plot.
+            feat_labels (List[str], optional): Optional labels for features.
+            n_features (int, optional): Top N features to plot.
+                (default: :obj:`20`)
```
wsad1, line 220: Update doc:
1. Use `explanation` instead of `node_feat_mask`.
2. rename `n_features` to `top_k` for consistency.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,54 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+        :rtype: :class:`matplotlib.axes.Axes`
+        """
+        import matplotlib.pyplot as plt
+        import pandas as pd
+
+        if 'node_feat_mask' not in explanation.available_explanations:
+            raise Exception("`node_feat_mask` not available in explanation.")
```
wsad1, line 227:
```suggestion
            raise ValueError("'node_feat_mask' not available in 'explanation'.")
```

wsad1 (commented):
Thanks for the updates. Can we update this example https://github.com/pyg-team/pytorch_geometric/blob/master/examples/gnn_explainer.py to use `visualize_feature_importance`.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,55 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
         return explanation
+
+    def visualize_feature_importance(
+        self,
+        explanation: Explanation,
+        feat_labels: Optional[List[str]] = None,
+        top_k: int = 20,
+    ):
+        r"""Creates a bar plot of the node features importance given
+        a :attr:`node_feat_mask`
```
wsad1, line 214:
```suggestion
        r"""Creates a bar plot of the node features importance by
        summing up :attr:`explanation.node_feat_mask` across all nodes.
```

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,55 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+
+        Args:
+            explanation (Explanation): The explanation object.
+            feat_labels (List[str], optional): Optional labels for features.
+            top_k (int, optional): Top K features to plot.
+                (default: :obj:`20`)
```
wsad1, line 220: Can we set default value to `None`. And if this is `None` plot all features.

examples/gnn_explainer.py
```diff
@@ -57,3 +58,7 @@ def forward(self, x, edge_index, edge_weight):
 explanation = explainer(data.x, data.edge_index, index=node_index,
                         edge_weight=data.edge_weight)
 print(f'Generated explanations in {explanation.available_explanations}')
+
+ax = explainer.visualize_feature_importance(explanation)
+print('Plotting the feature importance')
+plt.savefig("feature_importance.png")
```
wsad1, line 64: I'd like to see how this looks :). Could you paste it here?

wsad1 (commented):
Thanks for the updates @sebastian-montero . This looks good now. Will take a look again tomorrow and approve.

examples/gnn_explainer.py
sebastian-montero: Here's an example: 
feature_importance

sebastian-montero: FYI running this with a GNNExplainer of 10 epochs to expedite the process so importances are very similar

torch_geometric/explain/explainer.py
sebastian-montero: I think we should have a `top_k`, in the example case, there are >1400 features and the visualisation breaks when we try to plot all of them.

Another option would be to plot all feature importance if the number of features is below a default `k`. Else, plot importance of the the top 20 feats. 


Screenshot 2022-12-07 at 17 08 58

examples/gnn_explainer.py
wsad1: Great thanks for sharing.

torch_geometric/explain/explainer.py
```diff
@@ -203,3 +203,58 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
+        explanation: Explanation,
+        feat_labels: Optional[List[str]] = None,
+        top_k: int = 20,
+    ):
+        r"""Creates a bar plot of the node features importance by
+        summing up :attr:`explanation.node_feat_mask` across all nodes.
+
+        Args:
+            explanation (Explanation): The explanation object.
+            feat_labels (List[str], optional): Optional labels for features.
```
wsad1, line 218:
```suggestion
            feat_labels (List[str], optional): Optional labels for features.
                (default :obj:`None`)
```

torch_geometric/explain/explainer.py
wsad1: I agree that we should retain the `top_k` argument for reasons you just pointed out. But `top_k` can be optional, and if its not passed we just plot all features. With that the users does need to worry about `top_k` if they have few features.

wsad1 (approved):
Thanks @sebastian-montero for all the work.
Lets add a test. Simply calling `visualize_feature_importance` from a test in https://github.com/pyg-team/pytorch_geometric/blob/master/test/explain/test_explainer.py should do.

sebastian-montero:
thanks for the comments @wsad1!

PR #6100, opened 2022-11-30
Added `torch_geometric.optional_validation` context-manager 

hatemhelal:
This patch extends the feature added in #5290 with:

* `torch_geometric.is_validation_enabled` for querying the current state of the optional validation
* `torch_geometric.set_validation` used to globally toggle the optional validation state
* `torch_geometric.optional_validation` that creates a context manager to toggle the validation state in a scoped manner

### DIFF
--- a/torch_geometric/nn/aggr/base.py
+++ b/torch_geometric/nn/aggr/base.py
@@ -1,4 +1,3 @@
-from contextlib import contextmanager
 from typing import Optional, Tuple
 
 import torch
@@ -59,7 +58,6 @@ class Aggregation(torch.nn.Module):
         - **output:** graph features :math:`(|\mathcal{G}|, F_{out})` or node
           features :math:`(|\mathcal{V}|, F_{out})`
     """
-    _validate = __debug__
 
     # @abstractmethod
     def forward(self, x: Tensor, index: Optional[Tensor] = None,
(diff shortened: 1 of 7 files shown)


torch_geometric/nn/aggr/base.py
```diff
@@ -100,6 +106,24 @@ def set_validate_args(value: bool):
         """
         Aggregation._validate = value
 
+    @staticmethod
+    @contextmanager
+    def validation_context(value: bool):
```
rusty1s, line 111: Wondering if we can do this more broadly in PyG, e.g.,
```python
with torch_geometric.validate(False):
    out = model(data.x, data.edge_index)
```

hatemhelal: Yes I like that idea, we could coalesce other interfaces (`MessagePassing`, others? ) to have an optional validation step that can be toggled off for ahead-of-time compilation.

rusty1s: Yeah exactly. How do we wanna proceed with this. Do we still want to merge this PR as it is and do the re-factor later?

hatemhelal: I'd like feedback on the `validate` module I've added in the current version of this patch and if it looks ok we can merge this.  We can then refactor other interfaces (only `MessagePassing` is the other one I'm currently aware of) to introduce optional data validation.

rusty1s: Sounds good, the PR looks good to me :)

wsad1 (approved):
Thanks this looks good.

torch_geometric/validate.py
```diff
@@ -0,0 +1,41 @@
...
+
+@contextmanager
+def optional_validation(value: bool):
+    r"""Creates a context-manager for managing whether any optional validation
+    is enabled or disabled within a scope.
+
+    Example:
+
+        >>> with torch_geometric.validate(False):
+        ...     out = model(data.x, data.edge_index)
```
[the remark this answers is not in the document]
hatemhelal: Yes lets go for the shorter name

torch_geometric/nn/aggr/base.py
```diff
@@ -190,3 +173,9 @@ def expand_left(ptr: Tensor, dim: int, dims: int) -> Tensor:
     for _ in range(dims + dim if dim < 0 else dim):
         ptr = ptr.unsqueeze(0)
     return ptr
+
+
+def is_validation_enabled() -> bool:
+    from torch_geometric import is_validation_enabled
+
+    return is_validation_enabled()
```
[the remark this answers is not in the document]
hatemhelal: I think it ended up with a circular import (just confirming I'm not misremembering why I moved the import into the function)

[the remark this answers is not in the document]
hatemhelal: ah yes here is the stack:
```bash
$ pytest -s test/test_validate.py
ImportError while loading conftest '/localdata/hatemh/pytorch_geometric-fork/test/conftest.py'.
test/conftest.py:7: in <module>
    from torch_geometric.data import Dataset
torch_geometric/__init__.py:5: in <module>
    import torch_geometric.transforms
torch_geometric/transforms/__init__.py:53: in <module>
    from .feature_propagation import FeaturePropagation
torch_geometric/transforms/feature_propagation.py:5: in <module>
torch_geometric/nn/__init__.py:3: in <module>
    from .sequential import Sequential
torch_geometric/nn/sequential.py:8: in <module>
    from torch_geometric.nn.conv.utils.jit import class_from_module_repr
torch_geometric/nn/conv/__init__.py:1: in <module>
    from .message_passing import MessagePassing
torch_geometric/nn/conv/message_passing.py:26: in <module>
    from torch_geometric.nn.aggr import Aggregation, MultiAggregation
torch_geometric/nn/aggr/__init__.py:1: in <module>
    from .base import Aggregation
torch_geometric/nn/aggr/base.py:8: in <module>
    from torch_geometric import is_validation_enabled
E   ImportError: cannot import name 'is_validation_enabled' from partially initialized module 'torch_geometric' (most likely due to a circular import)
```

torch_geometric/nn/aggr/base.py
hatemhelal: > We can then refactor other interfaces (only `MessagePassing` is the other one I'm currently aware of) to introduce optional data validation.

I looked more into the `MessagePassing` and from what I can tell all the places where a `ValueError` is raised should be fine in a compiled context.  The main thing that needs to be avoided are places where there are conditional branches that depend on the input data (a good example of this is in the aggregation interface).

hatemhelal:
After a bit more thought it seems the entire feature of optional validation can be removed (see #6175)

The trick is to defer any conditional validation on input data until an exception is thrown.

PR #6102, opened 2022-11-30
Support `to_captum_input` + `to_captum_model` for batch explanations

### LINKED ISSUE #5917: Batch explanation with Captum

Currently, we can use `to_captum` to explain one prediction example only. However, it would be good to be able to explain predictions in batches as well. 
Currently, this does not work because Captum assumes that the dimension 0 of the inputs is equal to the number of prediction examples. For batch explanations, the target vector and the input vector are split based on the first dimension and passed to the forward function.

One way to solve this is to provide a method `to_captum_data` that transforms the data, given batch size and output indices, to the required captum format, in addition to having `to_captum_model`. The data transformation method is also needed for heterogeneous support. Optionally we could also provide both functionalities in `to_captum`. 
 

### Alternatives

_No response_

### Additional context

_No response_

RBendias:
Provides the possibility to get explanations batch-wise (fixes #5917).


torch_geometric/nn/models/captum.py
```diff
@@ -155,10 +194,17 @@ def _raise_on_invalid_mask_type(mask_type: str):
         raise ValueError(f"Invalid mask type (got {mask_type})")
 
 
-def to_captum_input(x: Union[Tensor, Dict[EdgeType, Tensor]],
-                    edge_index: Union[Tensor, Dict[EdgeType,
-                                                   Tensor]], mask_type: str,
-                    *args) -> Tuple[Tuple[Tensor], Tuple[Tensor]]:
+def _expand_input(input: Tensor, n_samples: int) -> Tensor:
+    return input.repeat(n_samples, *[1] * input.dim())
```
RBendias, line 198: expand, instead of repeat

torch_geometric/nn/models/captum.py
```diff
@@ -176,51 +222,54 @@ def to_captum_input(x: Union[Tensor, Dict[EdgeType, Tensor]],
...
     else:
         raise ValueError(
             "'x' and 'edge_index' need to be either"
             f"'Dict' or 'Tensor' got({type(x)}, {type(edge_index)})")
-    additional_forward_args.extend(args)
     return tuple(inputs), tuple(additional_forward_args)
 
 
+# TODO (jinu/ramona): Rethink batch hetero output dicts
 def captum_output_to_dicts(
```
RBendias, line 273: @wsad1 What do you think about having a general `captum_output_transform` method for both hetero and nonhetero graphs?

wsad1: Yeah thats a good idea.

wsad1 (commented):
Thanks for adding this feature. 
One concern I have is that this might lead to out of memory errors. For example if a user decides to explain all the node in a batch and `batch_size=1024` then this approach replicates the graph `1024` times. 
Instead of this can we assume that each node that needs explaining has a disjoint subgraph and split a single graph into disjoint subgraphs?

RBendias:
> Thanks for adding this feature.
   One concern I have is that this might lead to out of memory errors. For example if a user decides to 
   explain all the node in a batch and batch_size=1024 then this approach replicates the graph 1024 times.
   Instead of this can we assume that each node that needs explaining has a disjoint subgraph and split a 
   single graph into disjoint subgraphs?

Do you think this makes such a big difference? Because after all we want to get the importance of the whole graph. We would create a subgraph, calculate the importance values on that subgraph, then create the entire graph and assign the importance of the subgraphsto the entire graph.

torch_geometric/nn/models/captum.py
```diff
@@ -79,39 +96,55 @@ def __init__(self, model: torch.nn.Module, mask_type: str, output_id: int,
...
+            data_list.append(data)
+        return Batch.from_data_list(data_list)
 
     def forward(self, *args):
         # Validate args:
+        batch_size = args[0].shape[0]
+        if type(self.output_idx) == list:
+            msg = "`internal_batch_size` needs to be a multiple of the " \
+                  "number of objects to be evaluated."
+            assert args[0].shape[0] % (len(self.output_idx)) == 0, msg
```
wsad1, line 145:
```suggestion
            assert batch_size == len(self.output_idx), msg
```
Wouldn't this be true?

torch_geometric/nn/models/captum.py
```diff
@@ -124,24 +157,28 @@ def forward(self, *args):
...
         if 'edge' in self.mask_type:
             clear_masks(self.model)
 
         if self.output_idx is not None:
-            x = x[self.output_idx].unsqueeze(0)
+            # TODO (ramona): Remove this hacky solution
+            indices = torch.tensor(self.output_idx * batch_size, dtype=int)
+            x = x[torch.tensor(
+                [x.shape[0] / batch_size * (i)
+                 for i in range(batch_size)], dtype=int) + indices]
```
wsad1, line 181: Add a comment to explain this.

torch_geometric/nn/models/captum.py
```diff
@@ -176,59 +220,65 @@ def to_captum_input(x: Union[Tensor, Dict[EdgeType, Tensor]],
...
+    mask_type: str,
+    metadata: Metadata,
+) -> Tuple[Optional[List[Dict[NodeType, Tensor]]], Optional[List[Dict[
+        EdgeType, Tensor]]]]:
+    """Convert the output of `Captum.ai <https://captum.ai/>`_ attribution
+    methods which is a tuple of attributions to dictonaries with with node and
+    edge attribution tensors. If more than one object is explained, the
+    explanations are stored in a list of dictionaries. This function is used
+    while explaining :obj:`HeteroData` objects. See
+    :obj:`torch_geometric.nn.to_captum_model` for example usage.
```
wsad1, line 282:
```suggestion
    :obj:`~torch_geometric.nn.to_captum_model` for example usage.

```

torch_geometric/nn/models/captum.py
```diff
@@ -237,37 +287,52 @@ def captum_output_to_dicts(
...
+        elif mask_type == "node_and_edge":
+            assert len(edge_types) + len(node_types) == len(captum_attrs_i)
+            x_attr_dicts.append(
+                dict(zip(node_types, captum_attrs_i[:len(node_types)])))
+            edge_attr_dicts.append(
+                dict(zip(edge_types, captum_attrs_i[len(node_types):])))
+    if len(x_attr_dicts) == 1:
+        x_attr_dicts = x_attr_dicts[0]
+    if len(edge_attr_dicts) == 1:
+        edge_attr_dicts = edge_attr_dicts[0]
```
wsad1, line 334: Lets return a `List` even if it of length 1. Else we'll have to update the return type hint.

torch_geometric/nn/models/captum.py
```diff
@@ -237,37 +287,52 @@ def captum_output_to_dicts(
...
+    return x_attr_dicts, edge_attr_dicts
```
wsad1, line 335: Return `None` if a list is empty.

torch_geometric/nn/models/captum.py
```diff
@@ -365,10 +431,11 @@ def to_captum_model(
         mask_type (str, optional): Denotes the type of mask to be created with
             a Captum explainer. Valid inputs are :obj:`"edge"`, :obj:`"node"`,
             and :obj:`"node_and_edge"`. (default: :obj:`"edge"`)
-        output_idx (int, optional): Index of the output element (node or link
-            index) to be explained. With :obj:`output_idx` set, the forward
-            function will return the output of the model for the element at
-            the index specified. (default: :obj:`None`)
+        output_idx (Union[int, List[int]], optional): Index of the output
```
wsad1, line 434: Update code-block above to show `output_idx` with multiple nodes.

wsad1 (commented):
Thanks @RBendias. Left some initial comments.

> Do you think this makes such a big difference? 

I think we'll need to revisit this. Because the current implementation might break down if a user want to explain 1000s of nodes at once, which is a very common scenario, given a batch during training consists of 1000s of nodes.

torch_geometric/nn/models/captum.py
RBendias: No, because it can be more than the number of `output_idx`. E.g. you can set the batch_size to be 50 even if you want to explain only 2 node predictions. This can happen if the calculation of the explanation requires several runs trough the model.

torch_geometric/nn/models/captum.py
RBendias: What is `~` doing?

rusty1s: It will only maintain the method function name in the documentation.

PR #6073, opened 2022-11-26
Explainability Dataset Task: `ERGraph`

shenoynikhil:
Adds ERGraph based on @rfdavid's PR #6072. Fixes task #5817.

```python
from torch_geometric.datasets.generators import ERGraph

motif = Motif(structure='house')
generator = ERGraph(motif=motif, num_nodes=300, edge_prob=0.2, num_motifs=80)
dataset = ExplainerDataset(generator=generator)
```

#### TODO
- [x] Tests
- [x] ChangeLog

PR #6104, opened 2022-12-01
Explainability Benchmark Dataset Framework

rfdavid:
This PR implements benchmark dataset overall framework and API for generating benchmark datasets (#5817). Any feedback is appreciated, especially regarding the proposed architecture.

### Implementing a new graph generator

- Create a class inside generators inheriting from GraphGenerator
- Implement `generate_base_graph` using the provided methods (`generate_feature` and `attach_motif`)
- See #6072 and #6073 to check how it was implemented
- The current Motif generator used here will be replaced by #6179 

### Example for the final user

```
motif = Motif('house')
generator = BAGraph(num_nodes=300, num_motifs=80, motif=motif)
dataset = ExplainerDataset(generator)
```

TODO:

- [x] Finish graph generator (provide all necessary methods: feature generator, label generator)
- [x] Add tests
- [x] Add to Change log
- [x] Documentation


torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,57 @@
+from typing import Callable, Optional
+
+import torch
+
+from torch_geometric.data import Data
+
+
+class GraphGenerator:
+    def __init__(self, motif: Optional[Callable], num_nodes: int = 300):
```
RexYing, line 9: we could have an option to set the seed here. By default we'll use the same seed, for reproducibility. But if user wants to randomize, they are allowed to set the option and we could generate different graph datasets with the same generator.

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # TODO: FeatureGenerator (feature distribution...)
+    def generate_feature(self, num_features: int = 10):
+        self.x = torch.ones((self.num_nodes, num_features), dtype=torch.float)
+
+    def attach_motif(self, num_motifs=80,
+                     connection_distribution: str = 'random'):
+        if connection_distribution == 'random':
+            connecting_nodes = torch.randperm(self.num_nodes)[:num_motifs]
+        else:
```
RexYing, line 29: with random seed, i think we could also just use the "random" option and don't need the other uniform option (but we need to check if the random number generator will be deterministic given the seed here in attach_motif, after the generate_graph and generate_feature calls).

torch_geometric/datasets/generators/motif.py
```diff
@@ -0,0 +1,19 @@
+import torch
+
+
+class Motif:
+    def __init__(self, structure: str):
+        self.edge_index = None
```
RexYing, line 6: perhaps we could document what kind of "named" motifs are available so far (house). 

When people add more named motifs, they could modify in the documentation.

Also mention a todo that in future, we could allow edge_index or networkx objects for custom motifs other than named motifs.

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,57 @@
...
+        self.x = None
+
+    def generate_base_graph(self) -> Data:
+        raise NotImplementedError
+
+    # TODO: FeatureGenerator (feature distribution...)
+    def generate_feature(self, num_features: int = 10):
+        self.x = torch.ones((self.num_nodes, num_features), dtype=torch.float)
+
+    def attach_motif(self, num_motifs=80,
```
RexYing, line 25: we could have a test on attach_motif to e.g. make sure the correct number of motifs are attached (num nodes) etc.

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,57 @@
...
+                     connection_distribution: str = 'random'):
+        if connection_distribution == 'random':
+            connecting_nodes = torch.randperm(self.num_nodes)[:num_motifs]
+        else:
+            step = self.num_nodes // num_motifs
+            connecting_nodes = torch.arange(0, self.num_nodes, step)
```
RexYing, line 31: there's also a question of whether the attachment will be made on the same orbit of the motif (e.g. always the tip of house or a random node in house motif). Either way is ok for this PR, but might be a good idea to document it.

In future, if we see request to have the other option, we could have a different pr to address that.

RexYing (changes requested):
Thanks for the great pr:) We also had a great discussion and I just note down some comments I made.

rfdavid:
> Thanks for the great pr:) We also had a great discussion and I just note down some comments I made.

It was a great pleasure for me, thank you so much for the insights. All points are clear, I'll address them and commit my changes soon.

torch_geometric/datasets/generators/graph_generator.py
rfdavid: Thanks for all the comments!
I added a seed call before generating the base graph. I'm not sure if this approach is too naive, but I ran a couple of tests re-using the generator to create other datasets:

```
otif = Motif('house')
generator = BAGraph(num_nodes=300, num_motifs=80, motif=motif, seed=444)
dataset = ExplainerDataset(generator)
data1 = dataset[0]

np.random.rand(3,2)
torch.rand(2, 3)

generator.num_nodes = 300
generator.seed = 1234
dataset = ExplainerDataset(generator)
data2 = dataset[0]

generator.num_nodes = 300
generator.seed = 444 # initial seed
dataset = ExplainerDataset(generator)
data3 = dataset[0]

print("data1", data1.edge_index)
print("data2", data2.edge_index)
print("data3", data3.edge_index)
print(data1.edge_index == data3.edge_index)

Output:

data1 tensor([[  0,   0,   0,  ..., 699,  21, 695],
        [  2,   5,   7,  ..., 696, 695,  21]])
data2 tensor([[  0,   0,   0,  ..., 699, 240, 695],
        [  6,  10,  15,  ..., 696, 695, 240]])
data3 tensor([[  0,   0,   0,  ..., 699,  21, 695],
        [  2,   5,   7,  ..., 696, 695,  21]])
tensor([[True, True, True,  ..., True, True, True],
        [True, True, True,  ..., True, True, True]])

```

torch_geometric/datasets/generators/graph_generator.py
rfdavid: Removed the uniform option.

torch_geometric/datasets/generators/graph_generator.py
rfdavid: I added the description in the classes but I'm planning to work more on writing better documentation on those.

RexYing (approved):
Thanks overall it looks great now. I'll let Zecheng or Charles to have a look in case I miss anything.

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,84 @@
...
+                               int(connecting_nodes[i])]]))
+
+            edge_labels.append(
+                torch.ones(self.motif.edge_index.size(1), dtype=torch.long))
+            edge_labels.append(torch.zeros(2, dtype=torch.long))
+            node_labels.append(self.motif.label)
+            self.num_nodes += self.motif.num_nodes
+
+        self._expl_mask = torch.zeros(self.num_nodes, dtype=torch.bool)
+        self._expl_mask[torch.arange(self.motif.num_nodes * num_motifs,
```
zechengz, line 78: Just curious why we assign the nodes start from `self.motif.num_nodes * num_motifs` with the step `self.motif.num_nodes` with the explain mask `True`?

test/explain/test_graph_generator.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+import torch
+
+from torch_geometric.datasets import ExplainerDataset
+from torch_geometric.datasets.generators import Motif
+from torch_geometric.datasets.generators.graph_generator import GraphGenerator
+from torch_geometric.utils import barabasi_albert_graph
+
+
+class BAGraph(GraphGenerator):
```
zechengz, line 10: Should we define the `BAGraph` here or should we move it to the generators.

torch_geometric/datasets/explainer_dataset.py
```diff
@@ -0,0 +1,25 @@
...
+            to be used: :obj:`torch.geometric.datasets.generators.BAGraph`,
+            :obj:`torch.geometric.datasets.generators.ERGraph`.
+        transform (callable, optional): A function/transform that takes in an
+            :obj:`torch_geometric.data.Data` object and returns a transformed
+            version. The data object will be transformed before every access.
+            (default: :obj:`None`)
+    """
+    def __init__(
+        self,
+        generator,
```
zechengz, line 20:
```suggestion
        generator: GraphGenerator,
```

zechengz (commented):
Thank you :) ! In general the PR looks good to me. Left some comments.

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,84 @@
...
+        raise NotImplementedError
+
+    def generate_graph(self):
+        if self.seed:
+            seed_everything(self.seed)
+        self.generate_base_graph()
+
+    @property
+    def data(self):
+        return Data(x=self._x, edge_index=self._edge_index, y=self._node_label,
```
BlazStojanovic, line 47: Given that `data(self)` returns a `Data` class with an additional attribute `expl_mask`, it would make more sense to return an `Explanation` here. WDYT? 

We have created the `Explanation` class just for this purpose, to hold explanation masks (of different kinds, node mask in this case) along with other  `Data` attributes. This also makes the interface to ground truth explanation evaluation more clear (which is the purpose of these datasets anyway), e.g. we are really comparing two explanations there `gt_metics(explanation: Explanation, groundtruth: Explanation)`.

BlazStojanovic (commented):
Thank you for your amazing work @rfdavid!  👍🏻

I just have one caveat regarding the return type of benchmark datasets, I am of the opinion that it should be `Explanation` instead of `Data` (sorry for not chiming into the discussion earlier). This would make more sense because these benchmark datasets really return ground truth explanations, and it will make interfacing to evaluation of explanations more clear. Also interested what others think about this (mainly @RexYing)

test/explain/test_graph_generator.py
rfdavid: Good point. I assume it is better to remove this class and use the implemented one which is in the other branch. I'll update!

torch_geometric/datasets/explainer_dataset.py
rfdavid: For some reason, this annotation or `GraphGenerator` or `Callable` is breaking`test_config_store.py` test. I ended up removing and adding to my todo for further investigation:

```
              raise ValidationError(f"Unexpected type annotation: {type_str(ref_type)}")
E               omegaconf.errors.ValidationError: Unexpected type annotation: GraphGenerator
E                   full_key: generator
E                   object_type=ExplainerDataset
```

The annotation for `transform` doesn't raise anything on this particular test.

Do you have any idea why? Should I open an issue?

torch_geometric/datasets/generators/graph_generator.py
rfdavid: Thank you @BlazStojanovic  for bringing up those relevant points!
Since Explanation holds specific properties that Data is already returning here, as you pointed out, it makes sense. 
Is there any use case where GraphGenerator would be used to generate synthetic graphs not for explainability but for another purpose?

torch_geometric/datasets/generators/graph_generator.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class GraphGenerator:
+    r"""Base class for generating benchmark datasets. It contains
+        `generate_feature` and `attach_motif` methods used to generate
+        features and attach motifs to base graph. The motifs are
+        currently attached in random order.
+
+    Args:
+        num_nodes (int): The number of nodes used to attach the motifs.
```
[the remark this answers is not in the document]
rfdavid: yes, exactly.

torch_geometric/datasets/generators/graph_generator.py
rfdavid: This is a very good question, and I don't have an answer. I'm following the current implementation on `ba_shapes.py` dataset, which generates a base graph with 300 nodes + 80 houses (80 * 5 nodes), and assign the explain mask `True`, starting from 400.

torch_geometric/datasets/generators/graph_generator.py
BlazStojanovic: I don't thinks so, and even if there was another use, the Explanation is still a `Data` subclass, and can be used as one.

rfdavid: PR updated, thanks @BlazStojanovic

torch_geometric/datasets/explainer_dataset.py
zechengz: I see, then just leave it as TODO.

rusty1s:
Hey @rfdavid, thank you for this awesome PR. Really like it. I still made some modifications in order to separate concerns. For example, I think that `ExplainerDataset` should take care of attaching motifs, not the underlying graph generator. As such, I splitted the logic into `MotifGenerator`, `GraphGenerator`, and they all come together in the `ExplainerDataset`. I also added automatic resolving of generator names such that one can do `motif_generator="house"`. Hope the changes are okay to you.

rfdavid:
> Hey @rfdavid, thank you for this awesome PR. Really like it. I still made some modifications in order to separate concerns. For example, I think that `ExplainerDataset` should take care of attaching motifs, not the underlying graph generator. As such, I splitted the logic into `MotifGenerator`, `GraphGenerator`, and they all come together in the `ExplainerDataset`. I also added automatic resolving of generator names such that one can do `motif_generator="house"`. Hope the changes are okay to you.

Thank you, @rusty1s! That's way better. I hope my previous implementation was helpful somehow. I'll follow the pattern from your development for the next implementations.

rusty1s:
Yes, it was great :)

Issue #5958, opened 2022-11-11
[Explainability Evaluation] - Fidelity +/- metrics


BlazStojanovic:
### 🚀 The feature, motivation and pitch

Implement Fidelity +/- metrics (Fid+ and Fid-) for the purpose of evaluation explainability. This feature request is a part of broader explainability evaluation metrics (parent issue #5628) which will be crucial for the new revamp of explainability (see roadmap https://github.com/pyg-team/pytorch_geometric/issues/5520).

### Brief overview of Fidelity+/-
Fidelity measures check the explanations for their *faithfulness* to the model. This means that checking that that important inputs (nodes/edges/node features) according to the explanation are discriminative to the model. 

Formally fidelity is given with equations (from A.6 of [3])
![image](https://user-images.githubusercontent.com/22046931/201399747-4606da07-a8e3-4332-a72b-90590204f771.png)
For details and explanations see references, particularly [2-3].

### Implementation checklist
- Implement fidelity +/- for:
  - Phenomenon focus (measured with respect to ground truth labels)
  - Model focus (measured with respect to GNN outcome)
- Implement fidelity +/- for:
  - Probabilities
  - Accuracy
- Implement tests
- Implement examples

### References
1. [Explainability Methods for Graph Convolutional Neural Networks](https://openaccess.thecvf.com/content_CVPR_2019/papers/Pope_Explainability_Methods_for_Graph_Convolutional_Neural_Networks_CVPR_2019_paper.pdf)
2. [Explainability in Graph Neural Networks: A Taxonomic Survey](https://arxiv.org/abs/2012.15445)
3. [GraphFramEx: Towards Systematic Evaluation of Explainability Methods for Graph Neural Networks
](https://arxiv.org/abs/2206.09677)

AbdullahWasTaken:
`ExplainerAlgorithm` can produce `node_mask`, `node_feat_mask`, and `edge_mask`. But in case one of these is missing, what should be the default behavior? Should the mask contain all ones or all zeros?

BlazStojanovic:
@AbdullahWasTaken We’ve implemented some additional functionallity to `Explanation`, you can now use:
- `Explanation.get_explanation_subgraph()`, and
- `Explanation.get_complement_subgraph()`
to obtain the subgraphs you need!

CXX1113:
[...]

PR #6116, opened 2022-12-01
Fidelity metrics

BlazStojanovic:
## Introducing an interface for Explainability metrics [OUTDATED] 

**Due to the time constraints of the explainability sprint this PR will just implement Fidelity metrics and the Explanation method interface will be left as future work**

~~The main goal of this PR is to unify the behaviour of explainability metrics. The explainability metrics can broadly be used in two ways:~~
~~1. **To test the quality of explanation methods**, when developing new explainers. This requires measuring explanation metrics with respect to many different explanations (for many nodes, edges, subgraphs, etc.), usually at many different explanation configurations (e.g. many different explanation sparsities). In order to full-fill this need we provide PyG users with the abstract class `ExplanationMetric`.~~
~~2. **To test the quality of a single explanation**, e.g. when using explanations at inference time. For this purpose a single method suffices.~~

~~### The `ExplanationMetric` class~~

~~The proposed workflow of `ExplanationMetric` subclasses is as follows:~~
~~1. The `__init__` function specifies the range of explanations to test over, e.g. at different sparsities of masks, at different entropies, etc. Additional functionality to achieve this will be provided in the `util.py`.~~
~~2. After created, the class can be called as `explanation_metric(explainer)`, this returns the specific explanation metric computed over specified explanation settings in (1), done in (3) steps:~~
  ~~1. `ExplanationMetric.get_inputs()` generates all different inputs over which the metrics will be computed (e.g. explanation and complement graphs for all different explanation sparsities, for all nodes being explained)~~ 
  ~~2. `ExplanationMetric.compute_metrics()` computes the metrics over the inputs~~ 
  ~~3. `ExplanationMetric.aggregate()` aggregates all the evaluated metrics~~
  
~~### The `explanation_metric` method~~
~~Each explanation metric should have a corresponding `explanation_metric` method. Its signature is e.g.:~~

 ~~`explanation_metric(Explainer, Explanation, target, index)`~~

~~And it computes this metric for given explainer (as it is configured), the explanation (that explains index) it produced, given target (i.e. the target prediction with the full graph or ground truth prediction).~~

---

### Fidelity
Both implementations will be provided for Fidelity which will serve as a basis for implementing other `ExplanationMetric`s. Thus this also solves #5958.


torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+    # based on task level
+    if task_level == 'graph':
+        if index is not None:
+            ValueError(
+                f'Index for graph level task should be None, got (f{index})')
+        # evaluate whole entry
+        pass
+    elif task_level == 'edge':
+        # get edge prediction
+        pass  # TODO (blaz)
```
BlazStojanovic, line 102: Node level and graph level tasks make sense to me. For nodes we will have a clearly indexed output (each row of the prediction corresponding to nodes) and for full graphs we only have one prediction anyway. This makes it clear on what to compare to get fidelities (both in classification and regression tasks). 

But I am not certain for edge level tasks. Let say for example when obtaining the explanation for an edge level task with the explainer class, we use:
```
explanation = explainer(
        x,
        edge_index,
        target=target,
        index=index,
        edge_label_index=edge_label_index,
    )
```
What exactly does the `index` refer to, are we explaining the prediciton(s) for edge(s) with indices in `index` which correspond to edges in `edge_index`? Or something else? If so, how can we easily compare the predictions, as the `edge_index` changes for induced and complement graphs for an explanation?

BlazStojanovic (commented):
I've left an initial review with some comments where I am stuck. Would greatly appreciate some feedback so we can finish this and unblock other people in the sprint! :)

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+    if index != explanation.get('index'):
+        raise ValueError(f'Index ({index}) does not match '
+                         f'explanation.index ({explanation.index}).')
+
+    # get input graphs
+    explanation_graph = explanation.get_explanation_subgraph()  # for fid-
+    complement_graph = explanation.get_complement_subgraph()  # for fid+
+
+    # get target
+    target = explainer.get_target(x=explanation.x,
```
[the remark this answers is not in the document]
dufourc1: If I understand correctly, I think you have to pass the input that is explained with the model (i.e the one that was used to generate the explanation). I'm not sure if you can always make sure that an `Explanation` instance has a `x` or even an `edge_index` attributes. 

You would need to apply the masks of the explanation to the original input to get the `explanation_graph` and `complement_graph`, but maybe I'm missing something in what you want to do ?

torch_geometric/explain/metrics.py
dufourc1: Maybe the tests can help ? https://github.com/pyg-team/pytorch_geometric/blob/b2f9d0521f687e8f5341b998f52227540bab99fe/test/explain/algorithm/test_gnn_explainer.py#L84

I haven't taken a look at the edge level explainers in details, but it seems that for explaining the edge between node `0` and  `1` you would have `edge_label_index = torch.tensor([[0],1])`.

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+        explainer :obj:`~torch_geometric.explain.Explainer`
+            The explainer to evaluate
+        explanation :obj:`~torch_teometric.explain.Explanation`
+            The explanation to evaluate
+        target (Tensor, optional): The target prediction, if not provided it
+            is inferred from obj:`explainer`, defaults to obj:`None`
+        index (Union[int, Tensor]): The explanation target index, for node-
+            and edge- level task it signifies the nodes/edges explained
+            respectively, for graph-level tasks it is assumed to be None,
+            defaults to obj:`None`
```
wsad1, line 61:
```suggestion
        explanation :obj:`~torch_teometric.explain.Explanation`
            The explanation to evaluate
```
Do we need anything other than `explanation` now given other values are added in to `Explanation` in `Explainer`.

torch_geometric/explain/metrics.py
wsad1: @BlazStojanovic I think we should leave this as `NotImplemented` for now. Because I think `task_level=edge` needs some more work
1. We don't have an example to demonstrate `task_level=edge` on some real world data. We could have and example that explains [this](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/rgcn_link_pred.py#L20) edge prediction task. 
2. GNNExplainer needs to be updated to support `edge` level tasks, one line that needs to be updated is https://github.com/pyg-team/pytorch_geometric/blob/f622fee0d1ab71d1b2f97f9cc31d7990b45d99d9/torch_geometric/explain/algorithm/gnn_explainer.py#L116.

torch_geometric/explain/metrics.py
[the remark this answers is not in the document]
wsad1: @BlazStojanovic is this still blocking you?

PR #6118, opened 2022-12-01
integrate ToHeteroModule into ToHeteroTransformer

puririshi98:
Finally ready 🎉


torch_geometric/nn/dense/linear.py
```diff
@@ -276,6 +281,16 @@ def forward(self, x: Tensor, type_vec: Tensor) -> Tensor:
                 out[mask] = lin(x[mask])
         return out
 
+    @torch.no_grad()
+    def initialize_parameters(self, module, input):
+        if is_uninitialized_parameter(self.weight):
+            self.in_channels = max([i.size(-1) for i in input])
```
rusty1s, line 287: Can we actually test this? Why do we need the `max` here?

torch_geometric/nn/to_hetero_module.py
```diff
@@ -9,42 +9,96 @@
 from torch_geometric.typing import EdgeType, NodeType, OptTensor
 from torch_geometric.utils import scatter
 
+from .dense import HeteroLinear, Linear
```
rusty1s, line 12: Use absolute paths.

torch_geometric/nn/to_hetero_module.py
```diff
@@ -9,42 +9,96 @@
...
+
+
+class SubParam():
```
rusty1s, line 15:
```suggestion
class SubParam:
```

torch_geometric/nn/to_hetero_module.py
```diff
@@ -9,42 +9,96 @@
...
+    def set_data(self, data):
+        if self.param.data[self.idx].shape == data.t().shape:
+            self.param.data[self.idx] = data.t()
+        else:
+            self.param.data[self.idx] = data
+
+    data = property(get_data, set_data)
+
+
+class DummyLinear():
```
rusty1s, line 32:
```suggestion
class DummyLinear:
```

rusty1s, line 32: Can you clarify `DummyLinear` for me?

torch_geometric/nn/to_hetero_module.py
```diff
@@ -9,42 +9,96 @@
...
-            out_channels,
+            self.in_channels,
+            self.out_channels,
             num_types=len(types),
             is_sorted=True,
-            bias=bias,
+            bias=self.bias,
         )
 
+    def __getitem__(
```
rusty1s, line 90: This seems more like a hack to move this to the current structure. Wouldn't it make more sense to not expect working with dictionaries now whenever we see a `Linear` layer?

torch_geometric/nn/to_hetero_transformer.py
```diff
@@ -286,27 +289,47 @@ def call_global_pooling_module(self, node: Node, target: Any, name: str):
...
-
-        # Add calls to node type-wise or edge type-wise methods.
         self.graph.inserting_after(node)
+        if torch_geometric.typing.WITH_PYG_LIB:
+            # Addresses:
+            # "RuntimeError: Output 0 of SplitWithSizesBackward0
+            # is a view and is being modified inplace."
+            # Cause:
+            # split_with_sizes for pyg-lib.ops.segment_matmul returns a view.
+            # Using relu_ or other in place on this triggers the RuntimeError.
```
rusty1s, line 325: Then let's just not use `relu_`.

puririshi98: yes thats what my code does. if there is a relu_(or any in place will have the same issue) in the original homogeneous model then I replace it with the equivalent out of place in the final. this is needed to solve that RuntimeError. had mentioned in slack which is at the top of the comment. In general I believe my PR should accelerate to_hetero with zero code changes from the user needed or any changes in user experience and object behavior. I have put in a great deal of effort to make sure this is the case.

torch_geometric/nn/dense/linear.py
puririshi98: sorry u are correct i was working late into the evening and forgot for a sec that this heterolinear is taking in a single tensor input since the toheterolinear i was working on is taking dict. lol just a brain fry from late nights 😝

torch_geometric/nn/to_hetero_module.py
puririshi98: it is definitely a pythonic hack. took quite a bit of wrestling to get it working but my general belief is that my PR should accelerate to_hetero with zero code changes from the user needed or any changes in user experience and object behavior. I believe this kind of behavior you have set in the test is a valid thing to expect from a user POV and this python setup works.

torch_geometric/nn/to_hetero_module.py
puririshi98: it exists as a  pythonic hack to support this usage from `test_to_hetero_and_rgcn_equal_output`:
```
        model.lin[node_type].weight.data = conv.root.data.t()
        model.lin[node_type].bias.data = conv.bias.data
```
Since we have a 3D parameter in the pyg-lib heterolinear instead of a list of Linear modules, its very hard to access the weight and bias of each node type in the way done in the test but I believe a user should still be able to do this.

puririshi98:
there will be some changes in my HGT branch that ill need to rebase this onto once my HGT PR goes into main

puririshi98:
closing this since i rebased onto main and also going to add heteronorms

PR #6123, opened 2022-12-02
Faster `topk` implementation for large enough graphs

mova:
Topk needs quite some time to construct the mask in my application.
I noticed I can avoid the loop, if all of the graphs in the batch are large than `ratio`.


rusty1s (approved):
This is great, thank you!

Issue #5962, opened 2022-11-12
[Explainability Evaluation] - Explanability metrics for when ground truth is available


BlazStojanovic:
### 🚀 The feature, motivation and pitch

Implement AUC, accuracy, precision, recall metrics, which will be useful for when the groundtruth is available. This feature request is a part of broader explainability evaluation metrics (parent issue https://github.com/pyg-team/pytorch_geometric/issues/5628) which will be crucial for the new revamp of explainability (see roadmap https://github.com/pyg-team/pytorch_geometric/issues/5520).

### Implementation overview
The metric API could be `groundtruth_eval(explanation, groundtruth), where explanation is the `Explanation` class in #5804 and groundtruth is the groundtruth explanation (e.g. the motif structure in #5817). This may require hard (tresholded explanation) for now, but can potentially extend to soft explanations in the future. 

### Implementation checklist
* Accuracy,
* Precision,
* Recall,
* AUC

PR #6137, opened 2022-12-05
[Explainability Evaluation] Add accuracy metrics for evaluation with groundtruth

shhs29:
This is a PR for the issue [https://github.com/pyg-team/pytorch_geometric/issues/5962](https://github.com/pyg-team/pytorch_geometric/issues/5962). 
Accuracy metrics such as accuracy, recall, precision , auc and f1_score are available. 

TODOs 
- [x] Add docstring as needed.
- [x] Update CHANGELOG.md
- [x] Fix failing import of torchmetrics.


BlazStojanovic (commented):
Thank you @shhs29 for this PR! Here are a few high level comments:

- Please move the contents of `explain/evaluate/accuracy_metrics.py` into `explain/metrics.py`
- Let's try to refrain from using networkx in computing these metrics, given that you can access the masks directly from the `Explanation`, using `torch` for most of your computations will be much more efficient. 
- Moreover, try to use `torchmetrics` to evaluate ROC

**Other things to help you out:**
You can add this `masks` method to the `Explanation` class:
```
    @property
    def masks(self) -> dict[str, Tensor]:
        r"""Returns a dictionary of all masks available in the explanation"""
        mask_dict = {key:self[key] for key in self. keys if key.endswith('_mask') and self[key] is not None}
        return dict(sorted(mask_dict.items()))
```
This will allow you to access all the available masks in the explanation as
```
explanation.masks
```
then you can combine all the masks in the returned dictionary in to a single tensor with `torch.view(-1)` and `torch.cat`. This will make it easy to work with all masks at once, afterwards you can calculate number of true positives as
```
torch.sum(gt_mask_tensor == ex_mask_tensor)
```
and similarly for other metrics. Also with these improvements you should be left with considerably less code, so no need to split the function into two parts, lets just have `grountruth_metrics(explanation: Explanation, groundtruth: Explanation)`

shhs29:
Hi @BlazStojanovic,
Thanks a lot for the detailed comments. @venomouscyanide and I will work on these and update the PR soon.

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,46 @@
...
+import torch
+from torchmetrics import AUROC, ROC
+
+from torch_geometric.explain import Explanation
+
+
+def groundtruth_metrics(
+        explanation: Explanation,
+        groundtruth: Explanation) -> Tuple[float, float, float, float, float]:
+    """accuracy_scores: Compute accuracy scores when
```
BlazStojanovic, line 12: Describe thresholding behaviour in the docstring, i.e. *>0 thresholding* for TP, and FP.

BlazStojanovic: Also explain the order of returned metrics

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,46 @@
...
+    tp = torch.sum(gt_mask_tensor[correct_preds])
+    tn = torch.sum(correct_preds) - tp
+    fp = torch.sum(ex_mask_tensor[incorrect_preds])
+    fn = torch.sum(gt_mask_tensor[incorrect_preds])
+
+    precision = tp / (tp + fp)
+    accuracy = (tp + tn) / (tp + fp + tn + fn)
+    recall = tp / (tp + fn)
+    f1_score = 2 * (precision * recall) / (precision + recall)
+    roc = ROC(task="binary")
```
BlazStojanovic, line 41: Don't you need to do this before you threshold out the `ex_mask_tensor`? Because here `roc` receives a binary tensor, which it cannot threshold?

BlazStojanovic (commented):
Thank you @shhs29 and @venomouscyanide, this is looking much better. Left some more comments for you to address, otherwise we are close to closing this issue :)

test/explain/test_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+        node_feat_mask=gt_node_feat_mask,
+        edge_feat_mask=gt_edge_feat_mask,
+    )
+
+    accuracy_metrics = groundtruth_metrics(explanation, groundtruth)
+    assert accuracy_metrics[0] == 1.0
+    assert accuracy_metrics[1] == 1.0
+    assert accuracy_metrics[2] == 1.0
+    assert accuracy_metrics[3] == 1.0
+    assert accuracy_metrics[4] == 0.5
```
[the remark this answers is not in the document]
shhs29: Yes, you are right. Random explanations is not a good way to setup tests. I have updated the test to have hardcoded mask values. A few questions that I had in mind was: 1.

torch_geometric/explain/metrics.py
shhs29: Yes. That's right. I have updated the implementation to use original ``ex_mask_tensor`` in ROC.

test/explain/test_metrics.py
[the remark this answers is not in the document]
BlazStojanovic: 1. Great, using hardcoded masks here is a better way to test. Just a note when writing such tests in the future, always make sure to verify outcomes of hardcoded tests independently (i.e. by hand or other library)
2. Probably returning `Tensor`s is better, torch functions return them anyway so you don't need to change anything but the type hint
3. I think we shouldn't round anything, let the user do this if they so choose afterwards
4. Yes, this is a very good point! We need to handle edge cases, let's handle them in a way consistent with `sklearn`
  - when `true positive + false positive == 0`, precision returns 0
  - when `true positive + false negative == 0`, recall returns 0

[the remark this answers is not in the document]
shhs29: Thanks for the detailed comment. The status of each of these items are as follows:

1. I have verified all the values except for auroc which I need some clarification on.
2. I have updated the typehint.
3. I am currently not rounding the values.
4. I have added a couple of new conditions and added a new test for this flow. In addition to the conditions you mentioned, I added one more condition for f1_score where f1_score is 0 when `precision == 0.0 or recall == 0.0`.

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,69 @@
...
+        explanation: Explanation, groundtruth: Explanation
+) -> Tuple[Tensor, Tensor, Tensor, Tensor, Tensor]:
+    """Computes different accuracy metrics when groundtruth is available.
+
+    .. note::
+        Currently we perform hard thresholding (where the threshold value
+        is set to 0) on explanation and groundtruth masks to get true
+        positives, true negatives, false positives and false negatives.
+        I.e., all values in explanation masks and ground truth masks which
+        are greater than 0 is set to 1.
```
BlazStojanovic, line 20: I think it might actually be better to have an additional argument, `threshold=0.0` which defaults to 0.0, but can be set by the user to be the thresholding of both masks. This docstring will then describe the default behavior.

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,69 @@
...
+    ex_masks = explanation.masks
+    ex_mask_tensor = torch.cat(
+        list(map(lambda x: x.view(-1), ex_masks.values())))
+    gt_masks = groundtruth.masks
+    gt_mask_tensor = torch.cat(
+        list(map(lambda x: x.view(-1), gt_masks.values())))
+    threshold = 0
+    gt_mask_tensor[gt_mask_tensor > threshold] = 1.0
+    roc = ROC(task="binary")
+    fpr, tpr, thresholds = roc(ex_mask_tensor, gt_mask_tensor)
```
BlazStojanovic, line 41: You're right, we don't need the roc calculation for auroc. I think we have two options here, We can also return the full ROC curve, or we can remove these two lines and just return the AUROC, I leave the choice up to you.

BlazStojanovic (commented):
Just left two more minor comments, this is looking very good and once you incorporate these last two, we can approve and merge this! 👍🏻

venomouscyanide:
@BlazStojanovic Struggling a bit with reStructuredText. Can you please share some best practices regarding .rst files and how to make sure the formatting is correct?

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,78 @@
...
+from torch_geometric.explain import Explanation
+
+
+def get_groundtruth_metrics(
+    explanation: Explanation,
+    groundtruth: Explanation,
+    threshold: float = 0.0,
+) -> Tuple[Tensor, Tensor, Tensor, Tensor, Tensor]:
+    r"""Returns different accuracy metrics on explanation when
+    groundtruth is available.
```
BlazStojanovic, line 16:
```suggestion
    r"""Compares an explanation with the ground truth explanation. Returns basic evaluation metrics - accuracy, recall, precision, f1_score, and aucroc.
```

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+    .. note::
+        Currently we perform hard thresholding (where the threshold value
+        defaults to 0) on explanation and groundtruth masks to get true
+        positives, true negatives, false positives and false negatives.
+        I.e., all values in explanation masks and ground truth masks which
+        are greater than the threshold value is set to 1.
```
BlazStojanovic, line 23: You can remove this note

torch_geometric/explain/metrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Args:
+        explanation (Explanation): Explanation output generated by an
+            explainer.
+        groundtruth (Explanation): Groundtruth for explanation.
+        threshold (float): threshold value to perform hard thresholding.
+            (default: :obj:`0.0`)
```
BlazStojanovic, line 30:
```suggestion
        threshold (float): threshold value to perform hard thresholding of the `explanation` and `groundtruth` masks.
            (default: :obj:`0.0`)
```

torch_geometric/explain/metrics.py
shhs29: ROC curve gives three values: `fpr`, `tpr`, `thresholds`, whereas all other metrics return a single value. To maintain consistency, I have decided to calculate AUROC and return that. Moreover, with the inclusion of ROC curve, our return value will have nested values, in which case I believe returning a dict makes more sense.

BlazStojanovic (approved):
Thanks @shhs29 for including all the comments. This is looking good, so I think we can move towards merging this with other explainability code @rusty1s!

shhs29:
@BlazStojanovic @rusty1s For the fidelity metric in main now, the metrics are under a new folder. We have a different structure. Is that fine ?

rusty1s:
Yes, don't worry about it :)

rusty1s (approved):

PR #6141, opened 2022-12-05
Added `Dataset.to_datapipe` for converting PyG datasets into a torchdata `DataPipe`

hatemhelal:
There is some overlap in functionality in the torch DataPipe API with features that are already available in PyG's dataset implementations:

* downloading from remote filesystems
* on-disk and in-memory caching
* handling extraction of data from compressed files
* etc

This patch adds a `@classmethod` onto  the PyG `Dataset` interface to support converting any PyG dataset into a DataPipe. 

```python
from torch_geometric.datasets import QM9

dp = QM9.to_datapipe(root='./data/QM9/')
dp = dp.batch_graphs(batch_size=2, drop_last=True)

for batch in dp:
    pass
```

Todo:

- [x] decide on the right base class for `DatasetAdapter` : should it be `MapDataPipe` or is `IterDataPipe`
- [x] add a lazy `len` method 
- [x] unittests


torch_geometric/data/dataset.py
```diff
@@ -308,6 +308,12 @@ def print_summary(self):
         r"""Prints summary statistics of the dataset to the console."""
         return str(self.get_summary())
 
+    @classmethod
+    def to_datapipe(cls, **kwargs):
```
rusty1s, line 312: Can we add some documentation here?

hatemhelal: Indeed, still fleshing out a few things

torch_geometric/data/datapipes.py
```diff
@@ -64,6 +73,21 @@ def __iter__(self):
             yield data
 
 
+class DatasetAdapter(IterDataPipe):
```
hatemhelal, line 76: I'm not sure if it would make more sense to have this inherit from `MapDataPipe` since that is more natural for PyG's Dataset implementations.

Would also like to somehow lazily initialise the `len` to support sharding.

hatemhelal:
> Todo:
> 
> * [x]  decide on the right base class for `DatasetAdapter` : should it be `MapDataPipe` or is `IterDataPipe`
- `IterDataPipe` is preferred since there are more pipe implementations available for the iterable version (including PyG's batcher)
> * [x]  add a lazy `len` method
- decided to capture the length of the dataset when constructing the pipe adapter and using that.
> * [x]  unittests
- added a basic test in `test_dataset.py`

rusty1s (approved):
Sorry for taking so long to merge this. I changed the `to_datapipe` method to an instance method - I think this is a bit more intuitive.

hatemhelal:
Thank you for merging this!

> I changed the `to_datapipe` method to an instance method - I think this is a bit more intuitive.

I agree it is a bit more intuitive as a method but I will need to double check the behaviour with multiprocessing.  I think I wanted to avoid having the dataset as a class property of the pipe since I recall this as being more efficient when using multiprocessing since the dataset can be lazily initialised within each worker....at least in theory.  Will check the behaviour with the latest pyg-nightly.

rusty1s:
What's the benefit of initializing the dataset in each worker rather than making use of shared memory?

Sorry I went ahead with this. If it does not fit your needs, please feel free to adjust this :(

PR #6144, opened 2022-12-05
Add the unsupervised bipartite GraphSAGE model on the Taobao dataset

husimplicity:
This PR adds an implementation of unsupervised bipartite GraphSAGE on the Taobao User Behaviors dataset offered by Alibaba. The Taobao dataset contains a heterogeneous graph, where nodes represent users and items, and edges represent different types of behaviors between users and items. [](https://tianchi.aliyun.com/dataset/649)
We use the i2i co-occurrence matrix to construct the i2i-graph. When applying GraphSAGE, the model follows the i-i-i pattern to encode the item embedding and the i-i-u pattern to encode the user embedding. As the task is unsupervised, we use negative sampling and `binary_cross_entropy_with_logits` to compute the loss in the model.


rusty1s (approved):
Thank you!

PR #6155, opened 2022-12-06
Use `sampled_sub` for faster `utils.softmax`

rusty1s:
Also refactors `WITH_PYG_LIB` within TorchScript modules.


ZeynepP:
Final qualifiers are available since Python 3.8. This makes the package not compatible anymore with python 3.7.

rusty1s:
Thanks for letting me know. Will fix soon.

PR #6159, opened 2022-12-06
Correctly use edge weights in `GDC` example

dongyukang1:
If you want to use GDC,you will find accuracy drop,because T.GDC( ) save the calculated edge weight in data.edge_ attr,not data.edge_weight.

I directly change 'model(data.x, data.edge_index, data.edge_weight)' to 'model(data.x, data.edge_index, data.edge_attr)' in function train() and test().
When data.edge_weight is None ,like Cora,it's right.


rusty1s (approved):
Thanks for this catch :(

PR #6164, opened 2022-12-07
Infer correct CUDA device ID in `profileit` decorator

venomouscyanide:


torch_geometric/profile/profile.py
```diff
@@ -62,8 +62,22 @@ def wrapper(*args, **kwargs) -> Tuple[Any, Stats]:
                 raise AttributeError(
                     'First argument for profiling needs to be torch.nn.Module')
 
+            device = None
+            for arg in list(args) + list(kwargs.values()):
+                if isinstance(arg, torch.Tensor):
+                    device = arg.get_device()
+                    break
+            if not device:
```
rusty1s, line 70:
```suggestion
            if device is None:
```

rusty1s (approved):
This is great, thank you!

PR #6179, opened 2022-12-08
Custom motif generation

cuent:
This task is part of [GNN Explainability Dataset Generation](https://github.com/pyg-team/pytorch_geometric/issues/5817)

Implementation of the base class for motif generation. It depends on [the framework and API to generate benchmark datasets](https://github.com/pyg-team/pytorch_geometric/pull/6104).

The base class follows some of the structure from @rfdavid PRs. However, I changed a bit since I believe using Data is a better and cleaner approach to generate structures and create wrappers for other structures in PyG or NetworkX.

Once I have some feedback about the `MotifGenerator`, I will go ahead and 

- [x] update the `GraphGenerator.attach_motif()`
- [x] add tests 
- [x] documentation.


torch_geometric/datasets/generators/motif.py
```diff
@@ -0,0 +1,60 @@
...
+                    [1, 3, 4, 4, 2, 0, 1, 3, 2, 0, 0, 1],
+                ]),
+                y=torch.tensor([1, 1, 2, 2, 3]),
+            )
+        elif isinstance(self.structure, Graph):
+            return from_networkx(self.structure)
+        elif isinstance(self.structure, Data):
+            return Data
+        else:
+            return None
```
BlazStojanovic, line 60: Here I suggest we instead raise a `ValueError` surfacing that we only support documented motifs, networkx graphs and `Data`.

BlazStojanovic (approved):

rusty1s (approved):
Thank you! Aligned the code to the new interface :)

Issue #6182, opened 2022-12-08
Possible incorrect type hint in DataLoader


edwag:
### 🐛 Describe the bug

The documentation for `torch_geometric.loader.DataLoader` says that "Data objects can be either of type `torch_geometric.data.Data` or `torch_geometric.data.HeteroData`". But if you pass a `List[Data]` or a `List[HeteroData]`  to `DataLoader`, your type checker will complain. 

For example, Pylance complains about the final line of the following code (when configured with `"python.analysis.typeCheckingMode": "basic"`):

```python
from torch_geometric.data import Data
from torch_geometric.loader import DataLoader

dataset = [Data()]
loader = DataLoader(dataset)
```

The type hint for the `dataset` argument says that it has to be of type `Union[Dataset, List[BaseData]]` (this is line 63 of `torch_geometric/loader/dataloader.py`). The problem here is that the generic type `List` is invariant, so a `List[Data]` is _not_ a `List[BaseData]`.

I suspect that the correct fix here is to change the type hint on `dataset` to use a _covariant_ type, probably `typing.Sequence`. `Sequence` certainly satisfies the type checker, since a `Sequence[Data]` _is_ a `Sequence[BaseData]`. A `Sequence` is supposed to be immutable, so this is definitely _not_ the right fix if the `dataset` list needs to be mutated, but I don't think it does?

### Environment

Pylance version: 2022.11.40
VSCode version: 1.73.1
Python version: 3.9
PyG version: 2.2.0

PR #6183, opened 2022-12-08
Fix type hint in DataLoader

edwag:
Move from the invariant `List` generic to the covariant `Sequence` generic, to support passing a `List[Data]` or a `List[HeteroData]` to `DataLoader` without failing type checking.

Proposed fix for #6182

### DIFF
--- a/torch_geometric/loader/dataloader.py
+++ b/torch_geometric/loader/dataloader.py
@@ -1,5 +1,5 @@
-from collections.abc import Mapping, Sequence
-from typing import List, Optional, Union
+from collections.abc import Mapping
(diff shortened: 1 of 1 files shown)


torch_geometric/loader/dataloader.py
```diff
@@ -1,5 +1,5 @@
-from collections.abc import Mapping, Sequence
-from typing import List, Optional, Union
+from collections.abc import Mapping
+from typing import List, Optional, Sequence, Union
```
edwag, line 2: We need to use `typing.Sequence` instead of `collections.abc.Sequence` for compatibility with Python<=3.8.

Issue #5960, opened 2022-11-12
[Explainability Evaluation] - Characterization score, and other ways to combine Fid+/- scores


BlazStojanovic:
### 🚀 The feature, motivation and pitch

Implement functionality to combine Fidelity +/- metrics (https://github.com/pyg-team/pytorch_geometric/issues/5958) features for the purpose of evaluation explainability. This feature request is a part of broader explainability evaluation metrics (parent issue #5628) which will be crucial for the new revamp of explainability (see roadmap https://github.com/pyg-team/pytorch_geometric/issues/5520).

### Brief overview
Fidelity measures check the explanations for their *faithfulness* to the model. In [1] there are several ways proposed for combining both fidelity scores, like 

**The area under the** $\mathrm{Fid}+/(1-\text { Fid-) }$ **curve** - **AUC**

**The characterization score**: $\operatorname{charact} = \frac{w_{+}+w_{-}}{\frac{w_{+}}{f i d_{+}}+\frac{w_{-}}{1-f i d_{-}}}$

### Implementation checklist
- Implement the characterization score
- Implement the AUC

### References
1. [GraphFramEx: Towards Systematic Evaluation of Explainability Methods for Graph Neural Networks
](https://arxiv.org/abs/2206.09677)

PR #6188, opened 2022-12-08
[Explainability Evaluation] - Characterization score, and other ways to combine Fid+/- scores

seemanne:
This PR implements the features from #5960 

The tests and the math part is done, I'm somewhat lost with the sphinx documentation and very much open to suggestions for the docstrings. 

As #5962 also requires calculating an AUC and probably a weighted harmonic mean for F_1 type scores these parts have been implemented as separate helper functions.


torch_geometric/explain/utils/metrics.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from torch import Tensor
+
+
+def get_characterization_score(weights: Tensor,
+                               fidelities: Tensor) -> 'Tensor':
+    r"""Returns the componentwise characterization score of
+    fidelities[0] and fidelities[1]
```
RexYing, line 8: Would be nice to add here a bit more detail: do the two rows corresponding to fid+ and fid-? Which one is which?
And does the column corresponds to num nodes (explanation)?

torch_geometric/explain/utils/metrics.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from torch import Tensor
+
+
+def get_characterization_score(weights: Tensor,
```
RexYing, line 5: we can probably set a default weight here (1 : 1), since most of the time it's what people are going to use, unless there's prior knowledge.

torch_geometric/explain/utils/metrics.py
```diff
@@ -0,0 +1,63 @@
...
+                               fidelities: Tensor) -> 'Tensor':
+    r"""Returns the componentwise characterization score of
+    fidelities[0] and fidelities[1]
+
+
+    Args:
+        weights (Tensor): Tensor containing the two weights
```
RexYing, line 12: would be nice to include the formula for characterization and say what the weights correspond to?
Also we should have a dimension for weights. I'm ok with a pair of 2 numbers? It's ok as long as we document it.

RexYing (commented):
Thanks! A few comments on clarity.

torch_geometric/explain/utils/metrics.py
seemanne: I have added explanation to which row corresponds to what.
As for the columns I'm unsure but shouldn't it be the fidelities of different explanations? As far as I understand it the two fidelities are computed as the sum average over all nodes and serve as a way to gauge the "explanatory power" of different explanations, or am I mistaken here?

torch_geometric/explain/utils/metrics.py
seemanne: Makes sense, added :+1:

torch_geometric/explain/utils/metrics.py
seemanne: Added both, still unsure the latex compiles directly as I'm unable to get sphinx to build a link to the docs (require some help here for this)

torch_geometric/explain/utils/metrics.py
```diff
@@ -0,0 +1,83 @@
...
+def get_characterization_score(fidelity: Tensor,
+                               weights: Tensor = torch.tensor([1, 1])) \
+        -> 'Tensor':
+    r"""Returns the componentwise characterization score where fidelity[0]
+    is the tensor containing F_{+} and fidelity[1] is the tensor
+    containing F_{-}.
+
+    The characterization score is calculated as
+     .. math::
+      charac = \frac{w_+ + w_-}{\frac{w_+}{fid_+}+\frac{w_-}{fid_-}}
```
BlazStojanovic, line 14: I think there might be a typo in the formula here
```suggestion
      charac = \frac{w_+ + w_-}{\frac{w_+}{fid_+}+\frac{w_-}{1-fid_-}}
```

BlazStojanovic (approved):
This is great @seemanne, thank you! I'm not sure about the docstrings, but once we get confirmation that they compile, we can merge this 👍🏻

seemanne:
@BlazStojanovic The docs compile now, I've added a section about these evaluation metrics in `explain.rst`. Marked the branch as ready.

rusty1s:
Super, will take a final look tomorrow.

rusty1s (approved):

PR #6198, opened 2022-12-09
Restore correct order in the output of `HeteroLinear` in case of `is_sorted=False`

toenshoff:
If "is_sorted=False", the output of the forward pass will be unsorted to the original vertex order of the input.


rusty1s (approved):
Wow, this is a great catch. Thank you! What a bad mistake on our end :(

PR #6201, opened 2022-12-10
New utility function to get output of intermediate `MessagePassing` layer

wsad1:
Added `intermediate_mp_output` which will be used to get the output of some intermediate message passing layer in a `model`. 
This utility is needed to implement `PGExplainer` algorithm. Which uses the output of the last message passing layer to compute edge masks.


torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
+from typing import Callable
+
+import torch
+from torch import Tensor
+
+from .subgraph import get_num_hops
+
+
+def intermediate_mp_output(i: int, model: torch.nn.Module, **kwargs) -> Tensor:
```
rusty1s, line 9: I think it would be cleaner if you just return the outputs of all `MessagePassing` layers. Then you can simply do
```python
hs = get_intermediate_message_passing_embeddings(model, *args, **kwargs)
h = hs[-1]
```
Also makes implementation somewhat easier. WDYT?

rusty1s:
@wsad1 Do we desperately need this as a `utils`? Why not add as a helper function to `PGExplainer`?

wsad1:
@rusty1s it might be useful in other scenarios too, so it makes sense to keep it in `utils`.

torch_geometric/utils/get_mp_out.py
wsad1: Good idea. Done.

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
...
+    """Get output of all `MessagePassing` layers in
+    :obj:`model`.
+
+    Args:
+        **kwargs: Input to :obj:`model`.
+    """
+    from torch_geometric.nn import MessagePassing
+    intermediate_out = {}
+
+    def get_intermediate_output(name) -> Callable:
```
rusty1s, line 20: I am not totally convinced we need a nested function here. Why not make `intermediate_out` a list and just append the embeddings?

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
...
+        def hook(model, input, output):
+            intermediate_out[name] = output
+
+        return hook
+
+    num_mp_layers = get_num_hops(model)
+    if num_mp_layers == 0:
+        raise ValueError("'model' does not have message passing layer.")
```
rusty1s, line 28: No need to run this here. Just check for `len(hook_handles)` later.

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
...
+    for module in model.modules():
+        if isinstance(module, MessagePassing):
+            hook_handles.append(
+                module.register_forward_hook(get_intermediate_output(index)))
+            index += 1
+
+    # Run forward pass
+    model_state = model.training
+    model.eval()
+    _ = model(**kwargs)
```
rusty1s, line 42:
```suggestion
    model(**kwargs)
```

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
+from typing import Callable, List
+
+import torch
+from torch import Tensor
+
+from .subgraph import get_num_hops
+
+
+def get_intermediate_messagepassing_embeddings(model: torch.nn.Module,
```
rusty1s, line 9:
```suggestion
def get_message_passing_embeddings(model: torch.nn.Module,
```
Would be in favor of making this function name a little bit smaller :)

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,49 @@
...
+                                               **kwargs) -> List[Tensor]:
+    """Get output of all `MessagePassing` layers in
```
rusty1s, line 11:
```suggestion
    """Returns the output embeddings of all `MessagePassing` layers in
```

torch_geometric/utils/get_mp_out.py
rusty1s, line 42: Run with `torch.no_grad()`.

torch_geometric/utils/get_mp_out.py
wsad1: Yeah made `intermediate_out` a list.

test/utils/test_get_mp_out.py
```diff
@@ -0,0 +1,42 @@
...
+    with pytest.warns(UserWarning,
+                      match="does not have message passing layer"):
+        intermediate_outs = get_messagepassing_embeddings(
+            torch.nn.Linear(5, 5), x=x)
+    assert len(intermediate_outs) == 0
+
+    intermediate_outs = get_messagepassing_embeddings(model, x=x,
+                                                      edge_index=edge_index)
+    assert len(intermediate_outs) == 3
+    for index, intermediate_out in enumerate(intermediate_outs):
```
rusty1s, line 40:
```suggestion
    for expected, out in zip(expected_outs, outs):
```
reads nicer.

rusty1s (approved):
Looks good to me after comments have been resolved.

torch_geometric/utils/get_mp_out.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Args:
+        **kwargs: Input to :obj:`model`.
+    """
+    from torch_geometric.nn import MessagePassing
+    intermediate_out = []
+
+    def get_intermediate_output() -> Callable:
+        def hook(model, input, output):
+            intermediate_out.append(output)
```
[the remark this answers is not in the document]
wsad1: Good point.

PR #6204, opened 2022-12-10
`Explain`: Add `PGExplainer`

wsad1:
This PR makes the following changes.
1. Adds `PGExplainer` a new explainer algorithm. [Paper](https://arxiv.org/abs/2011.04573) , [Reference Implementation](https://openreview.net/forum?id=8JHrucviUf)
2. Updated the `Explainer` and `ExplainerAlgorithm` interface to support `PGExplainer`. Since `PGExplainer` uses a mlp to explain predictions that mlp needs to be trained before we call `Explainer.forward`.
The usage of `PGExplainer` would look like:
```
  explainer = Explainer(
      model=model,
      algorithm=PGExplainer(epochs=2),
      explanation_type=explanation_type,
      node_mask_type=None, # pg explainer only generates edge masks
      edge_mask_type=edge_mask_type,
      model_config=model_config,)
explainer.train_explainer_algorithm(x, edge_index , target)
# Forward with throw an error if 'train_explainer_algorithm' is not called
explanation = explainer.forward(x...)
```
Open to feedback on updates to the interface.

TODOs 
1. Add example in follow up PR.


torch_geometric/explain/algorithm/pg_explainer.py
```diff
@@ -0,0 +1,276 @@
...
+            task_level == ModelTaskLevel.graph) else 3 * out_channels
+        self.explainer_model = nn.Sequential(
+            nn.Linear(self.exp_in_channels, 64),
+            nn.ReLU(),
+            nn.Linear(64, 1),
+        )
+
+        assert x.shape[0] == z.shape[0]
+        batch = kwargs.get('batch', None)
+        assert ~(batch is None and index is None)
```
dufourc1, line 97: I'm not sure if I understand why we always need a `batch`: what happens if I want to explain a single graph for a graph regression problem ? 

If we always require a `batch`, should we add it as an explicit parameter ?

torch_geometric/explain/algorithm/pg_explainer.py
```diff
@@ -0,0 +1,276 @@
...
+        size_loss = mask.sum() * self.coeffs['edge_size']
+        mask_ent_reg = -mask * mask.log() - (1 - mask) * (1 - mask).log()
+        mask_ent_loss = mask_ent_reg.mean() if batch is None else scatter_mean(
+            mask_ent_reg, batch[edge_index[0]]).sum()
+        mask_ent_loss *= self.coeffs['edge_ent']
+
+        return loss + size_loss + mask_ent_loss
+
+    @staticmethod
+    def _drop_isolated_nodes(node_idxs: Tensor, edge_index: Tensor) -> Tensor:
```
dufourc1, line 227: This could be useful for other explainers, maybe we can move it to `utils` ?

torch_geometric/explain/algorithm/pg_explainer.py
```diff
@@ -0,0 +1,276 @@
...
+        is_hetero = self._is_hetero
+        if is_hetero:
+            logging.error("Heterogeneous graphs not supported.")
+            return False
+
+        return True
+
+    def train_explainer(self, model: torch.nn.Module, x: Tensor,
+                        edge_index: Tensor, target: Tensor = None,
+                        index: Tensor = None, **kwargs):
```
dufourc1, line 80: Just a quick question for the interface update: does the `PGExplainer` need to be trained for every input like the `GNNExplainer` or is the training done once and then the user can explain multiple inputs ? If we are in the first case, we could do the training in the `PGExplainer.forward` method ? 

From here it seems like the training has to be done for each input ($G$,$y$) we want to explain, but I'm not sure if I understand correctly.

torch_geometric/explain/algorithm/pg_explainer.py
```diff
@@ -0,0 +1,276 @@
...
+        **kwargs (optional): Additional hyper-parameters to override default
+            settings in
+            :attr:`~torch_geometric.explain.algorithm.PGExplainer.coeffs`.
+    """
+
+    coeffs = {
+        'edge_size': 0.05,
+        'edge_ent': 1.0,
+        'temp': [5.0, 2.0],
+        'bias': 0
```
dufourc1, line 42: Should we add here the temperature and bias of L188 ?

torch_geometric/explain/algorithm/pg_explainer.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            return torch.cat([z_i, z_j], 1)
+
+    def _compute_edge_mask(self, edge_weight, temperature=1.0, bias=0.0,
+                           training=True):
+
+        if training:  # noise is added to edge_weight.
+            bias += 0.0001
+            eps = (bias -
+                   (1 - bias)) * torch.rand(edge_weight.size()) + (1 - bias)
```
dufourc1, line 194:
```suggestion
            eps = (2*bias -1) * torch.rand(edge_weight.size()) + (1 - bias)
```

dufourc1 (commented):
Thanks for the work ! 

I added a few preliminary comments, but I will have to reread the paper in more details.

I have concerns about the change in interface for the `Explainer` class:
 - Do we need to expose the training of the explainer algorithm to the `Explainer` (unless we are using a surrogate model) ? 
 - If we do, can we hide it from the user by calling it immediately when `ExplainerAlgorithm.forward` is called (like in the `GNNExplainer` case)

What do you think ?

torch_geometric/explain/algorithm/pg_explainer.py
wsad1: In `PGExplainer` the training is done once and the user can explain multiple inputs with that. So its the second case, which is also the reason might need this interface change. Another thing we could do is have the user provide the training data while initializing `PGExplainer`.

torch_geometric/explain/algorithm/pg_explainer.py
wsad1: The function `compute_edge_mask` gets called in L123, the inputs to the function are set before that using these coefficients.

torch_geometric/explain/algorithm/pg_explainer.py
wsad1: Good point.

torch_geometric/explain/algorithm/pg_explainer.py
wsad1: Good point. Removed requirement for `batch`.

rusty1s (approved):
Thanks @wsad1. I made some slight change by moving the training loop outside of the explainer. I feel this allows the user more control, as the model might be training on a variety of graphs. Hope the changes are okay. Feel free to revise them if you feel they don't make sense.

PR #6215, opened 2022-12-12
`Explainer` and `Explanation` updates

BlazStojanovic:
Some updates which make explainer and explanation classes more easily usable:

- [x] Add `masks` property to `Explanation`, which returns all the non `None` masks in the explanation 
- [ ] Index between old graph node ids and newly reindexed nodes in the returned subgraphs (`get_explanation_subgraph()` and `get_complement_subgraph()`)
- [ ] Make it easy to get model predictions on the explanation induced subgraphs
- [x] Add `index` to `Explanation`, so now explanation carries information about which `indices` it is explaining
- [x] Add thresholding directly to `Explanation`
- [x] Move thresholding functionality into `util.py`
- [x] Add option to get `classification` probabilities and not only predicted classes
- [x] `get_target` method in `Explainer`


torch_geometric/explain/explainer.py
```diff
@@ -46,7 +44,8 @@ def __init__(
         self.algorithm.connect(self.explainer_config, self.model_config)
 
     @torch.no_grad()
-    def get_prediction(self, *args, **kwargs) -> torch.Tensor:
+    def get_prediction(self, *args, raw: bool = False,
+                       **kwargs) -> torch.Tensor:
```
rusty1s, line 48: Can you clarify why we need this? If we want to return raw predictions, we should apply the `argmax` in the loss function.

torch_geometric/explain/explainer.py
```diff
@@ -159,47 +181,6 @@ def _threshold(self, explanation: Explanation) -> Explanation:
...
-
-        else:
-            raise NotImplementedError
-
-        # Update the explanation with the thresholded masks:
-        for key, mask in mask_dict.items():
-            explanation[key] = mask
-
-        return explanation
+        return explanation.threshold(
```
rusty1s, line 184: Wouldn't we want to drop `threshold_config` in `Explainer` completely and move it to `Explanation`?

torch_geometric/explain/explanation.py
```diff
@@ -14,6 +16,10 @@ class Explanation(Data):
     also hold the original graph if needed.
 
     Args:
+        index (Union[int, Tensor], optional): The index of which part of the
+        model output the explanation is explaining. Can be a single index or
+        a tensor of indices, or None (for graph level explanations).
+        (default: :obj:`None`)
```
rusty1s, line 22:
```suggestion
        index (Union[int, Tensor], optional): The index of which part of the
            model output the explanation is explaining. Can be a single index or
            a tensor of indices, or None (for graph level explanations).
            (default: :obj:`None`)
```

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +136,45 @@ def _apply_masks(
             out = out.subgraph(node_mask)
 
         return out
+
+    @property
+    def masks(self) -> dict[str, Tensor]:
```
rusty1s, line 141:
```suggestion
    def masks(self) -> Dict[str, Tensor]:
```

rusty1s, line 141: Align with `available_explanations`, e.g., `available_explanation_dict`?

torch_geometric/explain/explanation.py
```diff
@@ -129,3 +136,45 @@ def _apply_masks(
...
+        mask_dict = {
+            key: self[key]
+            for key in self.keys
+            if key.endswith('_mask') and self[key] is not None
+        }
+        return dict(sorted(mask_dict.items()))
+
+    def threshold(self, threshold_value: Union[int, float],
+                  threshold_type: ThresholdType) -> 'Explanation':
+        r"""Returns a copy of explanation with applied threshold
```
rusty1s, line 152:
```suggestion
        r"""Returns a copy of explanation with applied threshold.
```

torch_geometric/explain/explainer.py
BlazStojanovic: The `get_prediction` method was already in the `Explainer` class, but it applied `argmax` by default. In explanations we might want to compare the difference in probability when using only the explanation graph as input, hence the `raw` flag. 

I am fine with moving `argmax` into the loss function and returning the raw value by default. Wdyt?

rusty1s:
Made some slight modifications while migrating to latest master. The `Explainer` now sets model inputs, `target` and `index` automatically when generating an `Explanation`. I thought this is very elegant, and `ExplainerAlgorithm` instantiations shouldn't worry about doing this.

rusty1s (approved):

PR #6222, opened 2022-12-13
Explainability Infection Benchmark

zechengz:
This PR implements Infection benchmark dataset as mentioned in #5817

Checklist
=========
- [x] CHANGELOG
- [x] Documentation

### DIFF
--- a/test/datasets/test_infection_dataset.py
+++ b/test/datasets/test_infection_dataset.py
@@ -1,0 +1,30 @@
+import pytest
+
+from torch_geometric.datasets import InfectionDataset
+from torch_geometric.datasets.graph_generator import ERGraph
+
+
+@pytest.mark.parametrize('graph_generator', [
+    ERGraph(num_nodes=500, edge_prob=0.004),
+])
+def test_infection_dataset_er(graph_generator):
+    dataset = InfectionDataset(graph_generator, seeds=[12345])
+    assert str(dataset) == ('InfectionDataset(1, graph_generator='
+                            'ERGraph(num_nodes=500, edge_prob=0.004)')
+    assert len(dataset) == 1
+    data1 = dataset[0]
+
+    assert len(data1) == 5
+    assert data1.num_nodes == 500
+    assert data1.edge_index.min() == 0
+    assert data1.edge_index.max() == data1.num_nodes - 1
+    assert data1.y.min() == 0 and data1.y.max() == 5
+    assert data1.node_mask.size() == (data1.num_nodes, )
+    assert data1.node_mask.min() == 0 and data1.node_mask.max() == 1
+    assert data1.node_mask.sum() == 242
+    assert len(data1.explain_path_edge_index) == 242
+
+    data2 = InfectionDataset(graph_generator, seeds=[12345])[0]
+    assert data1.node_mask.sum() == data2.node_mask.sum()
+    assert len(data1.explain_path_edge_index) == len(
+        data2.explain_path_edge_index)
--- a/test/datasets/test_infection_dataset.py
+++ b/test/datasets/test_infection_dataset.py
@@ -2,8 +2,10 @@ import pytest
 
 from torch_geometric.datasets import InfectionDataset
 from torch_geometric.datasets.graph_generator import ERGraph
+from torch_geometric.testing import withPackage
 
 
+@withPackage('networkx')
 @pytest.mark.parametrize('graph_generator', [
     ERGraph(num_nodes=500, edge_prob=0.004),
 ])
--- a/test/datasets/test_infection_dataset.py
+++ b/test/datasets/test_infection_dataset.py
@@ -1,32 +1,57 @@
-import pytest
+import torch
 
+from torch_geometric import seed_everything
+from torch_geometric.data import Data
 from torch_geometric.datasets import InfectionDataset
-from torch_geometric.datasets.graph_generator import ERGraph
-from torch_geometric.testing import withPackage
+from torch_geometric.datasets.graph_generator import ERGraph, GraphGenerator
 
 
-@withPackage('networkx')
-@pytest.mark.parametrize('graph_generator', [
-    ERGraph(num_nodes=500, edge_prob=0.004),
-])
-def test_infection_dataset_er(graph_generator):
-    dataset = InfectionDataset(graph_generator, seeds=[12345])
-    assert str(dataset) == ('InfectionDataset(1, graph_generator='
-                            'ERGraph(num_nodes=500, edge_prob=0.004)')
+class DummyGraph(GraphGenerator):
+    def __call__(self) -> Data:
+        edge_index = torch.tensor([
+            [0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9],
+            [1, 0, 2, 1, 3, 2, 4, 3, 5, 4, 6, 5, 7, 6, 8, 7, 9, 8],
+        ])
+        return Data(num_nodes=10, edge_index=edge_index)
+
+
+def test_infection_dataset():
+    seed_everything(12345)
+    graph_generator = DummyGraph()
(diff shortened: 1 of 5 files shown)


torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+                sp_length = [(idx, length) for idx, length in sp_length
+                             if length == distance]
+                if len(sp_length) == 0:
+                    continue
+                elif len(sp_length) == 1:
+                    # Add explanation for node with one unique path
+                    src_node = sp_length[0][0]
+                    sp_generator = nx.all_shortest_paths(G, src_node, node)
+                    sp = next(sp_generator)
+                    if next(sp_generator, None) is None:
```
RexYing, line 142: for nodes with multiple paths, is there no groundtruth explanation?

RexYing (approved):
Thanks! Overall it looks good to me. Just a few clarification questions.

test/datasets/test_infection_dataset.py
```diff
@@ -0,0 +1,32 @@
...
+from torch_geometric.datasets.graph_generator import ERGraph
+from torch_geometric.testing import withPackage
+
+
+@withPackage('networkx')
+@pytest.mark.parametrize('graph_generator', [
+    ERGraph(num_nodes=500, edge_prob=0.004),
+])
+def test_infection_dataset_er(graph_generator):
+    dataset = InfectionDataset(graph_generator, seeds=[12345])
```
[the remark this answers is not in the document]
zechengz: In the paper implementation mentioned in the issue https://dl.acm.org/doi/10.1145/3447548.3467283 there is no seed. But IMO we might need to add one to make the results reproducible?

torch_geometric/datasets/infection_dataset.py
zechengz: Yes, that's what the paper said

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+    r"""
+    Generates a synthetic infection dataset for evaluating explainabilty
+    algorithms, as described in the `"Explainability Techniques for Graph
+    Convolutional Networks" <https://arxiv.org/abs/1905.13686>`__ paper.
+    The :class:`~torch_geometric.datasets.InfectionDataset` creates synthetic
+    graphs coming from a
+    :class:`~torch_geometric.datasets.graph_generator.GraphGenerator` with
+    :obj:`num_infected` randomly assigned infected nodes. The dataset then
+    creates a node classification task of predicting length of the shortest
+    path from infected nodes with ground-truth node-level mask and explanation
```
RBendias, line 22: Let's add here, that there is a maximum path length.

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+            (or any string that automatically resolves to it).
+        num_graphs (int, optional): The number of graphs to generate.
+            (default: :obj:`1`)
+        graph_generator_kwargs (Dict[str, Any], optional): Arguments passed to
+            the respective graph generator module in case it gets automatically
+            resolved. (default: :obj:`None`)
+        seeds (List[int], optional): List of seeds to generate the infection
+            graphs. (default: :obj:`None`)
+        max_path_length (int, List[int]): The maximum shortest path length that
+            a node will not be infected. (default: :obj:`5`)
```
RBendias, line 50:
```suggestion
        max_path_length (int, List[int]): The maximum shortest path length that
            a node can be infected. (default: :obj:`5`)
```

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+        num_infected (List[int], optional): Number of randomly selected
+            infected nodes in the graph. (default: :obj:`50`)
+        transform (callable, optional): A function/transform that takes in an
+            :obj:`torch_geometric.data.Data` object and returns a transformed
+            version. The data object will be transformed before every access.
+            (default: :obj:`None`)
+    """
+    def __init__(
+        self,
+        graph_generator: Union[GraphGenerator, str],
```
RBendias, line 60: Why not using a default here as well?

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+    Args:
+        graph_generator (GraphGenerator or str): The graph generator to be
+            used, *e.g.*,
+            :class:`torch.geometric.datasets.graph_generator.BAGraph`
+            (or any string that automatically resolves to it).
+        num_graphs (int, optional): The number of graphs to generate.
+            (default: :obj:`1`)
+        graph_generator_kwargs (Dict[str, Any], optional): Arguments passed to
```
RBendias, line 44: Just an idea: We could also pass this as normal `kwargs`.

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+            the respective graph generator module in case it gets automatically
+            resolved. (default: :obj:`None`)
+        seeds (List[int], optional): List of seeds to generate the infection
+            graphs. (default: :obj:`None`)
```
RBendias, line 48:
```suggestion
        seeds (List[int], optional): List of seeds. Each seed is used to generate one
            infection graph. (default: :obj:`None`)
```
or adding something like:  `The number of seeds needs to be the same as the number of graphs`

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+        import networkx as nx
+        if seed is not None:
+            seed_everything(seed)
+        data = self.graph_generator()
+        num_nodes, edge_index = data.num_nodes, data.edge_index
+        rand_perm = torch.randperm(num_nodes)
+
+        x = torch.zeros((num_nodes, 2), dtype=torch.float32)
+        x[rand_perm[:num_infected], 1] = 1  # Infected
+        x[rand_perm[num_infected:], 0] = 1  # Healthy
```
RBendias, line 111: In the original implementation, they have two features, as you implemented. However, what is the advantage of having two features? Should we only use one binary feature?

torch_geometric/datasets/infection_dataset.py
```diff
@@ -0,0 +1,163 @@
...
+                # Get the subgraph with nodes that reach to node with k hops
+                sub_nodes, _, _, _ = k_hop_subgraph(node, distance, edge_index)
+                # All nodes in the subgraph are healthy
+                if sum(x[sub_nodes, 1]) == 0:
+                    continue
+                infected_indices = x[sub_nodes, 1].nonzero(as_tuple=True)[0]
+                infected_nodes = sub_nodes[infected_indices]
+                sp_length = [(idx.item(),
+                              nx.shortest_path_length(G, idx.item(), node))
+                             for idx in infected_nodes]
```
RBendias, line 131: Shouldn't this be clear depending on the distance? As soon as there is one infected node in the subgraph, that should be distance.

RBendias (commented):
Thank you so much for adding this. It looks good to me. In general, I noticed that the description of the arguments (seeds, max_path_lenght and num_infected) do not include the information that if a list is provided, that this list needs to be as long as the number of graphs and that each element is used per graph.

torch_geometric/datasets/infection_dataset.py
rusty1s: I think this models what we do in `ExplainerDataset` and it makes the assignment of arguments more clear. Would prefer to keep it.

torch_geometric/datasets/infection_dataset.py
rusty1s: I think it makes sense here to use two features, as otherwise, we need to learn about non-infected nodes via the bias term.

zechengz:
Thanks @rusty1s @RBendias @RexYing !

PR #6230, opened 2022-12-13
Add `IBMBBatchLoader` and `IBMBNodeLoader`

chendiqian:
add IBMB loaders of paper [Influence-Based Mini-Batching for Graph Neural Networks](https://openreview.net/forum?id=b9g0vxzYa_)


rusty1s:
Sorry for being silent here for so long. We are very interested to add this to PyG, and will integrate this soon. Intel team will help with benchmarking and improving performance.

chendiqian:
Thank you! Should I fix this _Read the Docs build failed!_ first? But I think there wasn't this issue

rusty1s:
I think the readthedocs build fails because `numba` is now a required dependency. Ideally, we keep this as optional IMO (similar to how we do it for `DimeNet` utilities).

rusty1s (approved):
@chendiqian Thanks! Can we merge the example in a separate PR?

PR #6233, opened 2022-12-13
Update `filter_*_store_()` returns in loader utils

JakubPietrakIntel:
Remove `return store` from `filter_node_store_` and `filter_edge_store_` as these values are not modified by the function, so the return can be `None`.


rusty1s (approved):

rusty1s:
Sounds good, thanks!

PR #6245, opened 2022-12-15
Implemented sparse variant of top-k sparsification for the GDC transform

marekdedic:
Hi,
the GDC transform was missing a sparse variant of the top-k sparsification method, so I share my implementation... Feedback welcome.


rusty1s (commented):
Thanks for the PR. Can we add a small test for this as well?

marekdedic:
Thanks, I'll try to add one.

PR #6246, opened 2022-12-15
Implemented an approximate variant of heat kernel diffusion for the GDC transform

marekdedic:
Depends on #6242 being merged first.

Previously, heat kernel diffusion was only implemented in the exact variant of GDC with just a note pointing to a paper by Kloster & Gleich for anybody who'd want an approximate version. I implemented the approximate variant based on that paper. The code started from the pseudo-code in the paper, but was heavily modified to preserve edge weight instead of just node weights an in order to enable numba. I'd be happy to walk any potential reviewers through the code.

The value of `N` can be computed in a more efficient way, however, this is just fine for the time being...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.3.0] - 2023-MM-DD
 ### Added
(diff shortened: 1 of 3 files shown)


marekdedic:
Hi, I am not really sure why the coverage is reported so low, on my machine the `calc_heat` function gets executed... Maybe something to do with numba?

rusty1s:
Feel free to add a `#pragma: no cover` comment to the `numba` calls to avoid tracking these code paths for code coverage.

PR #6248, opened 2022-12-16
fix for adj_type_to_edge_tensor_type & to_csc in graph_store.py

HunterTracer:
I'm not pretty sure that whether they are bugs in `graph_store.py`, but anyway, the original `adj_type_to_edge_tensor_type` & `to_csc` functions in `graph_store.py` seem to be unreasonable.


HunterTracer:
@rusty1s

rusty1s:
Yes, I agree this is a bit confusing. I still think this is correct since we assume that the `SparseTensor` we use here holds information in CSC format (`adj_t`).

HunterTracer:
OK. Thanks!

rusty1s:
I also have a PR out #5517 that cleans this up a little and makes this hopefully more understandable. If this resolves your concerns, shall we close this issue?

Issue #6255, opened 2022-12-16
Support BA2Motif Dataset


shenoynikhil:
### 🚀 BA2Motif Synthetic Graph Classification Dataset

For graph classification explanability tasks, we could make use of graph classification datasets like BA2Motifs Dataset. It is a synthetic dataset also implemented in [DGL library](https://docs.dgl.ai/en/0.8.x/generated/dgl.data.BA2MotifDataset.html) and used in several explainability papers. In my opinion, we would have to,
- [ ] Add support for five-node cycle motifs #6256 
- [ ] `BA2MotifDataset(num_graphs=n)`: Creates `n` BA graphs, `n/2` with house motifs and `n/2` with five-node cycle motifs #6257

shenoynikhil:
@rusty1s @RexYing Do you think this could be useful?

PR #6256, opened 2022-12-16
Adds Cycle Motif Support

shenoynikhil:
Support for cycle motif. Based on #6255 

- [x] Adds cycle motif class
- [x] Adds test for cycle motif
- [x] Updated CHANGELOG

##### Doubt [Remove if cleared]
I have not added node classification labels (I guess these motifs don't have them)

### DIFF
--- a/torch_geometric/datasets/motif_generator/cycle.py
+++ b/torch_geometric/datasets/motif_generator/cycle.py
@@ -7,18 +7,25 @@ from torch_geometric.datasets.motif_generator import CustomMotif
 class CycleMotif(CustomMotif):
     r"""Generates the cycle motif from the `"GNNExplainer:
     Generating Explanations for Graph Neural Networks"
-    <https://arxiv.org/pdf/1903.03894.pdf>`_ paper, containing n nodes and n
-    undirected edges.
+    <https://arxiv.org/pdf/1903.03894.pdf>`_ paper.
 
     Args:
-        n (int): Number of nodes (or edges) in the cycle.
+        num_nodes (int): The number of nodes in the cycle.
     """
-    def __init__(self, n: int = 6):
-        # construct edge_index based on n
+    def __init__(self, num_nodes: int):
+        self.num_nodes = num_nodes
+
+        row = torch.arange(num_nodes).view(-1, 1).repeat(1, 2).view(-1)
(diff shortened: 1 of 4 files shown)


torch_geometric/datasets/motif_generator/cycle.py
```diff
@@ -0,0 +1,24 @@
...
+    Args:
+        n (int): Number of nodes (or edges) in the cycle.
+    """
+    def __init__(self, n: int = 6):
+        # construct edge_index based on n
+        structure = Data(
+            num_nodes=n,
+            edge_index=torch.Tensor([[x for x in range(n)],
+                                     [y for y in range(1, n)] + [0]
+                                     ]).type(torch.int32),
```
wsad1, line 22:
```suggestion
            edge_index=to_undirected(torch.stack((torch.arange(n), torch.arange(1,n+1)),
```
This will make it undirected.

torch_geometric/datasets/motif_generator/cycle.py
```diff
@@ -0,0 +1,24 @@
...
+
+from torch_geometric.data import Data
+from torch_geometric.datasets.motif_generator import CustomMotif
+
+
+class CycleMotif(CustomMotif):
+    r"""Generates the cycle motif from the `"GNNExplainer:
+    Generating Explanations for Graph Neural Networks"
+    <https://arxiv.org/pdf/1903.03894.pdf>`_ paper, containing n nodes and n
+    undirected edges.
```
wsad1, line 11:
```suggestion
    <https://arxiv.org/pdf/1903.03894.pdf>`_ paper, containing :obj:`n` nodes and
    in a cycle.
```

torch_geometric/datasets/motif_generator/cycle.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Args:
+        n (int): Number of nodes (or edges) in the cycle.
```
wsad1, line 14:
```suggestion
        n (int): Number of nodes in the cycle.
```
There will be `2*n` edges since the graph is undirected.

wsad1 (commented):
Looks good mostly. Left some comments.

rusty1s (approved):

PR #6257, opened 2022-12-16
Adds the BA2Motif synthetic Dataset

shenoynikhil:
Part of #6255  (Builds on #6256)
Adds support for `BA2MotifDataset` from the paper [Parameterized Explainer for Graph Neural Network](https://arxiv.org/abs/2011.04573).

- [x] Add the dataset and update init
- [x] Add tests
- [x] Update changelog


wsad1 (commented):
This was probably not well communicated, apologies for that.
But can we use the new `ExaplainerDataset` Interface to achieve this. So instead of adding a new class, i think we could add an example of how to generate this dataset using the `ExplainerDataset` Interface.

rusty1s:
I think in this case it makes sense to directly use the provided dataset from https://github.com/flyingdoog/PGExplainer/blob/master/dataset/BA-2motif.pkl since it seems tight to the specific use-case of detecting houses vs 5-node cycles. Not sure if we can integrate this elegantly into the `ExplainerDataset`.

rusty1s (approved):

rusty1s:
@wsad1 You are right that we could also create this via the `ExplainerDataset`, by creating two datasets and concatenating them together. Will add a code example.

PR #6278, opened 2022-12-21
Add an `RGCNConv` model using the primitives in `cugraph-ops`

tingyu66:
This PR adds an RGCN model `torch_geometric.nn.RGCNConvCuGraph` that leverages the optimized aggregation functions in cugraph-ops.

Fixes https://github.com/rapidsai/cugraph-ops/issues/180.

It requires the package `pylibcugraphops >= 23.02`, which is going to be released by next February. I have tested it using the [nightly build](https://anaconda.org/rapidsai-nightly/pylibcugraphops).

The user interface largely resembles `RGCNConv`. Our aggregation function has two modes: full-graph and message flow graph (MFG) mode, with the latter being more performant. That's why our `forward` function has to have extra arguments like `num_nodes`, `from_neighbor_sampler` and `max_num_neighbors`, to help convert PyG edge_index to cugraph-ops graph type and specify which mode to opt for.

Here are some preliminary performance results using MAG dataset and 2-layer RGCN model, with batch_size=1024, num_neighbors=[25,20] in NeighborLoader. I set feature dim to 16 and did not use the provided paper features, as this is just a synthetic benchmark.

| time (s) per epoch | RGCNConv with pyg_lib | RGCNConvCuGraph | 
|-------|-----|-----|
|forward| 1.66 | 1.14 |
|backward | 2.62| 1.52 |

Right now, since cugraph-ops requires csc/csr representation of the graph, the forward time is dominated by the coo-to-csc conversion using the utility for `SparseTensor`, taking 50~70% of the forward time. We can expect a higher speedup if 1). dataloader can emit csc graph and 2). dataloader can output a list of MFG, instead of a fused subgraph.


tingyu66:
@rusty1s I have updated RGCNConv and tested it locally. Feel free to give it a go.

rusty1s (approved):

PR #6301, opened 2022-12-26
Add adaptive aggregators from the 'Graph Neural Networks with Adaptive Readouts' paper

davidbuterez:
In this PR I propose adding the neural network-based ('adaptive') aggregation/readout functions from the NeurIPS 2022 paper ['Graph Neural Networks with Adaptive Readouts'](https://arxiv.org/abs/2211.04952), with an example implementation available [here](https://github.com/davidbuterez/gnn-neural-readouts).

The paper/project has two main goals:
1. To motivate research into expressive aggregation functions, where the presence or lack of permutation invariance is an interesting trade-off
2. To benchmark different aggregation schemes and illustrate the benefit of neural-network based aggregators

This PR introduces 3 aggregation schemes:

1. **MLP aggregation** - not permutation invariant, but highly performant especially on certain molecular datasets. Currently, it appears that there are no similar aggregations in PyG.
2. **Set Transformer aggregation** - permutation invariant and based on the [Set Transformer](https://github.com/juho-lee/set_transformer) implementation. This shares some similarities with the unrelated work behind `GraphMultisetTransformer`, however our approach is not a hierarchical pooling method. Instead of graph coarsening, the focus is on regression/classification performance and the property of permutation invariance.
3. **GRU aggregation** - a simple counterpart to MLP aggregation but based on recurrent neural networks. The implementation is similar to the existing `LSTMAggregation`.

The proposed implementation is more customisable than the original implementation. MLP and GRU usually have a very small impact on the running time (at least on GPUs), with the Set Transformer incurring a more significant but not drastic increase (depending on its configuration). More details regarding the time and memory consumption are available in the paper (Appendix I).


rusty1s:
Thank you very much! Can we add some tests for this PR? This would be highly appreciated :)

davidbuterez:
Sure, I added some tests based on the other `Aggregations`. I also added a Deep Sets-based aggregator which was hinted at in the paper.

torch_geometric/nn/aggr/adaptive.py
```diff
@@ -0,0 +1,400 @@
...
+
+        reset(self.phi)
+        reset(self.rho)
+
+    def forward(self, x: Tensor, index: Optional[Tensor] = None,
+                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
+                dim: int = -2) -> Tensor:
+        x = self.phi(x)
+        x, _ = to_dense_batch(x, index, fill_value=0,
+                              max_num_nodes=self.max_num_nodes_in_dataset)
```
wsad1, line 392: does this functions really need the `self.max_num_nodes_in_dataset` argument. It looks like it'll work fine without that.

torch_geometric/nn/aggr/adaptive.py
```diff
@@ -0,0 +1,400 @@
...
+                f'{self.num_hidden_layers}, {self.hidden_dim}, '
+                f'{self.use_batch_norm}, {self.dropout}, '
+                f'Uses custom MLP: {has_custom_mlp})')
+
+
+# Set Transformer code as provided in
+# https://github.com/juho-lee/set_transformer
+
+
+class MAB(nn.Module):
```
wsad1, line 122: Do you think we can reuse [this](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/aggr/gmt.py).

torch_geometric/nn/aggr/__init__.py
```diff
@@ -19,26 +19,15 @@
...
-    'GraphMultisetTransformer',
-    'AttentionalAggregation',
-    'EquilibriumAggregation',
+    'Aggregation', 'MultiAggregation', 'SumAggregation', 'MeanAggregation',
+    'MaxAggregation', 'MinAggregation', 'MulAggregation', 'VarAggregation',
+    'StdAggregation', 'SoftmaxAggregation', 'PowerMeanAggregation',
+    'MedianAggregation', 'QuantileAggregation', 'LSTMAggregation', 'Set2Set',
+    'DegreeScalerAggregation', 'SortAggregation', 'GraphMultisetTransformer',
+    'AttentionalAggregation', 'EquilibriumAggregation', 'MLPAggregation',
+    'SetTransformerAggregation', 'GRUAggregation', 'DeepSetsAggregation'
```
wsad1, line 32:
```suggestion
    'Aggregation',
    'MultiAggregation',
    'SumAggregation',
    'MeanAggregation',
    'MaxAggregation',
    'MinAggregation',
    'MulAggregation',
    'VarAggregation',
    'StdAggregation',
    'SoftmaxAggregation',
    'PowerMeanAggregation',
    'MedianAggregation',
    'QuantileAggregation',
    'LSTMAggregation',
    'Set2Set',
    'DegreeScalerAggregation',
    'SortAggregation',
    'GraphMultisetTransformer',
    'AttentionalAggregation',
    'EquilibriumAggregation',
    'MLPAggregation',
    'SetTransformerAggregation',
    'GRUAggregation',
    'DeepSetsAggregation',
```

torch_geometric/nn/aggr/adaptive.py
```diff
@@ -0,0 +1,400 @@
...
+            self.mlp_agg = nn.Sequential(*modules)
+        else:
+            self.mlp_agg = self.custom_mlp
+
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        for module in self.mlp_agg.children():
+            if hasattr(module, 'reset_parameters'):
+                module.reset_parameters()
```
wsad1, line 100:
```suggestion
        reset(self.mlp_agg)
```

torch_geometric/nn/aggr/adaptive.py
```diff
@@ -0,0 +1,400 @@
+import math
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import Tensor, nn
+
+from torch_geometric.nn.aggr import Aggregation
+from torch_geometric.utils import to_dense_batch
+
```
wsad1, line 10:
```suggestion
from ..inits import reset
```

test/nn/aggr/test_adaptive.py
```diff
@@ -0,0 +1,131 @@
...
+    assert gru_aggr(x, index).size() == (2, out_channels)
+
+
+@pytest.mark.parametrize('num_hidden_layers_phi', [0, 1, 2])
+@pytest.mark.parametrize('num_hidden_layers_rho', [0, 1, 2])
+@pytest.mark.parametrize('max_num_nodes_in_dataset', [14, 50])
+def test_deep_sets_aggregation(num_hidden_layers_phi, num_hidden_layers_rho,
+                               max_num_nodes_in_dataset):
+    x = torch.randn(14, 9)
+    index = torch.tensor([0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1])
```
wsad1, line 116: Define this at the top of this file. That way all tests can share these tensors.

torch_geometric/nn/aggr/adaptive.py
davidbuterez: This is a good observation, the argument should not be necessary for Deep Sets.

torch_geometric/nn/aggr/adaptive.py
davidbuterez: I was not sure if it would be possible without refactoring the GMT code. The computation for the MAB itself should be the same but there is some additional logic in the GMT version. I haven't tested this but I think it may be possible with the right arguments.

rusty1s (approved):
Thanks! I moved each aggregator to its own file, and reused MAB, etc from the GTM code. These should be equivalent.

AntonCoon:
Hello everyone, I've been searching for an example of code that demonstrates the use of MLP aggregation within the SAGE model, but unfortunately, I haven't had any luck so far. I was wondering if any of you could kindly provide some guidance or point me in the right direction to find such an example?
I do have a code snippet that I'm working with, but I suspect it might be an incorrect usage of the MLP aggregation..
```python
self.convs.append(
    SAGEConv(
        in_channels, 
        hidden_channels_sizes[0], 
        aggr=aggr.MLPAggregation(
            in_channels=in_channels,
            out_channels=hidden_channels_sizes[0],
            max_num_elements=1,
            num_layers=1,
        )
    )
)
```
Any advice on how to correct this code snippet would be greatly appreciated.

rusty1s:
Just added a test for this, see https://github.com/pyg-team/pytorch_geometric/pull/8037/files. The main thing you need to change is that `MLPAggregation` should map to `in_channels` rather than `hidden_channels_sizes[0]` since `SAGEConv` currently expects that an aggregation does not change the feature dimension.

AntonCoon:
Hi, thanks for your helpful test and reply! I created a small example to check how it works, and I've encountered an issue that is not clear to me. Any advice is appreciated.. 
The layers:
```python
self.convs.append(
    SAGEConv(
        in_channels, 
        hidden_channels_sizes[0], 
        aggr=aggr.MaxAggregation()
    )
)

self.convs.append(
    SAGEConv(
        in_channels=hidden_channels_sizes[0],
        out_channels=output_size,
        aggr=aggr.MLPAggregation(
            in_channels=hidden_channels_sizes[0],
            out_channels=hidden_channels_sizes[0],
            max_num_elements=6,
            num_layers=1,
        ),
    )
)
```
After the training (typical contrastive learning) I evaluating the model to create an embeddings 
 ```python
with torch.no_grad():
    model.eval()
    embeddings = model.full_forward(x, edge_index).cpu()
```
and I've got this error message:
```pyton
    131     if index is not None and not torch.all(index[:-1] <= index[1:]):
--> 132         raise ValueError("Can not perform aggregation since the 'index' "
    133                          "tensor is not sorted")

ValueError: Can not perform aggregation since the 'index' tensor is not sorted
```
But here the weird part: `edge_index` is already sorted, moreover, a checked it with code from assert 
```python
if edge_index is not None and not torch.all(edge_index[:-1] <= edge_index[1:]):
    raise ValueError("Can not perform aggregation since the 'index' tensor is not sorted")
```
and there is no assertions here.
If someone have any ideas, what might be wrong, please, let me know

rusty1s:
Can you modify your code to
```python
if edge_index is not None and not torch.all(edge_index[1, :-1] <= edge_index[1, 1:]):
```
It looks like you are currently indexing in the first dimension.

PR #6306, opened 2022-12-27
Fix to save best models instead of worst

anton-bushuiev:
This PR Fixes a minor bug in PyTorchLightning examples. Currently, the worst models are saved instead of the best ones. See the [documentation of `pl.callbacks.ModelCheckpoint`](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.ModelCheckpoint.html#pytorch_lightning.callbacks.ModelCheckpoint.params.mode)


rusty1s (approved):
Thanks!

Issue #5340, opened 2022-09-02
NeighborLoader with loader worker processes fails on GPU


SebastianHurubaru:
### 🐛 Describe the bug

When using NeighborLoader with multi-process data loading on GPU, we get the following error:
```
RuntimeError: Caught RuntimeError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop
    data = fetcher.fetch(index)
  File "/usr/local/lib/python3.7/dist-packages/torch/utils/data/_utils/fetch.py", line 49, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/usr/local/lib/python3.7/dist-packages/torch/utils/data/_utils/fetch.py", line 49, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
RuntimeError: CUDA error: initialization error
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
```

Running on CPU works fine. The deprecated `NeighborSampler` seems to work fine on all devices!

Here is the **IPython (Jupyter notebook) code** used to reproduce the error for `NeighborLoader`. The latter part contains cells showing that `NeighborSampler` works fine:

```
import torch
import os
print("PyTorch has version {}".format(torch.__version__))

# Install torch geometric
!pip install torch-scatter -f https://pytorch-geometric.com/whl/torch-1.12.1+cu113.html
!pip install torch-sparse -f https://pytorch-geometric.com/whl/torch-1.12.1+cu113.html
!pip install torch-geometric

## Setup

import os
import copy
import torch
import pandas as pd
import torch.nn.functional as F
import torch_geometric.transforms as T

from torch_geometric.nn import SAGEConv
from torch_geometric.loader import NeighborLoader, NeighborSampler

# Install OGB
!pip install -q ogb

from ogb.nodeproppred import PygNodePropPredDataset, Evaluator

dataset_name = 'ogbn-arxiv'
dataset = PygNodePropPredDataset(name=dataset_name,
                                transform=T.ToSparseTensor())
data = dataset[0]
data.adj_t = data.adj_t.to_symmetric()
device = 'cuda' if torch.cuda.is_available() else 'cpu'

print('Device: {}'.format(device))

data = data.to(device)

[...]

PR #6307, opened 2022-12-27
Add DDP `NeighborLoader` examples

LukeLIN-web:
Since NeighborLoader is supported by pyg-lib. We could add an example in distributed_sampling_loader.py.
I refer https://github.com/pyg-team/pytorch_geometric/issues/5340#issuecomment-1235409005 to set `kwargs = {'batch_size': 1024, 'num_workers': 0}`


rusty1s:
Thank you! Can we replace the existing example with `NeighborLoader`?

LukeLIN-web:
> Thank you! Can we replace the existing example with `NeighborLoader`?

Yes, maybe you could replace the existing pytorch_geometric/examples/multi_gpu/distributed_sampling.py

rusty1s:
Yes, since we want to fully replace `NeighborSampler` with `NeighborLoader`, I think just replacing the example would be best.

LukeLIN-web:
> Yes, since we want to fully replace `NeighborSampler` with `NeighborLoader`, I think just replacing the example would be best.

Thank you. I replaced the example.

rusty1s (approved):

PR #6312, opened 2022-12-29
Fixed a bug in which `distributed_sampling_loader.py` mistakenly split data.

LukeLIN-web:
I am sorry that I made a mistake in that we don't split data into each GPU. Now I fixed it and tested it.


rusty1s (approved):
Good catch, thanks!

PR #6326, opened 2023-01-02
`GPSConv` Graph Transformer Layer

rusty1s:


Jerome-Cong:
Hi, thanks for adding this layer! The [paper](https://arxiv.org/abs/2205.12454) of GraphGPS mentioned that the use of linear transformers (e.g. [BigBird](https://github.com/google-research/bigbird) and [Performer](https://github.com/lucidrains/performer-pytorch)) dramatically improves the computation efficiency for big graphs (hundreds of nodes). It seems that `GPSConv` currently uses a vanilla transformer (i.e. the global attention). Will linear transformer options be one of the **TODOs** in the future?

rusty1s:
Yes, we are happy to extend `GraphGPS` to support other transformers. Also appreciate a PR if you would like to contribute :)

alexlyzhov:
https://github.com/pyg-team/pytorch_geometric/blob/1bb03064366dbda6e896990be26a4e16ba88c842/torch_geometric/nn/conv/gps_conv.py#L155

Seems like `norm3` is never used

rusty1s:
Can you clarify what you mean?

alexlyzhov:
It changes h, but h is not saved.

rusty1s:
You are absolutely, right, thanks! Fixed this :(

PR #6345, opened 2023-01-05
`connected` explainability thresholding

### LINKED ISSUE #5629: GNN Explanation Settings

soma2000-lang:
Fixes #5629 
@rusty1s @RexYing 
Please review my Pr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [2.3.0] - 2023-MM-DD
+
 ### Added
+
 - Added the `GNNFF` model ([#5866](https://github.com/pyg-team/pytorch_geometric/pull/5866))
 - Added `MLPAggregation`, `SetTransformerAggregation`, `GRUAggregation`, and `DeepSetsAggregation` as adaptive readout functions ([#6301](https://github.com/pyg-team/pytorch_geometric/pull/6301), [#6336](https://github.com/pyg-team/pytorch_geometric/pull/6336), [#6338](https://github.com/pyg-team/pytorch_geometric/pull/6338))
 - Added `Dataset.to_datapipe` for converting PyG datasets into a torchdata `DataPipe`([#6141](https://github.com/pyg-team/pytorch_geometric/pull/6141))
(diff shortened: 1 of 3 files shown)


wsad1 (commented):
@soma2000-lang thanks for including this. Could you explain which part Issue #5629 you are trying to address in this PR, is it the last item.
Also the `Explainer_threshold` function you added seems to be the `k_hop_subgraph` code, is this intentional?

soma2000-lang:
@wsad1 I wanted to implement the algorithm  mentioned in the  third part of the issue,but was not sure about the implementation.I thought the implementation of k_hop<subgraoh would be quite similar.Pardon me!.Any help regrading doing the issue on the correct way would be great.

wsad1:
From the Issue the english language description of the algorithm is 
```
1. Start from the seed node
2. At every step, identify all edges that are adjacent to any of the selected nodes
3. Pick the edge adjacent with the largest weight, and if the edge reaches a previously not selected node, add that into the set of selected nodes
4. Iterate 2 and 3, until a desired explanation size is achieved
```
Assuming the desired explanation size isn't too big we could initially implement this in python and move to c++ if needed later.  Roughly the implementation would look something like
```
from torch_geometric.sampler.utils import to_csc
get_consice_explanation(self, num_nodes_in_explnation : int) -> Explanation:
    concise_explanation = copy(self)
    colptr, row, _ = to_csc(self) # convert to csc so that its easy to traverse.
    get_neighbors(node_id: int)->Tensor:
        row[colptr[node_id]:colptr[node_id+1]]
    added_nodes = set(self.index)
    while len(added_nodes)<=num_nodes_in_explanation:
         # Add code similar to Prims algorithm, the only difference here is that self loops are allowed.
```

PR #6376, opened 2023-01-09
Change mutable default arguments to `None`

bwroblew:
Reading PyG code I accidently saw that class and noticed that it uses mutable default arguments what is generally considered a bad practice and can lead to a lot of problems. I believe it's not a desired behaviour (am I wrong?).

Problems can occur when the class is extended in the future such that it changes the value of any of those lists, the default value for new instances of this class would also be affected. The other possibility is that user changes any of the `follow_batch`, `exclude_keys` fields (those doesn't start with an underscore, so it's kinda possible) after the object is initialized, what's also going to change the default value for new instances.

### DIFF
--- a/torch_geometric/nn/data_parallel.py
+++ b/torch_geometric/nn/data_parallel.py
@@ -42,8 +42,8 @@ class DataParallel(torch.nn.DataParallel):
                  follow_batch=None, exclude_keys=None):
         super().__init__(module, device_ids, output_device)
         self.src_device = torch.device(f'cuda:{self.device_ids[0]}')
-        self.follow_batch = follow_batch if follow_batch is not None else []
-        self.exclude_keys = exclude_keys if exclude_keys is not None else []
+        self.follow_batch = follow_batch or []
+        self.exclude_keys = exclude_keys or []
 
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/data_parallel.py
```diff
@@ -34,16 +34,16 @@ class DataParallel(torch.nn.DataParallel):
...
     """
     def __init__(self, module, device_ids=None, output_device=None,
-                 follow_batch=[], exclude_keys=[]):
+                 follow_batch=None, exclude_keys=None):
         super().__init__(module, device_ids, output_device)
         self.src_device = torch.device(f'cuda:{self.device_ids[0]}')
-        self.follow_batch = follow_batch
-        self.exclude_keys = exclude_keys
+        self.follow_batch = follow_batch if follow_batch is not None else []
+        self.exclude_keys = exclude_keys if exclude_keys is not None else []
```
rusty1s, line 46:
```suggestion
        self.follow_batch = follow_batch or []
        self.exclude_keys = exclude_keys or []
```

rusty1s (approved):
Thank you!

Issue #6375, opened 2023-01-09
🚀 `CaptumExplainer` behind the new `Explainer` interface


rusty1s:
### 🚀 The feature, motivation and pitch

We want to move the `to_captum` functionality behind the new `Explainer` interface and remove any explainer code from `nn.models`.

- [ ] Remove the `Explainer` class in `nn/models/captum.py` completely. It is not longer used anyway
- [ ] Move the `CaptumModel`, `CaptumHeteroModel`, `to_captum_input`, `captum_outputs_to_dict` to `explain/algorithm/captum.py`, remove `to_captum`, move tests to `test/explain/algorithm/test_captum.py`
- [ ] Define `CaptumExplainer` that takes in a `captum.attr` class and any additional args/kwargs of the class (except for `model` -> this one will be passed automatically by `Explainer`)
- [ ] Define `Explainer.__call__` for homogeneous graphs
- [ ] Define `Explainer.__call__` for heterogeneous graphs
- [ ] Update the example script

cc: @RBendias

PR #6383, opened 2023-01-10
Adds `CaptumExplainer` class

RBendias:
#6375


torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -0,0 +1,63 @@
...
+                          f"'node_mask_type' in [None, 'attributes'], "
+                          f"but got '{node_mask_type}'")
+            return False
+
+        edge_mask_type = self.explainer_config.edge_mask_type
+        if edge_mask_type not in [None, MaskType.object]:
+            logging.error(f"'{self.__class__.__name__}' only supports "
+                          f"'edge_mask_type' in [None, 'object'], "
+                          f"but got '{edge_mask_type}'")
+            return False
```
rusty1s, line 61: This can be dropped since `edge_mask_type` can only be either `None` or `object`.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> Explanation:
+
+        return Explanation(node_mask=None, edge_mask=None)
+
+    def supports(self) -> bool:
+        node_mask_type = self.explainer_config.node_mask_type
+        if node_mask_type not in [None, MaskType.attributes]:
+            logging.error(f"'{self.__class__.__name__}' only supports "
+                          f"'node_mask_type' in [None, 'attributes'], "
+                          f"but got '{node_mask_type}'")
```
rusty1s, line 53:
```suggestion
                          f"'node_mask_type' in [None, 'attributes'] "
                          f"(got '{node_mask_type.value}')")
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -0,0 +1,63 @@
+import logging
+from typing import Optional, Union
+
+import torch
+from captum import attr  # noqa
```
rusty1s, line 5: This cannot be a top-level argument as it is an optional dependency.

torch_geometric/explain/algorithm/__init__.py
```diff
@@ -3,11 +3,13 @@
...
 from .attention_explainer import AttentionExplainer
+from .captum_explainer import CaptumExplainer
 
 __all__ = classes = [
     'ExplainerAlgorithm',
     'DummyExplainer',
     'GNNExplainer',
     'PGExplainer',
     'AttentionExplainer',
+    'CaptumExplainer',
```
rusty1s, line 14: Let's move it right after `GNNExplainer`.

rusty1s (approved):
We can merge this with `test_captum.py` and `captum.py` since Vid's PR just landed.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -0,0 +1,63 @@
...
+
+class CaptumExplainer(ExplainerAlgorithm):
+    """Captum-based explainer algorithm.
+
+    This explainer algorithm uses `Captum <https://captum.ai/>`_ to compute
+    attributions.
+
+    Args:
+        attribution_method (Union[str, Attribution]): The Captum
+            attribution method to use. Can be a string or a captum.attr method.
```
vid-koci, line 21:
```suggestion
            attribution method to use. Can be a string or a `captum.attr` method.
```

torch_geometric/explain/algorithm/captum_explainer.py
RBendias: Where do you see this? Because according to the doc in Explainer it can have more: https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/explain/explainer.py#L61

rusty1s: It is in the `explain/config.py`.

rusty1s (approved):

PR #6387, opened 2023-01-10
`CaptumExplainer` forward method

RBendias:
Defines Explainer.__call__ for homogeneous/heterogeneous  graphs (#6375)

### DIFF
--- a/torch_geometric/explain/algorithm/captum_explainer.py
+++ b/torch_geometric/explain/algorithm/captum_explainer.py
@@ -1,0 +1,56 @@
+import logging
+from typing import Any, Dict, Optional, Union
+
+import torch
+from torch import Tensor
+
+from torch_geometric.explain import Explanation
+from torch_geometric.explain.algorithm import ExplainerAlgorithm
+from torch_geometric.explain.config import MaskType
+from torch_geometric.typing import EdgeType, NodeType
+
+
+class CaptumExplainer(ExplainerAlgorithm):
+    """A Captum-based explainer for identifying compact subgraph structures and
+    node features that play a crucial role in the predictions made by a GNN.
+
+    This explainer algorithm uses `Captum <https://captum.ai/>`_ to compute
+    attributions.
+
+    Args:
+        attribution_method (Attribution or str): The Captum attribution method
+            to use. Can be a string or a :class:`captum.attr` method.
+        **kwargs: Additional arguments for the Captum attribution method.
+    """
+    def __init__(
+        self,
+        attribution_method: Union[str, Any],
+        **kwargs,
+    ):
+        super().__init__()
+
+        self.attribution_method = attribution_method
+        self.kwargs = kwargs
+
+    def forward(
+        self,
+        model: torch.nn.Module,
+        x: Union[Tensor, Dict[NodeType, Tensor]],
+        edge_index: Union[Tensor, Dict[EdgeType, Tensor]],
+        *,
+        target: Optional[Tensor] = None,
+        index: Optional[Tensor] = None,
+        **kwargs,
+    ) -> Explanation:
+
+        return Explanation(node_mask=None, edge_mask=None)
+
+    def supports(self) -> bool:
+        node_mask_type = self.explainer_config.node_mask_type
+        if node_mask_type not in [None, MaskType.attributes]:
+            logging.error(f"'{self.__class__.__name__}' only supports "
+                          f"'node_mask_type' None or 'attributes' "
+                          f"(got '{node_mask_type.value}')")
+            return False
+
+        return True
--- a/torch_geometric/explain/algorithm/captum_explainer.py
+++ b/torch_geometric/explain/algorithm/captum_explainer.py
@@ -79,8 +79,8 @@ class CaptumExplainer(ExplainerAlgorithm):
             *kwargs.values(),
         )
 
-        if hasattr(model, 'metadata'):
-            metadata = model.metadata
+        if isinstance(x, dict):
+            metadata = (x.keys(), edge_index.keys())
             captum_model = CaptumHeteroModel(
                 model,
                 mask_type,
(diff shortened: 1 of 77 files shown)


torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -32,6 +38,18 @@ def __init__(
         self.attribution_method = attribution_method
         self.kwargs = kwargs
 
+    def _get_mask_type(self):
+        "Based on the explainer config, return the mask type."
+        node_mask_type = self.explainer_config.node_mask_type
+        edge_mask_type = self.explainer_config.edge_mask_type
+        if node_mask_type is not None and edge_mask_type is not None:
+            mask_type = 'node_and_edge'
```
vid-koci, line 46: Could we introduce these three mask types as an Enum? Also in `captum.py`, possibly as a follow-up PR.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -42,8 +60,57 @@ def forward(
...
+
+        node_mask, edge_mask = convert_captum_output(
+            attributions,
+            mask_type,
+            metadata,
+        )
 
-        return Explanation(node_mask=None, edge_mask=None)
+        if hasattr(model, 'metadata'):
+            return HeteroExplanation(
```
rusty1s, line 108: I don't think this is correct. I think you would need to create similar to
```python
explanation = HeteroExplanation()
for node_type, node_mask in node_mask_dict.items():
    explanation[node_type].node_mask = node_mask

File: torch_geometric/explain/algorithm/captum_explainer.py [hunk 3]
```diff
@@ -42,8 +60,57 @@ def forward(
         index: Optional[Tensor] = None,
         **kwargs,
     ) -> Explanation:
+        import captum.attr  # noqa
```
rusty1s, line 63: Why do we need the import here?

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -42,8 +60,57 @@ def forward(
...
+
+        inputs, add_forward_args = to_captum_input(
+            x,
+            edge_index,
+            mask_type,
+            *kwargs.values(),
+        )
+
+        if hasattr(model, 'metadata'):
+            metadata = model.metadata
```
rusty1s, line 75: I think it is safer to use `x.keys()` and `edge_index.keys()`

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -42,8 +60,57 @@ def forward(
...
+            metadata = None
+            captum_model = CaptumModel(model, mask_type, index)
+
+        if isinstance(self.attribution_method, str):
+            attribution_method = getattr(
+                captum.attr,
+                self.attribution_method,
+            )(captum_model)
+        else:
+            attribution_method = self.attribution_method(captum_model)
```
rusty1s, line 92: Can we initialize that in `__init__`?

torch_geometric/explain/algorithm/captum_explainer.py
wsad1: +1.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -42,8 +60,57 @@ def forward(
...
+        )
+
+        node_mask, edge_mask = convert_captum_output(
+            attributions,
+            mask_type,
+            metadata,
+        )
 
-        return Explanation(node_mask=None, edge_mask=None)
+        if hasattr(model, 'metadata'):
```
wsad1:
```suggestion
        if isinstance(x, dict):
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -43,7 +70,51 @@ def forward(
...
+        mask_type = self._get_mask_type()
+
+        inputs, add_forward_args = to_captum_input(
+            x,
+            edge_index,
+            mask_type,
+            *kwargs.values(),
+        )
+
+        if hasattr(model, 'metadata'):
```
wsad1, line 82:
```suggestion
        if isinstance(x, dict):
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -43,7 +70,51 @@ def forward(
...
+            **self.kwargs,
+        )
+
+        node_mask, edge_mask = convert_captum_output(
+            attributions,
+            mask_type,
+            metadata,
+        )
+
+        if not hasattr(model, 'metadata'):
```
wsad1, line 109:
```suggestion
        if not isinstance(x, dict):
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -29,9 +35,30 @@ def __init__(
...
         self.kwargs = kwargs
 
+        if isinstance(attribution_method, str):
+            self.attribution_method = getattr(
+                captum.attr,
+                attribution_method,
+            )
+        else:
+            self.attribution_method = attribution_method
+
```
wsad1, line 49:
```suggestion
def supports():
    return True
```
It looks like all configs are supported. Except for probably a node mask `object`.

wsad1 (approved):
Looks good. 
Lets add an issue to update https://pytorch-geometric.readthedocs.io/en/latest/notes/explain.html, to show `HeteroExplanation`.

torch_geometric/explain/algorithm/captum_explainer.py
RBendias: `common_attributes` and `object` are not supported.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -43,7 +70,51 @@ def forward(
...
+            *kwargs.values(),
+        )
+
+        if isinstance(x, dict):
+            metadata = (x.keys(), edge_index.keys())
+            captum_model = CaptumHeteroModel(
+                model,
+                mask_type,
+                index,
+                model.metadata,
```
rusty1s, line 88:
```suggestion
                metadata,
```

rusty1s (approved):

PR #6388, opened 2023-01-10
Add a `SAGEConv` model using the primitives in `cugraph-ops`

tingyu66:
This PR adds a GraphSAGE model `torch_geometric.nn.SAGEConvCuGraph` that leverages the optimized aggregation function in cugraph-ops, similar to #6278.

Fixes rapidsai/cugraph-ops#178.


tingyu66:
@rusty1s This is now ready for review. Regarding the CI failure, should I disable the test for now as CI is cpu-only?

The interface of `SAGEConvCuGraph` largely resembles `SAGEConv`. It accepts `edge_index` in both `torch.Tensor` and `SparseTensor` type. As you will see in the code, we rely on the utilities in `torch_sparse` to generate a csc representation which is required by cugraph-ops. The conversion is actually the performance bottleneck, so it would be nice if the dataloader can create a csc layout in `SparseStorage`.

cugraph-ops provides aggregation primitives specialized for both full graphs and bipartite graphs, with the latter being more performant. That is why you see the extra argument `max_num_neighbors`, which allows the model to opt for the faster option when given bipartite graphs. 

CC: @stadlmax

rusty1s:
Thanks. You can use the `@onlyCUDA` decorator in `torch_geometric.testing` to only test this on GPU devices.

rusty1s:
Hey @tingyu66 Can you give me write access to the PR?

tingyu66:
> Hey @tingyu66 Can you give me write access to the PR?

Invited.

rusty1s:
@tingyu66 Thanks. I made a past. Can you confirm that everything runs through? I sadly still cannot test it :(

tingyu66:
> @tingyu66 Thanks. I made a past. Can you confirm that everything runs through? I sadly still cannot test it :(

Thank you for updating `to_csc`. The test has passed locally on my end.

rusty1s:
Super, then we are good to merge. Can we update the other modules to look similar?

tingyu66:
> Super, then we are good to merge. Can we update the other modules to look similar?

Sure. I will make changes accordingly.

tingyu66:
Shall we merge this in? Still need 1 approving review.

rusty1s (approved):

PR #6412, opened 2023-01-12
Add a `GATConv` model using the primitives in `cugraph-ops`

tingyu66:
This PR adds a GAT model `torch_geometric.nn.GATConvCuGraph` that leverages the optimized aggregation function in cugraph-ops, similar to https://github.com/pyg-team/pytorch_geometric/pull/6278.

Fixes https://github.com/rapidsai/cugraph-ops/issues/182.


tingyu66:
@rusty1s GATConv updated. I removed add_self_loops option here, since it is better for users to change the COO directly IMO.

rusty1s (approved):

PR #6420, opened 2023-01-14
Restructure documentation navigation

rusty1s:


wsad1 (approved):
Nice I like this new structure. Makes our docs more readable.

PR #6431, opened 2023-01-15
feat: create pyproject.toml

SauravMaheshkar:
This PR aims to minimize the number of configuration files for the project. Instead of a separate `.style.yapf`, `.coveragerc` and `pyrightconfig.json`, we now would have all the configurations within the `pyproject.toml` file leading to a overall minimal code structure. All the isort, pytest and pylint configurations were also moved to `pyproject.toml`.

I managed to transfer most of the content from `setup.cfg` into `pyproject.toml` but flake8 doesn't support it yet (https://github.com/PyCQA/flake8/issues/234) and I couldn't figure out if `aliases` are supported.

Some updates to the `.pre-commit-config.yaml` were needed such as the addition of a "check-toml" hook and the addition of `toml` as an additional dependency for the `yapf` hook.


rusty1s (approved):
Wow, what a great contribution. Thank you very much for the clean-up!

SauravMaheshkar:
Thank you 😄

PR #6433, opened 2023-01-16
Allow batch explanation on one mask for `CaptumExplainer`

RBendias:
Point 7 in #6375.


test/explain/algorithm/test_captum_explainer.py
```diff
@@ -103,12 +115,41 @@ def test_captum_explainer_multiclass_classification(
...
+
+@withPackage('captum')
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+def test_captum_explainer_multiclass_classification_batch(
+    node_mask_type,
+    edge_mask_type,
+):
+    index = [2, 3]
+    task_level = 'node'
```
rusty1s, line 129: Pass directly.

test/explain/algorithm/test_captum_explainer.py
```diff
@@ -103,12 +115,41 @@ def test_captum_explainer_multiclass_classification(
...
-        assert 'edge_mask' not in explanation
+
+@withPackage('captum')
+@pytest.mark.parametrize('node_mask_type', node_mask_types)
+@pytest.mark.parametrize('edge_mask_type', edge_mask_types)
+def test_captum_explainer_multiclass_classification_batch(
+    node_mask_type,
+    edge_mask_type,
+):
+    index = [2, 3]
```
rusty1s, line 128: Doesn't this need to be a `torch.Tensor`?

torch_geometric/explain/config.py
```diff
@@ -18,6 +18,13 @@ class MaskType(Enum):
     attributes = 'attributes'
 
 
+class MaskLevelType(Enum):
```
rusty1s, line 21: This feels kinda specific to `captum` - shall we move it there?

torch_geometric/nn/models/captum.py
```diff
@@ -13,8 +13,10 @@
 
 
 def to_captum_model(
-    model: torch.nn.Module, mask_type: str = "edge",
-    output_idx: Optional[int] = None, metadata: Optional[Metadata] = None
+    model: torch.nn.Module,
+    mask_type: str = "edge",
```
rusty1s, line 17: `MaskLevelType`?

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -59,6 +60,10 @@ def _get_mask_type(self):
             mask_type = 'edge'
         return mask_type
 
+    def _support_multiple_indices(self):
+        "Check if the method supports multiple indices."
```
rusty1s, line 64:
```suggestion
        r"""Checks if the method supports multiple indices."""
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -70,6 +75,20 @@ def forward(
         **kwargs,
     ) -> Explanation:
 
+        # Check if the method supports multiple indices.
+        if isinstance(index, list) and len(index) > 1:
```
rusty1s, line 79: Following the `Explainer` inteface, `index` needs to be `int` or `Tensor`, so we should check via `index.numel() > 1`

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -70,6 +75,20 @@ def forward(
...
+            if not self._support_multiple_indices():
+                raise ValueError(
+                    f"{self.attribution_method.__name__} does not "
+                    "support multiple indices. Please use a single "
+                    "index or a different method.")
+            if self.kwargs.get('internal_batch_size') != 1:
+                warnings.warn("Overriding 'internal_batch_size' to 1")
```
rusty1s, line 86: I don't think we want to warn in case `internal_batch_size` was not set.

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -70,6 +75,20 @@ def forward(
...
+                self.kwargs['internal_batch_size'] = 1
+            warnings.warn(
+                "For all indices, the explanations are computed on the"
+                "same mask. This can lead to overlapping explanations.")
+
```
rusty1s, line 91: That would be the case for every explainer, so I am not sure I understand the warning here.

rusty1s (approved):
Looks mostly good to me. Thanks!

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -70,6 +75,16 @@ def forward(
         **kwargs,
     ) -> Explanation:
 
+        if isinstance(index, torch.Tensor) and (index.numel() > 1):
```
rusty1s, line 78:
```suggestion
        if isinstance(index, torch.Tensor) and index.numel() > 1:
```

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -70,6 +75,16 @@ def forward(
...
+            if not self._support_multiple_indices():
+                raise ValueError(
+                    f"{self.attribution_method.__name__} does not "
+                    "support multiple indices. Please use a single "
+                    "index or a different method.")
+            if self.kwargs.get('internal_batch_size', 1) != 1:
```
rusty1s, line 84: We probably need to check if that argument is available for the given explainer. Shall we leave a TODO?

rusty1s (approved):
LGTM.

torch_geometric/explain/algorithm/captum_explainer.py
RBendias: Currently, we only support IG. When adding the support for more methods, I'll adjust those lines (added a reminder in the overall CaptumExplainer issue).

PR #6462, opened 2023-01-18
feat(ci): enable `pip` caching in CI 

SauravMaheshkar:
Set the `cache` argument to `pip` to enable caching in CI


rusty1s:
I tested this a while ago as well, but figured out it actually does not bring any gains (CI installation runtime got worse actually).

SauravMaheshkar:
> I tested this a while ago as well, but figured out it actually does not bring any gains (CI installation runtime got worse actually).

Interesting, the motivation behind [a628119](https://github.com/pyg-team/pytorch_geometric/pull/6462/commits/a6281192253daf58426c863a161a23a4872d837a), was to check if maybe the addition of `cache-dependency-path: setup.py` leads to any improvement.

rusty1s:
It doesn't look like it uses any caches in the "Install main package" section [here](https://github.com/pyg-team/pytorch_geometric/actions/runs/3958254514/jobs/6779598614), does it?

SauravMaheshkar:
The required logs for caching can be found under the "Setup packages" section [here](https://github.com/pyg-team/pytorch_geometric/actions/runs/3958254514/jobs/6779598614#step:4:23)

```
/opt/hostedtoolcache/Python/3.7.15/x64/bin/pip cache dir
/home/runner/.cache/pip
Received 41943040 of 287710919 (14.6%), 40.0 MBs/sec
Received 180355072 of 287710919 (62.7%), 85.8 MBs/sec
Received 287710919 of 287710919 (100.0%), 93.6 MBs/sec
Cache Size: ~274 MB (287710919 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/7770fbf4-3547-4b4f-886d-852607a1d65f/cache.tzst -P -C /home/runner/work/pytorch_geometric/pytorch_geometric
Cache restored successfully
```

rusty1s:
Yes, but this only caches PyTorch/extensions, not the dependencies of the main package (in "Install Main Packages"). Let me know if I am missing something.

SauravMaheshkar:
Yes, I see it now. I don't know why that is occurring though. It doesn't make sense if it only **partly** caches the dependencies. [`numpy`](https://github.com/pyg-team/pytorch_geometric/actions/runs/3971590261/jobs/6808655125#step:5:41) and [`scipy`](https://github.com/pyg-team/pytorch_geometric/actions/runs/3971590261/jobs/6808655125#step:5:30) seem to be cached somehow.

SauravMaheshkar:
@rusty1s I added some more paths in [b6cac4a](https://github.com/pyg-team/pytorch_geometric/pull/6462/commits/b6cac4a0856f3f5043b0b525eb19d54800efb180), maybe that helps ?

SauravMaheshkar:
@rusty1s I think I figured out the issue, The "Install main package" section of the pytest step doesn't cache the dependencies because it's doing a editable install of the package. (as opposed to pulling from a package index such as pypi)

rusty1s:
This is a great catch! It makes probably sense to move to a non-editable install. Can we change that?

SauravMaheshkar:
> This is a great catch! It makes probably sense to move to a non-editable install. Can we change that?

We could just change the install command to `pip install "torch-geometric[full, test]"` but wouldn't that install a pip wheel from pypi rather than the changed version in the branch ? 

I think we can install the packages required for testing (the `[test]` option) using the aforementioned change (just caching the test dependencies should speed up the "Minimal Installation" workflow), but we can't get around editable installs (unless I'm missing something here).

rusty1s:
Oh, I don't think it has to editable for testing. I thought of doing just `pip install .[full, test]`. Wouldn't this be sufficient?

SauravMaheshkar:
> Oh, I don't think it has to editable for testing. I thought of doing just `pip install .[full, test]`. Wouldn't this be sufficient?

Let's try ([0db7160](https://github.com/pyg-team/pytorch_geometric/pull/6462/commits/0db716017d086bbf4f25200e6f5c3a6e52388b48))

SauravMaheshkar:
@rusty1s any clue as to why code coverage is failing on this PR ?

rusty1s:
Looks like `pytest` fails to generate a coverage report without editable installation. Mmmh, not really sure why, but I guess there is a way to fix this. Let me look into as well.

SauravMaheshkar:
Based on the latest changes we can see that the cache is being hit, based on the `cache-hit` output of each step and also because of the use of `check-latest: true` property.

SauravMaheshkar:
@rusty1s more context on why the packages mentioned in the `setup.py` aren't being cached but those that are constrained (such as `numpy`, `psutil` and `protobuf`) are. 

> The requirements file format allows for specifying dependency versions using logical operators (for example chardet>=3.0.4) or specifying dependencies without any versions. In this case the pip install -r requirements.txt command will always try to install the latest available package version. To be sure that the cache will be used, please stick to a specific dependency version and update it manually if necessary.

So it seems like only those that are constrained with versions will be cached while those that aren't are reinstalled each time to be the latest version as mentioned in the docs.

rusty1s (approved):

Issue #6467, opened 2023-01-19
BitcoinOTC Dataset not working


andrinr:
### 🐛 Describe the bug

Ro reproduce: 

```
from torch_geometric.datasets import BitcoinOTC

BitcoinOTC(root="src/data")
```

Error message:

> Processing...
> Traceback (most recent call last):
>   File "/home/ar/projects/sign-prediction/src/tests.py", line 10, in <module>
>     BitcoinOTC(root="src/data")
>   File "/home/ar/miniconda3/envs/gnn/lib/python3.10/site-packages/torch_geometric/datasets/bitcoin_otc.py", line 42, in __init__
>     super().__init__(root, transform, pre_transform)
>   File "/home/ar/miniconda3/envs/gnn/lib/python3.10/site-packages/torch_geometric/data/in_memory_dataset.py", line 56, in __init__
>     super().__init__(root, transform, pre_transform, pre_filter, log)
>   File "/home/ar/miniconda3/envs/gnn/lib/python3.10/site-packages/torch_geometric/data/dataset.py", line 94, in __init__
>     self._process()
>   File "/home/ar/miniconda3/envs/gnn/lib/python3.10/site-packages/torch_geometric/data/dataset.py", line 211, in _process
>     self.process()
>   File "/home/ar/miniconda3/envs/gnn/lib/python3.10/site-packages/torch_geometric/datasets/bitcoin_otc.py", line 74, in process
>     edge_attr = torch.tensor(edge_attr, dtype=torch.long)
> TypeError: 'float' object cannot be interpreted as an integer

### Environment

* PyG version: pyg-nightly 2.2.0.dev2023011
* PyTorch version: torch 1.12.1
* OS: WSL Ubuntu 20.04
* Python version: Python 3.10.8
* CUDA/cuDNN version: =
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`):

rusty1s:
Thanks. Although it is working for me I pushed a potential fix in #6470. Let me know if it is working :)

rusty1s:
Oh, I wasn't aware you already sent a PR, I am so sorry :(

andrinr:
No worries, thanks for the fix :)

PR #6468, opened 2023-01-19
Fix BitcoinOTC dataset

andrinr:
Raw dataset has integer values as edge attributes, therefore using int() instead of float() is safe.

fixes: https://github.com/pyg-team/pytorch_geometric/issues/6467


rusty1s:
Duplicate of #6470.

PR #6478, opened 2023-01-20
fix examples `mnist_voxel_gri`d add keyword parameter

forest1040:
A sample `mnist_voxel_grid.py`, but I added it because it requires a keyword parameter in the `voxel_grid()` argument.


rusty1s (approved):
Thanks for the fix.

PR #6482, opened 2023-01-20
`RGCNConv`: Add information about dynamic shapes

bartekxk:


torch_geometric/nn/conv/rgcn_conv.py
```diff
@@ -59,6 +60,8 @@ class RGCNConv(MessagePassing):
         compensate.
         We advise to check out both implementations to see which one fits your
         needs.
+        RGCNConv can use dynamic shapes. If your device doesn't support dynamic
+        shapes, use FastRGCNConv.
```
wsad1, line 64: Could you elaborate on what you mean by `dynamic shapes`?

torch_geometric/nn/conv/rgcn_conv.py
```diff
@@ -102,6 +105,9 @@ def __init__(
         bias: bool = True,
         **kwargs,
     ):
+        warnings.warn('This RGCNConv can use dynamic shapes. If your device '
+                      'doesn\'t support dynamic shapes, use FastRGCNConv.')
+
```
wsad1, line 110: Lets not raise a warning, if its documented.

torch_geometric/nn/conv/rgcn_conv.py
bartekxk: Done.

torch_geometric/nn/conv/rgcn_conv.py
bartekxk: Done.

torch_geometric/nn/conv/rgcn_conv.py
rusty1s: I am also not sure what you mean by this. Do you have a pointer or some explanation?

torch_geometric/nn/conv/rgcn_conv.py
```diff
@@ -59,6 +59,9 @@ class RGCNConv(MessagePassing):
         compensate.
         We advise to check out both implementations to see which one fits your
         needs.
+        RGCNConv can use dynamic shapes. Dynamic shape means that the shape of
+        the interim tensors can be determined at runtime. If your device
+        doesn't support dynamic shapes, use FastRGCNConv.
```
wsad1, line 64: Still a little confused
1. `FastRGCNConv` also support dynamic shape, right? Like you can pass in `x` of shape `[in_channels, any shape]` and internally the shape of `x_i` depends on number of edges.
2. Have you tested this on some specific device where `FastRGCNConv` works and `RGCNConv` doesn't.

torch_geometric/nn/conv/rgcn_conv.py
bartekxk: Definition from https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#work_dynamic_shapes 
"Dynamic Shapes is the ability to defer specifying some or all tensor dimensions until runtime. "
For example You can see that torch.nonzero causes host synchronization for cuda device:
https://pytorch.org/docs/stable/generated/torch.nonzero.html

rusty1s: Got it. Can we add a link to the NVIDIA doc here to make this more understandable for readers?

torch_geometric/nn/conv/rgcn_conv.py
bartekxk: 1. I dont see any usage of dynamic shape. Can you send a place in code where it is used?
2. Yes, I tested it on device without support dynamic shapes. `FastRGCNConv` works and `RGCNConv` doesnt.

torch_geometric/nn/conv/rgcn_conv.py
bartekxk: Sure, added.

rusty1s (approved):

PR #6512, opened 2023-01-25
Change `DataLoader` workers with affinity to start at `cpu0` 

JakubPietrakIntel:
Changing default range for DataLoader workers with affinity to start at cpu0 instead of cpu1.
Latest results proved that there's no need to always leave a single core free for main process. Therefore changing it to default range starting at 0, simplifying the code.

### DIFF
--- a/torch_geometric/loader/node_loader.py
+++ b/torch_geometric/loader/node_loader.py
@@ -216,7 +216,7 @@ class NodeLoader(torch.utils.data.DataLoader):
         .. code-block:: python
 
             loader = NeigborLoader(data, num_workers=3)
-            with loader.enable_cpu_affinity(loader_cores=[1,2,3]):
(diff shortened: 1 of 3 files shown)


rusty1s (approved):

torch_geometric/loader/node_loader.py
```diff
@@ -216,7 +216,7 @@ def enable_cpu_affinity(self, loader_cores: Optional[List[int]] = None):
         .. code-block:: python
 
             loader = NeigborLoader(data, num_workers=3)
-            with loader.enable_cpu_affinity(loader_cores=[1,2,3]):
+            with loader.enable_cpu_affinity(loader_cores=[0,1,2]):
```
rusty1s, line 219:
```suggestion
            with loader.enable_cpu_affinity(loader_cores=[0, 1, 2]):
```

PR #6524, opened 2023-01-26
Automatic `n_id` and `e_id` attributes to mini-batches produced by `NodeLoader` and `LinkLoader`

yaoyaowd:
Keep the node index in the minibatch for more flexible model development and debug.

### DIFF
--- a/torch_geometric/loader/utils.py
+++ b/torch_geometric/loader/utils.py
@@ -191,6 +191,7 @@ def filter_custom_store(
     tensors = feature_store.multi_get_tensor(required_attrs)
     for i, attr in enumerate(required_attrs):
         data[attr.group_name][attr.attr_name] = tensors[i]
+        data[attr.group_name][f'{attr.attr_name}__index__'] = attr.index
 
     return data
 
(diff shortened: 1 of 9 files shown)


torch_geometric/loader/utils.py
```diff
@@ -191,6 +191,7 @@ def filter_custom_store(
     tensors = feature_store.multi_get_tensor(required_attrs)
     for i, attr in enumerate(required_attrs):
         data[attr.group_name][attr.attr_name] = tensors[i]
+        data[attr.group_name][f'{attr.attr_name}__index__'] = attr.index
```
mananshah99, line 194: We currently expose `input_id` on the returned `HeteroData` object that is equal to the node IDs that sampling begins from. I wonder if we may instead want to extend `input_id` or formalize it better to include this information as well.

If we went this route, the modifications would be in `node_loader` and `link_loader`, which may be cleaner as these loaders have the `HeteroSamplerOutput` and constructed `HeteroData` minibatch in their `filter_fn`.

Curious about what you feel is the most clean and extensible.

mananshah99 (commented):
Let's also add some tests?

torch_geometric/loader/utils.py
yaoyaowd: Add the test. I think it is tricky to move as you need to maintain two places. And different attr_name may have different index in the group. It is impossible to happen now but code interface is flexible in that way.

rusty1s (approved):

PR #6532, opened 2023-01-27
Add PyTorch SparseTensor support for `GINConv`, `SAGEConv`, and `GraphConv`

EdisonLeeeee:
This PR aims to add PyTorch SparseTensor support for `GINConv`, `SAGEConv`, and `GraphConv`, ~~and also update type hints in `message_and_aggregate` correspondingly~~.


rusty1s (approved):

EdisonLeeeee:
Testing failed because I changed the type hints from
```
def message_and_aggregate(self, adj_t: SparseTensor, x: Tensor) -> Tensor:
```
to 
```
def message_and_aggregate(self, adj_t: Adj, x: Tensor) -> Tensor:
```

@rusty1s Any solution?

rusty1s:
I think we should keep them for now since we don't support JIT for `torch.sparse.Tensor` anyway yet.

PR #6538, opened 2023-01-28
[Code Coverage] `NeighborSampler`

LukeLIN-web:
Add SparseTensor test in NeighborSampler

### DIFF
--- a/test/loader/test_neighbor_sampler.py
+++ b/test/loader/test_neighbor_sampler.py
@@ -7,43 +7,41 @@ from torch_geometric.nn.conv import GATConv, SAGEConv
 from torch_geometric.utils import erdos_renyi_graph
 
 
-def test_neighbor_sampler():
-    torch.manual_seed(12345)
+def test_neighbor_sampler_basic():
     edge_index = erdos_renyi_graph(num_nodes=10, edge_prob=0.5)
-    sparse_edge_index = SparseTensor.from_edge_index(edge_index)
+    adj_t = SparseTensor.from_edge_index(edge_index, sparse_sizes=(10, 10)).t()
     E = edge_index.size(1)
 
     loader = NeighborSampler(edge_index, sizes=[2, 4], batch_size=2)
-    sparse_loader = NeighborSampler(sparse_edge_index, sizes=[2, 4],
-                                    batch_size=2, collate_fn="test_init",
-                                    dataset="test_init")
     assert loader.__repr__() == 'NeighborSampler(sizes=[2, 4])'
     assert len(loader) == 5
-    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
 
     for batch_size, n_id, adjs in loader:
         assert batch_size == 2
         assert all(np.isin(n_id, torch.arange(10)).tolist())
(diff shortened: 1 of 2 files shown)


test/loader/test_neighbor_sampler.py
```diff
@@ -9,23 +10,32 @@
...
+    sparse_edge_index = SparseTensor.from_edge_index(edge_index)
     E = edge_index.size(1)
 
     loader = NeighborSampler(edge_index, sizes=[2, 4], batch_size=2)
+    sparse_loader = NeighborSampler(sparse_edge_index, sizes=[2, 4],
+                                    batch_size=2, collate_fn="test_init",
+                                    dataset="test_init")
     assert loader.__repr__() == 'NeighborSampler(sizes=[2, 4])'
     assert len(loader) == 5
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
wsad1, line 22:
```suggestion
```
Lets drop device.

test/loader/test_neighbor_sampler.py
```diff
@@ -9,23 +10,32 @@
...
+            edge_index.to(device)
             assert int(edge_index[0].max() + 1) <= size[0]
             assert int(edge_index[1].max() + 1) <= size[1]
             assert all(np.isin(e_id, torch.arange(E)).tolist())
             assert e_id.unique().size(0) == e_id.size(0)
             assert size[0] >= size[1]
 
+    for batch_size, n_id, adjs in sparse_loader:
+        for (edge_index, e_id, size) in adjs:
+            edge_index.to(device)
```
wsad1, line 38: add assert stmts to check `edge_index`.

LukeLIN-web: I am sorry, what is `stmts`?

test/loader/test_neighbor_sampler.py
LukeLIN-web: This is because I want to test  `edge_index.to(device)`?

test/loader/test_neighbor_sampler.py
wsad1: Sorry stmts means statements. I mean just like we check if the returned `edge_index` is valid for `loader` can we do something similar for `sparse_loader` 
```
assert int(edge_index[0].max() + 1) <= size[0]
assert int(edge_index[1].max() + 1) <= size[1]
```

test/loader/test_neighbor_sampler.py
wsad1: We don't need to test if code works on different devices.

test/loader/test_neighbor_sampler.py
LukeLIN-web: > Sorry stmts means statements. I mean just like we check if the returned `edge_index` is valid for `loader` can we do something similar for `sparse_loader`
> 
> ```
> assert int(edge_index[0].max() + 1) <= size[0]
> assert int(edge_index[1].max() + 1) <= size[1]
> ```

Thank you! I add assert statements

rusty1s (approved):

PR #6541, opened 2023-01-28
`BAMultiShapes` dataset

steveazzolin:
This PR integrates the new dataset proposed in [Azzolin et al.](https://arxiv.org/abs/2210.07147)
Specifically:
- BAMultiShapesDataset loads the original dataset as defined in the paper
- BAMultiShapesGraph is a generator to create a new instance of the dataset

My original goal was to incorporate BAMultiShapesGraph into ExplainerDataset. However, the code of such Dataset seems tailored to generate node classification explainability datasets, and I found that including it would result in major modifications to its original implementation, which I was not particularly liking from a modular-code point of view. Hence, I proposed an ad-hoc Generator


rusty1s (approved):
Thank you! Can we add the graph generator code in a separate PR? I am not totally sure if that should be done as a `GraphGenerator` - the underlying implementation should expect a `GraphGenerator` and a list of `Motif`s as input to generate the dataset. WDYT?

PR #6555, opened 2023-01-31
[Code Coverage] `test_profile`

LukeLIN-web:
fix the bug in test       https://github.com/pyg-team/pytorch_geometric/blob/0b1ecb3d7779a18720307b5dff0330a9bbff5a6f/test/profile/test_profile.py#L53 
 `TypeError: cannot unpack non-iterable float object`


LukeLIN-web:
profileit() is not support CPU 
```
if device == -1:
>           raise RuntimeError(
                “The profiling decorator does not support profiling ”
                “on non CUDA devices”)
E           RuntimeError: The profiling decorator does not support profiling on non CUDA devices
../../torch_geometric/profile/profile.py:75: RuntimeError
```

rusty1s (approved):

PR #6560, opened 2023-02-01
`EquilibriumAggregation`: Padding tensor was not created on correct device

HelgeS:
In the `forward` method of the equilibrium aggregation, if `dim_size > index_size` a new tensor is created for padding, but without specification of the target device.
I added the device to be the same as `self.lamb.device` like it is done in the `init_output` method.

Example to reproduce:

```
import torch
from torch_geometric.nn.aggr import EquilibriumAggregation

print("Has cuda:", torch.cuda.is_available())

x = torch.rand(8, 4)
index = torch.tensor([0, 0, 0, 0, 2, 2, 2, 2])

aggr = EquilibriumAggregation(4, 4, [8])

print(
    "This works",
    aggr(x.to(torch.device("cpu")), index=index, dim_size=3).shape
    == torch.Size([3, 4]),
)
print(
    "This crashes",
    aggr(x.to(torch.device("cuda")), index=index, dim_size=3).shape
    == torch.Size([3, 4]),
)
```

Error message:
```
torch_geometric/nn/aggr/equilibrium.py", line 39, in forward
    inp = torch.cat([x, y[index]], dim=1)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument tensors in method wrapper_cat)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Fix incorrect device in `EquilibriumAggregration` ([#6560](https://github.com/pyg-team/pytorch_geometric/pull/6560))
 - Added bipartite graph support in `dense_to_sparse()` ([#6546](https://github.com/pyg-team/pytorch_geometric/pull/6546))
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
Thanks!

PR #6570, opened 2023-02-02
Fix pyg-nightly workflow

jreniecki:
The following error shows up when trying to install pyg-nightly:
`Requested pyg-nightly==2.3.0.dev20230130  has inconsistent version: expected '2.3.0.dev20230130', but metadata has '2.3.0'`

This change should solve the problem.


rusty1s (approved):
Thanks for fixing :)

PR #6573, opened 2023-02-02
[Code Coverage] `test_profiler`

LukeLIN-web:
rename  `group_by(events, keyfn):` to  `_group_by(events, keyfn):`
rename `test_get_gpu_memory_from_gc():`
fix print function.
add `test_profiler`

### DIFF
--- a/test/profile/test_profiler.py
+++ b/test/profile/test_profiler.py
@@ -1,8 +1,8 @@
-from torch_geometric.profile.profiler import Profiler
 import torch
 import torch.nn.functional as F
 
 from torch_geometric.nn import GraphSAGE
+from torch_geometric.profile.profiler import Profiler
 from torch_geometric.testing import onlyFullTest, withCUDA
 
 
--- a/test/profile/test_profiler.py
+++ b/test/profile/test_profiler.py
@@ -1,0 +1,21 @@
+from torch_geometric.profile.profiler import Profiler
+import torch
+import torch.nn.functional as F
+
+from torch_geometric.nn import GraphSAGE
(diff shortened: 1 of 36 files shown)


LukeLIN-web:
I thought that ` __str__` will be tested by `print`, but not , why?

test/profile/test_profiler.py
```diff
@@ -0,0 +1,20 @@
...
+    dataset = get_dataset(name='Cora')
+    use_cuda = torch.cuda.is_available()
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+    data = dataset[0].to(device)
+    model = GraphSAGE(dataset.num_features, hidden_channels=64, num_layers=3,
+                      out_channels=dataset.num_classes).to(device)
+    model.train()
+    with Profiler(model, profile_memory=True, use_cuda=use_cuda) as prof:
+        model(data.x, data.edge_index)
+        print(prof)
```
lightaime, line 20: Should we assert that memory usage is larger than zero? WDYT?

LukeLIN-web: > Should we assert that memory usage is larger than zero? WDYT?

It seems that the profiler currently doesn't have any other output interfaces except `__str__` for print. Maybe I could add it. I refer    [Pytorch profiler](https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html). It also only has print API.

rusty1s (approved):
This module is sadly a bit unfinished, so it's quite hard to add meaningful tests to it. I'm sorry :( Still, this looks good. Thank you very much!

PR #6595, opened 2023-02-04
Fixed a bug in `AsymmetricConv`

gravins:
Hi, 
I fixed a bug in antisymmetric_conv.
Specifically, I changed the definition of `self.eye` (gradients don't need to be computed for this Tensor) and fixed the doc.

Best,
Alessio


rusty1s (approved):
Thanks for the catch. Moved `eye` to a `Buffer`.

Issue #6582, opened 2023-02-03
The `process` and `download` methods are not called for `Dataset` descendant named `Dataset`


amotta:
### 🐛 Describe the bug

When creating a `Dataset` class that is named `Dataset` itself, then the `process` and `download` methods are not executed.

```python
import torch_geometric

class Dataset(torch_geometric.data.InMemoryDataset):
    def __init__(self, root, transform=None, pre_transform=None, pre_filter=None):
        self.did_download = self.did_process = False
        super().__init__(root, transform, pre_transform, pre_filter, log=False)

        print("download called: " + str(self.did_download))
        print("process called: " + str(self.did_process))

        # Because `download` and `processed` are not called by the
        # parent constructors, the files at `self.processed_paths`
        # do not exist at this point. Thus, the following attempt
        # to load the data will fail.

        # self.data, self.slices = torch.load(self.processed_paths[0])

    @property
    def raw_file_names(self):
        return ['trigger-download']

    @property
    def processed_file_names(self):
        return ['trigger-processing']

    def download(self):
        print(f"{__class__.__name__}.download")
        self.did_download = True

    def process(self):
        print(f"{__class__.__name__}.process")
        self.did_process = True

Dataset('/path/to/root')
```

Running the above produces the following output:
```
download called: False
process called: False
```

The expected behavior can be triggered by changing the class name from `Dataset` to anything else (e.g., `MyDataset`):

```python
class MyDataset(torch_geometric.data.InMemoryDataset):
    def __init__(self, root, transform=None, pre_transform=None, pre_filter=None):
        self.did_download = self.did_process = False
        super().__init__(root, transform, pre_transform, pre_filter, log=False)

    @property
    def raw_file_names(self):
        return ['trigger-download']

    @property
    def processed_file_names(self):
        return ['trigger-processing']

    def download(self):
        print(f"{__class__.__name__}.download")
        self.did_download = True

    def process(self):
        print(f"{__class__.__name__}.process")
        self.did_process = True

MyDataset('/path/to/root')
```

This produces the following output (before failing because `self.processed_dir` does not exist):
```
MyDataset.download
MyDataset.process
```

This "bug" (in the sense of unexpected behavior) is caused by following if conditions in the constructor of `torch_geometric.data.Dataset`:
https://github.com/pyg-team/pytorch_geometric/blob/ab3efed107ed76a38a59b11dfd6744eace5e7204/torch_geometric/data/dataset.py#L90-L94

[...]

PR #6598, opened 2023-02-05
Fix `download` and `process` in `torch_geometric.data.Dataset` when using a duplicated class name `Dataset`

EdisonLeeeee:
To fix https://github.com/pyg-team/pytorch_geometric/issues/6582. Not sure if it is a "bug" and should be fixed.

### test example
```python
import torch_geometric
class Dataset(torch_geometric.data.Dataset):
    def __init__(self):
        self.enter_download = False
        self.enter_process = False
        super().__init__()

    def _download(self):
        self.enter_download = True

    def _process(self):
        self.enter_process = True

    def download(self):
        pass

    def process(self):
        pass

ds = Dataset()
assert ds.enter_download
assert ds.enter_process
```


rusty1s (approved):
Thank you! I changed it slightly and introduced `has_download` and `has_process` properties so that one can customize it in case it is necessary.

PR #6600, opened 2023-02-05
[Code Coverage] `PNAConv` and `DegreeScalerAggregation`

zechengz:
PNAConv test coverage


test/nn/conv/test_pna_conv.py
```diff
@@ -36,6 +36,16 @@ def test_pna_conv():
         assert torch.allclose(jit(x, adj.t()), out, atol=1e-6)
 
 
+def test_pna_divide_input():
+    x = torch.randn(4, 16)
+    edge_index = torch.tensor([[0, 0, 0, 1, 2, 3], [1, 2, 3, 0, 0, 0]])
+    deg = torch.tensor([0, 3, 0, 1])
+    conv = PNAConv(16, 32, aggregators, scalers, deg=deg, divide_input=True,
```
lightaime, line 43: Should we add `divide_input` to `__repr__` and test `str(conv)`?

test/nn/conv/test_pna_conv.py
```diff
@@ -36,6 +36,16 @@ def test_pna_conv():
...
+                   train_norm=True)
```
lightaime, line 44: We can move `train_norm` test in [nn/aggr/test_scaler.py](https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/aggr/test_scaler.py)

torch_geometric/nn/conv/pna_conv.py
```diff
@@ -192,17 +193,16 @@ def __repr__(self):
...
             d = degree(data.edge_index[1], num_nodes=data.num_nodes,
                        dtype=torch.long)
-            max_degree = max(max_degree, int(d.max()))
-        # Compute the in-degree histogram tensor
-        deg_histogram = torch.zeros(max_degree + 1, dtype=torch.long)
-        for data in loader:
-            d = degree(data.edge_index[1], num_nodes=data.num_nodes,
-                       dtype=torch.long)
-            deg_histogram += torch.bincount(d, minlength=deg_histogram.numel())
+            d_bincount = torch.bincount(d, minlength=deg_histogram.numel())
```
lightaime, line 201: Awesome! Like it. We can also split this fix in a separate PR.

lightaime (commented):
Great work!! Leave a few minor comments.

test/nn/conv/test_pna_conv.py
zechengz: I am also not sure about that. It seems for other convs usually we don't add this kind of bool args to `__repr__`

test/nn/conv/test_pna_conv.py
zechengz: Sure!

torch_geometric/nn/conv/pna_conv.py
zechengz: I see, I will split the PR.

lightaime (approved):
LGTM!

rusty1s (approved):

test/nn/conv/test_pna_conv.py
lightaime: We do bool args sometime like `SignedConv`. Maybe @rusty1s could provide some suggestions on when we should add bool args to `__repr__`.

test/nn/conv/test_pna_conv.py
```diff
@@ -12,16 +13,17 @@
 ]
 
 
-def test_pna_conv():
+@pytest.mark.parametrize("divide_input", [True, False])
```
rusty1s, line 16:
```suggestion
@pytest.mark.parametrize('divide_input', [True, False])
```

test/nn/conv/test_pna_conv.py
rusty1s: I don't have a strong opinion. Usually, I prefer to only add important args to `__repr__` since otherwise the output of `print(torch.nn.Module(...))` becomes hard to parse.

lightaime: Cool. `divide_input` is not that important. Let's not add it.

PR #6606, opened 2023-02-06
[Benchmark] Add throughput, latency and accuracy for inference and training

yanbing-j:
This PR is to add throughput, latency for inference benchmark and throughput, latency and accuracy for training benchmark.


benchmark/training/training_benchmark.py
```diff
@@ -166,14 +215,22 @@ def run(args: argparse.ArgumentParser) -> None:
...
                                               progress_bar=progress_bar,
                                               desc=f"Epoch={epoch}")
+                                        if args.evaluate:
+                                            # In evaluate mode, throughput and latency are not accurate.
+                                            val_acc = test(
+                                                model, val_loader, device,
+                                                hetero,
+                                                progress_bar=progress_bar)
+                                            print(
+                                                f'Val Accuracy: {val_acc:.4f}')
```
rusty1s, line 233:
```suggestion
                                            print(f'Val Accuracy: {val_acc:.4f}')
```

benchmark/training/training_benchmark.py
```diff
@@ -77,6 +113,9 @@ def run(args: argparse.ArgumentParser) -> None:
         hetero = True if dataset_name == 'ogbn-mag' else False
         mask = ('paper', data['paper'].train_mask
                 ) if dataset_name == 'ogbn-mag' else data.train_mask
+        if args.evaluate:
+            val_mask = ('paper', data['paper'].val_mask
+                        ) if dataset_name == 'ogbn-mag' else data.val_mask
```
rusty1s, line 118:
```suggestion
            if dataset_name == 'ogbn-mag':
                val_mask = ('paper', data['paper'].val_mask)
            else:
                val_mask = data.val_mask
```

yanbing-j:
@rusty1s ,thanks for the review. I have fixed the flake8 issue.

benchmark/inference/inference_benchmark.py
```diff
@@ -145,16 +174,39 @@ def run(args: argparse.ArgumentParser) -> None:
...
-                                                    progress_bar=True)
+                                    y = model.inference(
+                                        subgraph_loader, device,
+                                        progress_bar=True)
+                                    if args.evaluate:
+                                        test_acc = model.test(
+                                            y, test_loader, device,
+                                            progress_bar=True)
+                                        print(f'Mini Batch Test Accuracy: \
+                                            {test_acc:.4f}')
```
rusty1s, line 194: Maybe I am missing something, but why do we need to perform inference twice?

benchmark/inference/inference_benchmark.py
```diff
@@ -145,16 +174,39 @@ def run(args: argparse.ArgumentParser) -> None:
...
                                                 str(hidden_channels),
                                                 str(num_neighbors))
+                        total_time = time.duration
+                        if args.num_steps != -1:
+                            total_num_samples = args.num_steps * batch_size
+                        else:
+                            total_num_samples = num_nodes
+                        throughput = total_num_samples / total_time
+                        latency = total_time / total_num_samples * 1000
+                        print(f'Throughput: {throughput:.3f} fps')
```
rusty1s, line 208:
```suggestion
                        print(f'Throughput: {throughput:.3f} it/s')
```

benchmark/training/training_benchmark.py
```diff
@@ -122,6 +165,25 @@ def run(args: argparse.ArgumentParser) -> None:
...
                         sampler=sampler,
                     )
+                    if args.evaluate:
+                        val_loader = NeighborLoader(
+                            data,
+                            num_neighbors=num_neighbors,
+                            input_nodes=val_mask,
+                            batch_size=batch_size,
+                            shuffle=shuffle,
+                            num_workers=args.num_workers,
```
rusty1s, line 175: Can we consider grouping common args into some `kwargs` dictionary?

torch_geometric/nn/models/basic_gnn.py
```diff
@@ -250,6 +250,16 @@ def inference(self, loader: NeighborLoader,
 
         return x_all
 
+    @torch.no_grad()
```
rusty1s, line 253: Not totally sure why we need to add this method to a model. Can we avoid this?

benchmark/inference/inference_benchmark.py
yanbing-j: `y` will be passed to `model.test`, so it will not perform inference twice. Spliting them is to record e2e time precisely.

benchmark/inference/inference_benchmark.py
yanbing-j: Here the units should be `samples/s`. My mistake.

torch_geometric/nn/models/basic_gnn.py
yanbing-j: This function can be moved out of the model, to become a common function.

rusty1s (approved):

PR #6609, opened 2023-02-06
Fix bugs in `PNAConv` and `DegreeScalerAggregation`

zechengz:
Some fixes, including bugs fixes and minor documentation fixes

### `torch_geometric/nn/aggr/scaler.py`
The `deg` tensor should be clamped during the `inverse_linear` computation and thus avoid division by zero (in case the `deg` tensor contains a zero degree value). Previous implementation will make the zero degree value equal to two in the `amplification` and `attenuation` cases where we also add one directly. Though I am not sure whether this way is correct and whether we also need to `clamp(1)` in the `linear` scaler.

### `torch_geometric/nn/conv/pna_conv.py`
The previous implementation utilized two for loops to compute the `deg_histogram` tensor. The first loop calculates the `max_degree`, while the second loop updates the `deg_histogram` by adding each `bincount`, with a `minlength` of `max_degree + 1`. However, for certain loaders (e.g. the `NeighborLoader`), the degree in the second loop may exceed the `max_degree` computed in the first loop and thus raise following error. Also, current implementation that replaces the two loops with one loop may potentially improve computation speed, although I don't verify this...
```
    deg_histogram += torch.bincount(d, minlength=deg_histogram.numel())
RuntimeError: The size of tensor a (17) must match the size of tensor b (19) at non-singleton dimension 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Fixed bugs in `PNAConv` and `DegreeScalerAggregation` ([#6609](https://github.com/pyg-team/pytorch_geometric/pull/6609))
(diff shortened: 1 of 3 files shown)


rusty1s (approved):
Thanks!

PR #6613, opened 2023-02-06
Fixing Subgraph Methods

toenshoff:
This pull request addresses a bug in both ```Data.subgraph``` and ```HeteroData.subgraph```,  which yield incorrect results when an unsorted list of integers is used to select a subgraph. In this case, the attributes of nodes are reordered according to the given list of integers, but the ids used in ```edge_index``` are *not*. Therefore, the node features will be distributed incorrectly in the output subgraph.
Note that ```Data.subgraph``` does use the ```utils.subgraph``` method with the ```relabel_nodes=True``` option to compute the reduced edge list. However, this is insufficient to ensure a correct output, as this only shifts node ids downward to ensure consecutive ids. It has no effect on the actual order of nodes, which causes this problem.

Consider the following example:
```python
import torch

data = Data(x=torch.tensor([0, 0, 1, 1]), edge_index=torch.tensor([[0, 2], [1, 3]]))
sub = data.subgraph(torch.tensor([0, 2, 1]))
print(sub.x)
print(sub.edge_index)
```
With torch_geometric==2.3.0 (current nightly) we obtain the following output:
```
x: [0 1 0]
Edge Idx:
 [[0]
 [1]]
```
Note that in the original graph each edge has a source and destination with the same node label. In the subgraph, the remaining edge has a source with label 0 and a destination with label 1. This is therefore not a valid subgraph.

I have addressed this bug by re-mapping the node ids in ```edge_index``` to their correct new index if the input is indeed an unsorted integer list. ```HeteroData.subgraph``` had the same issue with an analog solution. I also extended the tests to cover these issues.

Beyond this, I was wondering what the behaviour of ```Data.subgraph``` should be if the argument is a list of  integers with duplicate entries. Should the method raise an Exception, ignore duplicates or create duplicate vertices? I have not implemented anything in this regard yet, but I do think this should be specified.

### DIFF
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -598,6 +598,11 @@ class Data(BaseData, FeatureStore, GraphStore):
         else:
             num_nodes = subset.size(0)
 
+            # check if selection is unordered and re-map edge_index if necessary
+            if (subset[:-1] > subset[1:]).any():
+                node_idx = torch.argsort(subset)
+                edge_index = node_idx[edge_index]
+
         data = copy.copy(self)
 
         for key, value in self:
--- a/torch_geometric/data/hetero_data.py
+++ b/torch_geometric/data/hetero_data.py
@@ -615,6 +615,16 @@ class HeteroData(BaseData, FeatureStore, GraphStore):
                 return_edge_mask=True,
             )
 
+            # reorder source nodes if needed
+            if src_subset.dtype == torch.long and (src_subset[:-1] > src_subset[1:]).any():
+                src_node_idx = torch.argsort(src_subset)
+                edge_index[0] = src_node_idx[edge_index[0]]
+
+            # reorder destination nodes if needed
+            if dst_subset.dtype == torch.long and (dst_subset[:-1] > dst_subset[1:]).any():
+                dst_node_idx = torch.argsort(dst_subset)
+                edge_index[1] = dst_node_idx[edge_index[1]]
+
             for key, value in self[edge_type].items():
                 if key == 'edge_index':
                     data[edge_type].edge_index = edge_index
--- a/test/data/test_data.py
+++ b/test/data/test_data.py
@@ -208,6 +208,15 @@ def test_data_subgraph():
     assert torch.equal(out.edge_weight, edge_weight[torch.arange(2, 6)])
     assert out.num_nodes == 3
 
+    # test for unordered selection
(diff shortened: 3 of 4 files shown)


torch_geometric/data/data.py
```diff
@@ -598,6 +598,11 @@ def subgraph(self, subset: Tensor) -> 'Data':
         else:
             num_nodes = subset.size(0)
 
+            # check if selection is unordered and re-map edge_index if necessary
+            if (subset[:-1] > subset[1:]).any():
+                node_idx = torch.argsort(subset)
+                edge_index = node_idx[edge_index]
```
wsad1, line 604: Would simply calling `torch.sort` at the start of the function resolve this issue?

torch_geometric/data/hetero_data.py
```diff
@@ -615,6 +615,18 @@ def subgraph(self, subset_dict: Dict[NodeType, Tensor]) -> 'HeteroData':
...
+            if src_subset.dtype == torch.long and (src_subset[:-1] >
+                                                   src_subset[1:]).any():
+                src_node_idx = torch.argsort(src_subset)
+                edge_index[0] = src_node_idx[edge_index[0]]
+
+            # reorder destination nodes if needed
+            if dst_subset.dtype == torch.long and (dst_subset[:-1] >
+                                                   dst_subset[1:]).any():
+                dst_node_idx = torch.argsort(dst_subset)
+                edge_index[1] = dst_node_idx[edge_index[1]]
```
wsad1, line 628: Same comment as above. Will sorting each tensor resolve this?

wsad1 (commented):
Thanks for catching and fixing this.

torch_geometric/data/data.py
toenshoff: This would fix the bug, too. 

The solution here has the additional utility of allowing a user to change the order of vertices to a new order of their choice. If this is not considered useful for these methods, we could also simply switch to always sorting all integer lists.

wsad1: Users can change the order outside of this function if needed. So lets switch to sorting.

toenshoff: Fair enough. I have switched to using torch.unique at the beginning of both methods. This also addresses my question regarding handling duplicate indices.

Perhaps we should add a `reorder_nodes` method as a utility function? But this is a separate PR I guess.

wsad1 (approved):

PR #6633, opened 2023-02-07
Pytorch Sparse tensor support: `AntiSymmetricConv`, `CGConv`, and `TransformerConv`

EdisonLeeeee:
A minor change on `nn.MessagePassing` layer, where `edge_attr` and `edge_type` are `None` by default if the sparse adjacency matrix is 2-dimensional.


torch_geometric/nn/conv/message_passing.py
```diff
@@ -317,9 +317,9 @@ def __collect__(self, args, edge_index, size, kwargs):
             if out.get('edge_weight', None) is None:
                 out['edge_weight'] = values
             if out.get('edge_attr', None) is None:
-                out['edge_attr'] = values
+                out['edge_attr'] = None if values.dim() == 1 else values
             if out.get('edge_type', None) is None:
-                out['edge_type'] = values
+                out['edge_type'] = None if values.dim() == 1 else values
```
rusty1s, line 322: Can you explain? Not sure I understand.

EdisonLeeeee: Unlike `torch_sparse.SparseTensor`, `edge_attr` will never be `None` for PyTorch SparseTensor, which is not applicable for layers that accept `edge_attr`, e.g. `TransformerConv`
https://github.com/pyg-team/pytorch_geometric/blob/a5e5a489f9124dad22141cb8e0c8ac833320bb22/torch_geometric/nn/conv/transformer_conv.py#L220-L221

rusty1s, line 322:
```suggestion
                out['edge_type'] = values
```

rusty1s (approved):

torch_geometric/nn/conv/message_passing.py
rusty1s: Ah, I see. This is tricky. I reverted the `edge_type` chance back since the values are always expected to be one-dimensional, but otherwise, I think the fix looks good to me.

PR #6635, opened 2023-02-07
Fix `gcn_norm` for PyTorch SparseTensor input with gradients

EdisonLeeeee:
Test case
```python
import torch
from torch_geometric.utils import to_torch_coo_tensor

edge_index = torch.tensor([[0, 1, 2], [3, 4, 5]])
edge_weight = torch.ones(3).requires_grad_()
adj = to_torch_coo_tensor(edge_index, edge_weight)

adj.requires_grad
>>> True

gcn_norm(adj)[0].requires_grad
>>> True


rusty1s:
Thanks for the catch, this is weird. Can we add a test and a comment here?

EdisonLeeeee:
According to PyTorch docs, calling `torch.Tensor._values()` will return a detached tensor. To track gradients, `torch.Tensor.coalesce().values()` must be used instead.

rusty1s (approved):

Issue #6622, opened 2023-02-06
failing test_graph_norm


puririshi98:
### 🐛 Describe the bug

```
_______________________________ test_graph_norm ________________________________
    def test_graph_norm():
        x = torch.randn(200, 16)
        batch = torch.arange(4).view(-1, 1).repeat(1, 50).view(-1)
    
        norm = GraphNorm(16)
        assert norm.__repr__() == 'GraphNorm(16)'
    
        if is_full_test():
            torch.jit.script(norm)
    
        out = norm(x)
        assert out.size() == (200, 16)
        assert torch.allclose(out.mean(dim=0), torch.zeros(16), atol=1e-6)
        assert torch.allclose(out.std(dim=0, unbiased=False), torch.ones(16),
                              atol=1e-6)
    
        out = norm(x, batch)
        assert out.size() == (200, 16)
        assert torch.allclose(out[:50].mean(dim=0), torch.zeros(16), atol=1e-6)
>       assert torch.allclose(out[:50].std(dim=0, unbiased=False), torch.ones(16),
                              atol=1e-6)
E       assert False
E        +  where False = <built-in method allclose of type object at 0x7fdfae261bc0>(tensor([1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000,\n        1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000],\n       grad_fn=<StdBackward0>), tensor([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]), atol=1e-06)
E        +    where <built-in method allclose of type object at 0x7fdfae261bc0> = torch.allclose
E        +    and   tensor([1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000,\n        1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000],\n       grad_fn=<StdBackward0>) = <built-in method std of Tensor object at 0x7fe10b57b680>(dim=0, unbiased=False)
[...]

PR #6642, opened 2023-02-08
fix seed in `GraphNorm` tests

puririshi98:
solves https://github.com/pyg-team/pytorch_geometric/issues/6622


test/nn/norm/test_graph_norm.py
```diff
@@ -1,10 +1,12 @@
 import torch
 
+from torch_geometric import seed_everything
 from torch_geometric.nn import GraphNorm
 from torch_geometric.testing import is_full_test
 
 
 def test_graph_norm():
+    seed_everything(42)
```
rusty1s, line 9:
```suggestion
from torch_geometric.nn import GraphNorm
from torch_geometric.testing import is_full_test


def test_graph_norm():
    torch.manual_seed(42)
```

rusty1s (approved):

PR #6652, opened 2023-02-09
[Code Coverage] utils/convert.py

ZeynepP:
Adding tests for from/to networkit, trimesh and cugraph. Needed some updates in convert.py for cugraph.


ZeynepP:
#6653 is correct one I closed this one FYI

PR #6654, opened 2023-02-09
Fixed bug in `utils.bipartite_subgraph`

toenshoff:
The PR fixes a bug in the `utils.bipartite_subgraph` method, which crashes if `src_subset` contains Boolean values but `dst_subset` contains integers.

Consider the following example:
```python
import torch
from torch_geometric.utils import bipartite_subgraph

subset = (torch.tensor([True, True, False]), torch.tensor([0, 3]))
edge_index = torch.tensor([[0, 1, 2], [0, 0, 3]])
sub_edge_index = bipartite_subgraph(subset, edge_index)
```
With torch_geometric==2.3.0 we obtain the following error:
```
  File "/home/jan/git/gerd/pygtest.py", line 8, in <module>
    sub_edge_index = bipartite_subgraph(subset, edge_index)
  File "/home/jan/miniconda3/envs/gerd/lib/python3.9/site-packages/torch_geometric/utils/subgraph.py", line 178, in bipartite_subgraph
    edge_mask = src_subset[edge_index[0]] & dst_subset[edge_index[1]]
IndexError: index 3 is out of bounds for dimension 0 with size 2
```
The method determines whether or not a cast of the masks is needed by only checking the dtype of the source mask, which causes this issue.

While this seems like a rare edge case, this issue does cause the `HeteroData.subgraph` method to crash when `subset_dict` uses Boolean masks and does not contain some ntypes as keys, as integer tensors are then added for the missing ntypes.

Since https://github.com/pyg-team/pytorch_geometric/pull/6218, `HeteroData.subgraph` does retain ntypes that are absent from `subset_dict`. I have also updated the doc string of the method to reflect this change, since it was still describing the previous behavior.


torch_geometric/utils/subgraph.py
```diff
@@ -170,6 +170,7 @@ def bipartite_subgraph(
     if src_subset.dtype != torch.bool:
         src_size = int(edge_index[0].max()) + 1 if size is None else size[0]
         src_subset = index_to_mask(src_subset, size=src_size)
+    if dst_subset.dtype != torch.bool:
```
wsad1, line 173: Lets add a test where `src_subset` is of type bool and `dst_subset` is of type int.

wsad1 (approved):
Thanks for fixing.

torch_geometric/utils/subgraph.py
toenshoff: Sounds good. I just added the test case. I also expanded the test of `HeteroData.subgraph` to cover this issue.

test/data/test_hetero_data.py
```diff
@@ -245,6 +245,28 @@ def test_hetero_data_subgraph():
                                   data[key].edge_attr[edge_mask[key]])
             assert out[key].name == 'cites'
 
+    # test for mixed selection with ints, bools and missing ntypes
```
wsad1, line 248:
```suggestion
    # Test for bool and index in `subset_dict`.
```

torch_geometric/data/hetero_data.py
```diff
@@ -545,6 +545,8 @@ def subgraph(self, subset_dict: Dict[NodeType, Tensor]) -> 'HeteroData':
         r"""Returns the induced subgraph containing the node types and
         corresponding nodes in :obj:`subset_dict`.
 
+        If a node type is not a key in :obj:`subset_dict` then all nodes of that type remain in the graph.
+
```
wsad1, line 549:
```suggestion
        If a node type is not a key in :obj:`subset_dict` then all nodes of that
        type remain in the graph.
```

torch_geometric/data/hetero_data.py
```diff
@@ -545,6 +545,8 @@ def subgraph(self, subset_dict: Dict[NodeType, Tensor]) -> 'HeteroData':
         r"""Returns the induced subgraph containing the node types and
         corresponding nodes in :obj:`subset_dict`.
 
+        If a node type is not a key in :obj:`subset_dict` then all nodes of that
+        type remain in the graph.
```
wsad1, line 549:
```suggestion
        If a node type is not a key in :obj:`subset_dict` then all nodes of
        that type remain in the graph.

```

PR #6661, opened 2023-02-10
Adding `trim_to_layer` utility function

mszarma:
Adding hierarchical graph adjacency matrix feature to significantly improve performance

Contributors: @rBenke @andreazanetti


torch_geometric/loader/neighbor_loader.py
```diff
@@ -167,6 +167,8 @@ class NeighborLoader(NodeLoader):
             (2) it may slown down data loading,
             (3) it requires operating on CPU tensors.
             (default: :obj:`False`)
+        use_hierarchical_tensor (bool, optional): If set to :obj:`True`, will
+            return hierarchical tensor info
```
rusty1s, line 171:
```suggestion
            return hierarchical tensor info. (default: :obj:`False`).
```
Can we also give some more information here what this means?

torch_geometric/loader/node_loader.py
```diff
@@ -174,11 +184,37 @@ def filter_fn(
...
         else:
             raise TypeError(f"'{self.__class__.__name__}'' found invalid "
                             f"type: '{type(out)}'")
 
-        return data if self.transform is None else self.transform(data)
+        data = data if self.transform is None else self.transform(data)
+        if self.use_hierarchical_tensor:
+            if isinstance(out, SamplerOutput) and data.get('adj_t',
+                                                           None) is not None:
+                data.adj_t = HierarchicalSparseTensor.fromSparseTensor(
```
rusty1s, line 199:
```suggestion
                data.adj_t = HierarchicalSparseTensor.from_sparse_tensor(
```

torch_geometric/loader/node_loader.py
```diff
@@ -174,11 +184,37 @@ def filter_fn(
...
+                data.node_num_dict = out.node_num_dict
+
         else:
             raise TypeError(f"'{self.__class__.__name__}'' found invalid "
                             f"type: '{type(out)}'")
 
-        return data if self.transform is None else self.transform(data)
+        data = data if self.transform is None else self.transform(data)
+        if self.use_hierarchical_tensor:
+            if isinstance(out, SamplerOutput) and data.get('adj_t',
```
rusty1s, line 197: So this currently works for `SparseTensor` only?

torch_geometric/sampler/base.py
```diff
@@ -134,6 +138,8 @@ class SamplerOutput(CastMixin):
     col: Tensor
     edge: OptTensor
     batch: OptTensor = None
+    edge_num: OptTensor = None
+    node_num: OptTensor = None
```
rusty1s, line 142: Why not just `List[int]` instead?

rusty1s (commented):
Thanks for this PR. I have a few questions:

* Can we split this PR into multiple such that we first integrate `sampled_info` into the `NeighborLoader` before we think about integration in `MessagePassing`/benchmark scripts?
* I am not entirely sure why we need the concept of a `HierarchicalSparseTensor`. Can't we just let `NeighborLoader` return a list of sliced edge indices/sparse tensors?
* I would like to avoid adding any logic of this to `MessagePassing` and its instances. IMO, any customization of `x` and `edge_index` should happen outside of it. Would that be possible?

mszarma:
> Thanks for this PR. I have a few questions:
> 
> * Can we split this PR into multiple such that we first integrate `sampled_info` into the `NeighborLoader` before we think about integration in `MessagePassing`/benchmark scripts?
> * I am not entirely sure why we need the concept of a `HierarchicalSparseTensor`. Can't we just let `NeighborLoader` return a list of sliced edge indices/sparse tensors?
> * I would like to avoid adding any logic of this to `MessagePassing` and its instances. IMO, any customization of `x` and `edge_index` should happen outside of it. Would that be possible?

Thanks @rusty1s  for the review. Let's proceed first with pyg-lib part.

torch_geometric/utils/trim_to_layer.py
```diff
@@ -0,0 +1,104 @@
+from typing import Dict, Tuple, Union
+
+import torch
+from torch import Tensor
+from torch_sparse import SparseStorage
+
+from torch_geometric.typing import Adj, EdgeType, NodeType, SparseTensor
+
+
+def trim_to_i_layer(
```
rusty1s, line 10: I think this is the function we would want to keep in the first iteration.

I think the interface should be something like
```python
def trim_to_layer(
    layer: int,
    edge_index: Tensor,
    node_attrs: Union[Tensor, List[Tensor]],
    edge_attrs: Union[Tensor, List[Tensor]],
    num_nodes_per_layer: List[int],
    num_edges_per_layer: List[int],
):
```
such that we can support both node and edge features, and possibly multiple of them (e.g., narrowing down both `x` and `pos`).

Otherwise, this LGTM. If you want to desperately include `SparseTensor` code here, I think it is the most straightforward way to make use of the `torch_sparse.narrow` functionality to implement this. Let me know if you see any blockers with this.

mszarma: @rusty1s updated PR accordingly to discussed changes.

rusty1s (approved):

PR #6664, opened 2023-02-10
[Code Coverage] `loader/imbalanced_sampler.py`

zechengz:
Improve code coverage for the `ImbalancedSampler`


test/loader/test_imbalanced_sampler.py
```diff
@@ -36,6 +37,29 @@ def test_dataloader_with_imbalanced_sampler():
 
     assert torch.allclose(y, torch.cat([batch.y for batch in loader]))
 
+    # Test with list of Data where each y is a tensor as input:
```
wsad1, line 40:
```suggestion
    # Test with list of Data where each y is a tensor:
```

PR #6671, opened 2023-02-11
[Code Coverage] `transforms/random_link_split.py`

venomouscyanide:
Currently three snippets are missing coverage

1) `rev_edge_types = [None] * len(edge_types)`
I have added coverage for this.

2) 
```
        if hasattr(store, self.key):
            edge_label = store[self.key]
            edge_label = edge_label[index]
            # Increment labels by one. Note that there is no need to increment
            # in case no negative edges are added.
            if neg_edge_index.numel() > 0:
                assert edge_label.dtype == torch.long
                assert edge_label.size(0) == store.edge_index.size(1)
                edge_label.add_(1)
            if hasattr(out, self.key):
                delattr(out, self.key)
   ``` 
I have tried adding coverage for this, but, I am not able to get `assert edge_label.size(0) == store.edge_index.size(1)` to evaluate.

3) `del rev_store[key]  # We delete all outdated attributes.`
This is pending right now

Any help is appreciated.


rusty1s:
I think there was indeed a bug in the code of `RandomLinkSplit`. The `assert` didn't make much sense :)

rusty1s (approved):

PR #6675, opened 2023-02-11
[Code Coverage] `data/dataset.py`

LukeLIN-web:


LukeLIN-web:
If we want to test `dataset.get_summary()`
It is not satisfy the Minimal Installation / import
```
    def __repr__(self) -> str:
>       from tabulate import tabulate
E       ModuleNotFoundError: No module named 'tabulate'

torch_geometric/data/summary.py:81: ModuleNotFoundError
```

rusty1s (approved):

PR #6699, opened 2023-02-14
High-level API for `torch.sparse.mm` with optimized `spmm_reduce` kernel using CSR format

JakubPietrakIntel:
Related to important optimization in Pytorch:
:white_check_mark: [port sparse_mm.reduce to pytorch and optimize it on CPU #83727
](https://github.com/pytorch/pytorch/pull/83727)
Updating simplified high-level API for `spmm_reduce()` kernel and tests.

The current kernel implementation has limitation to process `src` of type `torch.Tensor` in `torch.sparse_csr` format, therefore I've added an option to auto-convert `src` to CSR format using `src.to_sparse_csr()`, which is `False` by default and will result in `ValueError` if the input is not provided in the correct format.

The conversion from `SparseTensor` to `torch.Tensor` is enabled by default for Pytorch > 1.13.

Added transfrom to remove duplicated in ogbn-products dataset, because the new kernel can't handle duplicate entries (useful for benchmarks).

_Re-opened this PR because the draft (#6689) needed to be scrapped after a rebase._


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## [2.3.0] - 2023-MM-DD
 
 ### Added
-
+- Added high-level API for the CPU optimized `spmm_reduce` using CSR format ([#6689](https://github.com/pyg-team/pytorch_geometric/pull/6689))
```
rusty1s, line 9:
```suggestion

- Added high-level API for the CPU optimized `spmm_reduce` using CSR format ([#6689](https://github.com/pyg-team/pytorch_geometric/pull/6689))
```

torch_geometric/transforms/__init__.py
```diff
@@ -19,6 +19,7 @@
 from .local_degree_profile import LocalDegreeProfile
 from .add_self_loops import AddSelfLoops
 from .remove_isolated_nodes import RemoveIsolatedNodes
+from .remove_duplicate_self_loops import RemoveDuplicateSelfLoops
```
rusty1s, line 22: Can we move this to a separate PR? Not totally sure I get the purpose of this transform yet. Wouldn't `RemoveDuplicatedEdges` be sufficient?

torch_geometric/utils/spmm.py
```diff
@@ -1,23 +1,23 @@
...
-def spmm(src, other, reduce):
-    # type: (SparseTensor, Tensor, str) -> Tensor
+def spmm(src, other, reduce, convert_to_csr):
+    # type: (Tensor, Tensor, str, bool) -> Tensor
     pass
 
 
-def spmm(src: Adj, other: Tensor, reduce: str = "sum") -> Tensor:
+def spmm(src: Adj, other: Tensor, reduce: str = "sum",
+         convert_to_csr: bool = False) -> Tensor:
```
rusty1s, line 20: Do we desperately need this argument? I assume PyTorch will convert to CSR internally anyway?

torch_geometric/utils/spmm.py
```diff
@@ -31,18 +31,20 @@ def spmm(src: Adj, other: Tensor, reduce: str = "sum") -> Tensor:
...
+        src = src.to_torch_sparse_csr_tensor(dtype=other.dtype)
+
+    if convert_to_csr:
+        src = src.to_sparse_csr()
+    else:
+        if not src.layout == torch.sparse_csr:
+            raise ValueError("`src` must be a `torch.Tensor` in "
+                             f"`torch.sparse_csr` format (got {src.layout}). "
+                             "To auto-convert use `convert_to_csr=True`")
+    return torch.sparse.mm(src, other, reduce)
```
rusty1s, line 50: Does this also come with CUDA paths?

torch_geometric/utils/spmm.py
```diff
@@ -31,18 +31,20 @@ def spmm(src: Adj, other: Tensor, reduce: str = "sum") -> Tensor:
...
-                         f"or a `torch.sparse.Tensor` (got {type(src)}).")
+    reduce = 'sum' if reduce == 'add' else reduce
 
-    if reduce in ['sum', 'add']:
-        return torch.sparse.mm(src, other)
-
-    # TODO: Support `mean` reduction for PyTorch SparseTensor
-    raise ValueError(f"`{reduce}` reduction is not supported for "
-                     f"`torch.sparse.Tensor`.")
+    if isinstance(src, SparseTensor):
```
yanbing-j, line 40: Both `torch_geometric/utils/__init__.py` and here need to check torch version.

torch_geometric/utils/spmm.py
JakubPietrakIntel: I will remove it and make the conversion obligatory. I thought that it might have been better for the user to control this, but you're right it - won't be necessary.

torch_geometric/utils/spmm.py
JakubPietrakIntel: ok

torch_geometric/utils/spmm.py
```diff
@@ -31,7 +34,10 @@ def spmm(src: Adj, other: Tensor, reduce: str = "sum") -> Tensor:
...
     :rtype: :class:`Tensor`
     """
-    assert reduce in ['sum', 'add', 'mean', 'min', 'max']
+    reduce = 'sum' if reduce == 'add' else reduce
+
+    if reduce not in ['sum', 'mean', 'min', 'max']:
+        raise ValueError(f"`reduce` argument '{reduce}' not supported")
 
     if isinstance(src, SparseTensor):
         return torch_sparse.matmul(src, other, reduce)
```
yanbing-j, line 43: Hi @JakubPietrakIntel  @rusty1s , May I know the logic here? If `src` is `SparseTensor` which is the default data type, it will never go into the spmm optimized impl in PT.

rusty1s: Yes, I changed this back for now since `torch.sparse.mm` is missing CUDA support. I think we can either patch this in `SparseTensor` or patch this here such that we only call `torch.sparse.mm` in case PyTorch >= 2.0 and CPU. Otherwise, I think this solution is fine, as we support `torch.sparse.Tensor` now anyway (and are in the process of removing `torch-sparse` dependency).

yanbing-j: Understand. So for now, we need add a patch to call `torch.sparse.mm` when PT >= 2.0 and CPU. Do you know the time of adding CUDA support?

rusty1s: I don‘t have any insights into that. Let me add the optimized routine in a separate PR.

rusty1s: Fixed this in https://github.com/pyg-team/pytorch_geometric/pull/6759

yanbing-j: Thanks. The fix LGTM.

PR #6706, opened 2023-02-15
Added the `GPRGNN` model

thupchnsky:
Added the `GPRGNN` model (https://arxiv.org/pdf/2006.07988.pdf) into `torch_geometric.nn.conv`.


torch_geometric/nn/conv/gprgnn.py
```diff
@@ -0,0 +1,167 @@
...
+    edge weights via the optional :obj:`edge_weight` tensor. Here
+    :math:`\gamma_k` denotes the learnable generalized PageRank (GPR)
+    weight for layer :math:`k`. There are four ways to initialize the
+    weights: :obj:`Delta`, Personalized PageRank (:obj:`PPR`), Negative
+    Personalized PageRank (:obj:`NPPR`), and :obj:`Random`.
+
+    Args:
+        K (int): Number of propagation layers :math:`K`.
+        alpha (float): Initialization parameter :math:`\alpha`.
+        init (str, optional): Initialization method for GPR weights.
```
EdisonLeeeee, line 40: Can we list all the optional values of the `init` argument in the doc-string as well?

torch_geometric/nn/conv/gprgnn.py
```diff
@@ -0,0 +1,167 @@
...
+from torch import Tensor
+from torch.nn import Parameter
+
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.conv.gcn_conv import gcn_norm
+from torch_geometric.typing import Adj, OptPairTensor, OptTensor, SparseTensor
+from torch_geometric.utils import spmm
+
+
+class GPRGNN(MessagePassing):
```
EdisonLeeeee, line 14: Let's also add some tests.

torch_geometric/nn/conv/gprgnn.py
```diff
@@ -0,0 +1,167 @@
...
+        return h
+
+    def message(self, x_j: Tensor, edge_weight: OptTensor) -> Tensor:
+        return x_j if edge_weight is None else edge_weight.view(-1, 1) * x_j
+
+    def message_and_aggregate(self, adj_t: SparseTensor, x: Tensor) -> Tensor:
+        return spmm(adj_t, x, reduce=self.aggr)
+
+    def __repr__(self) -> str:
+        return '{}(K={}, temp={})'.format(self.__class__.__name__, self.K,
```
EdisonLeeeee, line 166: Use f-string instead would be better.

EdisonLeeeee (commented):
Thanks for the PR. Left some initial comments.

torch_geometric/nn/conv/gprgnn.py
thupchnsky: Done.

torch_geometric/nn/conv/gprgnn.py
thupchnsky: Added two unittests in the test folder.

torch_geometric/nn/conv/gprgnn.py
thupchnsky: Changed to f-string.

torch_geometric/nn/conv/gprgnn.py
```diff
@@ -0,0 +1,166 @@
...
+from torch import Tensor
+from torch.nn import Parameter
+
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.conv.gcn_conv import gcn_norm
+from torch_geometric.typing import Adj, OptPairTensor, OptTensor, SparseTensor
+from torch_geometric.utils import spmm
+
+
+class GPRGNN(MessagePassing):
```
EdisonLeeeee, line 14: No strong opinion. I think we should name it something like `GPRConv` rather than `GPRGNN` as it sounds like a model instead of a layer.

EdisonLeeeee (approved):

torch_geometric/nn/conv/gprgnn.py
thupchnsky: I think it would be better to keep the same name as in the paper and their original implementation. Would keep the name for now.

PR #6712, opened 2023-02-15
Make `LinkX` model jittable

ftxj:
This PR make the LinkX model to be jittable and get better performance.

The original LinkX model cann't be JIT due to many reasons:
1. The type of second parameter `edge_index` of LinkX.forward is `Adj`. So we need to add `@torch.jit._overload_method` for this method.
2. The `edge_norm` and `edge_mlp` in LinkX model isn't always initialized. So we modify the initialize logic.
3. After we add `@torch.jit._overload_method` for forward, we need a wrapper of model. Overwise, the TorchScript can't find the forward function.
4. The return type of MLP is an `Union`, which will caused type error in TorchScript when we use the return value of MLP.

I think there might be a better way to make it jittable, but I haven't found yet.


rusty1s (approved):
This is cool, thank you!

If we plan to do this for other models, I think it would be a nice idea to out-source the functionality of creating a jittable parent module.

PR #6718, opened 2023-02-15
Add `SimpleConv` to perform non-trainable propagation

rusty1s:


torch_geometric/nn/conv/simple_conv.py
```diff
@@ -0,0 +1,91 @@
...
+        \mathbf{x}_j
+
+    Args:
+        aggr (str or [str] or Aggregation, optional): The aggregation scheme
+            to use, *e.g.*, :obj:`"add"`, :obj:`"sum"` :obj:`"mean"`,
+            :obj:`"min"`, :obj:`"max"` or :obj:`"mul"`.
+            In addition, can be any
+            :class:`~torch_geometric.nn.aggr.Aggregation` module (or any string
+            that automatically resolves to it). (default: :obj:`"sum"`)
+        combine_root (str, optional): Specified how to combine the central node
```
yaoyaowd, line 33: I'm a little uncomfortable with name `combine_root`. I think we can simply call it `kernel`. The kernel is used to combine representations. The kernel could be a `sum` kernel or `concat` kernel.

yaoyaowd (approved):

torch_geometric/nn/conv/simple_conv.py
rusty1s: I think `kernel` is confusing, especially since we already "combine" representations of neighbors. Any other idea? :)

yaoyaowd: how about `concat (bool, default False)`? or `fuse= sum / concat`? Otherwise we go with `combine_root`.

rusty1s: Mh, `concat` wouldn't work since I think we want to support both sum/concat. `fuse` would be irritating since I think of operator fusion in this case :D What about `group`/`group_root`/`combine`/`combine_root`?

PR #6727, opened 2023-02-16
Add TorchScript support for `RECT_L`

ftxj:
This PR add the TorchScript support for `RECT_L` model.

The fail reason and our solution for original code is very similar with PR [#6721](https://github.com/pyg-team/pytorch_geometric/pull/6712), except that this model using the `torch.jit.export` on `embed` and `get_semantic_labels` methods. And another fail reason is `@torch.no_grad`, which bring some error msg I cann't understand.

Adding TorchScript support will bring a lot of extra code and reduce the code readability. I will consider how to do better in another PR.


rusty1s (approved):
Thanks. I like it. Your comment about code duplication makes a lot of sense to me, although I am not entirely sure how to do better here. Not sure if we can elegantly auto-generate these wrapper classes.

PR #6734, opened 2023-02-16
Added revised `MD17` dataset

RobDHess:
### Why?

The original MD17 dataset contains numerical noise and training on more than 1000 data points has been judged to be bad practice. Therefore, [revised MD17](https://archive.materialscloud.org/record/2020.82) was created. Simulations have been re-run using a better numerical approximation. It was introduced in [this paper](https://arxiv.org/abs/2007.09593).

There are currently three versions of the dataset:

1. The original
2. The revised version
3. The CCSD(T) version

Current papers tend to ignore the original MD17 dataset and only test on rMD17 or the CCSD(T) versions. Papers such as [NequIP](https://arxiv.org/pdf/2101.03164.pdf) and [GemNet](https://arxiv.org/abs/2106.08903) talk about the downsides of the original dataset, for reference. 

### What?
The new version of the MD17 dataset is fully backwards compatible, but also includes the revised version of MD17. All the revised trajectories are available by changing the name from e.g. `benzene` to `revised benzene`. A warning has been added to not train on more than 1000 training samples, following the advice of the authors of the revised dataset and in accordance with common practice. All new trajectories have been added to the table and the cheatsheet.

### Choices I Have Made
Since the revised version consists of a single `.tar.bz2` file, I have changed the logic slightly during the download stages and for checking whether the files already exist. I have also added `self.ccsd` and `self.revised` properties to easily distinguish the handing of files. The table and cheatsheet have been re-ordered to better reflect the different versions of the dataset. 

### Possible Improvements
The authors of the revised dataset include indices fo five fold cross-validation. We could opt to include these, so it is easier for users to stick to the preferred experimental setup of 1000 training samples, while improving comparability between different works. I would suggest an optional `fold` argument that ranges from one to five.


rusty1s (approved):
Thank you very much. This is super valuable.

Happy to include split options in a separate PR (if you are interested).

PR #6749, opened 2023-02-18
Update `LabelPropagation` docs and `post_step` default value

zechengz:
The typing of the arguments for the `LabelPropagation` `forward` function is currently incorrect due to the use of a default lambda function. To address this, changing the default value to `None`, and moving the lambda function into the function itself. Additionally, adding documentation explain the `post_step` for the output.

image


torch_geometric/nn/models/label_prop.py
```diff
@@ -21,6 +21,13 @@ class LabelPropagation(MessagePassing):
 
     where unlabeled data is inferred by labeled data via propagation.
 
+    .. note::
+
+        For an example of using the label propagation, see
```
rusty1s, line 26:
```suggestion
        For an example of using the :class:`LabelPropagation`, see
```

torch_geometric/nn/models/label_prop.py
```diff
@@ -57,7 +81,10 @@ def forward(
             out = self.propagate(edge_index, x=out, edge_weight=edge_weight,
                                  size=None)
             out.mul_(self.alpha).add_(res)
-            out = post_step(out)
+            if isinstance(post_step, Callable):
```
rusty1s, line 84:
```suggestion
            if post_step is not None:
```

test/nn/models/test_label_prop.py
```diff
@@ -22,3 +22,7 @@ def test_label_prop():
     out = model(y, edge_index, mask)
     assert out.size() == (6, 3)
     assert torch.allclose(model(y, adj.t(), mask), out)
+
+    # Test post step:
+    out = model(y, edge_index, mask, post_step=lambda y: y.clamp_(0., 0.))
```
rusty1s, line 27:
```suggestion
    out = model(y, edge_index, mask, post_step=lambda y: torch.zeros_like(y))
```

rusty1s (approved):
Thanks. Noticed this as well, thanks for taking care of it.

PR #6758, opened 2023-02-21
Remove adding `num_edges` directly to store in the `Pad` transform

bwroblew:
Handling `num_edges` as an attribute requires couple more changes and doesn't work well with the current implementation. For example, when `num_edges` is set on the `Data` objects and those objects are merged into a `Batch` the `num_edges` attribute is a tensor with values from all the objects. It works well when implemented as a property of `EdgeStorage` and not as a direct attribute, so removing it for now.


rusty1s (approved):
Makes sense, there is some custom logic in `Batch` to obtain the real number of nodes, which I assume we would need to copy?

bwroblew:
Yes, I can work on that later and extend that logic. I guess for now it's not necessary and it was a mistake on my side to set it explicitly.
There are couple of places where `num_nodes` has a separate logic (for example when creating a subgraph) and similar logic should be probably needed for `num_edges`.
I guess it's best to remove it now and I can work on that when it's really needed.

Issue #6528, opened 2023-01-27
[Community Sprint] Improving Code Coverage 🚀


rusty1s:
### 🚀 The feature, motivation and pitch

**We are kicking off our third community sprint!**

This community sprint resolves around **improving test coverage** across the PyG code base.
Currently, our [tests cover 85.68%](https://app.codecov.io/gh/pyg-team/pytorch_geometric/tree/master/torch_geometric) of all code in PyG. The goal of the community sprint is to bump this number into the high 90s (and to get yourself more familiar with the various parts of the code base). Each individual contribution is designed to only take **around 30 minutes to two hours** to complete.

The sprint begins **Friday Janurary 27th** and will **last 2 weeks**. If you are interested in helping out, please also join our **[PyG slack](https://data.pyg.org/slack.html) channel `#code-coverage-sprint`** for more information.

**You can assign yourself to the test you are planning to work on [here](https://docs.google.com/spreadsheets/d/1QpwREo8_iQ0sfnXCa22i1b1XEM9Udxu7X16Z77eyRg8/edit#gid=1220861398) (code-coverage tab)**.

# 🚀 Improving Code Coverage

## Example

Take a look at the current [code coverage report](https://app.codecov.io/gh/pyg-team/pytorch_geometric/tree/master/torch_geometric) of PyG. For example, we can see that we never test the [`copy()`](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.data.InMemoryDataset.html#torch_geometric.data.InMemoryDataset.copy) function of the [`InMemoryDataset`](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.data.InMemoryDataset.html#torch_geometric.data.InMemoryDataset) class, see [here](https://app.codecov.io/gh/pyg-team/pytorch_geometric/blob/master/torch_geometric/data/in_memory_dataset.py):

![Screenshot 2023-01-26 at 15 51 43](https://user-images.githubusercontent.com/6945922/215070701-5d23382a-b194-4cb4-964e-7e7356ee5197.png)

As such, we create a `test_in_memory_dataset_copy()` function in `test/data/test_dataset.py` to add a corresponding test:

```python
def test_in_memory_dataset_copy():
    data_list = [Data(x=torch.randn(5, 16)) for _ in range(4)]
    dataset = MyTestDataset(data_list)

    copied_dataset = dataset.copy()

    # Test that we actually do a copy:
[...]

PR #6781, opened 2023-02-23
[Code Coverage] `models/dimenet.py`

SauravMaheshkar:
Part of #6528, improves typing and code coverage for DimeNet.


SauravMaheshkar:
@rusty1s it seems like shifting from `"` to `'` has somehow lead to error with coverage patch. How do you suggest we proceed ?

rusty1s:
Will look into. I think you should disable black though, we are not using it in PyG and the diffs can get a bit annoying :)

rusty1s (approved):

SauravMaheshkar:
> Will look into. I think you should disable black though, we are not using it in PyG and the diffs can get a bit annoying :)

Yes ofcourse, will disable black globally on my system.

PR #6787, opened 2023-02-24
Fix hard masking in `GNNExplainer` for link prediction tasks

rusty1s:


wsad1 (approved):
Looks good. 
Can you add an `assert` to `test_gnn_explainer` to make sure that nodes/edges that farther than the `num_gnn_layer` actually have zero node_mask/edge_mask.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -124,6 +124,14 @@ def _train(
             loss.backward()
             optimizer.step()
 
+            # In the first iteration, we collect the nodes and edges that are
+            # involved into making the prediction. These are all the nodes and
+            # edges with gradient != 0 (without regularization applied).
+            if i == 0 and self.node_mask is not None:
+                self.hard_node_mask = self.node_mask.grad != 0.0
+            if i == 0 and self.edge_mask is not None:
+                self.hard_edge_mask = self.edge_mask.grad != 0.0
```
wsad1, line 133:
```suggestion
            if i == 0:
                self.hard_node_mask = self.node_mask.grad != 0.0
                if self.edge_mask is not None:
                    self.hard_edge_mask = self.edge_mask.grad != 0.0
```
`node_mask` cannot be None, see line 116

rusty1s: Oh, right. Not sure why. I changed it to also support `node_mask == None`.

wsad1: Yeah thanks for fixing that. We should support `node_mask == None`.
Can we update the test to include the condition where `node_mask = None`.

Issue #6162, opened 2022-12-07
List node features can not be sliced with subgraph


euerka:
How to pass subgraph Data as a node feature into main graph Data?
`test_data=Data(edge_index=edge_index,pos=torch.tensor(np.array(pos)), graph=subgraphList)`
subgraphList is as follows:
```[Data(x=[5, 3], edge_index=[2, 8], pos=[5, 2]),
 Data(x=[1, 3], edge_index=[2, 0], pos=[1, 2]),
 ...
 Data(x=[10, 3], edge_index=[2, 18])]```
The issue is that when I tried to mask some nodes:
`mask_test_data=test_data.subgraph(mask)`
mask_test_data.graph is the same as test_data.graph, while the size of mask_test_data.pos is reduced from test_data.pos.

After testing, I found the issue is caused by the type of list. May I know whether there is another type of container in which I can store the subgraph Data, and it can auto-align with  `test_data.subgraph(mask)`?

[Slack Message](https://torchgeometricco.slack.com/archives/C01DN0B3B1N/p1670346485240969?thread_ts=1670346485.240969&cid=C01DN0B3B1N)

**Comment by rusty1s** [comment]
Thanks for the issue.

**Comment by cemunds** [comment]
Hi, I think I have a similar problem. Each node in my graph has a whole point cloud associated with it. So, for example, one graph would look like this:
```
Data(edge_index=[2, 2234], y=[1561], pos=[1561, 3], point_cloud=[1561],  x=[1561, 51])
```
Where `point_cloud` is itself a list of Data objects representing the point cloud for each node. Obviously, this gets too big to load into my GPU memory for bigger graphs, so I am testing some sampling strategies. When experimenting with `RandomNodeLoader` for example, in my training loop, I get a batch like this:
```
Data(edge_index=[2, 120], y=[65], pos=[65, 3], point_cloud=[1561], x=[65, 51], batch=[65], ptr=[8])
```
The point_cloud attribute still contains all point clouds in my graph instead of containing only the point clouds for the sampled subgraph. I looked into the `RandomNodeLoader` (which uses the `subgraph` function of the `Data` object) and noticed that only `Tensor` objects are considered when overwriting the `Data` object's fields according to the sampled subset. So, since the point_cloud field in my `Data` object is a vanilla python list of `Data` objects, it is not processed.

It would be very helpful to also have vanilla lists be considered or have a special data structure for fields in a `Data` object that itself is a list of `Data` objects.

[...]

PR #6810, opened 2023-02-27
Add `utils.select` and `utils.narrow`

cemunds:
This adds a unified interface to perform filtering of node attributes in order to support attributes of type `list` inside `Data` objects. For now, this is only used in `Data.subgraph`, `ClusterData` and `ClusterLoader`, but should probably be extended in the future, as discussed in issue #6162.


rusty1s (approved):

cemunds:
Thanks for the quick merge. I had a look at the changes you made and noticed that `cat_dim` might be undefined [here](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/loader/cluster.py#L167).

rusty1s:
You are right, I am sorry. Thanks for the check, will fix.

rusty1s:
Fixed this here: https://github.com/pyg-team/pytorch_geometric/pull/6830

PR #6813, opened 2023-02-28
Ishan

IshanSabane:
MessageQuant Updates


ares89:
hello @IshanSabane , 
I noticed this pr was closed. I'm curious to understand the reason behind the closure and I think the idea of a quantizable PyTorch Geometric is very impressive!
Thank you.

PR #6847, opened 2023-03-03
Added PyTorch Sparse Tensor support for `remove_self_loops`, `add_self_loops`, and `maybe_num_nodes`

EdisonLeeeee:
This PR aims to add PyTorch Sparse Tensor support for `remove_self_loops`, `add_self_loops`, and `maybe_num_nodes`, which will be used for supporting PyTorch Sparse Tensor in `torch_geometric.nn.conv`, e.g. `AGNNConv`:
https://github.com/pyg-team/pytorch_geometric/blob/09dbb20176b3de850317277329c63d756504c783/torch_geometric/nn/conv/agnn_conv.py#L65-L69

There are some questions to be clarified:
+ Should I also add support for other functions like `add_remaining_self_loops`?
+ Should I also update the doc-strings correspondingly?


test/utils/test_loop.py
```diff
@@ -32,6 +33,10 @@ def test_remove_self_loops():
     assert out[0].tolist() == expected
     assert out[1].tolist() == [[1, 2], [3, 4]]
 
+    adj = to_torch_coo_tensor(edge_index)
```
wsad1, line 36: `to_troch_coo_tensor` returns a sparse `torch.Tensor`. It would be nice to add a test for `SparseTensor` too.

wsad1 (approved):

test/utils/test_loop.py
EdisonLeeeee: Do you mean to add tests with `torch_sparse.SparseTensor` as input for these functions as well?

wsad1: Actually skip that. I guess we only want to support sparse `torch.Tensor` here, right?

EdisonLeeeee: Yes, exactly.

PR #6857, opened 2023-03-06
Add missing `=None` to `: Optional[T]`-annotated pooling functions

ekalosak:
This PR adds the missing default values for `Optional`-annotated arguments in nn.pool.glob pooling functions.

Without this change, for example the following error is raised:
```python3
> nn.pool.global_max_pool(x)
...
TypeError: global_max_pool() missing 1 required positional argument: 'batch'
```

The code for these functions already accommodates the case in which `batch is None`, so this little fix to the default args is sufficient to rectify the discrepancy between the type annotation semantics (`Optional`) and the function signature (requires a positional arg).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+
 ## [2.3.0] - 2023-MM-DD
 
 ### Added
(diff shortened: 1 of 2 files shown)


rusty1s:
I personally would like to avoid adding a default argument here, since this IMO increases the chance to misuse this function. In most cases, the `batch` argument is actual required, and so it should also be treated as a required argument.

rusty1s:
Note that there is no discrepancy between required arguments and `Optional[T]` annotations. The latter just means that you can also input `None`.

PR #6858, opened 2023-03-06
[Code Coverage] `data/lightning/datamodule.py`

zechengz:
Code coverage improvement for `data/lightning/datamodule.py`

It seems that for `torchmetrics.Accuracy`, `task` is required after `0.11.0` [[see this](https://torchmetrics.readthedocs.io/en/v0.11.0/classification/accuracy.html)] Should we also update the torchmetrics to the newest version (in the workflow)?

Also, only tested on a single GPU, not sure following line of code is correct
```python
offset += devices * math.ceil(1000 / (devices * batch_size))
```


test/data/lightning/test_datamodule.py
```diff
@@ -52,26 +55,32 @@ def forward(self, x, batch):
         x = self.lin2(x)
         return x
 
-    def training_step(self, data, batch_idx):
+    def training_step(self, data: Data, batch_idx: int):
```
wsad1, line 58:
```suggestion
    def training_step(self, data: Data):
```
Since we don't use `batch_idx`, can we drop it?

wsad1 (approved):
Thanks.

test/data/lightning/test_datamodule.py
zechengz: Thanks. It seems that even though `batch_idx` is not used, PyTorch Lightning still need that to index the batch of data (https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#training-step). I try to remove that but then it will raise an error
```
TypeError: test_step() takes 2 positional arguments but 3 were given
```

PR #6868, opened 2023-03-07
Added `mean` reduction to `utils.spmm` for PyTorch Sparse Tensor

EdisonLeeeee:


torch_geometric/utils/spmm.py
```diff
@@ -65,7 +65,11 @@ def spmm(src: Adj, other: Tensor, reduce: str = "sum") -> Tensor:
 
     if reduce == 'sum':
         return torch.sparse.mm(src, other)
+    elif reduce == 'mean':
+        src = src.coalesce()
+        deg = degree(src.indices()[0], num_nodes=src.size(0))
```
rusty1s, line 70: Can we add a TODO here to utilize `rowptr` in case `src` is given as a CSR matrix?

EdisonLeeeee: Done.

Issue #6782, opened 2023-02-23
new CI fails


puririshi98:
### 🐛 Describe the bug

```
=========================== short test summary info ============================
FAILED test/datasets/test_elliptic.py::test_elliptic_bitcoin_dataset - TypeEr...
FAILED test/utils/test_convert.py::test_to_cugraph[False-True-edge_weight1]
FAILED test/utils/test_convert.py::test_to_cugraph[False-False-edge_weight1]
FAILED test/utils/test_convert.py::test_from_cugraph[True-True-None] - assert...
FAILED test/utils/test_convert.py::test_from_cugraph[True-True-edge_weight1]
FAILED test/utils/test_convert.py::test_from_cugraph[True-False-None] - asser...
FAILED test/utils/test_convert.py::test_from_cugraph[True-False-edge_weight1]
FAILED test/utils/test_convert.py::test_from_cugraph[False-True-None] - asser...
FAILED test/utils/test_convert.py::test_from_cugraph[False-True-edge_weight1]
FAILED test/utils/test_convert.py::test_from_cugraph[False-False-None] - asse...
FAILED test/utils/test_convert.py::test_from_cugraph[False-False-edge_weight1]
FAILED test/utils/test_spmm.py::test_spmm_basic[device1] - AssertionError: as...
==== 12 failed
```

```
=================================== FAILURES ===================================
________________________ test_elliptic_bitcoin_dataset _________________________
get_dataset = functools.partial(<function load_dataset at 0x7f5c16405ee0>, '/tmp/pyg_test_datasets')
    @onlyFullTest
    def test_elliptic_bitcoin_dataset(get_dataset):
>       dataset = get_dataset(name='EllipticBitcoinDataset')
test/datasets/test_elliptic.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/conftest.py:41: in load_dataset
    return EllipticBitcoinDataset(path, *args, **kwargs)
/usr/local/lib/python3.8/dist-packages/torch_geometric-2.3.0-py3.8.egg/torch_geometric/datasets/elliptic.py:60: in __init__
    super().__init__(root, transform, pre_transform)
/usr/local/lib/python3.8/dist-packages/torch_geometric-2.3.0-py3.8.egg/torch_geometric/data/in_memory_dataset.py:57: in __init__
[...]

PR #6872, opened 2023-03-07
Solve `cugraph` test failures

puririshi98:
address part of https://github.com/pyg-team/pytorch_geometric/issues/6782


test/utils/test_convert.py
```diff
@@ -436,9 +436,9 @@ def test_to_cugraph(edge_weight, directed, relabel_nodes):
     cu_edge_index, cu_edge_weight = sort_edge_index(cu_edge_index,
                                                     cu_edge_weight)
 
-    assert torch.equal(edge_index, cu_edge_index)
```
rusty1s, line 439: Can you clarify the fix for me? I would expect that both `edge_index` and `cu_edge_index` are on CPU?

puririshi98: yeah im pretty sure cu_edge_index is coming out as cuda thats the issue. "As far as I can tell, the CI failures are because you are expecting a CPU tensor, but the cugraph edgelist is on the GPU.  That is the correct behavior from our end." i could also move them both back to cpu up to you, either one will fix

rusty1s: I mean we are calling `to_pandas` that hold `numpy` values, so I would pretty amazed if they are not on the CPU.

puririshi98: investigating further:
before sort cu_edge_index= tensor([[1, 0, 2, 1],
        [2, 1, 1, 0]], device='cuda:0')
so this implies [sorting the indices](https://github.com/pyg-team/pytorch_geometric/blob/fdb79725f303a36c22a548d918204041e95a133a/test/utils/test_convert.py#L477-L478) causes the dst nodes to be dropped for whatever reason.

[sorting code](https://github.com/pyg-team/pytorch_geometric/blob/fdb79725f303a36c22a548d918204041e95a133a/torch_geometric/utils/sort_edge_index.py#L71-L85)

puririshi98: figured it out: https://github.com/pyg-team/pytorch_geometric/blob/fdb79725f303a36c22a548d918204041e95a133a/torch_geometric/utils/sort_edge_index.py#L80-L85

in these lines it should return the edge_index and then edge_attr but because the edge_attr is None then it only returns edge_index but we index into the output of the function w/  
cu_edge_index, cu_edge_weight = sort_edge_index(cu_edge_index, cu_edge_weight)
which splits src into `cu_edge_index` and dst into `cu_edge_weight`

soln:
cu_edge_index = sort_edge_index(cu_edge_index, cu_edge_weight).cpu()

rusty1s: Got it. Guess we need to revisit the `sort_edge_index` interface. This is dangerous 😅

test/utils/test_convert.py
```diff
@@ -433,12 +433,15 @@ def test_to_cugraph(edge_weight, directed, relabel_nodes):
...
-    cu_edge_index, cu_edge_weight = sort_edge_index(cu_edge_index,
-                                                    cu_edge_weight)
+    if edge_weight is None:
+        cu_edge_index = sort_edge_index(cu_edge_index, cu_edge_weight)
+    else:
+        cu_edge_index, cu_edge_weight = sort_edge_index(
+            cu_edge_index, cu_edge_weight)
 
-    assert torch.equal(edge_index, cu_edge_index)
+    assert torch.equal(edge_index, cu_edge_index.cpu())
```
rusty1s, line 442: Do we still need this?

rusty1s (approved):

test/utils/test_convert.py
```diff
@@ -433,12 +433,15 @@ def test_to_cugraph(edge_weight, directed, relabel_nodes):
         cu_edge_weight = edge_list['weights'].to_pandas().values
         cu_edge_weight = torch.from_numpy(cu_edge_weight)
 
-    cu_edge_index, cu_edge_weight = sort_edge_index(cu_edge_index,
-                                                    cu_edge_weight)
+    if edge_weight is None:
+        cu_edge_index = sort_edge_index(cu_edge_index, cu_edge_weight)
```
rusty1s, line 437:
```suggestion
        cu_edge_index = sort_edge_index(cu_edge_index)
```

test/utils/test_convert.py
```diff
@@ -473,12 +476,14 @@ def test_from_cugraph(edge_weight, directed, relabel_nodes):
     )
 
     cu_edge_index, cu_edge_weight = from_cugraph(G)
+    if edge_weight is None:
+        cu_edge_index = sort_edge_index(cu_edge_index, cu_edge_weight)
+    else:
+        cu_edge_index, cu_edge_weight = sort_edge_index(
+            cu_edge_index, cu_edge_weight)
```
rusty1s, line 483:
```suggestion
            cu_edge_index)
```

PR #6874, opened 2023-03-08
Pytorch Sparse tensor support: `ClusterGCN`, `SAGEConv`, `AGNNConv`, `APPNP`, and `FeaStConv`

EdisonLeeeee:


torch_geometric/nn/conv/cluster_gcn_conv.py
```diff
@@ -69,17 +72,27 @@ def reset_parameters(self):
     def forward(self, x: Tensor, edge_index: Adj) -> Tensor:
         edge_weight: OptTensor = None
         if isinstance(edge_index, Tensor):
+            is_sparse_tensor = is_torch_sparse_tensor(edge_index)
+            if is_sparse_tensor:
+                edge_index, _ = to_edge_index(edge_index)
```
rusty1s, line 77:
```suggestion
                edge_index, _ = to_edge_index(edge_index.t())
```
wouldn't that work as well? Would need to drop the requirement to treat `col, row` differently in the later part of the code.

torch_geometric/utils/loop.py
```diff
@@ -201,6 +201,7 @@ def add_self_loops(
     is_sparse = is_torch_sparse_tensor(edge_index)
     if is_sparse:
         assert edge_attr is None
+        size = edge_index.size()
```
rusty1s, line 204:
```suggestion
        size = (edge_index.size(0), edge_index.size(1))
```
seems to be needed for TorchScript.

torch_geometric/nn/conv/cluster_gcn_conv.py
EdisonLeeeee: Got it!

rusty1s (approved):

PR #6879, opened 2023-03-08
Changed the interface of `coalesce` to only return single `edge_index` information in case the `edge_attr` argument is not specified

rusty1s:


EdisonLeeeee:
@rusty1s thank you for the update. I've been struggling with this discrepancy between different functions, such as `coalesce` and `add_self_loops`. 

However, this is indeed a breaking change in PyG. I'm concerned that there may be compatibility issues for users who upgrade from older versions of PyG.  Most importantly, the returned `edge_index` is unpackable, which means that the same code may cause potential errors when `edge_attr` is `None`. For example:
```python
# for PyG<=2.2
edge_index, _ = coalesce(...) # correct

# for PyG >=2.3
edge_index, _ = coalesce(...) # wrong but work properly, equivalent to row, col = edge_index
```
The worst-case scenario is that code written with an older version will still work in the latest version, but return incorrect results. Should we take steps to avoid this situation?

rusty1s:
Yes, this is a breaking change and will be highlighted in the release notes. I am not totally convinced it will impact a lot of users, as `sort_edge_index` and `coalesce` are more used internally, and it would only cause problems if you do `edge_index = coalesce(edge_index, None)` before (which seems rare). If you have any idea on how to make this backward compatible, I would be happy to adapt it.

EdisonLeeeee:
Thank you for your clarification. You are right. But I think we can add a warning in the doc-string, WDYT?

rusty1s:
Yes, sg. Will take care of it.

rusty1s:
Done in https://github.com/pyg-team/pytorch_geometric/pull/6893.

PR #6894, opened 2023-03-10
[Transforms] Add `NodePropertySplit` transform

gvbazhenov:
New functional for creating node-level data splits, which allows to induce distributional shifts in graph and conduct experiments with graph models in more challenging setups. It is implemented via the `NodePropertySplit` transform and allows to use the class itself or its static method `mask_nodes_by_property` (the latter is for users who already knows their structural node property and just need to perform data splitting).

Test coverage is provided and passed successfully. I decided to edit `transforms.rst` in `docs/source/modules` to make the static method `mask_nodes_by_property` visible in documentation, but not sure if this is the right way.


gvbazhenov:
I see that `networkx` is not a necessary dependency for `torch-geometric`. How can I fix the implementation so that the library and documentation build successfully?

rusty1s:
You can move the imports of `networkx` to the corresponding functions that make use of it. In this case, you probably need to also replace the `nx.Graph` type hints with something more general such as `Any`.

gvbazhenov:
As I can see, test fails on `import networkx as nx` when `NodePropertySplit` instance calls `to_networkx`. This is rather strange, since I just rely on the existing utilities in `utils/convert`. Could you please explain me what I am doing wrong?

test/transforms/test_node_property_split.py
```diff
@@ -0,0 +1,64 @@
...
+from itertools import combinations
+
+import numpy as np
+import torch
+
+from torch_geometric.datasets import graph_generator
+from torch_geometric.transforms import NodePropertySplit
+
+
+def test_node_property_split():
```
EdisonLeeeee, line 11: You might need to add a decorator `@withPackage('networkx')` here to add the `networkx` package during tests.

gvbazhenov: It works, thank you!

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+    :obj:`"out_valid_mask"`, :obj:`"out_test_mask"`.
+
+    This class implements 3 particular strategies for inducing
+    distributional shifts in graph — based on
+    **popularity**, **locality** or **density**.
+
+    Args:
+        property_name (str): The name of the node property to be used,
+            which must be ``'popularity'``, ``'locality'`` or ``'density'``.
+
```
EdisonLeeeee, line 39: Remove the blank lines from the code.

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+    :obj:`"in_test_mask"`, and the remaining 2 — for the OOD nodes:
+    :obj:`"out_valid_mask"`, :obj:`"out_test_mask"`.
+
+    This class implements 3 particular strategies for inducing
+    distributional shifts in graph — based on
+    **popularity**, **locality** or **density**.
+
+    Args:
+        property_name (str): The name of the node property to be used,
+            which must be ``'popularity'``, ``'locality'`` or ``'density'``.
```
EdisonLeeeee, line 38:
```suggestion
            which must be :obj:`'popularity'`, :obj:`'locality'` or :obj:`'density'`.
```

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+        property_name = "popularity"
+        part_ratios = [0.3, 0.1, 0.1, 0.3, 0.2]
+        tranaform = NodePropertySplit(property_name, part_ratios)
+
+        data = transform(data)
+    """
+    def __init__(
+        self,
+        property_name: str,
+        part_ratios: list,
```
EdisonLeeeee, line 73:
```suggestion
        part_ratios: List[float],
```

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+        ascending: Optional[bool] = True,
```
EdisonLeeeee, line 74:
```suggestion
        ascending: bool = True,
```

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+        random_seed: Optional[int] = None,
+    ):
+        assert property_name in [
+            "popularity",
+            "locality",
+            "density",
+        ], "`property_name` has to be 'popularity', 'locality', or 'density'"
+
+        assert sum(part_ratios) == 1.0, "`part_ratios` must sum to 1.0"
```
EdisonLeeeee, line 83: Add an assert statement to verify that the length of `part_ratios` is equal to 5.

torch_geometric/transforms/node_property_split.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.property_name = property_name
+        self.part_ratios = part_ratios
+        self.ascending = ascending
+        self.random_seed = random_seed
+
+    def __repr__(self) -> str:
```
EdisonLeeeee, line 90: Move this section of the code to the bottom.

EdisonLeeeee (approved):
LGTM. Thanks for the PR!

gvbazhenov:
@EdisonLeeeee @rusty1s thanks for your PR review and help!

gvbazhenov:
@rusty1s @EdisonLeeeee excuse me, could you please share when it will be possible to merge this pull request? Is there anything else that requires my participation?

rusty1s:
Will take a final look soon, but overall this looks good. Sorry for the delay, I was quite busy with PyG 2.3 release :(

gvbazhenov:
> Will take a final look soon, but overall this looks good. Sorry for the delay, I was quite busy with PyG 2.3 release :(

Roger that, thank you!

rusty1s (approved):
Thank you :)

PR #6898, opened 2023-03-12
feat: Add `ComplEx` KGE

riyavsinha:
Hello! we are hoping to add the ComplEx knowledge graph embedding to PyG as part of our CS224W final project at Stanford.

Our implementation largely follows that of TransE. We implemented complex vectors by breaking the real and imaginary components into separate vectors, but are open to other approaches.

We'd also like to seek input on whether an example is necessary like for TransE [here](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/transe_fb15k_237.py), and if so, whether it would be preferred to be in the same file or a new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added the `ComplEx` KGE model ([#6898](https://github.com/pyg-team/pytorch_geometric/pull/6898))
 - Added option to write benchmark results to csv ([#6888](https://github.com/pyg-team/pytorch_geometric/pull/6888))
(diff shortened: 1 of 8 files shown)


rusty1s (commented):
This looks great. I suggest we add a argument parser to the initial `TransE` example to support a variety of models. Would that work for you?

riyavsinha:
Hi Matthias! I think this is ready for review again. The [PapersWithCode](https://paperswithcode.com/paper/complex-embeddings-for-simple-link-prediction) metric for ComplEx for Hits@10 on fb15k-237 is .42, and our current optimizer settings in the example are consistently about .36-37 (we didn't want to increase the hidden size when tuning though to keep it fast as an example). If you'd like us to find a better hyperparameter set please let us know! 

In the current set, we used Adagrad which ended up performing much better than Adam and which was the suggested optimizer in the paper, and we added some L2 weight regularization as also suggested by the paper.

We have also updated the loss function to use BCELogits since that is more numerically stable

PR #6903, opened 2023-03-13
FLAG module

### LINKED ISSUE #6795: FLAG Data Augmentation Approach

Myself and a few collaborators would like to implement [FLAG](https://arxiv.org/pdf/2010.09891.pdf), an adversarial data augmentation strategy, and are looking for feedback on our architecture. 

We propose a FLAG module:
```python
class FLAG(torch.nn.Module)
```
where `FLAG` is instantiated with a number of hyperparameters, and the FLAG algorithm is run in the `forward` method. A training loop that uses FLAG could look something like this:

```python
flag = FLAG(optimizer, loss_fn, device)

for epoch in range(num_epochs):
    loss, out = flag(x, edge_index, y_true, model, train_idx, step_size, n_ascent_steps)
```
Note that calling `optimizer.zero_grad()`, `loss.backward()`, and `optimizer.step()` are handled in FLAG's `forward` method. Our code would be adapted from:
https://github.com/devnkong/FLAG/blob/main/deep_gcns_torch/examples/ogb/ogbn_arxiv/main.py#L56
https://github.com/devnkong/FLAG/blob/main/deep_gcns_torch/examples/ogb/attacks.py

We are planning to add it to the `contrib` package but have thought about putting it in `torch_geometric.utils.augmentation`. 

Any suggestions on architecture or where this should live would be helpful! 

### Alternatives

We considered extending `BaseTransform` but believe that approach wouldn't work since FLAG iteratively updates the perturbation, requires access to the model, optimizer, and loss function, and performs gradient updates. 

### Additional context

Below is the FLAG algorithm in pseudocode:
algorithm

Note that for now, we are only planning to handle graphs with continuous node features. In the FLAG paper, to handle discrete node features, the authors project the features into a continuous space and then augment the hidden embeddings. However, doing this requires model architecture changes. If there are any suggestions on extending the architecture for graphs with discrete node features, please let us know!

reislerjul:
Hi PyG Team, 

We have created a module for the FLAG data augmentation strategy. This implementation follows the issue that we created here: https://github.com/pyg-team/pytorch_geometric/issues/6795. 

This PR includes the FLAG module, callbacks for FLAG, unit tests, and updates the documentation to include FLAG.


torch_geometric/contrib/flag/flag.py
```diff
@@ -0,0 +1,249 @@
...
+            self.perturb = torch.nn.Parameter(
+                torch.zeros(
+                    *x.shape,
+                    dtype=torch.float,
+                    device=self.device,
+                ).uniform_(-step_size, step_size))
+        else:
+            perturb_tensor = torch.zeros(*x.shape, dtype=torch.float,
+                                         device=self.device)
+            labeled_values = torch.zeros(len(train_idx), x.shape[1],
```
EdisonLeeeee, line 91:
```suggestion
            labeled_values = torch.zeros(len(train_idx), x.size(1),
```

torch_geometric/contrib/flag/flag.py
```diff
@@ -0,0 +1,249 @@
...
+                is set, `step_size` will only be applied to the labeled nodes.
+            step_size_unlabeled (Optional[float]): `-step_size_unlabeled` and
+                `step_size_unlabeled` are the ranges when initializing the
+                perturbation for unlabeled nodes. If this is not set,
+                `step_size` is applied to both labeled and unlabeled nodes.
+        """
+        if step_size_unlabeled is None:
+            self.perturb = torch.nn.Parameter(
+                torch.zeros(
+                    *x.shape,
```
EdisonLeeeee, line 84:
```suggestion
                    *x.size(),
```

EdisonLeeeee (commented):
Thank you for this PR. I have some questions/suggestions:
+ Do we really need `FLAGCallback`? Most of its functionalities will not be used in most cases, and it makes the code more complicated.
+ Can we move `loss.backward` and `optimizer.step()` out of the code? For example:
```python
optimizer.zero_grad()
loss, _ = flag(...) # FLAG forward
loss.backward()
optimizer.step()
```
This would make FLAG more flexible and enable users to make minimal modifications whether they are using it or not.

.gitignore
```diff
@@ -24,6 +24,8 @@ coverage.xml
 *.out
 *.pt
 *.onnx
+*.swp
```
EdisonLeeeee, line 27:
```suggestion
```

.gitignore
```diff
@@ -24,6 +24,8 @@ coverage.xml
...
+*.swo
```
EdisonLeeeee, line 28:
```suggestion
```

torch_geometric/contrib/flag/flag.py
```diff
@@ -0,0 +1,249 @@
...
+                    dtype=torch.float,
+                    device=self.device,
```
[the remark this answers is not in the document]
reislerjul: Great, will remove the `flag.device` argument then.

torch_geometric/contrib/flag/flag.py
```diff
@@ -0,0 +1,249 @@
...
+                " step_size_unlabeled should be None or a positive float.")
+
+        # Code below adapted from:
+        #   https://github.com/devnkong/FLAG/blob/main/deep_gcns_torch
+        #     /examples/ogb/ogbn_arxiv/main.py#L56
+        #   and
+        #   https://github.com/devnkong/FLAG/blob/main/deep_gcns_torch
+        #     /examples/ogb/attacks.py
+
+        training_labels = y_true.squeeze(1)[train_idx]
```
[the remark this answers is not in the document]
reislerjul: Got it -- we'll add in code to handle both cases then.

torch_geometric/contrib/flag/flag.py
```diff
@@ -0,0 +1,249 @@
...
+        self._set_perturb(x, train_idx, unlabeled_idx, step_size,
+                          step_size_unlabeled)
+
+        self.model.train()
+        self.optimizer.zero_grad()
+
+        [c.on_ascent_step_begin(0, None) for c in self.callbacks]
+
+        # Perturb the input and calculate the loss
+        out = self.model(x + self.perturb, edge_index)[train_idx]
```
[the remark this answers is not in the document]
reislerjul: @EdisonLeeeee instead of explicitly adding those arguments, would it make sense for `flag.forward` to take in `**kwargs` which get passed to the model's `forward` function? 

Also, just for my own understanding, what is the `batch` argument used for in a model's forward function?

PR #6910, opened 2023-03-14
`Explain`: Update `CaptumExplainer` and associated tests.

wsad1:


test/explain/algorithm/test_captum_explainer.py
```diff
@@ -142,9 +174,31 @@ def test_captum_hetero_data(node_mask_type, edge_mask_type, index, hetero_data,
...
 
+    kwargs = {}
+    if method == 'Occlusion':
+        if edge_mask_type is None:
+            sliding_window_shapes = ((3, 3), (3, 3))
+        elif node_mask_type is None:
+            sliding_window_shapes = ((5, ), (5, ), (5, ))
+        else:
+            sliding_window_shapes = ((3, 3), (3, 3), (5, ), (5, ), (5, ))
+        kwargs['sliding_window_shapes'] = sliding_window_shapes
```
RBendias, line 185: As `Occlusion` is an unsupported method, we do not need those lines, correct?

torch_geometric/explain/algorithm/captum_explainer.py
```diff
@@ -80,17 +123,10 @@ def forward(
...
-                raise ValueError(
-                    f"{self.attribution_method.__name__} does not support "
-                    "multiple indices. Please use a single index or a "
-                    "different attribution method.")
-
-        # TODO (matthias) Check if `internal_batch_size` can be passed.
-        if self.kwargs.get('internal_batch_size', 1) != 1:
-            warnings.warn("Overriding 'internal_batch_size' to 1")
-        self.kwargs['internal_batch_size'] = 1
+        if self._has_internal_batch_size():
```
RBendias, line 126: I think this check is not necessary. The method will anyways fail if we set `internal_batch_size` in `kwargs` if the captum attribution method doesn't have `internal_batch_size` as argument.

RBendias (approved):
LGTM! Thanks! :)

test/explain/algorithm/test_captum_explainer.py
wsad1: Yeah that true removed them.

test/explain/algorithm/test_captum_explainer.py
```diff
@@ -142,9 +174,21 @@ def test_captum_hetero_data(node_mask_type, edge_mask_type, index, hetero_data,
...
+                node_mask_type=node_mask_type,
+                model_config=model_config,
+                explanation_type='model',
+            )
+        return
+
     explainer = Explainer(
         model,
-        algorithm=CaptumExplainer('IntegratedGradients'),
+        algorithm=CaptumExplainer(method, **kwargs),
```
wsad1, line 191:
```suggestion
        algorithm=CaptumExplainer(method),
```

PR #6913, opened 2023-03-14
Fix distributed batching example

ssmmnn11:
fix loss computation (output comes only from one task) in distributed batching example following [the tutorial](https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html?utm_source=distr_landing&utm_medium=FSDP_getting_started)

Output before:
Let's use 1 GPUs!
.
Epoch: 050, **Loss: 0.0985**, Val: 0.7773, Test: 0.7450

Let's use 4 GPUs!
.
Epoch: 050, **Loss: 0.0249**, Val: 0.7242, Test: 0.7662

Output after fix:
Let's use 1 GPUs!
.
Epoch: 050, **Loss: 0.0977**, Val: 0.7751, Test: 0.7707

Let's use 4 GPUs!
.
Epoch: 050, **Loss: 0.1002**, Val: 0.7724, Test: 0.7507

### DIFF
--- a/examples/multi_gpu/distributed_batching.py
+++ b/examples/multi_gpu/distributed_batching.py
@@ -92,8 +92,8 @@ def run(rank, world_size: int, dataset_name: str, root: str):
             loss = criterion(logits, data.y.to(torch.float))
             loss.backward()
             optimizer.step()
-            total_loss[0] += loss.item() * logits.size(0)
-            total_loss[1] += len(data)
+            total_loss[0] += float(loss) * logits.size(0)
+            total_loss[1] += data.num_graphs
 
         dist.all_reduce(total_loss, op=dist.ReduceOp.SUM)
         loss = total_loss[0] / total_loss[1]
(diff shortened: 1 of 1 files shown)


examples/multi_gpu/distributed_batching.py
```diff
@@ -84,18 +84,19 @@ def run(rank, world_size: int, dataset_name: str, root: str):
...
             loss = criterion(logits, data.y.to(torch.float))
             loss.backward()
             optimizer.step()
-            total_loss += float(loss) * logits.size(0)
-        loss = total_loss / len(train_loader.dataset)
+            total_loss[0] += loss.item() * logits.size(0)
+            total_loss[1] += len(data)
 
-        dist.barrier()
+        dist.all_reduce(total_loss, op=dist.ReduceOp.SUM)
```
wsad1, line 98: Does `all_reduce` remove the need for calling `barrier`?

ssmmnn11: I assumed that this is the case because it is all to all communication and 

from https://pytorch.org/docs/stable/_modules/torch/distributed/distributed_c10d.html#all_reduce
```
def all_reduce(tensor, op=ReduceOp.SUM, group=None, **async_op=False**):
    """
    Reduces the tensor data across all machines in such a way that all get
    the final result.

    After the call ``tensor`` is going to be bitwise identical in all processes.

```

but after the all_reduce there is a block where only rank 0 does work and then there is another barrier in the end where the ranks should be synchronised again.

PR #6936, opened 2023-03-17
PyTorch Sparse Tensor support:  `LEConv`, `LGConv`, `NNConv`, `PANConv`, `SignedConv`, and `WLConv`

EdisonLeeeee:


torch_geometric/nn/conv/pan_conv.py
```diff
@@ -66,8 +66,12 @@ def forward(self, x: Tensor,
 
         adj_t: Optional[SparseTensor] = None
         if isinstance(edge_index, Tensor):
-            adj_t = SparseTensor(row=edge_index[1], col=edge_index[0],
-                                 sparse_sizes=(x.size(0), x.size(0)))
+            if is_torch_sparse_tensor(edge_index):
+                # TODO: handle PyTorch sparse tensor directly
+                adj_t = SparseTensor.from_torch_sparse_coo_tensor(edge_index)
```
wsad1, line 71: Can we assume the user will always pass a `coo` tensor?

wsad1 (commented):
It looks like we are making the assumption that if `is_torch_sparse_tensor` then the tensor is of format `coo`. Like for example in `wl_conv` we use `from_torch_sparse_coo_tensor`.

torch_geometric/nn/conv/pan_conv.py
wsad1: Instead of having an `if` statement here why not add a utitliy function 
```
from_torch_sparse_to_pyg_sparse(edge_index : torch.Tensor) -> SparseTensor:
"""
Converts a `edge_index` of type `torch.Tensor` to a `SparseTensor`. The `torch.Tensor`
can be sparse.
"""
```

EdisonLeeeee: Sounds good. Will take it in another PR.

wsad1: Assuming we address this in another PR. I am good with these changes. 
Thanks !

test/nn/conv/test_signed_conv.py
```diff
@@ -60,7 +69,7 @@ def test_signed_conv():
         t = '(PairTensor, SparseTensor, SparseTensor) -> Tensor'
         jit1 = torch.jit.script(conv1.jittable(t))
         jit2 = torch.jit.script(conv2.jittable(t))
-        assert torch.allclose(jit1((x, x[:2]), adj.t(), adj.t()), out1[:2],
+        assert torch.allclose(jit1((x, x[:2]), adj1.t(), adj1.t()), out1[:2],
                               atol=1e-6)
-        assert torch.allclose(jit2((out1, out1[:2]), adj.t(), adj.t()),
+        assert torch.allclose(jit2((out1, out1[:2]), adj1.t(), adj1.t()),
```
wsad1, line 74: Add test for `adj2`.

wsad1 (approved):

torch_geometric/nn/conv/pan_conv.py
rusty1s: I think an `if` statement here is totally fine rather than introducing `to_pyg_sparse`. The goal is to get rid of `SparseTensor` in the long run, and so there is no real need for this method IMO.

rusty1s (approved):

PR #6944, opened 2023-03-17
Fix typo in `AddRandomWalkPE` documentation

aurorarossi:
This PR fix the documentation of `AddRandomWalkPE` where it was written that the default attribute was `"laplacian_eigenvector_pe"` which is the one of `AddLaplacianEigenvectorPE`.

### DIFF
--- a/torch_geometric/transforms/add_positional_encoding.py
+++ b/torch_geometric/transforms/add_positional_encoding.py
@@ -105,7 +105,7 @@ class AddRandomWalkPE(BaseTransform):
         attr_name (str, optional): The attribute name of the data object to add
             positional encodings to. If set to :obj:`None`, will be
             concatenated to :obj:`data.x`.
-            (default: :obj:`"laplacian_eigenvector_pe"`)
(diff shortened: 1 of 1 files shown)


torch_geometric/transforms/add_positional_encoding.py
```diff
@@ -122,7 +122,7 @@ def __call__(self, data: Data) -> Data:
         adj = SparseTensor.from_edge_index(edge_index, edge_weight,
                                            sparse_sizes=(num_nodes, num_nodes))
 
-        # Compute D^{-1} A:
+        # Compute A D^{-1} :
```
EdisonLeeeee, line 125: `D^{-1} A` is correct.

EdisonLeeeee (commented):
Thanks for the catch!

torch_geometric/transforms/add_positional_encoding.py
rusty1s, line 125:
```suggestion
        # Compute D^{-1} A:
```

rusty1s (approved):

Issue #6871, opened 2023-03-07
Upgrade edge_weight to first-class citizens 🦾


adrianjav:
### 🛠 Proposed Refactor

We (@psanch21 and I) have been thinking on how to seamless integrate our last work ([LCAT](https://arxiv.org/abs/2211.11853)) into PyG. In short, we propose a way of combining GAT-like networks with GCN-like networks. However, we have struggled to find an open-closed implementation that frees us from implementing one LCAT implementation per existing GCN model.

One key point that we make in the paper (Eq. 13 of Appendix F) is that attention can be seen as a way of assigning weights different from 1 to the adjacency matrix of the graph, and that we can integrate it with other GCN models if we see the usual formulation:
![image](https://user-images.githubusercontent.com/25713831/223505667-ca9f5c4f-f90c-4cf7-9987-886942933e28.png)
to include the edge weights:
![image](https://user-images.githubusercontent.com/25713831/223505908-a6be7bac-c13a-49a9-a925-5a8bf5b9ed8a.png)
And, looking at #6867, we could go one step further and write it as:
![image](https://user-images.githubusercontent.com/25713831/223508134-633f6414-fe29-4918-a7bf-90f869e87b18.png)
were I have included the point x_i in the aggregator function, which now takes the edge weights as another argument.

From this point of view, and looking into the [GNN Cheatsheet](https://pytorch-geometric.readthedocs.io/en/latest/cheatsheet/gnn_cheatsheet.html), I have realized that PyG could benefit a lot from a simple refactor.  

### Suggest a potential alternative/fix

These are the API refactor I would propose:
- Remove `edge_updater` from `MessagePassing` (it is only used by GATConv).
- Call `edge_update` at the beginning of `propagate` instead to compute the weights of the edges.
- Implement a class `EdgeUpdater` similar to `Aggregation` that can be passed to a conv layer.
- Adapt the aggregators to take the edge_weight as an argument.

This would imply the following benefits (some need to adapt parts of the code):
- All message passing layers would work with edge_weight by default.
- Options to use the symmetric normalized adjacency matrix could be abstracted away to an EdgeUpdater class (instead of baked in, e.g., as in `GCNConv` and `GCN2Conv`) and it could be used with any layer.
[...]

PR #6969, opened 2023-03-19
Refactor `EdgeUpdater`

adrianjav:
## Intro

This is a PR implementing the changes described in #6871.

I am still working on the code, but I'd assumed I can just start already the pull request and keep working on the changes, so that we can discuss the changes as I work on them.  I am mostly new to contributing on GitHub, so I assume I can keep pushing on my fork and changes will appear here. If that's not the case, I'll close the PR and create a new one when the code is ready.

Since the idea is to add `edge_update` in the logic of `propagate`, I figured out that the most convenient interface would be to let any subclass of `EdgeUpdate` define which arguments they want (just like the `message` method), and return a dictionary with the updated objects that will update `kwargs` in `propagate`. 

So far this is working nicely and it has allowed me to greatly overcome issues I was having. For example:
- The existing code for `gcn_norm` needs to take `flow` as an argument, and my implementation declares `edge_index_i` and `edge_index_j` directly in the method signature. 
- Existing code (e.g. to add self loops) had to deal with all the cases of different types for `edge_index`. By relying on `__collect__` the code for `edge_update` only has to deal with normal tensors.

**This is work-in-progress, so don't pay too much attention to the code for now. There are parts that I have not yet adapted nor tested.**

## Interface summary

`EdgeUpdater` is as simple as possible:
```python
class EdgeUpdater(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, *args, **kwargs) -> Dict:
        raise NotImplementedError

    def reset_parameters(self) -> None:
        pass
```

and propagate now is of the form:
```python
def propagate(self, edge_index: Adj, size: Size = None, **kwargs):
    size = self.__check_input__(edge_index, size)

    ################ NEW CODE
    coll_dict = self.__collect__(self.__edge_user_args__, edge_index, size,
                                 kwargs)

    edge_kwargs = self.inspector.distribute('edge_update', coll_dict)
    out = self.edge_update(**edge_kwargs)

    size = out.pop('size', size)  # This I need to figure out yet
    kwargs.update(out)
    ################ UNTIL HERE

    coll_dict = self.__collect__(self.__user_args__, edge_index,
                                 size, kwargs)

    msg_kwargs = self.inspector.distribute('message', coll_dict)

    aggr_kwargs = self.inspector.distribute('aggregate', coll_dict)
    out = self.aggregate(out, **aggr_kwargs)

    update_kwargs = self.inspector.distribute('update', coll_dict)
    out = self.update(out, **update_kwargs)

    return out
```

Now some modules are greatly simplified. For example:
```python
class GCNNorm(EdgeUpdater):
    def forward(self, edge_index_i: LongTensor, edge_index_j: LongTensor,  
                edge_weight: Tensor, dim_size: int) -> Dict:
        deg = scatter(edge_weight, edge_index_i, dim=0, dim_size=dim_size, reduce='sum')
        deg_inv_sqrt = deg.pow_(-0.5)
        deg_inv_sqrt.masked_fill_(deg_inv_sqrt == float('inf'), 0)
        edge_weight = deg_inv_sqrt[edge_index_i] * edge_weight * deg_inv_sqrt[edge_index_j]

        return {
            'edge_weight': edge_weight
        }
```

and the previous behaviour of `gcn_norm` with `improved=True` and `add_self_loops=True` with caching could be reproduced using:
```python
edge_updater = Cache(Compose([
    AddRemainingSelfLoops(fill_value=2 if improved else 1),
    GCNNorm(),
]))
```
## RoadMap

- [x] Implemented `EdgeUpdater`.
- [ ] Implemented  `EdgeUpdater` subclasses
    - [x] `Fill` to set a constant as weights.
    - [x] `Cache` to cache a matrix.
    - [x] `GCNNorm` to reproduce `gcn_norm`.
    - [x] "Loop" edge updaters: `RemoveSelfLoops`, `SegregateSelfLoops`, `AddSelfLoops`, `AddRemainingSelfLoops`.
    - [x]  `Compose` subclass.
    - [ ] Attention edge updaters.
- [x] Adapt `MessagePassing` to use `EdgeUpdater`.
- [ ] Add weights to the aggregation functions.
- [ ] Add unit tests.
- [ ] Document code.
- [ ] Adapt existing classes.

## Things to fix/deal with
- [ ] Tests regarding "jittable" for `MessagePassing` are broken.
- [ ] Figure out how to deal with the case where `message_and_aggregate` is used.

## Other changes

I will add here a list of other changes I do in case you want to discuss any in detail:
- I plan to pass `edge_index_i` and `edge_index_j` as arguments to `__collect__` after calling `edge_update`, so I needed to:
    - keep the special args from `__user__args__` and variables alike. 
     - unify the behaviour of `__collect__` in the message passing class, simplifying the logic of `__lift__`.
     - compute `num_nodes` in `__collect__` exactly like `maybe_num_nodes`, so that the `EdgeUpdate` instances can access this info for free.
- I have temporarily changed `GATConv` to stop using `edge_updater` to stop tests from crashing.

### DIFF
--- a/test/nn/conv/test_message_passing.py
+++ b/test/nn/conv/test_message_passing.py
@@ -261,15 +261,12 @@ class MyEdgeConv(MessagePassing):
         super().__init__(aggr='add')
 
     def forward(self, x: Tensor, edge_index: Adj) -> Tensor:
-        # edge_updater_type: (x: Tensor)
-        edge_attr = self.edge_updater(edge_index, x=x)
-
         # propagate_type: (edge_attr: Tensor)
-        return self.propagate(edge_index, edge_attr=edge_attr,
+        return self.propagate(edge_index, x=x,
                               size=(x.size(0), x.size(0)))
 
-    def edge_update(self, x_j: Tensor, x_i: Tensor) -> Tensor:
-        return x_j - x_i
+    def edge_update(self, x_j: Tensor, x_i: Tensor) -> dict:
+        return {'edge_attr': x_j - x_i}
 
     def message(self, edge_attr: Tensor) -> Tensor:
         return edge_attr
--- a/test/nn/edge_updater/test_basics.py
+++ b/test/nn/edge_updater/test_basics.py
@@ -1,0 +1,65 @@
+import torch
+
+from torch_geometric.nn.conv.gcn_conv import gcn_norm
+from torch_geometric.nn.edge_updater import (
+    Compose, GCNNorm, AddRemainingSelfLoops
+)
+
+
+def test_gcn_norm():
+    torch.manual_seed(2)
+    edge_index = torch.tensor([[0, 0, 0, 1, 2, 3], [1, 2, 3, 0, 0, 0]])
+    row, col = edge_index
+    value = torch.rand(row.size(0))
+
+    updater = GCNNorm()
+    assert str(updater) == 'GCNNorm()'
+    out1 = updater(edge_index[1], edge_index[0], value,
+                   dim_size=4)['edge_weight']
+    out2 = gcn_norm(edge_index, value, 4,
+                    improved=False, add_self_loops=False)[1]
+    assert out1.size() == out2.size()
+    assert torch.allclose(out1, out2, atol=1e-6)
+
+    updater = Compose((
+        AddRemainingSelfLoops(),
+        GCNNorm()
+    ))
+    assert str(updater) == 'Compose([' \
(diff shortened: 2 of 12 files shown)


adrianjav:
I have one question @rusty1s: is there any use-case to add `edge_index` and `adj_t` to the dictionary in `__collect__`? I don't see any, and if we let edge updaters to modify `edge_index_i` and `edge_index_j` (among others), I only see issues arising from letting users access those two variables in `message` and others.

PR #6972, opened 2023-03-20
fix `to_csc` in cugraph models

tingyu66:
This PR:
- fixes the bug in `to_csc()` when `size` is not given, since PyG uses the same subgraph over all layers.
- absorb `num_src_nodes` into the `csc` tuple to make `get_cugraph` more concise
- misc: no longer allow identity node feature tensor (by setting `x=None`) in RGCN. It is an oversight from me in the previous PR to not remove it, since this is memory-inefficient

CC: @MatthiasKohl, @puririshi98

### DIFF
--- a/torch_geometric/nn/conv/cugraph/base.py
+++ b/torch_geometric/nn/conv/cugraph/base.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Optional, Tuple, Union
 
 import torch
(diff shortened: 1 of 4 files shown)


torch_geometric/nn/conv/cugraph/base.py
```diff
@@ -35,9 +36,10 @@ def reset_parameters(self):
     @staticmethod
     def to_csc(
         edge_index: Tensor,
-        size: Optional[Tuple[int, int]] = None,
+        size: Tuple[int, int] = None,
```
MatthiasKohl, line 39: This should remain optional if we allow `None` (which I think we should, so the rest of the logic looks good).

torch_geometric/nn/conv/cugraph/base.py
```diff
@@ -48,40 +50,48 @@ def to_csc(
...
+                :obj:`(row, colptr, num_src_nodes)`. Use the
+                :meth:`CuGraphModule.to_csc` method to convert an
+                :obj:`edge_index` representation to the desired format.
             max_num_neighbors (int, optional): The maximum number of neighbors
                 of a target node. It is only effective when operating in a
                 bipartite graph. When not given, will be computed on-the-fly,
                 leading to slightly worse performance. (default: :obj:`None`)
         """
-        row, colptr = csc
+        row, colptr, num_src_nodes = csc
```
MatthiasKohl, line 94: it would be good to add a comment in this function mentioning that since in both MFG and FG cases, we end up creating the structure based on this `num_src_nodes`, we will check the size of the inputs for an operator in the bindings correctly (we actually check the size of the input against `num_src_nodes` which was given independent of the size of the input here).

torch_geometric/nn/conv/cugraph/base.py
```diff
@@ -128,7 +137,7 @@ def get_typed_cugraph(
         if num_edge_types is None:
             num_edge_types = int(edge_type.max()) + 1
 
-        row, colptr = csc
+        row, colptr, num_src_nodes = csc
         edge_type = edge_type.int()
```
MatthiasKohl, line 141: similar to above, add a comment (simply refer to the comment in the above function)

torch_geometric/nn/conv/cugraph/gat_conv.py
```diff
@@ -61,10 +61,12 @@ def reset_parameters(self):
     def forward(
         self,
         x: Tensor,
-        csc: Tuple[Tensor, Tensor],
+        csc: Tuple[Tensor, Tensor, int],
         max_num_neighbors: Optional[int] = None,
     ) -> Tensor:
-        graph = self.get_cugraph(x.size(0), csc, max_num_neighbors)
+        assert x.size(0) == csc[
+            -1], "Size mismatch between node feature tensor and CuGraph graph."
```
MatthiasKohl, line 68: I believe this check is already done in the bindings afterwards. If needed, it would be better to improve the error messages there, since this leads to duplicated code (both the fact that the checks will be done twice and that you have to repeat this in every forward function of the different convolution modules, or create another abstraction that does this).

MatthiasKohl (changes requested):
Overall, looks good, thank you for the quick PR !
I'm not entirely sure about where to assert the size of the input features, but if it's already done as expected in the bindings, we should get rid of the additional `assert` here, IMO, since otherwise, we'll check it for every operator. Let me know what you think, too !

torch_geometric/nn/conv/cugraph/base.py
MatthiasKohl: I guess if we do the check for input feature size within the operators (before calling into this function), we should simply have a comment here saying that the size was already checked instead of what I mentioned above.
If we move that check to the bindings though, my comment above holds

PR #6982, opened 2023-03-21
Add default argument to `datapipe.py` example

Kh4L:
This change makes the `datapipe.py` example more consistent with other examples by providing a default arg.


rusty1s (approved):
Thanks, I like it.

PR #6990, opened 2023-03-21
Fixed some typos in doc-strings

EdisonLeeeee:


torch_geometric/transforms/to_sparse_tensor.py
```diff
@@ -57,7 +57,7 @@ def __init__(
         attr: Optional[str] = 'edge_weight',
         remove_edge_index: bool = True,
         fill_cache: bool = True,
-        layout: Optional[int] = None,
+        layout: Optional[torch.layout] = None,
```
rusty1s, line 60:
```suggestion
        layout: Optional[int] = None,
```
This needs to be `int` at the moment :(

rusty1s (approved):

Issue #6935, opened 2023-03-17
Implementing A Unified Lottery Ticket Hypothesis for Graph Neural Networks


PrathamSoni:
### 🚀 The feature, motivation and pitch

ULTH (https://arxiv.org/pdf/2102.06790.pdf) pruning on graphs allows for unified learned masks on both the graph and GNN weights that extend the notion of graph pruning. This allows significant improvements in computational efficiency without dropoffs in performance. 

We implement the ULTH GLT algorithm with three classes and define the interface as follows:
1) GLTMask: Generates an internal dictionary of masks that will be applied to the model. All named parameters will be added to the mask other than those specified by _ignore_keys_.
2) PrunedModel: Serves as a wrapper for PyG modules to inject the appropriate masks defined by GLTMask for adjacency data and model weights. 
3) GLTSearch: Performs the ULTH search algorithm as presented in the paper. First trains the masks using UGS loss (regularizes mask norms, each regularization param is individually set). Then 0s the mask to enforce sparsity to hit the desired level as specified by pruning rate arguments (again can have different rates for adjacency vs model masks). Final fine-tune of the resultant masked graph + model.

We also provide a testing script to compare results against the paper on small datasets.

### Additional context

CC: @ydalmia @nathankim7

rusty1s:
Thanks @PrathamSoni. I think this reads all good, let me give feedback directly in the PR.

PR #6992, opened 2023-03-21
Implementing Unified `GLTSearch`

PrathamSoni:
Implementing A Unified Lottery Ticket Hypothesis for Graph Neural Networks  ( #6935 )
ULTH (https://arxiv.org/pdf/2102.06790.pdf) pruning on graphs allows for unified learned masks on both the graph and GNN weights that extend the notion of graph pruning. This allows significant improvements in computational efficiency without dropoffs in performance. 

We implement the ULTH GLT algorithm with three classes and define the interface as follows:
1) GLTMask: Generates an internal dictionary of masks that will be applied to the model. All named parameters will be added to the mask other than those specified by _ignore_keys_.
2) PrunedModel: Serves as a wrapper for PyG modules to inject the appropriate masks defined by GLTMask for adjacency data and model weights. 
3) GLTSearch: Performs the ULTH search algorithm as presented in the paper. First trains the masks using UGS loss (regularizes mask norms, each regularization param is individually set). Then 0s the mask to enforce sparsity to hit the desired level as specified by pruning rate arguments (again can have different rates for adjacency vs model masks). Final fine-tune of the resultant masked graph + model.

We also provide a testing script to compare results against the paper on small datasets.

cc: @ydalmia @nathankim7
-----------------------------------
PR is a draft pending completion of testing


torch_geometric/contrib/nn/models/graph_lottery_ticket.py
```diff
@@ -0,0 +1,515 @@
...
+
+    def train(self, ticket: GLTModel,
+              ugs: bool) -> Tuple[float, Dict[str, Parameter]]:
+        """train loop. If ugs flag, use mask regularization."""
+        best_val_score = 0.0
+        final_test_score = 0.0
+        best_masks = {}
+        optimizer = self.optimizer(ticket.parameters(), **self.optim_args)
+
+        with tqdm.trange(self.max_train_epochs, disable=not self.verbose) as t:
```
wsad1, line 452: It looks like the parameters of the model might also get updated in this training loop, I guess that's why you have `rewind`. Instead wouldn't calling `self.module.eval()` before running this simply prevent model parameters from getting updated?

torch_geometric/contrib/nn/models/graph_lottery_ticket.py
```diff
@@ -0,0 +1,515 @@
...
+        super().__init__()
+        if ignore_keys is None:
+            ignore_keys = set()
+        self.module = module
+        self.graph = graph
+        self.ignore_keys = ignore_keys
+
+    @property
+    def model(self) -> Module:
+        return self.module
```
wsad1, line 56: Drop this property. We could call it `self.model` instead of `sefl.module`.

torch_geometric/contrib/nn/models/graph_lottery_ticket.py
```diff
@@ -0,0 +1,515 @@
...
+            self.lr_mask_graph = self.lr
+
+        if not self.lr_mask_model:
+            self.lr_mask_model = self.lr
+
+        self.optim_args = {"lr": self.lr, **self.optim_args}
+
+        self.mask = GLTMask(
+            self.module, self.graph,
+            torch.device("cuda" if torch.cuda.is_available() else "cpu"))
```
wsad1, line 400: Why is this on the GPU. It should be on the same device as `self.module`.

wsad1 (changes requested):
Thanks for the great work.
Will tak a look again once initial comments are addressed.

nathankim7:
> Thanks for the great work. Will tak a look again once initial comments are addressed.

Thank you for the generous review! Most of the low-hanging fruit are addressed in this commit. The major blocker remaining is the dependence on GNNs with `edge_weights`: despite the name, this method relies heavily on having access to a continuous analogue of the adjacency matrix to learn on, so I'm not entirely sure how we would work around this. Would appreciate some guidance on this matter.

torch_geometric/contrib/nn/models/graph_lottery_ticket.py
nathankim7: Addressed!

torch_geometric/contrib/nn/models/graph_lottery_ticket.py
```diff
@@ -0,0 +1,515 @@
...
+    """helper for getting accuracy on node classification"""
+    correct_val = (preds[val_mask] == targets[val_mask]).sum()
+    val_score = int(correct_val) / int(val_mask.sum())
+
+    correct_test = (preds[test_mask] == targets[test_mask]).sum()
+    test_score = int(correct_test) / int(test_mask.sum())
+    return val_score, test_score
+
+
+@dataclass
```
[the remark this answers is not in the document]
nathankim7: Addressed!

torch_geometric/contrib/nn/models/graph_lottery_ticket.py
nathankim7: The LTH relies on *jointly* updating model and mask parameters from their initialization to learn the best subnetwork, so the model should remain in train mode as far as I know.

PR #6993, opened 2023-03-21
Update type hints for `HeteroData`

binh-vu:
According to [PEP 484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values), variable positional arguments (*args) only need to specify the expected value for one such argument.

This PR will fix the warning from Pylance (in vscode) when assigning data in HeteroData as follow (code from tutorial):

```python
from torch_geometric.data import HeteroData

data = HeteroData()

data['paper'].x = ... # Pylance will complain but won't for data[('paper',)].x = . This PR will fix that
```


rusty1s (approved):
Thanks!

PR #6995, opened 2023-03-22
Implementing `ESAN` paper

TheJurman:
Hi PyG team, 

We have implemented DSnetwrok, DSSnetwork and 4 new subgraph policies (Node Deletion, Edge Deletion, EGO and EGO+) and also running some examples replicating the results from the original paper. The original issue was created [here](https://github.com/pyg-team/pytorch_geometric/issues/6808l)


juramatu (commented):
The submission meets PyG contribution reqs

Issue #7010, opened 2023-03-22
Temporal version for elliptic bitcoin dataset


rishiagarwal2000:
### 🚀 The feature, motivation and pitch

I am using the EllipticBitcoinDataset, and want to compare the performance of static GCNs (that do not change with time) with evolving GCNs. Hence, it would be useful if we have a temporal version for the dataset where a separate graph for each time step can be loaded.

akihironitta:
Closing this issue as #7011 added `EllipticBitcoinTemporalDataset`. It should be available in the next 2.4.0 release.

PR #7011, opened 2023-03-22
Temporal version of `EllipticBitcoinDataset`

rishiagarwal2000:
Hi PyG team,

We have implemented a temporal version of the EllipticBitcoinDataset to allow experimenting with models such as [EvolveGCN](https://arxiv.org/pdf/1902.10191.pdf), where the GCN evolves with time. The original issue was created [here](https://github.com/pyg-team/pytorch_geometric/issues/7010#issue-1636550901)


rusty1s (approved):
Thanks. I refactored this a bit such that `EllipticBitcoinTemporalDataset` inherits from `EllipticBitcoinDataset`. Hope the change works for you.

PR #7012, opened 2023-03-23
WIP: Implement SE(3)-Transformer

nealvaidya:
We implement the SE(3)-Equivariant Transformer network proposed in [SE(3)-Transformers: 3D Roto-Translation Equivariant Attention Networks](https://arxiv.org/abs/2006.10503). This implementation is a port of the DGL implementation [here](https://github.com/NVIDIA/DeepLearningExamples/tree/master/DGLPyTorch/DrugDiscovery/SE3Transformer).

We've also included an example training the model on the QM9 dataset.

Submitted for cs224w


examples/contrib/se3transformer_qm9.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class AsFiber(T.BaseTransform):
+    def __init__(self, node_fiber: Fiber, edge_fiber: Fiber) -> None:
+        self.node_fiber = node_fiber
+        self.edge_fiber = edge_fiber
+
+    def __call__(self, data: Data) -> Data:
+        data.node_feats = self.node_fiber.tensor_to_dict(data.x)
+        data.edge_feats = self.edge_fiber.tensor_to_dict(data.edge_attr)
```
wsad1, line 56: Why rename the variables here. Keep them as `data.x` and `data.edge_attr`?

wsad1 (changes requested):
Thanks for adding this.
To start of with could you please add tests for al the modules in `conv` and `models`.

torch_geometric/contrib/utils/fiber.py
```diff
@@ -0,0 +1,145 @@
...
+from itertools import product
+from typing import Dict
+
+import torch
+from torch import Tensor
+
+FiberEl = namedtuple("FiberEl", ["degree", "channels"])
+
+
+class Fiber(dict):
```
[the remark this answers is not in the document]
nealvaidya: Yes, it does come from the paper, there is some good explanation in this [blog post](https://fabianfuchsml.github.io/se3iterative/). I do think it would be possible to do this without the Fiber class, we would just need to keep track of the degrees of the different features in `data.x` in some other field of `data`

examples/contrib/se3transformer_qm9.py
nealvaidya: Sure, I don't think there's any good reason to rename

amorehead:
Any updates on this PR by chance? This looks like a great addition to the PyG model catalog!

PR #7016, opened 2023-03-23
Added `DegreeQuant`

laikhtewari:
We implemented the [DegreeQuant](https://arxiv.org/abs/2008.05000) method for Quantization-Aware Training for GNNs (for CS224W final project). We documented the files and included a README to describe the file structure. The changes are located in pytorch_geometric/torch_geometric/contrib/degree_quant


wsad1 (changes requested):
@laikhtewari Thanks for the amazing work. And sorry for the delay in reviews.
Given that the PR is over 2k lines I am finding it a little hard to review. Do you think we can break this down into multiple PRs.

PR #7026, opened 2023-03-24
feat: add `RotatE` KGE model

riyavsinha:
[No longer contributing affiliated with CS224W]

Paper reported FB15k-237 H@10: .533
Implementation Example H@10: .431

Of note, the paper only reports their score using their Adversarial Negative Sampling (ANS), which is not implemented as a part of this PR. Table 7 from the paper comparing the use of ANS vs. uniform random sampling shows a 5% increase in H@10 for FB15k-237, which may be a factor to the disparity in performance, combined with the small hidden size in the example, I believe. The margin=9 argument is from the paper's reported best performance in Table 12 on FB15k-237 as well.

Happy to try to implement the ANS as part of this PR if the current performance is below standard, let me know!


torch_geometric/nn/kge/rotate.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(
+        self,
+        num_nodes: int,
+        num_relations: int,
+        hidden_channels: int,
+        margin: float = 1.0,
+        sparse: bool = False,
+    ):
+        super().__init__(num_nodes, num_relations, hidden_channels, sparse)
+        self.margin = margin
```
wsad1, line 46:
```suggestion
        self.register_buffer('margin', margin)
```
This way `self.margin` will be on the same `device` as other parameters.

torch_geometric/nn/kge/rotate.py
```diff
@@ -0,0 +1,92 @@
...
+        tail_re = self.node_emb(tail_index)
+        tail_im = self.node_emb_im(tail_index)
+
+        rel_theta = self.rel_emb(rel_type)
+        rel_re, rel_im = torch.cos(rel_theta), torch.sin(rel_theta)
+
+        re_score = (rel_re * head_re - rel_im * head_im) - tail_re
+        im_score = (rel_re * head_im + rel_im * head_re) - tail_im
+        complex_score = torch.stack([re_score, im_score], dim=2)
+        score = torch.norm(complex_score, dim=(1, 2))
```
wsad1, line 73: Looks like `torch.norm` is depreciated https://pytorch.org/docs/stable/generated/torch.norm.html. Lets use `torch.linalg.vector_norm` instead.

torch_geometric/nn/kge/rotate.py
```diff
@@ -0,0 +1,92 @@
...
+class RotatE(KGEModel):
+    r"""The RotatE model from the `"RotatE: Knowledge Graph Embedding by
+    Relational Rotation in Complex Space" <https://arxiv.org/abs/
+    1902.10197>`_ paper.
+
+    :class:`RotatE` models relations as a rotation in complex space
+    from head to tail. The scoring function is thus:
+
+    .. math::
+        d(h, r, t) = - {\| \mathbf{e}_h \circ \mathbf{e}_r - \mathbf{e}_t \|}_p
```
wsad1, line 20:
```suggestion
    :class:`RotatE` models relations as a rotation in complex space
    from head to tail such that:
    
    .. math::
        - \mathbf{e}_t = \mathbf{e}_h \circ \mathbf{e}_r     
    
    Resulting in the scoring function:

    .. math::
        d(h, r, t) = - {\| \mathbf{e}_h \circ \mathbf{e}_r - \mathbf{e}_t \|}_p
```
Nitpick.

wsad1 (commented):
@riyavsinha thanks for all the work.
Changes look good. Lets add the Adversarial sampling in another PR to the `KGEModel` base class.
Will approve once comments are addressed.

torch_geometric/nn/kge/rotate.py
riyavsinha: Done, just an FYI had to wrap margin in `torch.Tensor`, wanted to make sure thats ok

torch_geometric/nn/kge/rotate.py
riyavsinha: Great catch, done

torch_geometric/nn/kge/rotate.py
riyavsinha: Done

wsad1 (approved):

PR #7048, opened 2023-03-27
Fix `bfloat16` conversion in `HeteroData` benchmark scripts

yanbing-j:
This PR is to fix the issues:

1. Re-use `test` function in training benchmark to avoid mismatch between `Data` and `HeteroData` in evaluation mode.
2. `HeteroData` to convert to `bfloat16` datatype in inference and training benchmarks.

### DIFF
--- a/benchmark/inference/inference_benchmark.py
+++ b/benchmark/inference/inference_benchmark.py
@@ -205,7 +205,7 @@ def run(args: argparse.ArgumentParser):
                                     if args.evaluate:
                                         test_acc = test(
                                             model, test_loader, device, hetero,
-                                            progress_bar=True)
+                                            progress_bar=True, )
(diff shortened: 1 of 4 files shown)


benchmark/inference/inference_benchmark.py
```diff
@@ -209,12 +203,9 @@ def run(args: argparse.ArgumentParser):
...
                                     if args.evaluate:
-                                        test_acc = model.test(
-                                            y,
-                                            test_loader,
-                                            device,
-                                            progress_bar=True,
-                                        )
+                                        test_acc = test(
+                                            model, test_loader, device, hetero,
+                                            progress_bar=True)
```
rusty1s, line 208:
```suggestion
                                            progress_bar=True, )
```

rusty1s (approved):

PR #7050, opened 2023-03-27
Fix `FastHGTConv` to correctly call the value module

toenshoff:
This fixes a small bug in the new `fast_hgt_conv` layer, which incorrectly called the module `self.k_rel` instead of `self.v_rel` to generate the value vectors `v`.


rusty1s (approved):
:(

Issue #6979, opened 2023-03-20
Bridging the gap between DGL and PyG


nowyouseemejoe:
### 🚀 The feature, motivation and pitch

DGL and PyG are two popular deep learning libraries that specialize in working with graph-structured data. Both DGL and PyG have become go-to libraries for researchers and practitioners working on graph deep learning problems, and have a growing community of contributors and users.

I would like to request a new feature for PyG that allows for converting DGL graphs to PyG instances (e.g., `Data`). This would make it easier for users who have existing graph structures in DGL to switch to PyG without having to recreate their graphs from scratch.

In addition, it would be great to have a tutorial added to the documentation, demonstrating how to mitigate from DGL to PyG using this new feature. The tutorial could introduce some equivalent message and aggregation operations for these two libraries, helping users navigate the differences between them. With this new feature and tutorial, users can seamlessly transition between DGL and PyG, unlocking the full potential of graph deep learning for their projects.

rusty1s:
I like the idea of providing conversion from DGL to PyG and vice versa. This should be pretty straightforward to get in. Do you have interest in contributing this?

rusty1s:
I am not sure if @nowyouseemejoe has the intention to work on this. I think we could split the work into two parts:
* [ ] `from_dgl(dgl.Graph) -> data: Union[Data, HeteroData]`
* [ ] `to_dgl(Union[Data, HeteroData]) -> dgl.Graph`

such that both of you could work on this in parallel. WDYT?

nowyouseemejoe:
@rusty1s Sorry. I'm afraid I don't have the time or expertise to contribute effectively to this at the moment. I appreciate your understanding.

Thank you @hbenedek . That would be great!

tjb-tech:
> 
Highly recommend this idea! Maybe having this api, the issue like https://github.com/pyg-team/pytorch_geometric/issues/7017, which I recently meet, will not arise.

akihironitta:
@Priyankagautam08 It's included in 2.4.0 release:
- https://pytorch-geometric.readthedocs.io/en/2.4.0/modules/utils.html#torch_geometric.utils.to_dgl
- https://pytorch-geometric.readthedocs.io/en/2.4.0/modules/utils.html#torch_geometric.utils.from_dgl

PR #7053, opened 2023-03-27
Add `to_dgl` and `from_dgl` conversions

hbenedek:
Issue https://github.com/pyg-team/pytorch_geometric/issues/6979

## Description

I add two functions `from_dgl(dgl.Graph) -> data: Union[Data, HeteroData]`and
 `to_dgl(Union[Data, HeteroData]) -> dgl.Graph` in `utils/convert.py`. Both function in both direction performs the same three operations, setting the node and edge attributes and copies the edges of the underlying graph(s).

### DIFF
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -479,7 +479,7 @@ def to_dgl(
         for attr in data.edge_attrs():
             if attr == "edge_index":
                 continue
-            g.edata[attr] = getattr(data, attr)
+            g.edata[attr] = data.get(attr)
 
     elif isinstance(data, HeteroData):
         data_dict = {}
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -525,7 +525,7 @@ def to_dgl(
                 g.nodes[node_type].data[attr] = value
         for edge_type, edge_store in zip(data.edge_types, data.edge_stores):
             for attr, value in edge_store.items():
-                if attr == "edge_index":
+                if attr in ["edge_index", "adj_t"]:
                     continue
                 g.edges[edge_type].data[attr] = value
 
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -467,6 +467,29 @@ def to_dgl(
     Args:
         data (torch_geometric.data.Data or torch_geometric.data.HeteroData):
             The data object.
+
+    Example:
+
+        >>> edge_index = torch.tensor([[0, 1, 1, 2, 3, 0], [1, 0, 2, 1, 4, 4]])
+        >>> x = torch.ones(5, 3)
+        >>> y = torch.tensor([1, 1, 1, 1, 1, 0], dtype=torch.float)
+        >>> data = Data(x=x, edge_index=edge_index, edge_attr=y)
+        >>> g = to_dgl(data)
+        >>> g
+        Graph(num_nodes=5, num_edges=6,
+            ndata_schemes={'x': Scheme(shape=(3,), dtype=torch.float32)}
+            edata_schemes={'edge_attr': Scheme(shape=(), dtype=torch.float32)})
+
+        >>> data = HeteroData()
+        >>> data['paper'].x = torch.ones(5, 128)
+        >>> data['authors'].x = torch.ones(5, 128)
+        >>> data['authors', 'cites', 'paper'].edge_index = torch.tensor(
+        ...                    [[0, 1, 2, 3, 4],[0, 1, 2, 3, 4]])
+        >>> g = to_dgl(data)
+        >>> g
+        Graph(num_nodes={'authors': 5, 'paper': 5},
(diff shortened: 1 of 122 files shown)


torch_geometric/utils/convert.py
```diff
@@ -455,3 +455,90 @@ def from_cugraph(g: Any) -> Tuple[Tensor, Optional[Tensor]]:
...
+            The data object.
+    """
+    import dgl
+
+    from torch_geometric.data import Data, HeteroData
+
+    if isinstance(data, Data):
+        g = dgl.graph((data.edge_index[0], data.edge_index[1]))
+        for attr in data.node_attrs():
+            g.ndata[attr] = getattr(data, attr)
```
EdisonLeeeee, line 478:
```suggestion
            g.ndata[attr] = data.get(attr)
```

torch_geometric/utils/convert.py
```diff
@@ -455,3 +455,90 @@ def from_cugraph(g: Any) -> Tuple[Tensor, Optional[Tensor]]:
...
+    Args:
+        data (torch_geometric.data.Data or torch_geometric.data.HeteroData):
+            The data object.
+    """
+    import dgl
+
+    from torch_geometric.data import Data, HeteroData
+
+    if isinstance(data, Data):
+        g = dgl.graph((data.edge_index[0], data.edge_index[1]))
```
EdisonLeeeee, line 476: We should also take `adj_t` into consideration.

torch_geometric/utils/convert.py
```diff
@@ -455,3 +455,90 @@ def from_cugraph(g: Any) -> Tuple[Tensor, Optional[Tensor]]:
...
+        for attr in data.node_attrs():
+            g.ndata[attr] = getattr(data, attr)
+        for attr in data.edge_attrs():
+            if attr == "edge_index":
+                continue
+            g.edata[attr] = getattr(data, attr)
```
EdisonLeeeee, line 482:
```suggestion
            g.edata[attr] = data.get(attr)
```

torch_geometric/utils/convert.py
```diff
@@ -455,3 +455,90 @@ def from_cugraph(g: Any) -> Tuple[Tensor, Optional[Tensor]]:
...
+
+def from_dgl(
+    g: Any,
+) -> Union["torch_geometric.data.Data", "torch_geometric.data.HeteroData"]:
+    r"""Converts a :obj:`dgl` graph object to a
+    :class:`torch_geometric.data.Data` or
+    :class:`torch_geometric.data.HeteroData` instance.
+
+    Args:
+        g (dgl.DGLGraph): The :obj:`dgl` graph object.
```
EdisonLeeeee, line 514: Add an example here :)

EdisonLeeeee (commented):
Thanks for the PR! Please also includes some test cases for these two functions.

torch_geometric/utils/convert.py
hbenedek: @EdisonLeeeee Thank you for your suggestions. Can you elaborate a bit on this one? Not sure if I understand what you mean by considering `adj_t`.

torch_geometric/utils/convert.py
```diff
@@ -455,3 +455,90 @@ def from_cugraph(g: Any) -> Tuple[Tensor, Optional[Tensor]]:
...
+    data: Union["torch_geometric.data.Data", "torch_geometric.data.HeteroData"]
+) -> Any:
+    r"""Converts a :class:`torch_geometric.data.Data` or
+    :class:`torch_geometric.data.HeteroData` instance to a :obj:`dgl` graph
+    object.
+
+    Args:
+        data (torch_geometric.data.Data or torch_geometric.data.HeteroData):
+            The data object.
+    """
```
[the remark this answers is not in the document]
hbenedek: I added two examples in both cases, one for HeteroData and one for Data, should I also create tests in `test/utils/test_convert.py`?

torch_geometric/utils/convert.py
EdisonLeeeee: As there may be cases where data only contains the transposed sparse adjacency matrix `adj_t`, we need to take this into consideration as well. For example, `data` was transformed by `ToSparseTensor`.

torch_geometric/utils/convert.py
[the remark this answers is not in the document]
EdisonLeeeee: Yep, that would be better :)

hbenedek:
@EdisonLeeeee I added the tests, but for some reason, codecov complains about not seeing them. I ran `pytest --cov` and the coverage of `torch_geometric/utils/convert.py` is only 9%. Am I doing something wrong, or just the external libraries causing this issue?

rusty1s:
@hbenedek Don't worry about it. It's good that we have a test but it's not called by CI due to missing dependency.

rusty1s (approved):

rusty1s:
Great PR, thanks for doing that :)

EdisonLeeeee:
@hbenedek thank you, too :D

PR #7060, opened 2023-03-28
Fix handling of missing node types in heterogeneous data loaders

toenshoff:
This addresses a bug in heterogeneous data loaders, which crash if a node type from the complete graph is not present in the sampled batch of nodes. This is can easily happen if a graph is disconnected or some node types are simply rare.

An unsafe dictionary access in the `filter_hetero_data` raises an exception in this case. I solved this by calling `HeteroData.node_type_subgraph` on the graph before the batch of nodes is filtered. I added a test for `HGTLoader` which covers this case and did not pass previously.

Note that the behavior of the data loaders now differs slightly from `HeteroData.subgraph`, as the loader will handle missing keys by not including any node of that type in the batch, while the `subgraph` method would retain all of the nodes. However, it seems more convenient for data loading if batches only contain nodes that where actually returned by the underlying sampler.

### DIFF
--- a/test/loader/test_hgt_loader.py
+++ b/test/loader/test_hgt_loader.py
@@ -181,3 +181,32 @@ def test_hgt_loader_on_cora(get_dataset):
     out2 = hetero_model(hetero_batch.x_dict, hetero_batch.edge_index_dict,
                         hetero_batch.edge_weight_dict)['paper'][:batch_size]
     assert torch.allclose(out1, out2, atol=1e-6)
+
+
+@withPackage('torch_sparse')
+def test_hgt_loader_disconnected():
+    torch.manual_seed(12345)
+    data = HeteroData()
+
+    data['paper'].x = torch.arange(100)
+    data['author'].x = torch.arange(100, 300)
+
+    # enforce two disconnected clusters of papers, only the first cluster is connected to authors
+    data['paper', 'paper'].edge_index = torch.cat([
+        get_random_edge_index(50, 50, 250),
+        get_random_edge_index(50, 50, 250) + 50,
+    ], dim=1)
+    data['paper', 'paper'].edge_attr = torch.arange(500)
+    data['paper', 'author'].edge_index = get_random_edge_index(50, 200, 1000)
+    data['paper', 'author'].edge_attr = torch.arange(500, 1500)
+    data['author', 'paper'].edge_index = get_random_edge_index(200, 50, 1000)
+    data['author', 'paper'].edge_attr = torch.arange(1500, 2500)
+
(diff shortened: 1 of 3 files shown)


test/loader/test_hgt_loader.py
```diff
@@ -181,3 +181,34 @@ def forward(self, x, edge_index, edge_weight):
...
+    # enforce two disconnected clusters of papers, only the first cluster is connected to authors
+    data['paper', 'paper'].edge_index = torch.cat([
+        get_random_edge_index(50, 50, 250),
+        get_random_edge_index(50, 50, 250) + 50,
+    ], dim=1)
+    data['paper', 'paper'].edge_attr = torch.arange(500)
+    data['paper', 'author'].edge_index = get_random_edge_index(50, 200, 1000)
+    data['paper', 'author'].edge_attr = torch.arange(500, 1500)
+    data['author', 'paper'].edge_index = get_random_edge_index(200, 50, 1000)
+    data['author', 'paper'].edge_attr = torch.arange(1500, 2500)
```
wsad1, line 203:
```suggestion
    data['paper'].x = torch.arange(10)
    data['author'].x = torch.arange(10, 3)

    # Paper is disconnected from author
    data['paper', 'paper'].edge_index = get_random_edge_index(10, 10, 15)
    data['paper', 'paper'].edge_attr = torch.arange(15)
```

test/loader/test_hgt_loader.py
```diff
@@ -181,3 +181,34 @@ def forward(self, x, edge_index, edge_weight):
...
+
+    loader = HGTLoader(data, num_samples=[5] * 4, batch_size=1,
+                       input_nodes='paper')
```
wsad1, line 206:
```suggestion
    loader = HGTLoader(data, num_samples=2, batch_size=2,
                       input_nodes='paper')
```

torch_geometric/loader/utils.py
```diff
@@ -130,13 +130,13 @@ def filter_hetero_data(
 ) -> HeteroData:
     # Filters a heterogeneous data object to only hold nodes in `node` and
     # edges in `edge` for each node and edge type, respectively:
-    out = copy.copy(data)
+    out = data.node_type_subgraph(node_dict.keys())
```
wsad1, line 133:
```suggestion
    out = data.node_type_subgraph(node_dict.keys())
    out = data.edge_type_subgraph(edge_dict.keys())
```

torch_geometric/loader/utils.py
```diff
@@ -130,13 +130,14 @@ def filter_hetero_data(
...
+    out = data.edge_type_subgraph(edge_dict.keys())
```
wsad1, line 134:
```suggestion
```

PR #7065, opened 2023-03-28
Make edges in `HeterohilousGraphDataset` undirected

OlegPlatonov:
Hi! The graphs [here](https://github.com/yandex-research/heterophilous-graphs) are meant to be undirected. Since PyG treats all graphs as directed, I've added a call to transforms.ToUndirected to double the edges.

### DIFF
--- a/torch_geometric/datasets/heterophilous_graph_dataset.py
+++ b/torch_geometric/datasets/heterophilous_graph_dataset.py
@@ -5,6 +5,7 @@ import numpy as np
 import torch
 
 from torch_geometric.data import Data, InMemoryDataset, download_url
+from torch_geometric.transforms import ToUndirected
 
 
 class HeterophilousGraphDataset(InMemoryDataset):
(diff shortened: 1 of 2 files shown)


torch_geometric/datasets/heterophilous_graph_dataset.py
```diff
@@ -118,6 +119,8 @@ def process(self):
         data = Data(x=x, y=y, edge_index=edge_index, train_mask=train_mask,
                     val_mask=val_mask, test_mask=test_mask)
 
+        data = ToUndirected()(data)
```
EdisonLeeeee, line 122:
```suggestion
        data.edge_index = to_undirected(data.edge_index)
```

torch_geometric/datasets/heterophilous_graph_dataset.py
```diff
@@ -5,6 +5,7 @@
 import torch
 
 from torch_geometric.data import Data, InMemoryDataset, download_url
+from torch_geometric.transforms import ToUndirected
```
EdisonLeeeee, line 8:
```suggestion
from torch_geometric.utils import to_undirected
```

rusty1s (approved):

PR #7104, opened 2023-04-02
Fix `data.num_edges` for `torch.sparse.Tensor`

### LINKED ISSUE #7103: `data.num_edges` is 0 when using native PyTorch sparse tensor
### 🐛 Describe the bug

When using the native PyTorch sparse tensor for representing the adjacency matrix, the number of edges reported by the `Data` object is always zero. The following code reproduces the issue:

```python

transform = T.ToSparseTensor(layout=torch.sparse_csr)
dataset = Planetoid("Planetoid", name="Cora", transform=transform)
data = dataset[0]
print(data.num_edges)
```

### Environment

* PyG version: 2.3.0
* PyTorch version: 2.0.0
* OS: Linux (Debian)
* Python version: 3.10.9

sisaman:
Fixes zero `num_edges` when using native PyTorch sparse tensor (#7103).

### DIFF
--- a/torch_geometric/data/storage.py
+++ b/torch_geometric/data/storage.py
@@ -418,6 +418,8 @@ class EdgeStorage(BaseStorage):
         for value in self.values('adj', 'adj_t'):
             if isinstance(value, SparseTensor):
                 return value.nnz()
+            elif isinstance(value, Tensor) and (value.is_sparse or value.is_sparse_csr):
+                return value._nnz()
         return 0
 
     @property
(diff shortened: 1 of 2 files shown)


torch_geometric/data/storage.py
```diff
@@ -418,6 +418,9 @@ def num_edges(self) -> int:
         for value in self.values('adj', 'adj_t'):
             if isinstance(value, SparseTensor):
                 return value.nnz()
+            elif isinstance(value, Tensor) and (value.is_sparse
```
EdisonLeeeee, line 421: Can we use `torch_geometric.utils.is_torch_sparse_tensor` instead?

EdisonLeeeee (commented):
Thank you for catching!

rusty1s (approved):

Issue #6077, opened 2022-11-27
Implement Neural Fingerprint from Duvenaud et al.


j-adamczyk:
### 🚀 The feature, motivation and pitch

Neural Fingerprint, continous alternative for molecular fingerprints via convolutional GNNs, is a well known approach to graph classification. PyG already has [the appropriate layer implemented](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.MFConv), and extending this for a full model would be simple, and it would make making benchmarks easier. At the same time it would be very similar to the paper, more so than [DeepChem implementation, which features some changes](https://github.com/deepchem/deepchem/issues/1160). Thus it would be a useful and easy contribution.

rusty1s:
Definitely happy to take this in :)

kkothari2001:
Hi @rusty1s 
I have prior experience in PyG, but am new to contributing to PyTorch Geometric. Could I work on this issue? Some directions would be appreciated!

j-adamczyk:
@kkothari2001 starting with [GCN](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/models/basic_gnn.html#GCN) would probably be a good start. You could probably even make this almost identically, inheriting from `BasicGNN`. This would not only implement Neural Fingerprint, but also allow users to extend it, e.g. with dropout.

Kumbong:
@kkothari2001 this looks interesting. Not sure if you are still working on this. Can I pick this up ?

Kumbong:
Feel free to. Currently working on a different issue

EulerPascal404:
> @Kumbong @rusty1s can I also start working on this? I have a few holidays coming up so I plan to work on this issue this week. I am new to PyTorch Geometric and open source contributions so working on this will provide me with a valuable experience.

@harshit5674 I am in the same boat as you. Would you be interested in collaborating/splitting up the work. It might make it easier for both of us.

harshit5674:
@EulerPascal404 Yeah sure, i will be happy to collaborate. So how would this work, would we discuss through on this on some platform?

EulerPascal404:
> @EulerPascal404 Yeah sure, i will be happy to collaborate. So how would this work, would we discuss through on this on some platform?

@harshit5674 Great. I think we can just collaborate on this thread(unless you have a better way).

EulerPascal404:
[...]

PR #7105, opened 2023-04-02
Implementing Neural Fingerprint from Duvenaud et al.

harshit5674:
I have tried to implement the neural fingerprint model as suggested in the paper "Convolutional Networks on Graphs for Learning Molecular Fingerprints" and suggested in the issue #6077.

I am new to PyTorch geometric and open source in general so there may be many trivial bugs/errors in the program. So I kindly request the developers/community to guide me. Your guidance will help me learn so much more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added support for floating-point slicing in `Dataset`, *e.g.*, `dataset[:0.9]` ([#7915](https://github.com/pyg-team/pytorch_geometric/pull/7915))
 - Added nightly GPU tests ([#7895](https://github.com/pyg-team/pytorch_geometric/pull/7895))
 - Added the `HalfHop` graph upsampling augmentation ([#7827](https://github.com/pyg-team/pytorch_geometric/pull/7827))
 - Added the `Wikidata5M` dataset ([#7864](https://github.com/pyg-team/pytorch_geometric/pull/7864))
--- a/test/datasets/test_enzymes.py
+++ b/test/datasets/test_enzymes.py
@@ -18,6 +18,7 @@ def test_enzymes(get_dataset):
     assert len(dataset.shuffle()) == 600
     assert len(dataset.shuffle(return_perm=True)) == 2
     assert len(dataset[:100]) == 100
+    assert len(dataset[0.1:0.2]) == 60
     assert len(dataset[torch.arange(100, dtype=torch.long)]) == 100
     mask = torch.zeros(600, dtype=torch.bool)
     mask[:100] = 1
--- a/torch_geometric/data/dataset.py
+++ b/torch_geometric/data/dataset.py
@@ -275,6 +275,14 @@ class Dataset(torch.utils.data.Dataset, ABC):
         indices = self.indices()
 
         if isinstance(idx, slice):
+            start, stop, step = idx.start, idx.stop, idx.step
+            # Allow floating-point slicing, e.g., dataset[:0.9]
+            if isinstance(start, float):
+                start = round(start * len(self))
+            if isinstance(stop, float):
+                stop = round(stop * len(self))
+            idx = slice(start, stop, step)
+
             indices = indices[idx]
 
         elif isinstance(idx, Tensor) and idx.dtype == torch.long:


torch_geometric/nn/models/neural_fingerprint.py
```diff
@@ -0,0 +1,73 @@
...
+from torch_geometric.utils import add_self_loops
+
+
+class NeuralConv(MessagePassing):
+    def __init__(self, num_features, fingerprint_length):
+        super(NeuralConv, self).__init__(aggr="add")
+        self.H = nn.ModuleList()
+        # H[i] would correspond to the hidden weight which would be used
+        # when num_neighbors=i+1
+        for i in range(5):
```
EulerPascal404, line 14: The max degree(5) shouldn't be hard-coded in. You should pass it in as a parameter.

torch_geometric/nn/models/neural_fingerprint.py
```diff
@@ -0,0 +1,73 @@
...
+
+        return self.propagate(edge_index, size=(x.size(0), x.size(0)), x=x)
+
+    def message(self, x_i, x_j, edge_index, size):
+        num_neighbors = x_j.size(0)
+
+        return {"msg": x_j, "num_neighbors": num_neighbors}
+
+    def update(self, aggr_out):
+        num_neigbors = aggr_out["num_neighbors"] - 2
```
EulerPascal404, line 30: Why is this -2, and not -1? We added only 1 self-loop, right?

EulerPascal404 (changes requested):
Nice implementation. Couple minor things that you should take a look at.

torch_geometric/nn/models/neural_fingerprint.py
```diff
@@ -0,0 +1,73 @@
+import torch
+import torch.nn as nn
+
+from torch_geometric.nn import MessagePassing
+from torch_geometric.utils import add_self_loops
+
+
+class NeuralConv(MessagePassing):
+    def __init__(self, num_features, fingerprint_length):
```
EulerPascal404, line 9: add parameter max_degree

torch_geometric/nn/models/neural_fingerprint.py
harshit5674: You are right, but in the paper when they were implementing the model they hardcoded it as 5, because max neighbours of an atom in an organic molecule can go up to 5, so just for the sake of following the paper I hardcoded it as 5.

torch_geometric/nn/models/neural_fingerprint.py
harshit5674: Yes the number of neighbours would be with -1, but since self.H is zero indexed meaning self.H[ i ] would correspond to i+1 neighbours.

torch_geometric/nn/models/neural_fingerprint.py
EulerPascal404: Still think you should change it, in case a molecular graph is being passed in that has a max degree higher than 5.

torch_geometric/nn/models/neural_fingerprint.py
wsad1: +1. Add parameter `max_degree` and set default value to 5.

torch_geometric/nn/models/neural_fingerprint.py
```diff
@@ -0,0 +1,73 @@
...
+
+class NeuralFingerprint(nn.Module):
+    """The NeuralFingerprint is the model suggested in the paper
+    <https://arxiv.org/pdf/1509.09292.pdf>
+    The model generates fingerprints for molecules.
+
+    Args:
+        num_features(int): The number of features of the atom in the data.
+        fingerprint_length(int): Length of the fingerprint generated.
+        radius(int): Radius is the number of layers in the model.
```
wsad1, line 46:
```suggestion
        in_channels(int): The number of features of the atom in the data.
        out_channels(int): Length of the fingerprint generated.
        num_layers(int): Radius is the number of layers in the model.
```
Make docs consistent with other classes in `models` like https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.models.GCN.html#torch_geometric.nn.models.GCN

wsad1 (commented):
Really appreciate your work @harshit5674. 
Usually every PR has a test too. Could you please add a new test file to `pytorch_geometric/test/nn/models `for `neural_fingerprint`

torch_geometric/nn/models/neural_fingerprint.py
harshit5674: @wsad1 Thanks for your review, I will change the max_degree thing and set it to default 5. Sorry for the trivial question but can you please guide me as to what should be in a test file, I am new to contributing.

wsad1: For this model you can test two things:
1. Test the model with random inputs and 2 different `num_layers`. 
2. Add a `__repr__` function and test the output of that too.
You can refer to these tests: https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/models/test_rect.py and https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/models/test_rev_gnn.py

harshit5674: @wsad1 Thanks for the guidance, I will start working on the tests.

harshit5674: I apologise for the delay, currently I am into my final exams of the semester, I will try to add the tests as soon as possible.

harshit5674: Sorry for the delay, I am now done with my exams, I Will try to add the tests by this weekend.

rusty1s: Yes, this is what we usually do. You can look up other test files of models as guidance. It doesn't need to be complex.

harshit5674: @rusty1s @wsad1 I am starting to realise that there may be many problems with my implementation. First of all I think I cannot return a dictionary from the message method.
The paper requires to apply a different weight depending on the number of neighbours of each node, due to my lack of understanding I thought that the message function deals with a single node at a time so I assumed that number of neighbours would be x_j.size(0), now I get it it's wrong.
Since I am beginner, I would request to please guide me as to how I can apply different weights to different nodes based on their degree.
I will do my best to implement this correctly, I just need some help.

rusty1s: Oh wait, isn't this what we already do in https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/mf_conv.py?

harshit5674: Thank you, I will look into this!

harshit5674:
Apologies for the inactivity on this PR(had a hectic summer).
I by mistake closed the PR, I will try to resolve the issue this week itself and probably open a new PR and then link it here.

PR #7109, opened 2023-04-03
Update datamodule.py

JakubPietrakIntel:
Bugfix for datamodule.py
Shouldn't line 102 say `kwargs.pop('batch_sampler', None)` ?


EulerPascal404 (approved):
Yes, that seems to be an error. Good catch.

PR #7117, opened 2023-04-04
Replacing `HGTConv` with `FastHGTConv`

puririshi98:
we have proven its correctness and speed, lmk if anything else needed to merge fasthgt->hgt


EulerPascal404 (approved):

torch_geometric/nn/conv/hgt_conv.py
```diff
@@ -146,54 +170,58 @@ def forward(
...
+        # Transform output node embeddings:
+        a_dict = self.out_lin({
+            k: torch.nn.functional.gelu(v) if v is not None else v
+            for k, v in out_dict.items()
+        })
 
-            if out is None:
+        # Iterate over node types:
+        for node_type, out in out_dict.items():
+            if out is None or node_type not in self.dst_node_types:
```
wsad1, line 208: When would `out is None` be True?

wsad1 (commented):
Thanks for the update.
The changes look good to me. How did you access the correctness of `fast_hgt_conv`?

torch_geometric/nn/conv/hgt_conv.py
puririshi98: I ran hgt and fast hgt for the hgt_dblp.py example and get matching results:
[hgt_vs_fasthgt.txt](https://github.com/pyg-team/pytorch_geometric/files/11187738/hgt_vs_fasthgt.txt)
also checked that their outputs matched w/ matching params
https://github.com/puririshi98/rgcn_pyg_lib_forward_bench/blob/main/hgt_correctness.py

puririshi98:
@rusty1s lmk if anything else needed to merge

puririshi98:
thanks for approving @wsad1, i will merge this and add additional testing in a sep PR

Issue #6964, opened 2023-03-19
Unnecessarily large memory usage in `from_networkx()`


pi314ever:
### 🐛 Describe the bug

[`from_networkx()`](https://github.com/pyg-team/pytorch_geometric/blob/209268a4e48c9c9f701f02f98520d5b96c53468e/torch_geometric/utils/convert.py#L169) appears to copy the data for `edge_index` and `x` multiple times throughout the creation of the Data object. For large graphs, this results in a high likelihood of OOM errors. 

In particular, here are the places `edge_index` is copied:
- lines 209-214 makes 2 copies
- line 239 makes a fragmented copy (using python `int` objects)
- line 278 creates another copy since `torch.cat` copies the data

And here are the places `x` is copied:
- line 232 makes a fragmented copy (using python `int` objects)
- line 266 creates another copy since `torch.cat` copies the data

In total, there are 5 copies of `edge_index` and 3 copies of `x` (including the original) from what I can tell, and more copies of other data as well. 

### Environment

* PyG version: 2.2.0
* PyTorch version: 1.13.1
* OS: Arch Linux
* Python version: 3.7.1
* CUDA/cuDNN version: 12.1
* How you installed PyTorch and PyG (`conda`, `pip`, source): `pip install ...` for cu117
* Any other relevant information (*e.g.*, version of `torch-scatter`): 
```
torch-scatter==2.1.0+pt113cu117
torch-sparse==0.6.16+pt113cu117
torch-spline-conv==1.2.1+pt113cu117
networkx==2.6.3
```

rusty1s:
Thanks for this issue. You are right that we are indeed copying here multiple times. The only way to optimize this would be to directly write edges into a `torch.Tensor` rather than first converting to a list and then using `torch.tensor()` on top, but IMO this feels like a negligible optimization since the data is freed afterwards anyway.

If you have interest in making this more memory-friendly, I welcome a PR though :)

rusty1s:
Done via https://github.com/pyg-team/pytorch_geometric/pull/7119.

PR #7119, opened 2023-04-04
Optimize `from_networkx()`

EulerPascal404:
Addresses issue #6964 by deleting unneeded copies by directly writing edges into a torch.Tensor rather than first converting to a list and then using torch.tensor() on top.

### DIFF
--- a/torch_geometric/utils/convert.py
+++ b/torch_geometric/utils/convert.py
@@ -204,10 +204,13 @@ def from_networkx(
 
     G = nx.convert_node_labels_to_integers(G)
     G = G.to_directed() if not nx.is_directed(G) else G
-    edge_index = torch.empty(G.number_of_edges(), 2, dtype=torch.long)
-    for i, e in enumerate(G.edges):
-        edge_index[i, 0], edge_index[i, 1] = e
-    edge_index = edge_index.t().contiguous()
+    if isinstance(G, (nx.MultiGraph, nx.MultiDiGraph)):
+        edges = list(G.edges(keys=False))
+    else:
+        edges = list(G.edges)
+
+    edge_index = torch.tensor(edges, dtype=torch.long).t().contiguous()
+    del edges
 
     data = defaultdict(list)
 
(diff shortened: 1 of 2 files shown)


EulerPascal404:
FYI, the unit test for this file seems to be broken(I tested it before I made any changes). The code changes I implemented didn't change the status of the test.

EdisonLeeeee:
Thank you for the PR!

EdisonLeeeee (approved):

torch_geometric/utils/convert.py
```diff
@@ -204,13 +204,10 @@ def from_networkx(
...
-    if isinstance(G, (nx.MultiGraph, nx.MultiDiGraph)):
-        edges = list(G.edges(keys=False))
-    else:
-        edges = list(G.edges)
-
-    edge_index = torch.tensor(edges, dtype=torch.long).t().contiguous()
+    edge_index = torch.empty(G.number_of_edges(), 2, dtype=torch.long)
+    for i, e in enumerate(G.edges):
+        edge_index[i, 0], edge_index[i, 1] = e
+    edge_index = edge_index.t().contiguous()
```
wsad1, line 210: this is a for loop over edges, which can be slow if there are many edges. 
What if instead we call `del edges` after creating `edge_index` and allow the garbage collector to free up that memory if needed, this should make the OOM less likely.
```suggestion
    if isinstance(G, (nx.MultiGraph, nx.MultiDiGraph)):
        edges = list(G.edges(keys=False))
    else:
        edges = list(G.edges)

    edge_index = torch.tensor(edges, dtype=torch.long).t().contiguous()
    del edges
```

EulerPascal404 (commented):
Sure, this is also a feasible solution - I just implemented the solution @rusty1s suggested in the issue. We should do this at other places in the file too, as per the initial issue.

rusty1s (approved):
Thanks! I changed it back to the original implementation of yours since we are iterating over all edges anyway here.

PR #7126, opened 2023-04-05
Additional GPU tests

puririshi98:


test/data/test_feature_store.py
```diff
@@ -22,21 +26,23 @@ def __init__(self):
         self._tensor_attr_cls = MyTensorAttrNoGroupName
 
 
-def test_feature_store():
+@pytest.mark.parametrize('device', DEVICES)
+def test_feature_store(device):
```
mananshah99, line 30: Do we need to parameterize on device here (in this file and in the graph store test file as well)? Seems like we are not really changing what we test based on the device.

test/nn/conv/test_rgcn_conv.py
```diff
@@ -4,31 +4,43 @@
...
+
+DEVICES = [torch.device('cpu')]
+if torch.cuda.is_available():
+    DEVICES.append(torch.device('cuda'))
+    import os
+    os.environ['NVIDIA_TF32_OVERRIDE'] = '0'
+    torch.backends.cuda.matmul.allow_tf32 = False
+    minor_vers = str(torch.__version__).split('.')[1]
+    if WITH_PT2 or minor_vers >= 12:  # This only exists after 1.12
+        torch.set_float32_matmul_precision('highest')  # Enforce FP32
```
mananshah99, line 17: Curious, what is happening here? Why is this needed?

mananshah99 (commented):
Shall we use the `withCUDA` decorator instead of repeating the `DEVICES` logic?

Kh4L (approved):

test/nn/conv/test_rgcn_conv.py
puririshi98: in nvidia containers, default matmuls act in tf32 and we were getting numerical mismatches. I have the same code in pyg-lib (thats where i copypasted from): https://github.com/pyg-team/pyg-lib/blob/master/test/ops/test_matmul.py#L7-L19

test/data/test_feature_store.py
puririshi98: valid point, will adjust

rusty1s (approved):

PR #7156, opened 2023-04-11
Fix `Batch.from_data_list` only store node/edge types of first element for `HeteroData`

### LINKED ISSUE #7147: `Batch.from_data_list` for `HeteroData` omits all node types that don't appear in the first item
### 🐛 Describe the bug

When constructing a batch of heterogeneous data, it seems that all node types must appear in the first item passed to `Batch.from_data_list`. Otherwise, the missing types would be omitted from the batch.

**First case (b1 is the first item)**
```python
import torch
from torch_geometric.data import HeteroData, Batch
b1 = HeteroData({"agent1":{"x":torch.randn(3,2)}})
b2 = HeteroData({"agent2":{"x":torch.randn(3,2)}})
print(Batch.from_data_list([b1,b2]).node_types)
``` 

Output:
```
['agent1']
```

**Second case (b2 is the first item)**
```python
import torch
from torch_geometric.data import HeteroData, Batch
b1 = HeteroData({"agent1":{"x":torch.randn(3,2)}})
b2 = HeteroData({"agent2":{"x":torch.randn(3,2)}})
print(Batch.from_data_list([b2,b1]).node_types)
``` 

Output:
```
['agent2']
```

**Third case (b1 is the first item and contains both types)**
```python
import torch
from torch_geometric.data import HeteroData, Batch
b1 = HeteroData({"agent1":{"x":torch.randn(3,2)}, "agent2":{"x":torch.randn(3,2)}})
b2 = HeteroData({"agent2":{"x":torch.randn(3,2)}})
print(Batch.from_data_list([b1,b2]).node_types)
``` 

Output:
```
['agent1', 'agent2']
```

**Fourth case (b1 is the second item and contains both types)**
```python
import torch
from torch_geometric.data import HeteroData, Batch
b1 = HeteroData({"agent1":{"x":torch.randn(3,2)}, "agent2":{"x":torch.randn(3,2)}})
b2 = HeteroData({"agent2":{"x":torch.randn(3,2)}})
print(Batch.from_data_list([b2,b1]).node_types)
``` 

Output:
```
['agent2']
```

Note that the agent

### Environment

* PyG version: 2.3.0
* PyTorch version: 1.13.1
* OS: Ubuntu
* Python version: 3.8
* CUDA/cuDNN version:
* How you installed PyTorch and PyG (`conda`, `pip`, source): conda
* Any other relevant information (*e.g.*, version of `torch-scatter`):

EdisonLeeeee:
Fix #7147. Will this modification be acceptable when the length of `data_list` is long?


rusty1s:
Hey @EdisonLeeeee, sorry for late reply here. I don't think this fix is necessarily correct. For example `batch['a'].batch` gives you a zero-vector, although it should be a vector of ones (since only node type `a` is present in `data2`). I personally don't think this necessarily needs fixing though at all. This seems more like a user-error, since we expect that all node types and its attributes are identical across different batches (the same holds true in the homogeneous case). If we allow such behavior, this can get confusing pretty fast. For example, if you are unlucky and produce a mini-batch for which all data objects are missing a certain node type, then your GNN may not necessarily work as expected anymore because the expected metadata does not align.

A simple fix to that is to just add empty tensors to node types, e.g.:
```python
data1['a'].x = torch.empty(0, 128)
```
and everything works as expected.

EdisonLeeeee:
Ok, you're right. Thank you for the clarification! Will close it now.

ekosman:
> Hey @EdisonLeeeee, sorry for late reply here. I don't think this fix is necessarily correct. For example `batch['a'].batch` will give you a zero-vector with this PR, although it should be a vector of ones (since only in `data2` the node type `a` is present). I personally don't think this necessarily needs fixing though at all. This seems more like a user-error, since we expect that all node types and their attributes are exactly identical across different batches (the same holds true in the homogeneous case). If we allow such behavior, this can get confusing pretty fast. For example, if you are unlucky and produce a mini-batch for which all data objects are missing a certain node type, then your GNN may not necessarily work as expected anymore because the expected metadata does not align. If we allow that some attributes can be missing, then `data.x` may not be defined at all.
> 
> A simple fix to that is to just add empty tensors to node types, e.g.:
> 
> ```python
> data1['a'].x = torch.empty(0, 128)
> ```
> 
> and everything works as expected.

Ok, thanks for clarifying.

PeterDavison01:
Hi folks - apologies for reviving a stale req. I agree with @rusty1s, but I've been struggling with the lack of a clear failure. This behaviour does not appear to be documented in the public docs for `DataLoader` / `Batch.from_data_list()` / `HeteroData` or the Heterogeneous Graph learning tutorial as far as I am aware. I couldn't find a built-in validation check in the docs as well, which leads to silent data loss. With `shuffle=True` it is especially painful; which subset of stores survives depends on whichever graph happens to land in `data_list[0]` for that batch (assuming no static seed is set), so the same dataset produces different `x_dict.keys()` from step to step.

For datasets with a wide, sparse type vocabulary (industrial timelines in my case, but the same shape shows up in molecular datasets with sparse atom types, etc.), a given graph naturally only touches a subset of the schema, so this failure mode is easy to hit without realising it. Would you consider adding a docs note about the store key requirement and the “pad with empty tensors” workaround, or a warning in `collate()`?

PR #7185, opened 2023-04-16
Fix `HeteroDictLinear` crashes if some node types are absent

berkekisin:
This pr fixes a bug in HeteroDictLinear forward function, which crashes if the input x_dict has some node types missing.


rusty1s:
Currently, we need to iterate over `lins` due to TorchScript limitations. One thing we can probably do is skip any `lins` that are not part of `x_dict`:
```python
weights = [lin.weight.t() for key, lin in self.lins.items() if key in x_dict]
```

rusty1s (approved):
I think I fixed it.

PR #7191, opened 2023-04-17
add an example to do graph partition(support Homogeneous graph first)

kaixuanliu:
@rusty1s We create this PR as a preparation step for distributed GNN training. It is a common practice to do graph partition when we have a big dataset. Although pyG has already have a ClusterData class to do this, it saves all the partition data into one single file. We think it may be a better choice to save graph topology and node/edge features for each partition separately. Hence we give an example to show how to implement this based on existing ClusterData class. For now it only supports homogeneous graph partition, and we will add heterogeneous graph support soon later. 
To show you what we want to do, and get your feedback quickly, here we just write a simple example. After aligning design direction with you, we plan to encapsulate the core partition function into a new class (support both homo/hetero).  Your opinions will help us a lot to get this work move on. Thanks!!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,34 @@
+# About code owners
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+#
+# Code owners recieve review requests when parts of code they "own" change.
+
+*.py @rusty1s @wsad1
+
+/torch_geometric/nn/ @EdisonLeeeee
+
+/torch_geometric/explain/ @RexYing @RBendias @dufourc1
+
+/torch_geometric/data/ @mananshah99
+
+/torch_geometric/loader/ @mananshah99 @pyg-team/intel-team
+
+/torch_geometric/sampler/ @mananshah99 @pyg-team/intel-team
+
+/torch_geometric/transforms/ @wsad1
+
+/torch_geometric/utils/ @wsad1
+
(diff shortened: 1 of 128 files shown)


yao-matrix:
@kding1 @yao-matrix

kaixuanliu:
Hi @rusty1s , besides this homo graph partition example, we have added hetero graph partition support, and wrapped them into a separate class. But we do not know if you think this design is OK or not. Please let us know your opinions so that we can move on to submit our later PRs. Add @ZhengHongming888

rusty1s:
I think this looks good, although I would prefer to disentangle partitioning from `ClusterLoader`. Maybe we can move `ClusterData` to a separate file in `torch_geometric.data`?

kaixuanliu:
> I think this looks good, although I would prefer to disentangle partitioning from `ClusterLoader`. Maybe we can move `ClusterData` to a separate file in `torch_geometric.data`?

The reason we do not disentangle partition data in `ClusterLoader` is we think when the size of graph data is very large, and for distributed training, if one of the machines does not have enough memory, it will meet challenge to load the entire graph data. But in this design, we will separate the large graph data file into several small parts, and it will avoid this case.

For the second question, we wrapped the implementation of homo/hetero graph partition into a new folder called `partition`, and it has two files: `generate_partition.py` and `load_partition.py`. Maybe we can move `Clusterdata` into this new folder. And we can move the new folder `partition` into `torch_geometric.data`? Just a suggestion, you make the decision~

PR #7194, opened 2023-04-17
Fix `HGTConv` `edge_type_vec` construction 

berkekisin:
This pr fixes the utility function https://github.com/pyg-team/pytorch_geometric/blob/3d4836bc24dbb1b180f29cbbbdbcd18b94116dd7/torch_geometric/nn/conv/hgt_conv.py#L123, which constructs the type_vec of edges wrong and also crashes if some edge_types are not present in the current edge_index_dict. 
Consider the following scenario:

```python
# N =2, D=2, H=2 (2 nodes, head_dim 2, 2 heads)
k = [
     [0,0,1,1],
     [2,2,3,3]
    ]
```

after calling this line:
https://github.com/pyg-team/pytorch_geometric/blob/3d4836bc24dbb1b180f29cbbbdbcd18b94116dd7/torch_geometric/nn/conv/hgt_conv.py#L141

the matrix k looks like this:

```python
k= [
     [0,0],
     [1,1],
     [2,2],
     [3,3]]

# the type vec should look like this
type_vec = [0,1,0,1]
# but at current implementation it would look like this
type_vec = [0,0,1,1]
```

After the reshape the attention heads are interleaved but  the type vector that  is currently constructed is sorted. 
We fixed this issue by constructing interleaved type vec. Alternatively we can transpose the k before the reshape to ensure that we can use sorted type vec. This will also allow us to set `is_sorted=True` for the heterolinear `k_rel` which would be more efficient.
Also note that we added a test case for missing edge type in edge_index_dict.

### DIFF
--- a/test/nn/conv/test_hgt_conv.py
+++ b/test/nn/conv/test_hgt_conv.py
@@ -204,9 +204,7 @@ def test_hgt_conv_missing_input_node_type():
          'paper'].edge_index = get_random_edge_index(4, 6, 20)
 
     metadata = (['author', 'paper',
-                 'university'], [('author', 'writes', 'paper'),
-                                 ('university', 'employs', 'author'),
-                                 ('author', 'works at', 'university')])
+                 'university'], [('author', 'writes', 'paper')])
     conv = HGTConv(-1, 64, metadata, heads=1)
 
     out_dict = conv(data.x_dict, data.edge_index_dict)
--- a/test/nn/conv/test_hgt_conv.py
+++ b/test/nn/conv/test_hgt_conv.py
@@ -230,7 +230,7 @@ def test_hgt_conv_missing_edge_type():
     out_dict = conv(data.x_dict, data.edge_index_dict)
     assert out_dict['author'].size() == (4, 64)
     assert out_dict['paper'].size() == (6, 64)
-    assert 'university' not  in out_dict
+    assert 'university' not in out_dict
 
 
 if __name__ == '__main__':
--- a/test/nn/conv/test_hgt_conv.py
+++ b/test/nn/conv/test_hgt_conv.py
@@ -196,25 +196,6 @@ def test_hgt_conv_missing_dst_node_type():
     assert out_dict['university'] is None
 
 
-def test_hgt_conv_missing_src_node_type():
-    data = HeteroData()
-    data['author'].x = torch.randn(4, 16)
-    data['paper'].x = torch.randn(6, 32)
-
-    data['author', 'writes', 'paper'].edge_index = get_random_edge_index(4, 6, 20)
-
-    metadata = (
-        ['author', 'paper', 'university'],
-        [('author', 'writes', 'paper'), ('university', 'employs', 'author')]
-    )
-    conv = HGTConv(-1, 64, metadata, heads=1)
-
-    out_dict = conv(data.x_dict, data.edge_index_dict)
-    assert out_dict['author'].size() == (4, 64)
-    assert out_dict['paper'].size() == (6, 64)
-    assert out_dict['university'] == (10, 64)
-
-
 def test_hgt_conv_missing_edge_type():
     data = HeteroData()
     data['author'].x = torch.randn(4, 16)
--- a/test/nn/conv/test_hgt_conv.py
+++ b/test/nn/conv/test_hgt_conv.py
@@ -203,6 +203,8 @@ def test_hgt_conv_missing_input_node_type():
     data['author', 'writes',
          'paper'].edge_index = get_random_edge_index(4, 6, 20)
 
+    # Some nodes from metadata are missing in data.
+    # This might happen while using NeighborLoader.
     metadata = (['author', 'paper',
                  'university'], [('author', 'writes', 'paper')])
     conv = HGTConv(-1, 64, metadata, heads=1)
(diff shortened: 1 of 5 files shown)


test/nn/conv/test_hgt_conv.py
```diff
@@ -196,6 +196,26 @@ def test_hgt_conv_missing_dst_node_type():
...
+    data['author'].x = torch.randn(4, 16)
+    data['paper'].x = torch.randn(6, 32)
+    data['university'].x = torch.randn(10, 32)
+
+    data['author', 'writes',
+         'paper'].edge_index = get_random_edge_index(4, 6, 20)
+
+    metadata = (['author', 'paper',
+                 'university'], [('author', 'writes', 'paper'),
+                                 ('university', 'employs', 'author')])
```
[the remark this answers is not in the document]
toenshoff: The input graph could be a small subgraph sampled from a much bigger graph on which the model was defined. Depending on graph characteristics not all edge types from the full graph may be present in every subgraph of interest. This is similar to the use case covered in [#7060](https://github.com/pyg-team/pytorch_geometric/pull/7060).

On the same note, src and dst node types may also be absent in the input graph. However, I am not sure what the defined output behavior of HGT is supposed to be in this case. Should the output dict map these node types to None (as it does for none-destination types) or should it simply not contain anything for missing inputs? I think the None values can be a problem when several layers are stacked.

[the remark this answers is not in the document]
toenshoff: Sounds good, I'll make this change to the out_dict.

berkekisin:
> > Alternatively we can transpose the k before the reshape to ensure that we can use sorted type vec. This will also allow us to set is_sorted=True for the heterolinear k_rel which would be more efficient.
> 
> @berkekisin @toenshoff , I like this proposal much more, would make the code more readable too.

After trying every possible version :) we decided to design the function this way.

berkekisin:
@wsad1 we decided to design the function this way and find this to be the most efficient version.

berkekisin:
We just realized a small bug will update that as soon as possible

test/nn/conv/test_hgt_conv.py
```diff
@@ -193,7 +193,46 @@ def test_hgt_conv_missing_dst_node_type():
...
+    data = HeteroData()
+    data['author'].x = torch.randn(4, 16)
+    data['paper'].x = torch.randn(6, 32)
+    data['author', 'writes',
+         'paper'].edge_index = get_random_edge_index(4, 6, 20)
+
+    metadata = (['author', 'paper',
+                 'university'], [('author', 'writes', 'paper'),
+                                 ('university', 'employs', 'author'),
+                                 ('author', 'works at', 'university')])
```
wsad1, line 209:
```suggestion
                 'university'], [('author', 'writes', 'paper')])
```

test/nn/conv/test_hgt_conv.py
```diff
@@ -193,7 +193,44 @@ def test_hgt_conv_missing_dst_node_type():
...
+
+    metadata = (['author', 'paper',
+                 'university'], [('author', 'writes', 'paper'),
+                                 ('university', 'employs', 'author')])
+    conv = HGTConv(-1, 64, metadata, heads=1)
+
+    out_dict = conv(data.x_dict, data.edge_index_dict)
+    assert out_dict['author'].size() == (4, 64)
+    assert out_dict['paper'].size() == (6, 64)
+    assert not 'university' in out_dict
```
wsad1, line 233:
```suggestion
    assert 'university' not  in out_dict
```

test/nn/conv/test_hgt_conv.py
```diff
@@ -193,7 +193,44 @@ def test_hgt_conv_missing_dst_node_type():
...
+    data['paper'].x = torch.randn(6, 32)
+    data['author', 'writes',
+         'paper'].edge_index = get_random_edge_index(4, 6, 20)
+
+    metadata = (['author', 'paper',
+                 'university'], [('author', 'writes', 'paper')])
+    conv = HGTConv(-1, 64, metadata, heads=1)
+
+    out_dict = conv(data.x_dict, data.edge_index_dict)
+    assert out_dict['author'].size() == (4, 64)
```
wsad1, line 211:
```suggestion
```

test/nn/conv/test_hgt_conv.py
```diff
@@ -193,7 +193,43 @@ def test_hgt_conv_missing_dst_node_type():
...
+
+
+def test_hgt_conv_missing_input_node_type():
+    data = HeteroData()
+    data['author'].x = torch.randn(4, 16)
+    data['paper'].x = torch.randn(6, 32)
+    data['author', 'writes',
+         'paper'].edge_index = get_random_edge_index(4, 6, 20)
+
+    metadata = (['author', 'paper',
```
wsad1, line 206:
```suggestion
    # Some nodes from metadata are missing in data.
    # This might happen while using NeighborLoader.
    metadata = (['author', 'paper',
```

wsad1 (approved):
Thanks for fixing this.

berkekisin:
@wsad1 `def test_hgt_conv_missing_input_node_type():` currently fails because of a small bug in Heterodictlinear, we will create a pr for this tomorrow.

PR #7195, opened 2023-04-17
[Code Coverage] `data/datapipes.py`

akihironitta:
Part of #6528.
Completes #6799.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,8 @@ full = [
     "pytorch-memlab",
     "pgmpy",
     "opt_einsum",
-    "statsmodels"
+    "statsmodels",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -86,7 +86,8 @@ full = [
     "pytorch-memlab",
     "pgmpy",
     "opt_einsum",
-    "statsmodels"
+    "statsmodels",
+    "rdkit",
```
akihironitta, line 90: Added this dependency to run the corresponding test case in CI. The coverage is shown less because the test case is skipped with `@withPackage('rdkit')`.

rusty1s (approved):

PR #7216, opened 2023-04-20
Add `batch_size` argument to `avg_pool_x` and `max_pool_x`

kjkozlowski:
It can speedup runtime because:
1. We do not need to go through the batch dimension and look for max value.
2. We do not have to read tensor value which is placed on the device.

Besides dim_size can be used if a user is using fixed size datasets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Added an optional `batch_size` argument to `avg_pool_x` and `max_pool_x` ([#XXXX](https://github.com/pyg-team/pytorch_geometric/pull/XXXX))
+- Added an optional `batch_size` argument to `avg_pool_x` and `max_pool_x` ([#7216](https://github.com/pyg-team/pytorch_geometric/pull/7216))
 - Fixed `subgraph` on unordered inputs ([#7187](https://github.com/pyg-team/pytorch_geometric/pull/7187))
 - Allow missing node types in `HeteroDictLinear` ([#7185](https://github.com/pyg-team/pytorch_geometric/pull/7185))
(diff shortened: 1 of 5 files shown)


torch_geometric/nn/pool/avg_pool.py
```diff
@@ -34,14 +35,17 @@ def avg_pool_x(
         batch (torch.Tensor): The batch vector
             :math:`\mathbf{b} \in {\{ 0, \ldots, B-1\}}^N`, which assigns each
             node to a specific example.
+        batch_size (int, optional): The number of examples :math:`B`.
+                Automatically calculated if not given. (default: :obj:`None`)
```
rusty1s, line 39:
```suggestion
            Automatically calculated if not given. (default: :obj:`None`)
```

torch_geometric/nn/pool/max_pool.py
```diff
@@ -33,6 +34,8 @@ def max_pool_x(
         batch (torch.Tensor): The batch vector
             :math:`\mathbf{b} \in {\{ 0, \ldots, B-1\}}^N`, which assigns each
             node to a specific example.
+        batch_size (int, optional): The number of examples :math:`B`.
+                Automatically calculated if not given. (default: :obj:`None`)
```
rusty1s, line 38:
```suggestion
            Automatically calculated if not given. (default: :obj:`None`)
```

rusty1s (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Added an optional `batch_size` argument to `avg_pool_x` and `max_pool_x` ([#XXXX](https://github.com/pyg-team/pytorch_geometric/pull/XXXX))
```
rusty1s, line 30:
```suggestion
- Added an optional `batch_size` argument to `avg_pool_x` and `max_pool_x` ([#7216](https://github.com/pyg-team/pytorch_geometric/pull/7216))
```

PR #7221, opened 2023-04-22
Test `return_emb` in `MLP`

vstenby:
There was a bug in the .forward method of the MLP class. If return_emb=False, then it would still return a tuple containing the embeddings. The line with the bug was:

```python
return (x, emb) if isinstance(return_emb, bool) else x
```
since `isinstance(False, bool)` is True.

### DIFF
--- a/test/nn/models/test_mlp.py
+++ b/test/nn/models/test_mlp.py
@@ -2,7 +2,6 @@ import pytest
 import torch
 
 from torch_geometric.nn import MLP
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
We currently need this hack to be compatible with TorchScript. Added some tests and comments to make this more clear.

vstenby:
Ah, that makes sense! Thanks @rusty1s! 💪🏻

PR #7227, opened 2023-04-24
Add `all_padding_self_loops` argument to `Pad` transform

bwroblew:
This PR adds an optional `all_padding_self_loops` argument to `Pad` transform. This option changes the way new edges are generated for `Data` objects and stores of `HeteroData` objects with the same `src` and `dst` node type.

Before, all the edges were self-loops on the first padding nodes. It's a bit problematic in case when such a data is later processed by a (mostly) convolution layer that requires it to have all the self-loops. Calling `add_remaining_self_loops` on such an object changes the dimensions of `edge_index` tensor -- in general, it changes the number of edges by removing duplicated self-loops on the first padding node and adding missing ones on other nodes.

This change adds an option that fixes this behavior by adding exactly one self-loop on every padding nodes. When combined with `AddRemainingSelfLoops` transform, `Pad` transform after this change changes graphs in such a way that calling `add_remaining_self_loop` (or `AddRemainingSelfLoops`) on them doesn't change the number of edges, and the dimensions of the tensors.

The argument is optional since it adds two new constraints to the data objects and `Pad`'s options -- there must be enough padding edges to create all the self-loops and if there is more padding edges than possible self-loops, we need to have at least 2 padding nodes so we can create the rest of padding edges between them (they cannot be self-loops as those would be removed by `add_remaining_self_loops` calls).


bwroblew:
@rusty1s, could you take a look, please?

bwroblew:
Closing as the team does not need it anymore.

PR #7244, opened 2023-04-27
It adds a minimal example of Hierarichial Sampling on one Epoch.

andreazanetti:
It compares one epoch of training with and without Hierarchical Sampling.
With pyg-lib>0.1.0 we return the sampled number of nodes/edges in [neighbor_sampler.py](https://github.com/pyg-team/pytorch_geometric/blob/e3e63d66e52aa9ca4553274f0572f1f066d99c41/torch_geometric/sampler/neighbor_sampler.py#L241) 
Leveraging this, the [training_benchmark.py](https://github.com/pyg-team/pytorch_geometric/blob/master/benchmark/training/training_benchmark.py) refers to `BasicGNN` base class, in which [the forward pass does the trimming if required](https://github.com/pyg-team/pytorch_geometric/blob/e3e63d66e52aa9ca4553274f0572f1f066d99c41/torch_geometric/nn/models/basic_gnn.py#L201) (using the `--trim` flag with `training_benchmark.py`). 
Therefore, this is an example that mimics what is being done in the `training_benchmark.py,` to make evident for the user what this trimming/Hierarchical Sampling is about, how to test it, and have an idea of the advantage.

### DIFF
--- a/examples/basic_training_with_hierarchical_sampling.py
+++ b/examples/basic_training_with_hierarchical_sampling.py
@@ -1,0 +1,59 @@
+import os.path as osp
+import torch
+import torch.nn.functional as F
+from tqdm import tqdm
+from torch_geometric.datasets import Reddit
+from torch_geometric.loader import NeighborLoader
+from torch_geometric.nn.models.basic_gnn import GraphSAGE
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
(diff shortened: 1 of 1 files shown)


andreazanetti:
@rusty1s  do you have comments on this simple example? Thank you!

rusty1s (approved):

PR #7246, opened 2023-04-27
Introduce `disable_dynamic_shapes` experimental flag; adding its use into `to_dense_batch function`

piotrchmiel:
There are devices that do not support dynamic shapes - (compiling and optimizing only static graphs). The ability to set and read the "disable_dynamic_shapes" flag allows implementors to provide static shape-friendly implementations and report user-friendly messages if it is impossible to avoid using dynamic shapes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added the `disable_dynamic_shape` exeperimental flag ([#7246]https://github.com/pyg-team/pytorch_geometric/pull/7246)
+- Added the `disable_dynamic_shape` experimental flag ([#7246](https://github.com/pyg-team/pytorch_geometric/pull/7246))
 - Added the `MovieLens-1M` heterogeneous dataset ([#7479](https://github.com/pyg-team/pytorch_geometric/pull/7479))
 - Added a CPU-based and GPU-based `map_index` implementation ([#7493](https://github.com/pyg-team/pytorch_geometric/pull/7493))
 - Added the `AmazonBook` heterogeneous dataset ([#7483](https://github.com/pyg-team/pytorch_geometric/pull/7483))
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -26,7 +26,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for `torch.jit.script` within `MessagePassing` layers without `torch_sparse` being installed ([#7061](https://github.com/pyg-team/pytorch_geometric/pull/7061), [#7062](https://github.com/pyg-team/pytorch_geometric/pull/7062))
 - Added unbatching logic for `torch.sparse` tensors ([#7037](https://github.com/pyg-team/pytorch_geometric/pull/7037))
 - Added the `RotatE` KGE model ([#7026](https://github.com/pyg-team/pytorch_geometric/pull/7026))
-
+- Added `disable_dynamic_shape` exeperimental flag ([#7246]https://github.com/pyg-team/pytorch_geometric/pull/7246)
```
rusty1s, line 29:
```suggestion
- Added `disable_dynamic_shape` experimental flag ([#7246]https://github.com/pyg-team/pytorch_geometric/pull/7246)
```

torch_geometric/utils/to_dense_batch.py
```diff
@@ -92,17 +93,30 @@ def to_dense_batch(x: Tensor, batch: Optional[Tensor] = None,
...
     if max_num_nodes is None:
-        max_num_nodes = int(num_nodes.max())
-    elif num_nodes.max() > max_num_nodes:
+        if dynamic_shapes_disabled:
+            raise RuntimeError(
+                "Dynamic shapes disabled. `max_num_nodes` argument should be "
+                "set.")
+        else:
+            max_num_nodes = int(num_nodes.max())
+    elif not dynamic_shapes_disabled and num_nodes.max() > max_num_nodes:
```
rusty1s, line 119: I am wondering if we can make this a decorator instead? For example,
```python
@disable_dynamic_shapes(required_args=['batch_size', 'max_num_nodes'])
def from_dense_batch(...):
    ...
```
seems a bit cleaner than adding this logic to all functions manually.

piotrchmiel: The idea with the decorator looks nice, especially for the validation of the parameters. But what about the possibility of conditional blocking some piece of code or replacing the implementation ? For example, in 119 I want to disable the execution of that part of the code.

rusty1s: Is there a particular reason you want to disable this code-branch here at all? I am okay with only using the decorator for basic validation, and then use `dynamic_shapes_disabled` for some more advanced custom logic (although I am still not sure it is needed here).

piotrchmiel: I'm sorry for not getting back to you sooner. I was on sick leave.
1. I'm okay with the decorator solution. I will prepare a patch with that.
2. I want to disable that part of code because filtering_nodes produces a dynamic mask, which can produce a dynamic shape on the x. Disabling dynamic shapes in torch dynamo will produce an error here during torch compile, it is also a problem for static compiler backends like ours (GC).

piotrchmiel: @rusty1s I pushed change with a decorator.

rusty1s (approved):
Looks beautiful, thank you!

PR #7272, opened 2023-05-02
Minor equation update in documentation

vedal:
Node index here should probably be "i" rather than "v"


EdisonLeeeee (approved):
Thank you for catching!

rusty1s (approved):

PR #7275, opened 2023-05-02
Option to preserve directed graphs in `CitationFull` datasets

rusty1s:


joeloskarsson:
Hi, does this work correctly? The argument `to_undirected` seems to be unused in `read_npz`, in particular it is not fed forward to the `parse_npz` function. I'm assuming `parse_npz` will always make the graphs undirected then, but let me know if I am missing something!

EdisonLeeeee:
Thank you for pointing it out. You are right and it has been fixed via #7843

joeloskarsson:
Cool, thanks for fixing it!

Issue #7260, opened 2023-04-29
Impossibility of two class multilabel classification using GraphGym


RemyLau:
### 🐛 Describe the bug

The current GraphGym model builder makes it impossible to set up multilabel tasks with **exactly** two classes. This is because of the implicit modification of `dim_out` in the model builder function.

https://github.com/pyg-team/pytorch_geometric/blob/8b37ad571b6e08d700f344cd9965724939f4bd4c/torch_geometric/graphgym/model_builder.py#L77-L78

I understand that in *some* cases, it is desirable to truncate the two-class binary task into a single binary classification task when it is ***actually*** a binary classification task. But there is currently no mechanism to ensure whether this is the case and *ALL* two-class classification tasks are treated as a single binary classification task.

I think it would be great to give the option at least to bypass this implicit modification. Or is there any way to work around this?

### Environment

* PyG version: latest
* PyTorch version: 2.0.0
* OS: Linux
* Python version: 3.10.9
* CUDA/cuDNN version: cuda 11.7
* How you installed PyTorch and PyG (`conda`, `pip`, source): `conda`
* Any other relevant information (*e.g.*, version of `torch-scatter`):

rusty1s:
Do you mean that you cannot train a multi-label problem (both classes can be present at the same time) due to this? Wouldn't you want to set `task_type: classification_multi` in this case?

RemyLau:
Hi @rusty1s, thanks for your reply! I did try to use `task_type: classification_multi`. But the issue persists because of the logic I linked earlier, which then turns `dim_out` to 1.

rusty1s:
Oh, wow. I guess this should be "'classification' == cfg.dataset.task_type`. Does that resolve your issues?

RemyLau:
Yes exactly. I think that should work

PR #7284, opened 2023-05-03
fix graphgym two-class multilabel classification logic #7260

RemyLau:


rusty1s (approved):

EdisonLeeeee:
There seems to be a need to make a corresponding change as well:
https://github.com/pyg-team/pytorch_geometric/blob/1b6c40f1bd6ed668a131f0f6e0dcf10f9bf54cb9/torch_geometric/graphgym/logger.py#L229-L238

RemyLau:
Hi @EdisonLeeeee, thanks for pointing this out! I think this is a bit trickier to solve. Because `num_label = 2` could be both multi or binary classifications. My personal preference would be explicitly setting `task_type` to either `classification_binary` or `classification_multi` to eliminate the confusion.

EdisonLeeeee:
Ok, that makes sense :)

Issue #6455, opened 2023-01-18
Graph Pooling Framework :rocket:


rusty1s:
### 🚀 The feature, motivation and pitch

The ["Understanding Pooling in Graph Neural Networks"](https://arxiv.org/abs/2110.05292) paper introduces a simple framework to unify (most) graph pooling approaches:

1. **Select:** Selects input nodes to map to supernodes
2. **Reduce:** Reduces the supernodes to singletons
3. **Connect:** Decides how the new nodes are connected

Currently, the pooling operators included in PyG are implemented isolated from each other, and there exists a lot of repetitive code and different interfaces, which makes their application confusing and challenging. By following the above approach, we can unify existing implementations and can accelerate new research on graph pooling. For example, we can introduce base classes for each of the aforementioned steps:

```python
class Select(torch.nn.Module):
    def forward(self, *args, **kwargs) -> torch.Tensor:
         """Returns a bipartite `edge_index` mapping input nodes to supernodes."""

class Reduce:  # No need -> we can utilize `nn.aggr` for this
    pass

class Connect(torch.nn.Module):
    def forward(self, cluster_index: torch.Tensor, edge_index: torch.Tensor, *args, **kwargs):
        """Returns a coarsened graph."""
```

With this, *e.g.*, `graclus` can be moved to a `Select` operator and `TopK` pooling can be moved to a `Connect` operator.

Relevant twitter thread: https://twitter.com/riceasphait/status/1447867635442585601

Tasks:

- [x] Add `Select` and `Connect` base classes [#6540]
- [x] Add `Pooling` base class [#6540]

Add concrete `Select` and `Connect` classes and update implementations of.
- [x] `TopKPooling` [#7308]
- [x] #7617

wsad1:
Would it be useful to add a `Pooling` class that all pooling operators inherit from.
```
class Pooling(torch.nn.Module):
    def forward(self, x, edge_index, *args, **kwargs):
          mapping = self.select(x, edge_index, *args, **kwargs)
          x = self.reduce(x, mapping)
          edge_index = self.connect(mapping, edge_index, *args, **kwargs)
```
`self.select`, `reduce` and `connect` would be objects defined in the child class like `graclus` or `TopK`.

rusty1s:
Yes, definitely :)

danielegrattarola:
Hey, first author of the paper here!

This is super, if help is still wanted (as I see on the tags) I'd be happy to participate! 

[...]

PR #7308, opened 2023-05-06
Refactor `TopkPooling` into `SelectTopK`

wsad1:
Towards #6455.
Do not review before #7307 is merged.


rusty1s:
Hi @wsad1, I modified this PR quite a bit. Hope the changes work for you:
* I moved attention score computation to `SelectTopK` - I think this belongs there.
* Added tests for `SelectTopK`
* I reverted the addition of `TopKConnect` - Can we add that separately with corresponding test cases? I would also be in favor of renaming that to something more general, e.g., `FilterGraph`? The name should make it clear that we only maintain the nodes present in `select_output.node_index`. It may also require changes since it should respect the assignment matrix coming from `SelectOutput`
* Can you add documentation of `SelectTopK` in a follow-up?

Issue #7226, opened 2023-04-24
add_self_loops() tracing constant with torch.jit.trace()


Vuenc:
### 🐛 Describe the bug

When a model calls `add_self_loops` in its forward function and does not explicitly pass the `num_nodes` parameter, this leads to wrong behavior when tracing the model with `torch.jit.trace`: The number of nodes in the example input to `torch.jit.trace`  will be saved as a constant.

This may lead to wrong output when forwarding a bigger graph through the traced model (due to self loops not being added), and possibly an error being thrown when forwarding a smaller graph through the traced model (due to out-of-range self loops being added).

Here's example code, roughly along the lines of `test_message_passing.py`. I trace the model with `(x, edge_index)` and call the traced model with a smaller and bigger graph. When adding self loops is enabled, the traced model returns wrong outputs on the bigger graph and crashes on the smaller graph. This is because the number of nodes 4 is traced as a constant for the `add_self_loops` part.

```python
from torch import Tensor
from torch_geometric.nn import MessagePassing
import torch
import torch_geometric

class MyConv(MessagePassing):
    def __init__(self, use_self_loops: bool, aggr: str = 'add'):
        super().__init__(aggr=aggr)
        self.use_self_loops = use_self_loops

    def forward(self, x: Tensor, edge_index: torch.Tensor) -> Tensor:        
        if self.use_self_loops:
            edge_index, _ = torch_geometric.utils.add_self_loops(edge_index)

        # propagate_type: (x: OptPairTensor, edge_weight: OptTensor)
        return self.propagate(edge_index, x=x, size=None)
        
    def message(self, x_j: Tensor) -> Tensor:
        return x_j

def test_my_conv_jittable():
    x = torch.randn(4, 16)
    x_smaller = torch.randn(2, 16)
    x_bigger = torch.randn(8, 16)
    edge_index = torch.stack([torch.arange(0, 3), torch.arange(1, 4)])
    edge_index_bigger = torch.stack([torch.arange(0, 7), torch.arange(1, 8)])
    edge_index_smaller = torch.stack([torch.arange(0, 1), torch.arange(1, 2)])

    for conv, name in [(MyConv(use_self_loops=False), "MyConv (no self loops)"), (MyConv(use_self_loops=True), "MyConv (with self loops)")]:
        out = conv.forward(x, edge_index)
        traced_conv = torch.jit.trace(conv, ((x, edge_index)))
        out_traced = traced_conv.forward(x, edge_index)
[...]

PR #7330, opened 2023-05-09
Fixing `jit.trace` tracing a constant number of nodes with `add_self_loops`

Vuenc:
Fix of issue #7226.

The problem I described came from the `maybe_num_nodes` function which computed `int(edge_index.max()) + 1`. The `int()` call made the computed number of nodes a constant for the `torch.jit.trace` function.

I fixed it with an if branch that is only executed while tracing. (I could not get the same code for all cases to work: the workaround used only for tracing now would also work for normal operation, but then would break when using `torch.jit.script`).

### DIFF
--- a/test/nn/aggr/test_basic.py
+++ b/test/nn/aggr/test_basic.py
@@ -74,6 +74,17 @@ def test_var_aggregation():
     assert torch.allclose(out, expected, atol=1e-6)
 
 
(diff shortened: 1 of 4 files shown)


rusty1s (approved):
Thank you! I somehow missed the issue, so thanks for fixing it :) Cool solution.

PR #7354, opened 2023-05-12
Exclude `num_nodes` and `num_edges` from keys lists returned from `node_types` and `node_edges`

dsciebu:
Keys `num_nodes` and `num_edges` are special fields storing number of nodes/edges explicitly. They should not be included when listing node/edge_types and node/edge_stores as the user expects to get the nodes/edges, not the extra fields used internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Exclude special fields `num_nodes` and `num_edges` from listing nodes/edges stores [#7354](https://github.com/pyg-team/pytorch_geometric/pull/7354)
 - Added an example for hierarichial sampling ([#7244](https://github.com/pyg-team/pytorch_geometric/pull/7244))
 - Added Kùzu remote backend examples ([#7298](https://github.com/pyg-team/pytorch_geometric/pull/7298))
 - Fixed tracing of `add_self_loops` for a dynamic number of nodes ([#7330](https://github.com/pyg-team/pytorch_geometric/pull/7330))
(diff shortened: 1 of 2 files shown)


rusty1s:
Can you clarify for me why we need this? When is `num_nodes` a `NodeType`/`EdgeType`?

rusty1s:
Yes, `num_nodes` is a special property in a `Store`, but it should never be saved in `node_store_dict` or `edge_store_dict` - that's why I am confused.

dsciebu:
> Yes, `num_nodes` is a special property in a `Store`, but it should never be saved in `node_store_dict` or `edge_store_dict` - that's why I am confused.

Ok, so why is it read and returned in the `num_nodes` function I pasted?

rusty1s:
If the user specifies it via `store.num_nodes = ...`, we return the user-specified value. Otherwise, we infer it from other attributes, e.g., `x.size(0)` if present.

dsciebu:
> If the user specifies it via `store.num_nodes = ...`, we return the user-specified value. Otherwise, we infer it from other attributes, e.g., `x.size(0)` if present.

Awesome, so it gets added to the storage and exists there like it was a regular node type. If, after doing that, the user lists all node types, using the function `node_types`, all the fields, including `num_nodes`, is returned, which is error prone. That's why I state, both `num_nodes` and `num_edges` should be treated as special keys and excluded from listing them to users in functions `node/edge_types`.

rusty1s:
I am still not sure I understand. It gets added as a attribute inside a storage, but it is not a storage (which would be needed to get it listed in `node_types`/`node_stores`).

dsciebu:
Ok, FYI - I tried to make use of a dirty hack, namely set up `num_nodes` in storage directly, so when it's read in storage.py it gets my injected value. All this was needed because I didn't know how to override the `num_nodes` property getter, but I finally found the solution for this, so this particular PR can get closed.

PR #7377, opened 2023-05-17
Update GraphGPS example

kasper-piskorski:
Addressing training problems of the GraphGPS example. It is now more aligned with the reference implementation by:
- having a LR scheduler,
- concatenating initial node features and PEs
- having an extra post-GPS MLP layer.
I plotted loss and test-MAE for the example before and after changes together with results from PNA example (same dataset) for reference.
loss
mae_gps
mae_updated


examples/graph_gps.py
```diff
@@ -37,24 +46,29 @@ def __init__(self, channels: int, num_layers: int):
             conv = GPSConv(channels, GINEConv(nn), heads=4, attn_dropout=0.5)
             self.convs.append(conv)
 
-        self.lin = Linear(channels, 1)
+        self.mlp = Sequential(Linear(channels, channels // 2), ReLU(),
+                              Linear(channels // 2, channels // 4), ReLU(),
+                              Linear(channels // 4, 1))
```
rusty1s, line 51:
```suggestion
                              Linear(channels // 4, 1), )
```

rusty1s (approved):
Thank you!

PR #7398, opened 2023-05-19
Add MovieLens-100K heterogeneous dataset

ThomasKLY:


torch_geometric/datasets/movie_lens_100k.py
```diff
@@ -0,0 +1,133 @@
...
+            every access. (default: :obj:`None`)
+        pre_transform (callable, optional): A function/transform that takes in
+            an :obj:`torch_geometric.data.HeteroData` object and returns a
+            transformed version. The data object will be transformed before
+            being saved to disk. (default: :obj:`None`)
+    """
+
+    url = 'https://files.grouplens.org/datasets/movielens/ml-100k.zip'
+
+    def __init__(self, root, split="train",
```
EdisonLeeeee, line 46:
```suggestion
    def __init__(self, root: str, split: str = "train",

```

torch_geometric/datasets/movie_lens_100k.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Process user data
+        df = pd.read_csv(
+            self.raw_paths[1], sep="|", header=None,
+            names=["userId", "age", "gender", "occupation",
+                   "zipCode"], index_col='userId', encoding="ISO-8859-1")
+        user_mapping = {idx: i for i, idx in enumerate(df.index)}
+
+        max_age = df["age"].values.max()
+        age = df["age"].values / max_age
```
EdisonLeeeee, line 96: Why do we need to normalize `age`? Isn't it a category feature?

torch_geometric/datasets/movie_lens_100k.py
```diff
@@ -0,0 +1,133 @@
...
+                path, sep="\t", header=None,
+                names=['userId', 'movieId', 'rating', 'timestamp'])
+
+            src = [user_mapping[idx] for idx in df['userId']]
+            dst = [movie_mapping[idx] for idx in df['movieId']]
+            edge_index = torch.tensor([src, dst])
+
+            rating = torch.from_numpy(df['rating'].values).to(torch.long)
+            data['user', 'rates', 'movie'].edge_index = edge_index
+            data['user', 'rates', 'movie'].edge_label = rating
```
EdisonLeeeee, line 125: Should we also include `timestamp` as an additional edge feature, e.g., `edge_time`?

EdisonLeeeee (approved):
Thank you for the PR. Left some minor comments.

torch_geometric/datasets/movie_lens_100k.py
ThomasKLY: ML-100K's `age` column is numeric, whereas ML-1M's `age` column is categorical. Initially, I considered binning it for consistency, but ultimately I opted for the normalization approach used in the [reference design](https://github.com/muhanzhang/IGMC/blob/5fa0e3cb1bdc46ffcb97fa6e0249a553e2e0b438/preprocessing.py#L516).

rusty1s (approved):

Issue #7420, opened 2023-05-23
Add support to create approximate nearest neighbour graph using node features


sky-2002:
### 🚀 The feature, motivation and pitch

PyG currently supports creating knn-graphs that use node features and perform knn algorithm on them to create a knn-graph. Earlier there was [this](https://pytorch-geometric.readthedocs.io/en/1.3.0/_modules/torch_cluster/knn.html) code which made use of [`scipy.spatial.cKDTree`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html). 

I had used this code in one of my projects where I had around 26000 nodes with feature dimension 501. This code worked very slow and I had to switch to approx nearest neighbours. In some use cases, it suffices to have approximate nearest neighbour graphs. I had used [`pynndescent`](https://pynndescent.readthedocs.io/en/latest/) for this which gave good and speedy results.

So it will be useful to have approximate nearest neighbour graph in PyG.
As per discussion with @rusty1s on slack, I am creating a PR to add approx-nn support.

### Alternatives

I haven't used these, but there are some alternatives in approx nn domain, like annoy and annlite.

PR #7421, opened 2023-05-23
Add support for approximate k-NN

sky-2002:
As mentioned in [this](https://github.com/pyg-team/pytorch_geometric/issues/7420) issue, I am adding code to support approximate nearest neighbour graphs.

### DIFF
--- a/torch_geometric/nn/pool/approx_knn.py
+++ b/torch_geometric/nn/pool/approx_knn.py
@@ -2,10 +2,11 @@
 # nearest neighbors. We can call it approx-knn.
 
 import torch
-from pynndescent import NNDescent
 
 
 def approx_nn(x, y, k, batch_x=None, batch_y=None):
+    from pynndescent import NNDescent
+
     if batch_x is None:
         batch_x = x.new_zeros(x.size(0), dtype=torch.long)
 
--- a/torch_geometric/nn/pool/approx_knn.py
+++ b/torch_geometric/nn/pool/approx_knn.py
@@ -34,8 +34,7 @@ def approx_nn(x: Tensor, y: Tensor, k: int, batch_x: Tensor = None,
     x = torch.cat([x, 2 * x.size(1) * batch_x.view(-1, 1).to(x.dtype)], dim=-1)
     y = torch.cat([y, 2 * y.size(1) * batch_y.view(-1, 1).to(y.dtype)], dim=-1)
 
-    # Default metric is euclidean
-    # More can be found in the
+    # More can be found about pynndescent in the
     # documentation here - https://pynndescent.readthedocs.io/en/latest/
     index = NNDescent(x.detach().numpy())
     col, dist = index.query(y.detach().cpu(), k=k)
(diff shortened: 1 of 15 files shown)


torch_geometric/nn/pool/approx_knn.py
```diff
@@ -0,0 +1,55 @@
+# This file has code to create a KNN graph with approximate
+# nearest neighbors. We can call it approx-knn.
+
+import torch
+from pynndescent import NNDescent
```
rusty1s, line 5: This likely needs to be made an optional dependency. You can achieve this by moving the import into the respective function.

torch_geometric/nn/pool/approx_knn.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def approx_nn(x, y, k, batch_x=None, batch_y=None):
```
rusty1s, line 8: Can we add type hints here?

torch_geometric/nn/pool/approx_knn.py
```diff
@@ -0,0 +1,55 @@
...
+    x, y, = x / max_xy, y / max_xy
+
+    # Concat batch/features to ensure no cross-links between examples exist.
+    x = torch.cat([x, 2 * x.size(1) * batch_x.view(-1, 1).to(x.dtype)], dim=-1)
+    y = torch.cat([y, 2 * y.size(1) * batch_y.view(-1, 1).to(y.dtype)], dim=-1)
+
+    # Default metric is euclidean
+    # More can be found in the
+    # documentation here - https://pynndescent.readthedocs.io/en/latest/
+    index = NNDescent(x.detach().numpy())
```
rusty1s, line 37: Missing `cpu()` call?

torch_geometric/nn/pool/approx_knn.py
```diff
@@ -0,0 +1,55 @@
...
+    max_xy = max(x.max().item(), y.max().item())
+    x, y, = x / max_xy, y / max_xy
+
+    # Concat batch/features to ensure no cross-links between examples exist.
+    x = torch.cat([x, 2 * x.size(1) * batch_x.view(-1, 1).to(x.dtype)], dim=-1)
+    y = torch.cat([y, 2 * y.size(1) * batch_y.view(-1, 1).to(y.dtype)], dim=-1)
+
+    # Default metric is euclidean
+    # More can be found in the
+    # documentation here - https://pynndescent.readthedocs.io/en/latest/
```
rusty1s, line 36: I think this scheme here only works for Euclidean due to how we integrate `batch` into `x`/`y`.

rusty1s (commented):
Thanks for this PR. Added some comments. The next steps would be
* Add a line to `CHANGELOG.md` that describes the feature you add here
* Add a corresponding test in `test/nn/pool/test_approx_knn.py`. This can be similar to https://github.com/rusty1s/pytorch_cluster/blob/master/test/test_knn.py#L14-L48. Note that you likely need to add the `@withPackage('pynndescent')` decorator to let CI pass.

torch_geometric/nn/pool/approx_knn.py
sky-2002: Done

torch_geometric/nn/pool/approx_knn.py
sky-2002: ok, can you elaborate?

rusty1s (approved):

PR #7430, opened 2023-05-24
`HeteroDictLinear` heuristic

puririshi98:
```
Loop Times: {4: 0.00011243343353271484, 8: 0.00021158695220947266, 16: 0.0007293605804443359, 20: 0.0013348865509033203, 21: 0.0015002059936523437, 22: 0.001708526611328125, 23: 0.0018759918212890624, 24: 0.0020225000381469726, 32: 0.003490099906921387, 40: 0.004931988716125488, 45: 0.005780115127563477, 50: 0.006725430488586426, 55: 0.007553043365478515, 60: 0.008383927345275878, 64: 0.009244050979614258, 128: 0.01996610164642334, 256: 0.04032662868499756, 512: 0.0806538200378418, 1024: 0.16130864143371582}
Dict Times: {4: 0.00010602951049804688, 8: 0.00021188735961914063, 16: 0.0007947540283203125, 20: 0.0012229681015014648, 21: 0.0012968969345092773, 22: 0.0014571046829223633, 23: 0.0015320730209350585, 24: 0.0016141510009765624, 32: 0.002482771873474121, 40: 0.0033274221420288084, 45: 0.0039056396484375, 50: 0.004541130065917969, 55: 0.005016393661499023, 60: 0.0055756521224975585, 64: 0.006009225845336914, 128: 0.012227153778076172, 256: 0.024424519538879395, 512: 0.048862857818603514, 1024: 0.09928223609924317}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Set `output_size` in the `repeat_interleave` operation in `QuantileAggregation` ([#7426](https://github.com/pyg-team/pytorch_geometric/pull/7426))
 - Fixed gradient computation of edge weights in `utils.spmm` ([#7428](https://github.com/pyg-team/pytorch_geometric/pull/7428))
 - Re-factored `ClusterLoader` to integrate `pyg-lib` METIS routine ([#7416](https://github.com/pyg-team/pytorch_geometric/pull/7416))
 - Fixed an index-out-of-range bug in `QuantileAggregation` when `dim_size` is passed ([#7407](https://github.com/pyg-team/pytorch_geometric/pull/7407))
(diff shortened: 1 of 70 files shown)


rusty1s:
Thanks @puririshi98. Why do we need a different heuristic here? Internally, both `segment_matmul` and `grouped_matmul` call the same kernel, right?

puririshi98:
> Thanks @puririshi98. Why do we need a different heuristic here? Internally, both `segment_matmul` and `grouped_matmul` call the same kernel, right?

the difference is that the shape of each tensor can differ in grouped matmul where as segment matmul has the sizes constant. so for segmatmul i can just check the dimensions of the tensor quickly, but for gmm i have to iterate over the list of tensors and check the shapes(since for example some could be large and some could be small). this iteration as part of the hueristic makes gmm too slow since this hueristic adds too much extra time. the <10  type hueristic works great and is lightweight(adds no extra runtime)

rusty1s (approved):

PR #7442, opened 2023-05-26
Add `GDELTLite` dataset

akihironitta:


rusty1s (approved):
Thank you @akihironitta, looks wonderful. I added the reduced dataset version to `data.pyg.org`, which should make downloading and processing much more lightweight.

PR #7451, opened 2023-05-27
Add `LocalGraphStore` for distributed training

ZhengHongming888:

This class extends from GraphStore and use the store dict to save the graph topology and also use edge_ids dict to save the global eids. 

We also include the unit test under /test folder to show how the graph edge_index/ edge_ids will be saving/getting and verified. 


rusty1s (approved):

ZhengHongming888:
Thanks Matthias!

PR #7452, opened 2023-05-27
Add `LocalFeatureStore` for distributed training

ZhengHongming888:

This class extends from FeatureStore and use the store dict to save the node/edge features and also use global_idx & id2index dict to save the global node ids and the id mapping between node ids and local feature index, these of which will be used by sampling and feature lookup stage.

We also include the unit test under /test/distributed folder to show how the node/edge feature will be saving/getting and verified by two kinds of method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added hierarichial heterogeneous GraphSAGE example on OGB-MAG ([#7425](https://github.com/pyg-team/pytorch_geometric/pull/7425))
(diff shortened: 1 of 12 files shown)


ZhengHongming888:
@rusty1s anything i need change here? :-) Thanks.

rusty1s (approved):
@ZhengHongming888 I made some changes, PTAL. Especially, I introduced `get_tensor_from_global_id` such that `global_id_to_index` is solely a private attribute. I think this makes the logic a bit nicer.

Also left a TODO such that we could potentially compute the mapping much more memory-friendly, e.g., via `utils.map_index` functionality from PyG. PTAL.

ZhengHongming888:
Thanks @rusty1s yes it is very clear for me on these correction ! Thanks.

PR #7458, opened 2023-05-29
Implementation of Directed Graph Neural Networks (Dir-GNN)

emalgorithm:
Implementation of Directed Graph Neural Networks (Dir-GNN), as introduced in the paper "[Edge Directionality Improves Learning on Heterophilic Graphs](https://arxiv.org/abs/2305.10498)". 

We start by implementing DirSageConv.

We note the following regarding the implementation:

- It's not trivial to implement a generic DirGNN module which just takes a conv as input, and deep copies it to have convolutions in both directions. This is because if we take eg. SageConv as base conv, then we would have the root weights twice, whereas we need them only once. Or for GCN, we actually need to change the normalization to account for both in- and out-degree, and cannot just use two standard GCNConv layers.
- As it is implemented now, DirSageConv uses two SageConv layers with different flows. However, the flow direction "target_to_source" does not support message propagation via torch_sparse.SparseTensor or torch.sparse.Tensor . Should we instead have a generic utility function which is able to transpose any Adj  object, and always use SageConv with source_to_target flow, and feed once the original Adj, and the other time the tranpose?


emalgorithm:
Thanks @rusty1s, I agree with your point. I implemented a general DirGNN class. In the example I tried it with both GCN and SAGE internal convolution and it works well.

If we generalize the gcn_norm function to handle directed graphs properly, then we should be all set.

rusty1s (approved):
This is pretty cool :)

PR #7459, opened 2023-05-29
Add `LinkEncoder` of GraphMixer

akihironitta:
This PR introduces the "link-encoder" defined in 3.1 of the GraphMixer paper, [Do We Really Need Complicated Model Architectures For Temporal Networks?](https://openreview.net/forum?id=ayPPc0SyLv1).


test/nn/test_encoding.py
```diff
@@ -20,3 +24,31 @@ def test_temporal_encoding(device):
...
+                            f'num_edge_features={num_edge_features}, '
+                            f'hidden_channels={hidden_channels}, '
+                            f'out_channels={out_channels}, '
+                            f'time_channels={time_channels})')
+
+    num_nodes = 17
+    num_edges = num_nodes * num_edges_per_node
+    edge_attr = torch.rand((num_edges, num_edge_features), device=device)
+    edge_time = torch.rand((num_edges, ), device=device)
+    assert encoder(edge_attr, edge_time).size() == (num_nodes, out_channels)
```
weihua916, line 54: Quick question; how do we pass the edge connectivity information? Do we assume the padding is already done, so `edge_attr[0:K-1]` is for the first node, `edge_attr[K:2K-1]` is for the second node, and so on? Instead of assuming this, I think it'd be nice if `LinkEncoder` gets the following three arguments
`edge_attr`, `edge_time`, `node_batch`,  `num_nodes`, where `node_batch` indicates the node index each edge belongs to, and `num_nodes` indicates the number of nodes in this batch (can be optional, we can infer that via `node_batch.max()+1`)

so i-th edge has `edge_attr[i]` feature with `edge_time[i]` time and connected to `node_batch[i]` node.

weihua916 (commented):
Thanks for working on this! Quick initial comment.

test/nn/test_encoding.py
weihua916: You may find https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/utils/to_dense_batch.html to be useful.

weihua916: Something like
```python
your_edge_time, _ = to_dense_batch(edge_time, node_batch, max_num_nodes=K)
your_edge_attr, _ = to_dense_batch(edge_attr, node_batch, max_num_nodes=K)
```

akihironitta: @weihua916 Thank you for providing me with the reference! Please let me know if 5468010 has addressed your comment.

test/nn/test_encoding.py
```diff
@@ -20,3 +24,35 @@ def test_temporal_encoding(device):
...
+
+@withCUDA
+def test_link_encoding(device):
+    num_nodes = 3
+    num_edges = 6
+    num_edge_features = 10
+    edge_attr = torch.rand((num_edges, num_edge_features), device=device)
+    node_batch = torch.randint(low=0, high=num_nodes, size=(num_edges, ),
+                               device=device)
+    edge_time = torch.rand(num_edges, device=device)
```
weihua916, line 37: This is just for reference but according to the paper, the magnitude of `edge_time` is in the order of millions. In the testing, you are sampling from normal distribution. Might be good to align that in the testing code, or at least leave some comments.
Screenshot 2023-06-02 at 6 47 14 PM

test/nn/test_encoding.py
weihua916: This part looks great. Thank you. Will continue reviewing the rest next week.

wsad1: Nice use of `to_dense_batch`. Why do we need `node_batch` at all? Does't `edge_index[1]` give you the `dst` node for each edge?, can't we just use that for `to_dense_batch`?

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +97,178 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+    Args:
+        K (int): The number of most recent teomporal links to use to construct
+            an intermediate feature representation for each node.
+        num_edge_features (int): The number of edge features used to construct
+            a linear layer encoding the output of :class:`TemporalEncoding`
+            concatenated with ``edge_attr``.
+        time_channels (int): dims to encode each timestamp into with
+            :class:`TemporalEncoding`.
+        out_channels (int): Size of each output sample.
```
wsad1, line 189:
```suggestion
        in_channels (int): The number of edge features used to construct
            a linear layer encoding the output of :class:`TemporalEncoding`
            concatenated with ``edge_attr``.
        time_channels (int): dims to encode each timestamp into with
            :class:`TemporalEncoding`.
        out_channels (int): Size of each output sample.
```

test/nn/test_encoding.py
weihua916: Yes, we can assume `edge_index` to be given, and replace `node_batch` with `edge_index[1]`. I think that makes it look more like GNNs :)

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +97,178 @@ def forward(self, x: Tensor) -> Tensor:
...
+            in_channels=hidden_channels,
+            out_channels=out_channels,
+        )
+
+    def forward(
+        self,
+        edge_attr: Tensor,
+        edge_time: Tensor,
+        node_batch: Tensor,
+        num_nodes: Optional[int] = None,
```
wsad1, line 237: I don't think we need these two arguments, see above comment.
But we should add an argument `is_sorted` to indicate if the `edge_index` is sorted by `dst` node.

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+        # zero-pad each node's edges:
+        # [num_edges, hidden_channels] -> [num_nodes*K, hidden_channels]
+        edge_attr_time, _ = to_dense_batch(
+            edge_attr_time,
+            edge_index[1],
+            max_num_nodes=self.K,
+        )
+        return self.mlp_mixer(
+            edge_attr_time.view(-1, self.K, self.hidden_channels))
+
```
wsad1, line 270: If the `edge_index` did get sorted then they need to be put back in their original order.
```suggestion
        out= self.mlp_mixer(
            edge_attr_time.view(-1, self.K, self.hidden_channels))
        if indices is not None:
            out_unsorted = torch.empty_like(out_attr)
            out_unsorted[indices] = out
            out = out_unsorted
       return out

```

akihironitta: @wsad1 Thank you for actively reviewing my PR!

One thing that's not obvious is that the input and output of `forward()` don't share the same shape in any dim because edges of each node are squashed into a single vector:
```
[num_edges, in_channels]   # edge_attr
[num_nodes, out_channels]  # out
```
and I'm not sure what the unsorted tensor would look like.

One reasonable way I can think of is to keep the output sorted and just document that the output tensor is always sorted by the node index.

What do you think?

wsad1: Apologies at @akihironitta , you don't need to make any changes. I forgot that the output if shape `num_nodes, out_channels`.

akihironitta:
Thank you for all your fruitful comments! Please let me know if there are any further changes needed. I will complete an implementation of `GraphMixer` on top of this component once this PR gets merged.

test/nn/test_encoding.py
akihironitta: This comment should be addressed now. PTAL!

test/nn/test_encoding.py
akihironitta: Do you mean we should use a large tensor within this unit test?

weihua916: I think either way is fine in terms of unit-test, but I think it's better to leave some note so that end-users are aware of the scale of the timestamps.

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+    def forward(self, x: Tensor) -> Tensor:
+        """
+        Args:
+            x (torch.Tensor): Tensor of size ``[N, num_tokens, in_channels]``
+
+        Returns:
+            Tensor of size ``[N, out_channels]``
+        """
+        # token mixing
+        h = self.token_layer_norm(x).mT
```
weihua916, line 140: what is mT?

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+        h = self.token_lin_1(h)
+        h = torch.nn.functional.gelu(h)
+        h = torch.nn.functional.dropout(h, p=0.5, training=self.training)
+        h = self.token_lin_2(h)
+        h = torch.nn.functional.dropout(h, p=0.5, training=self.training)
```
weihua916, line 145: Why apply dropout? why 0.5?

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+        h_token = h.mT + x
+
+        # channel mixing
+        h = self.channel_layer_norm(h_token)
+        h = self.channel_lin_1(h)
+        h = torch.nn.functional.gelu(h)
+        h = torch.nn.functional.dropout(h, p=0.5, training=self.training)
+        h = self.channel_lin_2(h)
+        h = torch.nn.functional.dropout(h, p=0.5, training=self.training)
```
weihua916, line 154: Why dropout? Why 0.5?

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+        h_channel = h + h_token
+
+        # head
+        h_channel = self.head_layer_norm(h_channel)
+        t = torch.mean(h_channel, dim=1)
+        return self.head_lin(t)
```
weihua916, line 160: nit: Are `self.head_layer_norm` and `self.head_in` needed? They are not mentioned in the paper.

weihua916 (commented):
Looks great! thank you. Left some comments/questions.

torch_geometric/nn/encoding.py
akihironitta: It's a transpose op over the last two dims:
```python
x    # [N, num_tokens, in_channels]
x.mT # [N, in_channels, num_tokens]
```
https://pytorch.org/docs/2.0/tensors.html#torch.Tensor.mT

torch_geometric/nn/encoding.py
akihironitta: I originally didn't have these dropout layers as I precisely followed the paper, but when I finished implementing this on my own and looked at the official impl. for validation, it turned out that they have these dropout layers with 0.5 as their default value: https://github.com/CongWeilin/GraphMixer/blob/bbdd9923706a02d0619dfd00ef7880911f003a65/model.py#L179

I can parameterize the `0.5` if preferred, but should I remove these dropout layers completely?

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+        self.out_channels = out_channels
+        self.time_channels = time_channels
+        self.is_sorted = is_sorted
+
+        # teomporal encoder
+        self.temporal_encoder = TemporalEncoding(time_channels)
+        self.temporal_encoder_head = torch.nn.Linear(
+            time_channels + in_channels,
+            hidden_channels,
+        )
```
akihironitta, line 228: @weihua916 Just FYI, this linear layer wasn't mentioned in the paper AFAIK, but I added it to match this implementation with their official code: https://github.com/CongWeilin/GraphMixer/blob/bbdd9923706a02d0619dfd00ef7880911f003a65/model.py#LL170C25-L170C25

torch_geometric/nn/encoding.py
akihironitta: I didn't have this originally, but I added this to match my implementation with theirs:  https://github.com/CongWeilin/GraphMixer/blob/bbdd9923706a02d0619dfd00ef7880911f003a65/model.py#L239

torch_geometric/nn/encoding.py
akihironitta: Commented here: https://github.com/pyg-team/pytorch_geometric/pull/7459/files#r1220524099

torch_geometric/nn/encoding.py
weihua916: I see. Thanks!

torch_geometric/nn/encoding.py
weihua916: Got it, thanks for being careful! I think we can parameterize.

torch_geometric/nn/encoding.py
weihua916: Got it. thanks for clarifying. Then let's leave it.

torch_geometric/nn/encoding.py
weihua916: Makes sense. Thanks!

test/nn/test_encoding.py
akihironitta: Documented in the `LinkEncoding` module!

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +97,193 @@ def forward(self, x: Tensor) -> Tensor:
...
+            edge_index[1], indices = index_sort(edge_index[1])
+            edge_attr_time = edge_attr_time[indices]
+
+        # zero-pad each node's edges:
+        # [num_edges, hidden_channels] -> [num_nodes*K, hidden_channels]
+        edge_attr_time, _ = to_dense_batch(
+            edge_attr_time,
+            edge_index[1],
+            max_num_nodes=self.K,
+            batch_size=num_nodes,
```
wsad1, line 278: It looks like we are choosing `self.K` links, but it doesn't look like they are the latest.

torch_geometric/nn/encoding.py
```diff
@@ -2,6 +2,9 @@
 
 import torch
 from torch import Tensor
+from troch_geometric.utils.num_nodes import maybe_num_nodes
+
+from torch_geometric.utils import index_sort, to_dense_batch
```
wsad1, line 7:
```suggestion
from torch_geometric.utils import index_sort, to_dense_batch
```

wsad1, line 7:
```suggestion

from torch_geometric.utils.num_nodes import maybe_num_nodes
from torch_geometric.utils import index_sort, to_dense_batch
```

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +98,193 @@ def forward(self, x: Tensor) -> Tensor:
...
+class LinkEncoding(torch.nn.Module):
+    r"""The link-encoding function from the `"Do We Really Need Complicated
+    Model Architectures for Temporal Networks?"
+    <https://openreview.net/forum?id=ayPPc0SyLv1>`_ paper.
+    :class:`LinkEncoding` is composed of two components. The first component is
+    :class:`TemporalEncoding` that maps each edge timestamp to a
+    :args:`time_channels` dimensional vector.
+    The second component a 1-layer MLP that maps each encoded timestamp
+    feature concatenated with its corresponding link feature to a
+    :args:`out_channels` dimensional vector.
```
wsad1, line 192:
```suggestion
    :obj:`time_channels` dimensional vector.
    The second component a 1-layer MLP that maps each encoded timestamp
    feature concatenated with its corresponding link feature to a
    :obj:`out_channels` dimensional vector.
```

torch_geometric/nn/encoding.py
akihironitta: I'm not sure what you mean by "latest". The parameter `K` is a fixed parameter that represents the max number of edges each node can have. It is not dependent on each mini-batch size.

torch_geometric/nn/encoding.py
akihironitta: Done!

test/nn/test_encoding.py
```diff
@@ -20,3 +24,35 @@ def test_temporal_encoding(device):
...
+        time_channels=time_channels,
+    ).to(device)
+    assert str(encoder) == (f'LinkEncoding(K={K}, '
+                            f'in_channels={num_edge_features}, '
+                            f'hidden_channels={hidden_channels}, '
+                            f'out_channels={out_channels}, '
+                            f'time_channels={time_channels})')
+
+    out = encoder(edge_attr, edge_time, edge_index)
+    assert out.size() == (num_nodes, out_channels)
```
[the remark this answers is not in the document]
akihironitta: I'm still unsure when that would be a case, but I have added the `num_nodes` as an optional forward pass arg.

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+            edge_index[1], indices = index_sort(edge_index[1])
+            edge_attr_time = edge_attr_time[indices]
+
+        # zero-pad each node's edges:
+        # [num_edges, hidden_channels] -> [num_nodes*K, hidden_channels]
+        edge_attr_time, _ = to_dense_batch(
+            edge_attr_time,
+            edge_index[1],
+            max_num_nodes=self.K,
+        )
```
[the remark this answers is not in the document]
akihironitta: Done! https://github.com/pyg-team/pytorch_geometric/pull/7459#discussion_r1225793012

torch_geometric/nn/encoding.py
```diff
@@ -94,3 +96,182 @@ def forward(self, x: Tensor) -> Tensor:
...
+            in_channels=hidden_channels,
+            out_channels=out_channels,
+        )
+
+    def forward(
+        self,
+        edge_attr: Tensor,
+        edge_time: Tensor,
+        edge_index: Tensor,
+    ) -> Tensor:
```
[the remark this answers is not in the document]
akihironitta: Done! https://github.com/pyg-team/pytorch_geometric/pull/7459#discussion_r1225793012

test/nn/test_encoding.py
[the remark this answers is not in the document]
weihua916: Without passing `num_nodes`, `out.size(1)` will be `edge_index[1].max() + 1`. So If `edge_index[1]` does not contain last node in , the shape of `out` won't be `num_nodes.`

So I suggest 
```python
out = encoder(edge_attr, edge_time, edge_index, num_nodes)
```

torch_geometric/nn/encoding.py
weihua916: That's a good point. If a destination node has degree larger than `K`, we will need to include `K` more recent edges, i.e., largest `K` elements in `edge_time` for the destination node.

test/nn/test_encoding.py
weihua916: Thank you!

weihua916 (commented):
Thanks for the hard work. My final round of comments to address.

torch_geometric/nn/encoding.py
weihua916: Thanks!

weihua916: I personally prefer it to be set to 0., since it's a hyper-parameter and the default dropout value should be 0.

akihironitta: Done!

test/nn/test_encoding.py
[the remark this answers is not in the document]
akihironitta: > If edge_index[1] does not contain last node , the shape of out will be num_nodes-1, which is different from num_nodes.

That totally makes sense! Thank you for clarifying this :)

wsad1 (approved):
@akihironitta Thanks for the great work.
@weihua916 I updated the PR, the latest K edge attributes are used to create the node embedding now. Appreciate review and approval from you before we merge.

torch_geometric/nn/models/graph_mixer.py
```diff
@@ -47,3 +52,223 @@ def forward(
...
+                f"out_channels={self.out_channels}, "
+                f"dropout={self.dropout})")
+
+
+def get_latest_k_edge_attrs(K: int, edge_index: Tensor, edge_time: Tensor,
+                            edge_attr: Tensor, num_nodes: int) -> Tensor:
+    r"""Returns the latest :obj:`K` incoming edge attributes by
+    :obj:`edge_time` for each node. The shape
+    of the output tensor is :obj:`[num_nodes, K, edge_attr_dim]`.
+    Nodes with fewer than :obj:`K` incoming edges are zero-padded.
```
weihua916, line 142: Worth mentioning that we return the `edge_attr` sorted by `edge_time`.

weihua916 (approved):
LGTM! Thanks for the great work.

torch_geometric/nn/models/graph_mixer.py
```diff
@@ -47,3 +52,223 @@ def forward(
...
+        self.channel_lin_2 = torch.nn.Linear(4 * in_channels, in_channels)
+
+        # head
+        self.head_layer_norm = torch.nn.LayerNorm((in_channels, ))
+        self.head_lin = torch.nn.Linear(in_channels, out_channels)
+
+    def forward(self, x: Tensor) -> Tensor:
+        """
+        Args:
+            x (torch.Tensor): Tensor of size ``[N, num_tokens, in_channels]``
```
wsad1, line 97:
```suggestion
            x (torch.Tensor): Tensor of size :obj:`[N, num_tokens, in_channels]`
```

torch_geometric/nn/models/graph_mixer.py
```diff
@@ -47,3 +52,223 @@ def forward(
...
+
+        Returns:
+            Tensor of size ``[N, out_channels]``
```
wsad1, line 100:
```suggestion
            Tensor of size `[N, out_channels]`.
```

torch_geometric/nn/models/graph_mixer.py
```diff
@@ -47,3 +52,223 @@ def forward(
...
+        self.dropout = dropout
+
+        # teomporal encoder
+        self.temporal_encoder = TemporalEncoding(time_channels)
+        self.temporal_encoder_head = torch.nn.Linear(
+            time_channels + in_channels,
+            hidden_channels,
+        )
+
+        # temporal information summariser
```
wsad1, line 226:
```suggestion
        # MLP that summarises temporal embedding.
```

torch_geometric/nn/models/graph_mixer.py
```diff
@@ -47,3 +52,223 @@ def forward(
...
+        in_channels (int): Input channels.
+        out_channels (int): Output channels.
+        dropout (float, optional):
+    """
+    def __init__(
+        self,
+        num_tokens: int,
+        in_channels: int,
+        out_channels: int,
+        dropout: float,
```
wsad1, line 72:
```suggestion
        dropout (float, optional): The dropout probability. (default: :obj:`0`)
    """
    def __init__(
        self,
        num_tokens: int,
        in_channels: int,
        out_channels: int,
        dropout: float = 0
```

PR #7465, opened 2023-05-30
Add `Performer` to `GPSConv`

zechengz:
### Description

* Most code is adopted and simplified from https://github.com/lucidrains/performer-pytorch/blob/main/performer_pytorch/performer_pytorch.py and https://github.com/rampasek/GraphGPS/blob/main/graphgps/layer/gps_layer.py Should we add some comments to cite? Checked that both of them have the MIT License.
* There is a projection matrix redrawer that redraws the projection matrix in the performer. It seems to be not required so I just put it into the `examples/graph_gps.py`. Not sure where should I put it.
* There are also some other params and kernels to be added to the performer, such as softmax kernel etc. For now I added some TODOs.

### ZINC Example
```python
python3 examples/graph_gps.py --attn_type performer
```
I directly replace the multihead attention with performer fast attention with the `ReLU` kernel. The performance is similar (there are some randomness for each run and for the graph classification task) Following is the `test_mae` plot for each epoch.

training


torch_geometric/nn/conv/gps_conv.py
```diff
@@ -159,4 +181,142 @@ def forward(
...
+            q = generalized_kernel(q, self.projection_matrix, self.kernel)
+            k = generalized_kernel(k, self.projection_matrix, self.kernel)
+        else:
+            # TODO: Support kernel such as the softmax kernel
+            raise ValueError('Only the generalized kernels are supported')
+        out = linear_attention(q, k, v)
+        return out
+
+
+class PerformerAttention(torch.nn.Module):
```
wsad1, line 278: @zechengz , can `PerformerAttention` be used by any PyG model that needs multihead attention. If so can we move this to some common place like `torch_geometric/nn/models` or something?

wsad1, line 278: Add docs.

torch_geometric/nn/conv/gps_conv.py
```diff
@@ -55,32 +54,49 @@ class GPSConv(torch.nn.Module):
...
+        attn_kwargs (Dict[str, Any], optional): Arguments passed to the
+            attention layer. (default: :obj:`None`)
     """
     def __init__(
         self,
         channels: int,
         conv: Optional[MessagePassing],
         heads: int = 1,
         dropout: float = 0.0,
-        attn_dropout: float = 0.0,
```
wsad1, line 65: Why is this argument dropped? I think we should keep this and forward the argument to both `MultiheadAttention` and `PerformerAttention`.

examples/graph_gps.py
```diff
@@ -65,8 +78,35 @@ def forward(self, x, pe, edge_index, edge_attr, batch):
...
 
+class RedrawProjection(torch.nn.Module):
+    def __init__(self, model: torch.nn.Module,
+                 redraw_interval: Optional[int] = None):
+        super().__init__()
+        self.model = model
+        self.redraw_interval = redraw_interval
+        self.register_buffer('num_last_redraw', torch.tensor(0))
+
+    def redraw_projections(self):
```
wsad1, line 89: Looks like this is called before every batch. Why not just call it inside `PerformerAttention`?

wsad1 (commented):
Left some initial comments. Will take a look at it again.

examples/graph_gps.py
zechengz: Right, I just moved it into the `PerformerAttention`

torch_geometric/nn/conv/gps_conv.py
zechengz: It seems that moving to `torch_geometric/nn/models` will cause some circular import. Any suggestions?

torch_geometric/nn/conv/gps_conv.py
zechengz: IMO this dropout is specific to different attention mechanism, so it might be better to put this into the `attn_kwargs`. Also, it's possible that some other attention mechanism might not need `attn_dropout`. But drop this may break some current script if user is using the `GPSConv` with `attn_dropout`. WDYT?

torch_geometric/nn/conv/gps_conv.py
wsad1: Ah I see. Since we plan to add more attention modules. Lets create a folder called `torch/nn/attention`, and put all the attention modules there.

torch_geometric/nn/conv/gps_conv.py
```diff
@@ -159,4 +181,180 @@ def forward(
...
+            q = generalized_kernel(q, self.projection_matrix, self.kernel)
+            k = generalized_kernel(k, self.projection_matrix, self.kernel)
+        else:
+            # TODO: Support kernel such as the softmax kernel
+            raise ValueError('Only the generalized kernels are supported')
+        out = linear_attention(q, k, v)
+        return out
+
+
+class PerformerAttention(torch.nn.Module):
```
wsad1, line 282: Could you add separate unit tests for `PerformerAttention`

torch_geometric/nn/conv/gps_conv.py
```diff
@@ -159,4 +181,180 @@ def forward(
...
+    def __init__(self, num_cols: int, kernel: Optional[Callable] = None):
+        super().__init__()
+        num_rows = int(num_cols * math.log(num_cols))
+        self.num_rows = num_rows
+        self.num_cols = num_cols
+        # Generate an orthogonal projection matrix
+        # with the shape (num_rows, num_cols)
+        projection_matrix = orthogonal_matrix(self.num_rows, self.num_cols)
+        self.register_buffer('projection_matrix', projection_matrix)
+        self.kernel = kernel
```
wsad1, line 269:
```suggestion
        kernel (Callable): Kernels for generalized attention.
            (default: :obj:`torch.nn.ReLU`)
    """
    def __init__(self, num_cols: int, kernel: Callable = torch.nn.ReLU()):
        super().__init__()
        num_rows = int(num_cols * math.log(num_cols))
        self.num_rows = num_rows
        self.num_cols = num_cols
        # Generate an orthogonal projection matrix
        # with the shape (num_rows, num_cols)
        projection_matrix = orthogonal_matrix(self.num_rows, self.num_cols)
        self.register_buffer('projection_matrix', projection_matrix)
        assert kernel is not None
        self.kernel = kernel
```
Why not use the `ReLU` kernel by default for now.

torch_geometric/nn/conv/gps_conv.py
```diff
@@ -159,4 +181,180 @@ def forward(
...
+
+    def forward(self, q: Tensor, k: Tensor, v: Tensor) -> Tensor:
+        if self.kernel:
+            q = generalized_kernel(q, self.projection_matrix, self.kernel)
+            k = generalized_kernel(k, self.projection_matrix, self.kernel)
+        else:
+            # TODO: Support kernel such as the softmax kernel
+            raise ValueError('Only the generalized kernels are supported')
```
wsad1, line 277:
```suggestion
        q = generalized_kernel(q, self.projection_matrix, self.kernel)
        k = generalized_kernel(k, self.projection_matrix, self.kernel)
```

torch_geometric/nn/conv/gps_conv.py
```diff
@@ -159,4 +181,180 @@ def forward(
...
+    num_batches, num_heads, *_ = x.shape
+    # Expand projection matrix to number of batches and number of heads
+    projection = mat.expand(num_batches, num_heads, *mat.shape)
+    # "Inner" product x with projection matrix
+    x = torch.einsum('...id,...jd->...ij', x, projection)
+    out = kernel(x) + epsilon
+    return out
+
+
+class PerformerFastAttention(torch.nn.Module):
```
wsad1, line 249: Can we rename this . `PerformerFastAttention` seems like a faster version of `PerformerAttention` :).

wsad1 (commented):
@zechengz I am not sure how many attention models we plan to add. **But if there are potentially several out there and multiple pyg modules could use them**, we should define a `Attention` base class and have all the attention modules we add use the same interface, similar to `torch_geometric/nn/aggr`. With this the user can simply pass any attention module as an argument to the GNN modules that support attention. WDYT? cc: @rusty1s 

I don't want to gate this PR, with the above suggestion though. I've left some more comments and it looks good.

torch_geometric/nn/conv/gps_conv.py
zechengz: Ok

torch_geometric/nn/conv/gps_conv.py
zechengz: Sure

zechengz:
@wsad1 Current plan is to add this `performer` and I am also working on another one called `bigbird`. So, with normal multihead attention we at least have three different global attentions. I also observed that there are also relative a lot of attention methods other than `performer` and `bigbird` that can be useful to add to PyG. So creating the folder `torch/nn/attention` makes sense to me. Thanks!

examples/graph_gps.py
```diff
@@ -65,8 +79,33 @@ def forward(self, x, pe, edge_index, edge_attr, batch):
         return self.mlp(x)
 
 
+class RedrawProjection(torch.nn.Module):
+    def __init__(self, model: torch.nn.Module,
+                 redraw_interval: Optional[int] = None):
```
wsad1, line 84: Why allow `redraw_interval` to be `None`?

wsad1 (approved):
Thanks @zechengz !
Good to merge after the comments are addressed.

examples/graph_gps.py
zechengz: Here we set the `redraw_interval` to `None` when we use multihead attention in this example as we only do the redraw for performer.

torch_geometric/nn/attention/performer.py
```diff
@@ -0,0 +1,177 @@
...
+            lambda t: t.reshape(b, n, self.heads, self.head_channels).permute(
+                0, 2, 1, 3), (q, k, v))
+        if mask is not None:
+            mask = mask[:, None, :, None]
+            v.masked_fill_(~mask, 0.)
+        out = self.fast_attn(q, k, v)
+        out = out.permute(0, 2, 1, 3).reshape(b, n, -1)
+        out = self.attn_out(out)
+        out = self.dropout(out)
+        return out
```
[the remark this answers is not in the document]
zechengz: Thanks

torch_geometric/nn/attention/performer.py
```diff
@@ -0,0 +1,177 @@
...
+    projection = mat.expand(num_batches, num_heads, *mat.shape)
+    # "Inner" product x with projection matrix
+    x = torch.einsum('...id,...jd->...ij', x, projection)
+    out = kernel(x) + epsilon
+    return out
+
+
+class PerformerProjection(torch.nn.Module):
+    r"""The fast attention that uses a projection matrix
+    from the `"Rethinking Attention with Performers"
```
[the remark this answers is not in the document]
zechengz: Sure!

PR #7470, opened 2023-05-31
[Benchmark] Add compile option in benchmark kernel, citation and points

yanbing-j:
Citation uses `Data` as input, which can be extended in `profile/benchmark.py`.

Kernel and points use `DataLoader`. They only compile(model), does not use `profile/benchmark.py` structure.

### DIFF
--- a/benchmark/citation/train_eval.py
+++ b/benchmark/citation/train_eval.py
@@ -5,7 +5,8 @@ import torch.nn.functional as F
 from torch import tensor
 from torch.optim import Adam
 
-from torch_geometric.profile import timeit, torch_profile
+import torch_geometric
+from torch_geometric.profile import benchmark, timeit, torch_profile
 from torch_geometric.utils import index_to_mask
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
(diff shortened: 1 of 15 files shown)


benchmark/citation/train_eval.py
```diff
@@ -96,10 +97,21 @@ def run_train(dataset, model, runs, epochs, lr, weight_decay, early_stopping,
...
+    if use_compile:
+        compiled_model = torch_geometric.compile(model)
+        benchmark(
+            funcs=[model, compiled_model],
+            func_names=['Vanilla', 'Compiled'],
+            args=data,
+            num_steps=30 if device == torch.device('cpu') else 50,
+            num_warmups=5 if device == torch.device('cpu') else 10,
+            backward=True,
+        )
```
rusty1s, line 109: Not sure why we are using the `benchmark` function here. Wouldn't we first want to compile the model and then train it regularly?

yanbing-j: I have updated citation as kernel and points, not using `benchmark` anymore.

benchmark/citation/train_eval.py
```diff
@@ -96,10 +97,18 @@ def run_train(dataset, model, runs, epochs, lr, weight_decay, early_stopping,
         with torch_profile():
             train(model, optimizer, data)
 
+    if use_compile:
```
rusty1s, line 100: Thanks for the update. I am still not sure why we added it this way. I would expect that the `compile` flag converts the model to a compiled model at the start of this `train` function, rather than making a separate training run. I think I would prefer this, but if you have strong opjections we can also merge it in the current state.

yanbing-j: I understand your point. I was influenced by `benchmark.py` to compare eager mode and compiled mode in the same run.

I will remove this logic among `kernel`, `citation` and `points`.

rusty1s (approved):
Thank you :)

PR #7258, opened 2023-04-28
Heuristic for `segment_matmul`

puririshi98:
good to go

PR #7474, opened 2023-05-31
Add option to override the `segmm` heuristic decision in `HeteroLinear`

toenshoff:
We added an option to override the heuristic that decides whether segmm should be used in HeteroLinear ([#7258](https://github.com/pyg-team/pytorch_geometric/pull/7258)).

We also added the option to the constructor of HGTConv which is passed to the projections for keys and values, respectively.

We found that on one use case we are working on the heuristic incorrectly selects the for-loop implementation even though it runs around 3 times slower than the fused segmm kernel.
This happens with a standard 4-layer HGT model with dimension 128 and 4 heads when training on a heterogeneous graph with 16 node types, 26 edge types and around 1 Million vertices.  We can not share the dataset, but the user should be able to force a certain selection if the heuristic makes a sub-optimal choice on a given configuration.

Interestingly, we use an A40 GPU which should be very similar to the A100 that was used to fit the heuristic. It is not clear to us why a wrong selection is made. We were wondering if the training of the heuristic should also incorporate the time needed for a backwards pass through the layer as well as imbalanced node types?


rusty1s:
Thanks for the PR. I am wondering if it makes sense to control this through a global flag, or whether we want to propagate this to every single operator that uses `HeteroLinear`. I think both have their advantages and disadvantages :(

toenshoff:
> Thanks for the PR. I am wondering if it makes sense to control this through a global flag, or whether we want to propagate this to every single operator that uses `HeteroLinear`. I think both have their advantages and disadvantages :(

A global flag would probably be easier. Should it be a static attribute of the `HeteroLinear` class or is there a standard place for global flags in PyG? I also think it would be good if the constructor of `HeteroLinear` itself still accepts an optional binary flag directly.

rusty1s:
We don't have a standard place for this (yet). I suggest that we could simply add it to `nn/__init__.py`? I agree that `HeteroLinear` should accept this flag directly as well.

rusty1s (approved):
I removed the change to `HGTConv` for now. Feel free to bring it back via a global flag.

PR #7477, opened 2023-05-31
Making `RevGNN` example work w/o `torch-sparse`

puririshi98:


rusty1s (commented):
I think this works for me although I am a bit skeptical about the runtime. I assume this decreases quite a lot?

rusty1s:
Thanks for the update. Would it make sense to just use the `ToSparseTensor` transform in this case?

PR #7482, opened 2023-06-01
Add `LocalFeatureStore` and `LocalGraphStore` helper initializations

ZhengHongming888:

This class will initialize the graph & node/edge feature data based on LocalGraphStore/LocalFeatureStore data structure. You can based on these initialize API to define your graph and node/edge features for your homo/hetero datasets.

We also include the unit test under /test/distributed folder to show how one OGB_MAG dataset will be initialize into the graph and feature by LocalDataset APIs.


test/distributed/test_local_dataset.py
```diff
@@ -0,0 +1,61 @@
...
+import torch_geometric.transforms as T
+from torch_geometric.datasets import OGB_MAG
+from torch_geometric.distributed import LocalDataset
+
+
+def test_local_dataset():
+
+    path = osp.join(osp.dirname(osp.realpath(__file__)), './ogbn-mags')
+    transform = T.ToUndirected(merge=True)
+    dataset = OGB_MAG(path, preprocess='metapath2vec', transform=transform)
```
rusty1s, line 16: You cannot use `OGB_MAG` in a test. I suggest to use a small custom dataset, or use `FakeHeteroDataset` for this.

torch_geometric/distributed/local_dataset.py
```diff
@@ -0,0 +1,216 @@
...
+from torch_geometric.data import EdgeAttr, TensorAttr
+from torch_geometric.distributed.local_feature_store import \
+    LocalFeatureStore as Feature
+from torch_geometric.distributed.local_graph_store import \
+    LocalGraphStore as Graph
+from torch_geometric.typing import EdgeType, NodeType, TensorDataType
+from torch_geometric.utils import convert_to_tensor, squeeze
+
+
+class LocalDataset(object):
```
rusty1s, line 14: Can we increase code coverage in this file? For me, it would be fine to drop support for `edge_features` completely for now, especially since `FeatureStore` doesn't support them yet anyway.

test/distributed/test_local_dataset.py
ZhengHongming888: Sure. Will change it. thanks.

torch_geometric/distributed/local_dataset.py
ZhengHongming888: Will try. :-) Thanks.

rusty1s (approved):
Hey @ZhengHongming888. Thank you! I am personally not super sure why we need a `LocalDataset` wrapper to hold `GraphStore` and `FeatureStore` instances. I modified this PR quite a bit and instead introduced `FeatureStore.from_data` and `FeatureStore.from_hetero_data` (and respectively for `GraphStore`) functions. I personally think this makes for a cleaner design. If you desperately need `LocalDataset`, you should be able to implement their functions with these helpers.

PR #7494, opened 2023-06-02
Use `GraphSAGE` model in Reddit example

LukeLIN-web:


rusty1s (commented):
Thank you for the PR. Currently, the inference pass is approximately four times slower than the existing solution, so there seems to be something fishy going on. Need to understand the differences first before merging :(

examples/reddit.py
```diff
@@ -99,7 +55,8 @@ def train(epoch):
 @torch.no_grad()
 def test():
     model.eval()
-    y_hat = model.inference(data.x, subgraph_loader).argmax(dim=-1)
```
LukeLIN-web, line 102: @rusty1s  This is because the previous version set all data.x in GPU.
And current `GraphSAGE` class  will set all data.x in CPU, `x_all = loader.data.x.cpu() `
When I use `    y_hat = model.inference(data.x.cpu(), subgraph_loader).argmax(dim=-1)` in the previous version(6.96it/s), it is slower as the newer version (7.31it/s).

Kh4L: This should fix the slow inference #7548

wsad1 (approved):
Thanks @LukeLIN-web. 
Have you verified that the accuracy metric is same before and after the change?

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Use the pre-defined `GraphSAGE` model in the Reddit example ([#7494](https://github.com/pyg-team/pytorch_geometric/pull/7494))
```
wsad1, line 74:
```suggestion
```

LukeLIN-web:
> Thanks @LukeLIN-web. Have you verified that the accuracy metric is same before and after the change?

Yeah, the accuracy metric is same after the change. 
Screenshot 2023-07-08 at 9 53 21 PM

examples/reddit.py
```diff
@@ -21,54 +20,11 @@
...
-        pbar.close()
-        return x_all
-
-
-model = SAGE(dataset.num_features, 256, dataset.num_classes).to(device)
+subgraph_loader = NeighborLoader(data, input_nodes=None, num_neighbors=[-1],
+                                 shuffle=False, **kwargs)
+
+model = GraphSAGE(dataset.num_features, 256, num_layers=2,
+                  out_channels=dataset.num_classes).to(device)
```
rusty1s, line 27:
```suggestion
                  out_channels=dataset.num_classes, ).to(device)
```

examples/reddit.py
```diff
@@ -21,54 +20,11 @@
...
-        pbar = tqdm(total=len(subgraph_loader.dataset) * len(self.convs))
-        pbar.set_description('Evaluating')
-
-        # Compute representations of nodes layer by layer, using *all*
-        # available edges. This leads to faster computation in contrast to
-        # immediately computing the final representations of each batch:
-        for i, conv in enumerate(self.convs):
-            xs = []
-            for batch in subgraph_loader:
-                x = x_all[batch.n_id.to(x_all.device)].to(device)
```
rusty1s, line 60: I am still a bit confused about this. Here, in the previous example, we explicitly convert intermediate embeddings to CPU, while we are now storing them on GPU.

examples/reddit.py
```diff
@@ -21,54 +20,15 @@
...
-        return x_all
-
-
-model = SAGE(dataset.num_features, 256, dataset.num_classes).to(device)
+subgraph_loader = NeighborLoader(data, input_nodes=None, num_neighbors=[-1],
+                                 shuffle=False, **kwargs)
+
+model = GraphSAGE(
+    dataset.num_features,
+    256,
```
rusty1s, line 28:
```suggestion
    in_channels=dataset.num_features,
    hidden_channels=256,
```

examples/reddit.py
LukeLIN-web: Do you mean it may not be as fast as the previous example if we store intermediate embeddings to the CPU? Or do we need to maintain the semantics?

rusty1s: Sorry for late reply. I mean that the previous example stored intermediate embeddings on CPU, and we can only maintain the same runtime if we now store them on GPU. I think we still need to look into what causes the real slowdown.

Issue #7475, opened 2023-05-31
knn_graph is not working on cuda


qilinli:
### 🐛 Describe the bug

The knn_graph works fine with x on the CPU but returns no edge if x is on CUDA.

```
import torch
from torch_geometric.nn import knn_graph

x = torch.rand((10, 2))

edge_index_cpu = knn_graph(x.cpu(), k=2)
edge_index_cuda = knn_graph(x.cuda(), k=2)

print('cpu:', edge_index_cpu.shape)
print('cuda:', edge_index_cuda.shape)

cpu: torch.Size([2, 20])
cuda: torch.Size([2, 0])
```

### Environment

* PyG version: 2.3.1
* PyTorch version: 2.0.1
* torch-cluster version: 1.6.1
* torch-scatter version: 2.1.1
* OS: Windows 10
* Python version: 3.10.10
* CUDA/cuDNN version: 11.8
* How you installed PyTorch and PyG (`conda`, `pip`, source): conda (PyTorch), pip (PyG)

rusty1s:
Seems related: https://github.com/rusty1s/pytorch_cluster/issues/181

PR #7497, opened 2023-06-02
Ensure data is on the same device in `knn_graph`

mzamini92:
we can directly move the tensors to the desired device using the `to()` method provided by PyTorch. This ensures that all tensors within the data object are on the same device before calling `knn_graph`.
This seems to be a solution for #7475 issue.

### DIFF
--- a/torch_geometric/nn/pool/__init__.py
+++ b/torch_geometric/nn/pool/__init__.py
@@ -165,8 +165,8 @@ def knn_graph(
     if batch is not None and x.device != batch.device:
         warnings.warn("Input tensor 'x' and 'batch' are on different devices "
                       "in 'knn_graph'. Performing blocking device transfer")
-
         batch = batch.to(x.device)
+
     if not torch_geometric.typing.WITH_TORCH_CLUSTER_BATCH_SIZE:
         return torch_cluster.knn_graph(x, k, batch, loop, flow, cosine,
                                        num_workers)
--- a/torch_geometric/nn/pool/__init__.py
+++ b/torch_geometric/nn/pool/__init__.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Optional
 from torch import Tensor
 
@@ -161,6 +162,11 @@ def knn_graph(
 
     :rtype: :class:`torch.Tensor`
     """
+    if batch is not None and x.device != batch.device:
+        warnings.warn("Input tensor 'x' and 'batch' are on different devices "
+                      "in 'knn_graph'. Performing blocking device transfer")
+
+        batch = batch.to(x.device)
     if not torch_geometric.typing.WITH_TORCH_CLUSTER_BATCH_SIZE:
         return torch_cluster.knn_graph(x, k, batch, loop, flow, cosine,
                                        num_workers)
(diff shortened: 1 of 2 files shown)


rusty1s:
Thanks for the PR. I am not totally sure though why we need it. I don't think we should do any implicit device transfers, and `knn_graph` should already fail on tensors on different devices. Can you clarify?

mzamini92:
> Thanks for the PR. I am not totally sure though why we need it. I don't think we should do any implicit device transfers, and `knn_graph` should already fail on tensors on different devices. Can you clarify?

Thanks for going through my solution. The reason for this is that `knn_graph` might not raise an error when the input tensors are on different devices. Instead, it may silently produce incorrect results or unexpected behavior. By explicitly transferring the input tensor to the correct device before calling knn_graph, we ensure that the function operates correctly and returns valid edge indices.
While it's possible that `knn_graph` should raise an error when the input tensors are on different devices, the proposed fix provides a more robust solution by handling the device transfer explicitly. This ensures that the function works as intended and avoids potential issues with mixed-device tensors.

rusty1s:
I cannot really reproduce this via
```python
import torch

from torch_geometric.nn import knn_graph

x = torch.rand((10, 2))
batch = torch.tensor([0, 0, 0, 0, 0, 1, 1, 1, 1, 1])

edge_index_cpu = knn_graph(x, k=2, batch=batch.cuda())
print(edge_index_cpu)
edge_index_cuda = knn_graph(x.cuda(), k=2, batch=batch)
print(edge_index_cuda)
```
Both calls fail as expected. I am a bit hesitant with automatic device transfers in this function, as it will slow down your program if the inputs are not on the correct device. This is a user error, and I think failing out here is the correct approach.

mzamini92:
> I cannot really reproduce this via
> 
> ```python
> import torch
> 
> from torch_geometric.nn import knn_graph
> 
> x = torch.rand((10, 2))
> batch = torch.tensor([0, 0, 0, 0, 0, 1, 1, 1, 1, 1])
> 
> edge_index_cpu = knn_graph(x, k=2, batch=batch.cuda())
> print(edge_index_cpu)
> edge_index_cuda = knn_graph(x.cuda(), k=2, batch=batch)
> print(edge_index_cuda)
> ```
> 
> Both calls fail as expected. I am a bit hesitant with automatic device transfers in this function, as it will slow down your program if the inputs are not on the correct device. This is a user error, and I think failing out here is the correct approach.

yes this error makes sense to show up. since this was mentioned as bug I thought maybe it would be a good idea that we explicitly transferring the data. but I agree with you. The code would be more robust but a little bit might be slower. it is not considerable though.

rusty1s:
Thanks for your response. In that case, shall we close this PR? Alternatively, we can add a warning whenever we perform memory transfers in this function. This would work for me as well.

mzamini92:
> Thanks for your response. In that case, shall we close this PR? Alternatively, we can add a warning whenever we perform memory transfers in this function. This would work for me as well.

The Alternative sounds like a good option too. I modified the code accordingly to show the Warning. In this modified version, a warning message is printed when the input data `(data.pos)` is on CUDA. The data is then transferred to the CPU `(data.pos.cpu())` before calling the knn_graph function. This ensures that the knn_graph function is always called with CPU data. if that's not what you meant then you can go ahead and close the PR. Thank you for your time.

PR #7520, opened 2023-06-05
Fix for `NeighborSampler` for `mag240m`

puririshi98:
to solve:
```
self.time= ('paper', tensor([False, False, False,  ..., False, False, False]))

    self.time['paper'][index] if self.time is not None else None,
TypeError: tuple indices must be integers or slices, not str
```

https://github.com/snap-stanford/ogb/pull/449

### DIFF
--- a/torch_geometric/sampler/base.py
+++ b/torch_geometric/sampler/base.py
@@ -112,20 +112,12 @@ class EdgeSamplerInput(CastMixin):
     def __getitem__(self, index: Union[Tensor, Any]) -> 'EdgeSamplerInput':
         if not isinstance(index, Tensor):
             index = torch.tensor(index, dtype=torch.long)
-        if self.time is None:
-            time = None
-        else:
-            if isinstance(self.time, tuple):
-                # for case where self.time = ('paper', tensor)
-                time = self.time[-1][index]
-            else:
-                time = self.time[index]
         return EdgeSamplerInput(
             self.input_id[index] if self.input_id is not None else index,
             self.row[index],
             self.col[index],
             self.label[index] if self.label is not None else None,
-            time,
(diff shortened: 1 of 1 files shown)


torch_geometric/sampler/base.py
```diff
@@ -105,13 +105,20 @@ class EdgeSamplerInput(CastMixin):
     def __getitem__(self, index: Union[Tensor, Any]) -> 'EdgeSamplerInput':
         if not isinstance(index, Tensor):
             index = torch.tensor(index, dtype=torch.long)
-
+        if self.time is None:
```
rusty1s, line 108: When is `time` a `tuple`? It should always be of type `Optional[Tensor]`.

puririshi98: it is for the mag240m PR i am working on based off of yanbing's original PR
https://github.com/snap-stanford/ogb/pull/449

If i print time before the error:
```
self.time= ('paper', tensor([False, False, False,  ..., False, False, False]))

    self.time['paper'][index] if self.time is not None else None,
TypeError: tuple indices must be integers or slices, not str
```

rusty1s: Where and how do you pass `time`. Wouldn't it be sufficient to just pass a `Tensor` there? I don't think there is a need to "fix" this here.

puririshi98: basing off of yanbings PR here: https://github.com/snap-stanford/ogb/pull/449
not sure why the time is like that but ill investigate and try to fix in ogb instead of here

puririshi98:
https://github.com/snap-stanford/ogb/pull/449 PR works w/ main now, no issues

PR #7533, opened 2023-06-06
Update local_cartesian.py

mzamini92:
The intermediate tensors `cart` and `max_value` in the original code were replaced with in-place operations to reduce memory usage. This was done by directly operating on the `cart` tensor and computing the maximum value iteratively without creating a separate `max_value` tensor. In-place operations `(torch.sub, cart.div_, cart.mul_, cart.add_)` were used to perform computations directly on tensors, reducing memory usage and eliminating the need for intermediate tensors. To compute the maximum value in a streaming fashion, a loop was introduced which iterates over the edges and updates the maximum value tensor (max_value) accordingly.

### DIFF
--- a/torch_geometric/transforms/local_cartesian.py
+++ b/torch_geometric/transforms/local_cartesian.py
@@ -1,41 +1,50 @@
 import torch
-
 from torch_geometric.data import Data
 from torch_geometric.data.datapipes import functional_transform
 from torch_geometric.transforms import BaseTransform
-from torch_geometric.utils import scatter
-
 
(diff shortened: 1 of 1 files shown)


rusty1s (commented):
Thanks. I like the `interval` argument and I think it would be great to add this to other transforms as well, so please feel free to go ahead and modify other transforms.

Note that I removed the in-place refactoring in this PR. I don't think this is needed, and there were some for-loops added here that unnecessarily slowed down this transformation. Hope this works for you!

rusty1s (approved):

PR #7537, opened 2023-06-07
Use `repeat` instead of `new_full` to avoid graph breaks in `to_dense_batch`

piotrchmiel:
New full if fill value is tensor call item() on it under the hood which causes graph break in torch.compile

### DIFF
--- a/torch_geometric/utils/to_dense_batch.py
+++ b/torch_geometric/utils/to_dense_batch.py
@@ -94,7 +94,6 @@ def to_dense_batch(
                 [ True, False, False, False],
                 [ True,  True,  True, False]])
     """
-
     if batch is None and max_num_nodes is None:
         mask = torch.ones(1, x.size(0), dtype=torch.bool, device=x.device)
         return x.unsqueeze(0), mask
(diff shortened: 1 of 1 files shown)


torch_geometric/utils/to_dense_batch.py
```diff
@@ -126,7 +125,7 @@ def to_dense_batch(
         x, idx = x[mask], idx[mask]
 
     size = [batch_size * max_num_nodes] + list(x.size())[1:]
-    out = x.new_full(size, fill_value)
+    out = torch.as_tensor(fill_value).repeat(size)
```
rusty1s, line 128:
```suggestion
    out = torch.as_tensor(fill_value, dtype=x.dtype, device=x.device).repeat(size)
```

CHANGELOG.md
```diff
@@ -94,6 +94,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added an optional `batch_size` and `max_num_nodes` arguments to `MemPooling` layer ([#7239](https://github.com/pyg-team/pytorch_geometric/pull/7239))
 - Fixed training issues of the GraphGPS example ([#7377](https://github.com/pyg-team/pytorch_geometric/pull/7377))
 - Allowed `CaptumExplainer` to be called multiple times in a row ([#7391](https://github.com/pyg-team/pytorch_geometric/pull/7391))
+- Use repeat instead new_full to create out with fill_value inside `to_dense_batch` function ([#7537](https://github.com/pyg-team/pytorch_geometric/pull/7537))
```
rusty1s, line 97:
```suggestion
```

rusty1s (approved):

PR #7539, opened 2023-06-07
Avoid transforming original data in `ToUndirected` 

denadai2:
This PR clones the original data object in `ToUndirected` so that the method does not change the original data


rusty1s:
I don't think this is needed anymore with the addition of #7429.

denadai2:
yeah I do not think so!

PR #7543, opened 2023-06-08
Add an example PMLP on Cora dataset

n-patricia:


examples/cora_pmlp.py
wsad1, line 1: This file is empty. Probably forgot to add it?

n-patricia: is it still empty? I had it in my branch

rusty1s: Works now :)

rusty1s (approved):
Thank you :)

PR #7548, opened 2023-06-08
Allow GPU inference in `BasicGNN`

Kh4L:


torch_geometric/nn/models/basic_gnn.py
```diff
@@ -262,7 +262,7 @@ def inference(self, loader: NeighborLoader,
             pbar = tqdm(total=len(self.convs) * len(loader))
             pbar.set_description('Inference')
 
-        x_all = loader.data.x.cpu()
```
rusty1s, line 265: Can we instead add an additional argument to the function that dictates on which device embeddings should be stored? For example, `embedding_device: torch.device = 'cpu'`.

Kh4L: Done

rusty1s (approved):

PR #7573, opened 2023-06-13
Added `edge_index` property to `TemporalData`

SalvishGoomanee:
Added the `edge_index` property in order to make it easier to visualize time series of temporal by simply doing a `test_data.edge_index` and using `networkx` which is useful for large multiple graph datasets.

### DIFF
--- a/torch_geometric/data/temporal.py
+++ b/torch_geometric/data/temporal.py
@@ -212,6 +212,10 @@ class TemporalData(BaseData):
         r"""Alias for :meth:`~torch_geometric.data.TemporalData.num_events`."""
         return self.num_events
 
+    @property
+    def edge_index(self) -> Any:
+        return self['edge_index'] if 'edge_index' in self._store else None
+
     def size(
         self, dim: Optional[int] = None
     ) -> Union[Tuple[Optional[int], Optional[int]], Optional[int]]:
--- a/torch_geometric/data/temporal.py
+++ b/torch_geometric/data/temporal.py
@@ -213,8 +213,14 @@ class TemporalData(BaseData):
         return self.num_events
 
     @property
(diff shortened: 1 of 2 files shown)


torch_geometric/data/temporal.py
```diff
@@ -212,6 +212,10 @@ def num_edges(self) -> int:
         r"""Alias for :meth:`~torch_geometric.data.TemporalData.num_events`."""
         return self.num_events
 
+    @property
+    def edge_index(self) -> Any:
```
EdisonLeeeee, line 216:
```suggestion
    def edge_index(self) -> Optional[Tensor]:
```

torch_geometric/data/temporal.py
```diff
@@ -212,6 +212,10 @@ def num_edges(self) -> int:
...
+        return self['edge_index'] if 'edge_index' in self._store else None
```
EdisonLeeeee, line 217: `edge_index` can be also obtained by stacking `src` and `dst`, WDYT?

SalvishGoomanee: This is in order to build a graph based on temporal edges by creating a TemporalData that networkx can understand just like with the Data object. I agree, one can stack `src` and `dst' but this is just a bit cleaner.

EdisonLeeeee: Yep, I mean we can do something like:
```python
    @property
    def edge_index(self):
        edge_index = self['edge_index'] 
        if edge_index is None and not (self.src is None or self.dst is None):
            edge_index = torch.stack([self.src, self.dst], dim=0)
        return edge_index
```

rusty1s (approved):

PR #7357, opened 2023-05-12
`GLGExplainer`

steveazzolin:
PR implementing  `GLGExplainer` (https://openreview.net/forum?id=OTbRTIY4YS)

PR #7577, opened 2023-06-14
Added `GLGExplainer`

steveazzolin:
As discussed in #7357, I'm uploading a refined version of the main implementation of `GLGExplainer` in which only the essential components are kept to avoid clutter.
Future PRs will include the visualization module, the relative metrics along with the two classes `LogicExplanation` and `LogicExplanations`

### DIFF
--- a/torch_geometric/explain/glg_explainer.py
+++ b/torch_geometric/explain/glg_explainer.py
@@ -16,7 +16,7 @@ import torch.nn.functional as F
 
 class GLGExplainer(torch.nn.Module):
     r"""Implementation of GLGExplainer `"Global Explainability of GNNs via
-    Logic Combination of Learned Concepts" 
+    Logic Combination of Learned Concepts"
     <https://openreview.net/forum?id=OTbRTIY4YS>`_
 
     Importantly, the :class:`GLGExplainer` needs to be trained via
--- a/torch_geometric/explain/glg_explainer.py
+++ b/torch_geometric/explain/glg_explainer.py
@@ -1,0 +1,621 @@
+from typing import Dict, Optional, Tuple
+from collections import defaultdict
+import math
+
+import torch
+from sympy import to_dnf, lambdify
+from torch import Tensor
+from torch_geometric.explain.metric import (
+    formulas_accuracy
+)
+from torch_geometric.explain import LogicExplanation, LogicExplanations
+from torch_geometric.loader import DataLoader
+from torch_scatter import scatter
+import torch.nn.functional as F
+
+
+class GLGExplainer(torch.nn.Module):
+    r"""Implementation of GLGExplainer `"Global Explainability of GNNs via
+    Logic Combination of Learned Concepts" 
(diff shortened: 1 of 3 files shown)


torch_geometric/explain/glg_explainer.py
```diff
@@ -0,0 +1,621 @@
...
+from torch_geometric.explain.metric import (
+    formulas_accuracy
+)
+from torch_geometric.explain import LogicExplanation, LogicExplanations
+from torch_geometric.loader import DataLoader
+from torch_scatter import scatter
+import torch.nn.functional as F
+
+
+class GLGExplainer(torch.nn.Module):
```
wsad1, line 17: Can we have this explainer inherit from `ExplainerAlgorithm` as other algorithms in [torch_geometric/explain/algorithm](https://github.com/pyg-team/pytorch_geometric/tree/master/torch_geometric/explain/algorithm) does? And have it conform to the same interface as other `ExplainerAlgorithm`s .

torch_geometric/explain/glg_explainer.py
```diff
@@ -0,0 +1,621 @@
...
+        self.optimizer.add_param_group({
+            'params': self.len_model.parameters(),
+            'lr': len_lr
+        })
+        self.optimizer.add_param_group({
+            'params': self.prototype_vectors,
+            'lr': prototypes_lr
+        })
+
+    def do_epoch(self, loader: DataLoader, train: bool = False) -> Dict:
```
wsad1, line 96: Can we create a `train` interface like [PGExplainer](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/explain/algorithm/pg_explainer.py), where its expects inputs `x`, `edge_index` and `batch` and runs once batch of training.

wsad1 (commented):
My high level comment would be to conform to the [ExplainerAlgorithm](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/explain/algorithm/base.py) interface if possible, because that helps this new algorithm integrate better with the existing [explain](https://pytorch-geometric.readthedocs.io/en/latest/modules/explain.html) module.

torch_geometric/explain/glg_explainer.py
steveazzolin: Thanks for your comment. At first I actually tried to conform better to the interface, but then I realized that the differences between the proposed approach and the other Explainers available are substantial. Indeed, `ExplainerAlgorithm` seems heavily tailored for supporting Local Explainers (aka Instance-level Explainers), while the proposed one is a Global Explainer, and as such is radically different from the already implemented explainers.

For example, the forward function in `ExplainerAlgorithm` is meant to extract the explanations for every input instance, while for `GLGExplainer` the explanations are extracted during training, then you can simply inspect what the model has learned.

So, after chatting with @rusty1s, I opted for differentiating the two approaches, in order to avoid confusion about different frameworks sharing the same functions and the same input parameters (Local Explainers take as input graphs, while Global takes as input graphs which are actually local explanations)

PR #7591, opened 2023-06-16
`NBFNet` implementation for homogeneous graphs

daniel-unyi-42:
Initial version of NBFNet - works for homogeneous graphs.

Possible modifications/extensions of this version:
- support for knowledge graphs (I'm working on this already)
- support for other aggregations, especially PNA - right now, only sum aggregation can be used
- support for other message functions - right now, only distmult can be used
- the users can specify which graph conv layer they want to use, including the incorporation of node features and edge attributes (this would also extend the methodology detailed in the NBFNet paper)
- adding documentation & testing

### DIFF
--- a/torch_geometric/nn/models/nbfnet.py
+++ b/torch_geometric/nn/models/nbfnet.py
@@ -1,0 +1,191 @@
+from functools import reduce
+
+import torch
+from torch import LongTensor, Tensor
+from torch.nn import Embedding, LayerNorm, Module, ModuleList, ReLU
+
+from torch_geometric.nn.conv import MessagePassing
+from torch_geometric.nn.dense.linear import Linear
+from torch_geometric.nn.models import MLP
+from torch_geometric.typing import Optional, OptTensor, Tuple
+from torch_geometric.utils import add_self_loops
+
+
+def edge_match(edge_index, query_index):
+    # source of algorithm:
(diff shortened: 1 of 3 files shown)


torch_geometric/nn/models/nbfnet.py
```diff
@@ -0,0 +1,191 @@
...
+        self.linear.reset_parameters()
+        self.relation.reset_parameters()
+
+    def forward(self, x: Tensor, boundary: Tensor, edge_index: LongTensor,
+                edge_weight: OptTensor = None) -> Tensor:
+        x0 = x.clone()
+        x = self.propagate(edge_index, x=x, edge_weight=edge_weight)
+        x = x + boundary
+        x = torch.cat([x0, x], dim=-1)
+        x = self.linear(x)
```
wsad1, line 72:
```suggestion
        out = self.propagate(edge_index, x=x, edge_weight=edge_weight)
        out = out + boundary
        out = torch.cat([out, x], dim=-1)
        out = self.linear(out)
```

wsad1 (commented):
Thanks for adding this @daniel-unyi-42. 
Could we add units tests and docs for all the new classes and functions.

torch_geometric/nn/models/nbfnet.py
daniel-unyi-42: Sorry, I've been quite busy lately. I'm planning to add the tests and docs in late September.
(If you ask whether you're allowed to add the test and docs instead of me, yes absolutely!)

Issue #1773, opened 2020-10-29
Request to add two graph self-supervised methods


yyou1996:
## 🚀 Feature

Request to add two graph self-supervised methods to Pytorch Geometric:
(1) heuristic designed graph self-supervised tasks (ICML'20): https://arxiv.org/abs/2006.09136;
code: https://github.com/Shen-Lab/SS-GCNs;
(2) contrastive learning as self-supervision (NeurIPS'20): https://arxiv.org/abs/2010.13902;
code: https://github.com/Shen-Lab/GraphCL.

## Motivation

Self-supervision as an emerging technique has been employed to train graph neural networks (GNNs) for more transferrable, generalizable, and robust representation learning of graphs.

rusty1s:
Thanks, I will look into it. Feel free to contribute if you like :)

kou18n:
@rusty1s Hello, I implemented the GraphCL model with sparse_tesnor support. I use the [PyGCL](https://github.com/PyGCL/PyGCL) code, and rewrite some code. Currently, I think PyG can not support the contrastive learning model. How should I contribute to PyG?

rusty1s:
That sounds cool :) We are very happy to take a pull request in on that one, including both the model and an example. Let me know if you have any questions.

SauravMaheshkar:
@kou18n I'd like to work on this again. Can I take it up?

kou18n:
> @kou18n I'd like to work on this again. Can I take it up?

Of course. I haven't submitted the code. Thank you for your help. Maybe we can finish it together.

SauravMaheshkar:
> > @kou18n I'd like to work on this again. Can I take it up?
> 
> Of course. I haven't submitted the code. Thank you for your help. Maybe we can finish it together.

Sounds good. I'll create a new PR and we can discuss async.

SauravMaheshkar:
PS: totally forgot about this. If someone wants to take this up

PR #7593, opened 2023-06-16
`GraphCL` implementation with example

kou18n:
#1773 I don't know how to better organize the code into PyG, please check it.

### DIFF
--- a/examples/GCL/augmentors/augmentor.py
+++ b/examples/GCL/augmentors/augmentor.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
-import torch
 from abc import ABC, abstractmethod
-from typing import Optional, Tuple, NamedTuple, List
+from typing import List, NamedTuple, Optional, Tuple
+
+import torch
 from torch_sparse import SparseTensor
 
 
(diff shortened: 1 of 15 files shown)


examples/GCL/augmentors/augmentor.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import List, NamedTuple, Optional, Tuple
+
+import torch
+from torch_sparse import SparseTensor
+
+
+class Graph(NamedTuple):
```
wsad1, line 10: Lets not create a new data object. Reuse `torch_geometric.data.Data`.

wsad1 (changes requested):
Hi @kou18n thanks for making these changes. I have two hight level comments
1. Reuse the [Data](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.data.Data.html#torch_geometric.data.Data) object instead of defining a new data object `Graph`.  All of pyg uses this data object, and it doesn't make sense to define a new data object. Plus the `Data` object should have all the properties needed to implement the augmentors/transforms.
2. All transformations in `augmentors` can be moved to [torch_goemetric/transforms](https://github.com/pyg-team/pytorch_geometric/tree/master/torch_geometric/transforms) 
3. Move models to [torch_geometric/nn](https://github.com/pyg-team/pytorch_geometric/tree/master/torch_geometric/nn)

Overall there isn't a need to define new interfaces like `Graph` \ `Augmentor` \ `Sampler`. Lets use the existing interfaces like `Data` \ `Transform` in this PR.

examples/GCL/augmentors/augmentor.py
kou18n: Thank you for advices. I will check the code!

PR #7594, opened 2023-06-16
Documentation Page for HGAM

andreazanetti:
Adding a documentation page for HGAM. 
The idea is to pass the following information to the user:

a) high level perspective of what HGAM is
b) role of BFS ordering of the graph batch for HGAM to work
c) support in the neighbor loader
d) practical working example of TrimToLayer class
e) pointers to external resources

This is a draft, asking for preliminary feedback.
Thanks for any input!

Contributors: @mszarma,  @rBenke (robert.benke@intel.com), manos.farsarakis@intel.com


rusty1s (approved):
I think this looks good, thank you :) Eventually, we can extend this tutorial by some figures to show the difference in GNN computation visually as well.

PR #7595, opened 2023-06-16
Fixed `SelectTopK` with x.dim() == 1

marshka:
Fixed unsqueezing when x is unidimensional (reassignment was missing). From

```python
x.view(-1, 1) if x.dim() == 1 else x
```
To:
```python
x = x.view(-1, 1) if x.dim() == 1 else x
```


EdisonLeeeee (approved):
Thank you!

PR #7598, opened 2023-06-16
Fix typo in documentation: adivisable -> advisable

djm93dev:
I noticed a typo in the docs where "adivisable" was used instead of "advisable."


EdisonLeeeee (approved):
Thanks for spotting!

PR #7603, opened 2023-06-18
`LightGCN` example

happykygo:


rusty1s (approved):

hankaj:
Hi @rusty1s and @happykygo!
I have a question - why was validation set deleted? I was doing experiments with LightGCN on my own and it seems weird to me to use test set during training but I suppose there is a reason for that. Could you tell more about it?

happykygo:
Hi @hankaj, I would normally use a val dataset for finding the optimal hyperparameters and use the test set only at the very end to test the final model performance. In this example, I used the val set to figure out the optimal number of epochs during training. I guess @rusty1s removed that to simplify the code and highlight how to use LightGCN. Using a val set or not is not relevant in this context. In a real project, I suggest you use a val set to find the optimal hyperparameters, including the number of epochs and so on.

PR #7604, opened 2023-06-18
Add `DistDataset` for distributed training

ZhengHongming888:

This class will do
1) load the distributed partition of graph/node_feats/edge_feats/label/partition books from different partition files in different multiple nodes; 
2) initialize the partition data into LocalGraphStore/LocalFeatureStore format for later use
3) also keep the original PyG graph Data format for homo/hetero. 
4) keep the interface for other partition format for import, like dgl/glt, etc

The files related -
1)  distributed/dist_dataset.py
2)  distributed/partition.py  --- load_partition()
3)  test/test_dist_dataset.py to cover the home/hetero both format based on FakeDataset/FakeHeteroDataset

### DIFF
--- a/test/distributed/test_dist_dataset.py
+++ b/test/distributed/test_dist_dataset.py
@@ -1,0 +1,102 @@
+import os.path as osp
+
+import torch
+
+from torch_geometric.datasets import FakeDataset, FakeHeteroDataset
+from torch_geometric.distributed import Partitioner
+from torch_geometric.typing import EdgeTypeStr
+from torch_geometric.distributed import DistDataset
+
+def test_dist_dataset_for_homo():
+    num_partitions = 2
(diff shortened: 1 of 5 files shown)


rusty1s (commented):
Hey @ZhengHongming888, I am not totally sure why we need `DistDataset` here. I think it is very important for us on PyG side to keep the disentanglement of `FeatureStore` and `GraphStore` and not unify them again. That is, we should probably create like `FeatureStore.from_partition()` and `GraphStore.from_partition()` functionalities for this. With that, you can even think about just distributing the feature stores, while keeping the graph in memory (and vice versa).

rusty1s:
Hey @ZhengHongming888, do we still need this?

ZhengHongming888:
No thanks Matthias!

PR #7611, opened 2023-06-19
Hypergraph data object

CodeTal:
Hello. I don't know the definition of a subgraph for a hypergraph (not mentioned in papers). When getting a subgraph, a hyperedge can be cut. For example, if a hyperedge connects vertices [1, 2, 3], but only [1, 3] are selected for the subgraph, then we have the option to: 1) remove this hyperedge; 2) change the hyperedge and make it only connects [1, 3]. So, in this version, I choose the second option.


wsad1 (commented):
Thanks for the great work @CodeTal. 
Will look at your code in detail and leave more comments. For now, could you add tests for the `HyperGraphData` and all the new methods. Also add docs.

CodeTal:
> Thanks for the great work @CodeTal. Will look at your code in detail and leave more comments. For now, could you add tests for the `HyperGraphData` and all the new methods. Also add docs.

Sure.

CodeTal:
Hi. I found that the modified `HyperGraphData` class should include a different `to_heterogeneous` function. Since the format of `edge_index` in HyperGraphData has changed (from (2, edge_num) to (2, edge_num*nodes_per_edge)), then the format of `edge_type` should change. Do you need to implement this function in this version? If so, I was thinking about how to express the `edge_type` of `HyperGraphData`. We can either assign each node-edge pair an `edge_type` or just assign the unique edge numbers an `edge_type`.

wsad1:
Sorry for the late reply @CodeTal. For this PR we can skip the implementation of `to_heterogenous` and raise a `NotImplemntedError`. Lets implement this in a follow up PR, since reviewing smaller PRs is easier.

CodeTal:
Hello! In this PR, I have added test cases to test `subgraph()` and `__inc__()`. I would like to add some docs to these code pieces, but I am not sure what are the docs. Do you mean a post on the [PyG documentation](https://pytorch-geometric.readthedocs.io/en/latest/), or a README file under the directory of the source code?

torch_geometric/data/hypergraph_data.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __inc__(self, key: str, value: Any, *args, **kwargs) -> Any:
+        if 'batch' in key:
+            return int(value.max()) + 1
+        elif key == 'face':
+            return self.num_nodes
+        elif key == 'edge_index':
+            return torch.tensor([[self.num_nodes], [self.num_edges]])
+        else:
+            return 0
```
wsad1, line 30:
```suggestion
        if key == 'edge_index':
            return torch.tensor([[self.num_nodes], [self.num_edges]])
        else:
            return super().__inc__(key, value, *args, **kwargs)
```

test/data/test_hypergraph_data.py
```diff
@@ -0,0 +1,158 @@
...
+    del data[value]
+    del data[value]  # Deleting unset attributes should work as well.
+
+    assert data.get(key) is None
+    assert data.get('title') == 'test'
+
+    torch_geometric.set_debug(False)
+
+
+def test_hypergraphdata_subgraph():
```
wsad1, line 129: Can we add tests with dataloader and `HypyerGraphData` object.

torch_geometric/data/hypergraph_data.py
```diff
@@ -0,0 +1,118 @@
...
+import torch
+from torch import Tensor
+
+from torch_geometric.data import Data
+from torch_geometric.typing import EdgeType, NodeType, OptTensor
+from torch_geometric.utils import hyper_subgraph, select
+
+
+class HyperGraphData(Data):
+    def __init__(self, x: OptTensor = None, edge_index: OptTensor = None,
```
wsad1, line 14: Add comments like done in `Data` object [here](https://github.com/pyg-team/pytorch_geometric/blob/a4447449a7ab185a643dba878e8548c452a62a26/torch_geometric/data/data.py#L397). Try to mention the `edge_index` and how it should look for hyper graphs.

torch_geometric/data/hypergraph_data.py
```diff
@@ -0,0 +1,118 @@
...
+                 edge_attr: OptTensor = None, y: OptTensor = None,
+                 pos: OptTensor = None, **kwargs):
+        super().__init__(x=x, edge_index=edge_index, edge_attr=edge_attr, y=y,
+                         pos=pos)
+        for key, value in kwargs.items():
+            setattr(self, key, value)
```
wsad1, line 20:
```suggestion
        super().__init__(x=x, edge_index=edge_index, edge_attr=edge_attr, y=y,
                         pos=pos, **kwargs)
```

wsad1 (commented):
Thanks for the great work @CodeTal. 

The `num_edges` function also needs to overridden, because the current `data.num_edges` function returns `edge_index.shape[1]`, but for hyper graphs the number of edges is `max(edge_index[1])`. Assuming `edge_index` for hypergraphs is define as done here: https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HypergraphConv.html#torch_geometric.nn.conv.HypergraphConv.

CodeTal:
@wsad1 Thanks for your careful review and advice! I have added documentation comments, as well as fixing the old code.

wsad1:
@CodeTal I updated the the tests for `subgraph`  by adding `edge_attr`. This is when I realised there is an inconsistency in the way we define edge attributes. For hyper edges should an edge attribute be of shape `[num_edges, num_attrs]` ie. a single attribute per hyper edge OR of shape `[num_edges*num_nodes_per_edge, num_attrs]` ie. an attribute for each connection of a hyper edge. I think it should be the latter. Given that `hyper_subgraph` should return a masks of shape `num_edges` to filter out, and not a mask of shape  ` num_edges*num_nodes_per_edge`.
I'll try to make this update today. But if I am not able to please go ahead with it.

wsad1 (approved):
@CodeTal thanks for the great work.

CodeTal:
> @CodeTal thanks for the great work.

And thank you for your careful review and enlightening update! Sorry for not following the issue last week, I was busy with other things on last week and had little spare time.

mehular0ra:
@CodeTal  I really liked your implementation. Just a question I had. Will this Hypergraph data object support multimodality? By that what I mean is: if let's say Incidence Matrix has a size of not [n,n], but [n, m] where m = k*n. k=1, 2, 3,...
That is, will your implementation support when there more number of hyperedges than the nodes?

PR #7614, opened 2023-06-20
Add `interval` argument to `Cartesian`

mzamini92:
I've added the `interval` argument to the `LineGraph` class and modified the code to apply the normalization to the `data.x` (edge_attr) based on the specified interval.
@rusty1s


rusty1s (approved):
Thanks. I moved this functionality to `Cartesian` instead since this is where I think it is the most fitting. I don't think we need it in any transform since one can usually achieve the same by composing different transforms together, e.g., via `NormalizeFeatures()` transform.

PR #7615, opened 2023-06-20
Update local_degree_profile.py

mzamini92:
I've added the `interval` argument to the `LocalDegreeProfile` class and modified the code to apply normalization to the `data.x` based on the specified interval.

@rusty1s


rusty1s:
Please see my comment in #7614. I don't think adding this argument is needed here.

PR #7633, opened 2023-06-23
Add some default NVTX ranges

Kh4L:
This PR leverages torch's NVTX API to add default nvtx ranges to help [profiling with nsys](https://developer.nvidia.com/nsight-systems). Those can be enabled with the `NVIDIA_NVTX_RANGES` environment variable.

## How to use:
```
NVIDIA_NVTX_RANGES=1 nsys profile -s none -t cuda,nvtx,osrt,cudnn,cublas --force-overwrite=true python train.py 
```
## Result
nsys_profile_example_pyg

It adds ranges for the following functions:
- `Dataloaders`'s `collate_fn`
- `Dataset`'s `__getitem__`
- `MessagePassing`'s main member  functions
- `NeighborSampler`'s


puririshi98:
LGTM!

torch_geometric/nn/conv/message_passing.py
```diff
@@ -184,6 +184,34 @@ def __init__(
         self._edge_update_forward_pre_hooks = OrderedDict()
         self._edge_update_forward_hooks = OrderedDict()
 
+        # Init NVTX Ranges for this op
+        if os.environ.get('NVIDIA_NVTX_RANGES',
+                          "0") == "1" and torch.cuda.is_available():
+            self._nvtx_handles = dict()
+
+            def get_hooks_for(func_name):
```
akihironitta, line 192: I think this is not an issue if `NVIDIA_NVTX_RANGES` is not meant to be used by the community, but I prefer to define the hook at the top-level to avoid `PicklingError` in a multiprocessing setting. https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled

PR #7644, opened 2023-06-26
Clean-up `TGN` example

puririshi98:
checked that accuracys match before and after my PR


torch_geometric/nn/models/tgn.py
```diff
@@ -98,10 +99,10 @@ def forward(self, n_id: Tensor) -> Tuple[Tensor, Tensor]:
 
         return memory, last_update
 
-    def update_state(self, src: Tensor, dst: Tensor, t: Tensor,
-                     raw_msg: Tensor):
+    def update_state(self, batch: Batch):
         """Updates the memory with newly encountered interactions
         :obj:`(src, dst, t, raw_msg)`."""
+        src, dst, t, raw_msg = batch.src, batch.dst, batch.t, batch.msg
```
wsad1, line 105:
```suggestion
    def update_state(self, src: Tensor, dst: Tensor, t: Tensor,
                     raw_msg: Tensor):
        """Updates the memory with newly encountered interactions
        :obj:`(src, dst, t, raw_msg)`."""
```
I would keep this as is. Because it doesn't break things, most models and model functions in pyg have this kind of signature, and this is explicit.

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -17,7 +18,9 @@ class TemporalDataLoader(torch.utils.data.DataLoader):
         **kwargs (optional): Additional arguments of
             :class:`torch.utils.data.DataLoader`.
     """
-    def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
+    def __init__(self, data: TemporalData, batch_size: int = 1,
+                 device: Union[torch.device, str] = 'cpu',
+                 negative_sampling=False, neighbor_loader_size=10, **kwargs):
```
wsad1, line 23: update doc string, with `neighbor_loder_size` , and also explain how last neighbor loader is used.

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -26,13 +29,44 @@ def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
...
+        self.assoc = torch.empty(data.num_nodes, dtype=torch.long,
+                                 device=self.device)
         if kwargs.get('drop_last', False) and len(data) % batch_size != 0:
             arange = range(0, len(data) - batch_size, batch_size)
         else:
             arange = range(0, len(data), batch_size)
 
         super().__init__(arange, 1, shuffle=False, collate_fn=self, **kwargs)
 
+    def _get_iterator(self):
```
wsad1, line 51: Does this get called at the start of each iteration?

torch_geometric/loader/temporal_dataloader.py
wsad1: Also update tests for `TemporalDataloader` to test `negative_sampling=True`.

wsad1 (approved):
Thanks for the update @puririshi98. 
Are the val/train/test metrics  of `examples/tgn.py` same before and after this change?

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -17,7 +18,9 @@ class TemporalDataLoader(torch.utils.data.DataLoader):
         **kwargs (optional): Additional arguments of
             :class:`torch.utils.data.DataLoader`.
     """
-    def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
+    def __init__(self, data: TemporalData, batch_size: int = 1,
+                 device: Union[torch.device, str] = 'cpu',
```
rusty1s, line 22: Mh, I feel this implicit device transfer is kinda breaking the usual computation flow of PyTorch. Why do we need this here? Can we drop it?

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -26,13 +29,44 @@ def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
 
         self.data = data
         self.events_per_batch = batch_size
-
+        self.device = device
+        self.negative_sampling = negative_sampling
+        if self.negative_sampling:
```
rusty1s, line 34: Can we pass a `neg_sampling_fn` here instead? I feel this assumption of `min_dst_idx` and `max_dst_idx` is not very general and very dataset specific.

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -26,13 +29,44 @@ def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
...
+            # Ensure to only sample actual destination nodes as negatives.
+            self.min_dst_idx, self.max_dst_idx = int(self.data.dst.min()), int(
+                self.data.dst.max())
+        self.neighbor_loader = LastNeighborLoader(data.num_nodes,
```
rusty1s, line 38: I don't think this is correct. In the example, you will create multiple temporal loaders, each with there own neighbor loader. However, you wanna share the neighbor loader state across all of them.

torch_geometric/loader/temporal_dataloader.py
```diff
@@ -26,13 +29,44 @@ def __init__(self, data: TemporalData, batch_size: int = 1, **kwargs):
...
+                                 device=self.device)
         if kwargs.get('drop_last', False) and len(data) % batch_size != 0:
             arange = range(0, len(data) - batch_size, batch_size)
         else:
             arange = range(0, len(data), batch_size)
 
         super().__init__(arange, 1, shuffle=False, collate_fn=self, **kwargs)
 
+    def _get_iterator(self):
+        self.neighbor_loader.reset_state()  # Start with an empty graph.
```
rusty1s, line 52: As mentioned earlier: You don't want to start with an empty graph for validation/testing.

rusty1s (changes requested):
Thank you, really appreciate the clean-up. In general, this looks good except for the integration of `LastNeighborLoader` inside `TemporalDataLoader`. I think we should revert this, and then come up with a more principled approach of how data loading should look like in temporal scenarios.

torch_geometric/loader/temporal_dataloader.py
puririshi98: yeah

puririshi98:
@wsad1 @rusty1s thanks for the review, I have addressed your comments and checked that accuracys match before and after my PR. lmk if anything else needed to merge

puririshi98:
@rusty1s @wsad1  anything else needed to merge?

rusty1s (approved):

PR #7672, opened 2023-07-01
Add distributed graph info for distributed training

ZhengHongming888:

This PR originally designed for the DistGraph class and now merged with LocalGraphStore -

1. Add partition info into LocalGraphStore like num_partition, partition_idx, node_pb (node_partitionbook), edge_pb, partition_meta
2. Two api functions to return the partition idx based on the global ids (node/edge ids) are included. 

Now we combined the local graph and dist graph properties into one GraphStore. later on we will change the class name from LocalGraphStore into PartitionGraphStore with another PR. 

Any comments please let us know. thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Made `FieldStatus` enum picklable to avoid `PicklingError` in a multi-process setting ([#7808](https://github.com/pyg-team/pytorch_geometric/pull/7808))
 - Fixed `edge_label_index` computation in `LinkNeighborLoader` for the homogeneous+`disjoint` mode ([#7791](https://github.com/pyg-team/pytorch_geometric/pull/7791))
 - Fixed `CaptumExplainer` for `binary_classification` tasks ([#7787](https://github.com/pyg-team/pytorch_geometric/pull/7787))
 - Warn user when using the `training` flag in `to_hetero` modules ([#7772](https://github.com/pyg-team/pytorch_geometric/pull/7772))
--- a/test/data/test_feature_store.py
+++ b/test/data/test_feature_store.py
@@ -4,14 +4,13 @@ import pytest
 import torch
 
 from torch_geometric.data import TensorAttr
-from torch_geometric.data.feature_store import AttrView, _field_status
+from torch_geometric.data.feature_store import AttrView, _FieldStatus
 from torch_geometric.testing import MyFeatureStore
 
 
 @dataclass
 class MyTensorAttrNoGroupName(TensorAttr):
-    def __init__(self, attr_name=_field_status.UNSET,
-                 index=_field_status.UNSET):
+    def __init__(self, attr_name=_FieldStatus.UNSET, index=_FieldStatus.UNSET):
         # Treat group_name as optional, and move it to the end
         super().__init__(None, attr_name, index)
 
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -20,7 +20,7 @@ import torch
 from torch import Tensor
 
 from torch_geometric.data import EdgeAttr, FeatureStore, GraphStore, TensorAttr
-from torch_geometric.data.feature_store import _field_status
+from torch_geometric.data.feature_store import _FieldStatus
 from torch_geometric.data.graph_store import EdgeLayout
 from torch_geometric.data.storage import (
     BaseStorage,
@@ -372,7 +372,7 @@ class DataTensorAttr(TensorAttr):
     r"""Tensor attribute for `Data` without group name."""
     def __init__(
         self,
-        attr_name=_field_status.UNSET,
+        attr_name=_FieldStatus.UNSET,
         index=None,
     ):
         super().__init__(None, attr_name, index)
--- a/torch_geometric/data/feature_store.py
+++ b/torch_geometric/data/feature_store.py
@@ -32,11 +32,13 @@ import torch
 from torch_geometric.typing import FeatureTensorType, NodeType
 from torch_geometric.utils.mixin import CastMixin
 
-_field_status = Enum("FieldStatus", "UNSET")
-
 # We allow indexing with a tensor, numpy array, Python slicing, or a single
 # integer index.
 IndexType = Union[torch.Tensor, np.ndarray, slice, int]
+
+
+class _FieldStatus(Enum):
+    UNSET = None
 
 
 @dataclass
@@ -52,20 +54,20 @@ class TensorAttr(CastMixin):
     """
 
     # The group name that the tensor corresponds to. Defaults to UNSET.
-    group_name: Optional[NodeType] = _field_status.UNSET
+    group_name: Optional[NodeType] = _FieldStatus.UNSET
 
     # The name of the tensor within its group. Defaults to UNSET.
-    attr_name: Optional[str] = _field_status.UNSET
+    attr_name: Optional[str] = _FieldStatus.UNSET
 
     # The node indices the rows of the tensor correspond to. Defaults to UNSET.
-    index: Optional[IndexType] = _field_status.UNSET
+    index: Optional[IndexType] = _FieldStatus.UNSET
 
     # Convenience methods #####################################################
 
     def is_set(self, key: str) -> bool:
         r"""Whether an attribute is set in :obj:`TensorAttr`."""
         assert key in self.__dataclass_fields__
-        return getattr(self, key) != _field_status.UNSET
+        return getattr(self, key) != _FieldStatus.UNSET
 
     def is_fully_specified(self) -> bool:
         r"""Whether the :obj:`TensorAttr` has no unset fields."""
@@ -137,7 +139,7 @@ class AttrView(CastMixin):
         # Find the first attribute name that is UNSET:
         attr_name: Optional[str] = None
         for field in out._attr.__dataclass_fields__:
-            if getattr(out._attr, field) == _field_status.UNSET:
+            if getattr(out._attr, field) == _FieldStatus.UNSET:
                 attr_name = field
                 break
 
--- a/torch_geometric/distributed/local_feature_store.py
+++ b/torch_geometric/distributed/local_feature_store.py
@@ -8,7 +8,7 @@ import torch
 from torch import Tensor
 
 from torch_geometric.data import FeatureStore, TensorAttr
-from torch_geometric.data.feature_store import _field_status
+from torch_geometric.data.feature_store import _FieldStatus
 from torch_geometric.typing import EdgeType, NodeType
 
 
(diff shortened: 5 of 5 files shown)


torch_geometric/distributed/local_graph_store.py
```diff
@@ -18,6 +18,13 @@ def __init__(self):
         self._edge_attr: Dict[Tuple, EdgeAttr] = {}
         self._edge_id: Dict[Tuple, Tensor] = {}
 
+        # for partition info related to graph
```
mananshah99, line 21: Would recommend documenting `node_pb`, `edge_pb`, and `meta`. Meta is a bit concerning since it seems to be a Dict[Any, Any] with no additional information.

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+
+    def set_edge_pb(
+            self, edge_pb: Union[torch.Tensor, Dict[NodeType,
+                                                    torch.Tensor]]) -> bool:
+        self.edge_pb = edge_pb
+        return True
+
+    def set_partition_meta(self, partition_meta: Dict) -> bool:
+        self.meta = partition_meta
+        return True
```
mananshah99, line 85: * Why do these functions return `bool`?
* We should either define setters and have private instance variables or just use the public instance variables. No need to have setters on public instance variables.

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+
+    def get_partition_ids_from_nids(self, ids: torch.Tensor,
+                                    ntype: Optional[NodeType] = None):
+        # Get the local partition ids of node ids with a specific node type.
```
mananshah99, line 89: nit, move this to a function-level doc, with `r"""<doc>"""`

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+                                                    torch.Tensor]]) -> bool:
+        self.edge_pb = edge_pb
+        return True
+
+    def set_partition_meta(self, partition_meta: Dict) -> bool:
+        self.meta = partition_meta
+        return True
+
+    def get_partition_ids_from_nids(self, ids: torch.Tensor,
+                                    ntype: Optional[NodeType] = None):
```
mananshah99, line 88:
```suggestion
    def get_partition_ids_from_nids(self, ids: torch.Tensor,
                                    node_type: Optional[NodeType] = None):
```
nit

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+        # Get the local partition ids of node ids with a specific node type.
+        if (self.meta["is_hetero"]):
+            assert ntype is not None
+            return self.node_pb[ntype][ids]
+        return self.node_pb[ids]
+
+    def get_partition_ids_from_eids(self, eids: torch.Tensor,
+                                    etype: Optional[EdgeType] = None):
+        # Get the partition ids of edge ids with a specific edge type.
+        if (self.meta["is_hetero"]):
```
mananshah99, line 98:
```suggestion
        if self.meta["is_hetero"]:
```

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+        return True
+
+    def set_partition_meta(self, partition_meta: Dict) -> bool:
+        self.meta = partition_meta
+        return True
+
+    def get_partition_ids_from_nids(self, ids: torch.Tensor,
+                                    ntype: Optional[NodeType] = None):
+        # Get the local partition ids of node ids with a specific node type.
+        if (self.meta["is_hetero"]):
```
mananshah99, line 90:
```suggestion
        if self.meta["is_hetero"]:
```

torch_geometric/distributed/local_graph_store.py
```diff
@@ -51,6 +58,48 @@ def _remove_edge_index(self, edge_attr: EdgeAttr) -> bool:
...
+            assert ntype is not None
+            return self.node_pb[ntype][ids]
+        return self.node_pb[ids]
+
+    def get_partition_ids_from_eids(self, eids: torch.Tensor,
+                                    etype: Optional[EdgeType] = None):
```
mananshah99, line 96:
```suggestion
                                    edge_type: Optional[EdgeType] = None):
```

mananshah99 (commented):
Mostly looks good. Thanks for unifying the data into the `LocalGraphStore` class, I'm optimistic this will be cleaner going forward.

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: the current style is aligned with _edge_attr and _edge_id and all to be set later by set_xxx(). The bool return is also the same as from the original set_edge_index().

torch_geometric/distributed/dist_context.py
mananshah99, line 1: Can we rebase this PR on the other one to avoid this duplicated diff?

torch_geometric/distributed/local_graph_store.py
mananshah99: I don't think we need that here. These are just instance variables, and don't need setters and getters as they are not private and these setters/getters do not perform any additional mutation. Would be good to clean that up.

torch_geometric/distributed/local_graph_store.py
```diff
@@ -126,15 +184,18 @@ def from_partition(cls, root: str, pid: int) -> 'LocalGraphStore':
...
             graph_store.put_edge_id(graph_data['edge_id'], **attr)
 
         if meta['is_hetero']:
             for edge_type, data in graph_data.items():
                 attr = dict(edge_type=edge_type, layout='coo',
                             size=data['size'])
-                graph_store.put_edge_index((data['row'], data['col']), **attr)
+                graph_store.put_edge_index(
+                    torch.stack((data['row'], data['col']), dim=0), **attr)
+                #graph_store.put_edge_index((data['row'], data['col']), **attr)
```
mananshah99, line 198:
```suggestion
```

torch_geometric/distributed/local_graph_store.py
```diff
@@ -126,15 +184,18 @@ def from_partition(cls, root: str, pid: int) -> 'LocalGraphStore':
 
         if not meta['is_hetero']:
             attr = dict(edge_type=None, layout='coo', size=graph_data['size'])
-            graph_store.put_edge_index((graph_data['row'], graph_data['col']),
-                                       **attr)
+            graph_store.put_edge_index(
+                torch.stack((graph_data['row'], graph_data['col']), dim=0),
+                **attr)
```
mananshah99, line 189: Why are we stacking here? put_edge_index expects a tuple. Same below.

mananshah99 (commented):
Mostly LGTM, apart from these few comments.

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: The reason :  edge_index (Tuple[torch.Tensor, torch.Tensor]) defined here can not pass through mp.spawn due to Tuple and used with stack to convert into tensor which can pass through mp.spawn.  We need send PartitionGraphStore/.. into subprocess.

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: removed

torch_geometric/distributed/dist_context.py
ZhengHongming888: sure will do!

torch_geometric/distributed/local_graph_store.py
```diff
@@ -18,10 +18,38 @@ def __init__(self):
         self._edge_attr: Dict[Tuple, EdgeAttr] = {}
         self._edge_id: Dict[Tuple, Tensor] = {}
 
+        self.num_partitions: int = 1
+        self.partition_idx: int = 0
+        self.node_pb: Union[torch.Tensor, Dict[NodeType, torch.Tensor]] = None
```
rusty1s, line 23:
```suggestion
        self.node_pb: Union[Tensor, Dict[NodeType, Tensor]] = None
```

torch_geometric/distributed/local_graph_store.py
```diff
@@ -18,10 +18,38 @@ def __init__(self):
...
+        r""" node_partition_book:  mapping between node ids and partition idx """
```
rusty1s, line 24: Move the comment up. Also use `#` instead of `"""`.

torch_geometric/distributed/local_graph_store.py
```diff
@@ -18,10 +18,38 @@ def __init__(self):
...
+        self.edge_pb: Union[torch.Tensor, Dict[EdgeType, torch.Tensor]] = None
+        r""" edge_partition_book:  mapping between edge ids and partition idx """
+        self.meta: Optional[Dict[Any, Any]] = None
+        r""" meta information related to partition and graph store info """
+        self.labels: Union[torch.Tensor, Dict[EdgeType, torch.Tensor]] = None
```
rusty1s, line 29: What is `self.labels` and why is it related to edge type information?

torch_geometric/distributed/local_graph_store.py
rusty1s: First time I hear that tuples are a problem with `mp.spawn`. Are you sure this info is correct?

ZhengHongming888: sure. i will check again and let my colleague check together. Also in from torch_geometric.testing import get_random_edge_index,  #L22 edge_index = torch.stack([row, col], dim=0) also used stack to convert into tensor instead of EdgeTensorTuple = Tuple[Tensor, Tensor]

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: normally the dataset will also have the labels y and when we do partition we also can move this part into partition file but also we can put outside. if outside for label we need input this label as argument in NeighborLoader and also for the following sub-funcs like dist_neighbor_sampler, etc

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

torch_geometric/distributed/local_graph_store.py
ZhengHongming888: done

PR #7671, opened 2023-06-30
Add `dist_context`/RPC for distributed training

ZhengHongming888:
This code belongs to the part of the whole distributed training for PyG.

This class will do

1. dist_context is to setup the distributed mode like worker mode or other mode and also setup the information for distributed context like role, rank, group_name, world_size,etc.
2. based on pytorch rpc.api to setup the api to check is_rpc_initialized and wrapper func, init_rpc(), RpcCall(), RpcRouter, rpc_async_request() and rpc_sync_request(), etc

These basic rpc functionality will be used in later feature lookup function after node sampling. 

Any comments please let us know. thanks.

PR #7678, opened 2023-07-02
Add distributed feature info for distributed training

ZhengHongming888:

This PR originally designed for the DistFeature class and now merged with LocalFeatureStore -

1) Add partition/rpc info into LocalFeatureStore like num_partition, partition_idx, feature_pb (feature_partitionbook), partition_meta, RpcRouter, etc
2) Add one new class (RpcCallFeatureLookup) to do real remote rpc feature_lookup work
3) Add one api ( .lookup_features() ) to do feature lookup in local node and remote nodes based on sampled global node ids/edge ids based on torch rpc apis
4) one unit test to verify the function of local/remote feature lookup under .test/distributed/. folder

Now we combined the local feature store and distributed feature properties (partition info and rpc remote access apis) into one FeatureStore. later on we will change the class name from LocalFeatureStore into PartitionFeatureStore with another PR.

Any comments please let us know. thanks.

### DIFF
--- a/torch_geometric/distributed/dist_context.py
+++ b/torch_geometric/distributed/dist_context.py
@@ -1,0 +1,51 @@
+
+from enum import Enum
+from typing import Optional
+
+
+class DistRole(Enum):
+    WORKER = 1  # for worker mode
+
+_WORKER_GROUP_NAME = '_def_worker'
+
+class DistContext(object):
+    r""" Context info for distributed Info in the current process."""
+    def __init__(self,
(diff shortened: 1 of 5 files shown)


torch_geometric/distributed/dist_context.py
mananshah99, line 1: Can we rebase this PR on https://github.com/pyg-team/pytorch_geometric/pull/7671 to avoid the duplicated diff?

PR #7689, opened 2023-07-04
Update get_mesh_laplacian.py

mzamini92:
If you have multiple meshes to compute the Laplacian for, you can consider batch processing them. This involves stacking the `pos` and `face` tensors along the batch dimension and performing computations in a batched manner. This can provide significant performance improvements, especially when dealing with a large number of small meshes. You can consider using inplace operations (`tensor.pow_()`, `tensor.fill_()`) instead of creating new tensors.

### DIFF
--- a/torch_geometric/utils/get_mesh_laplacian.py
+++ b/torch_geometric/utils/get_mesh_laplacian.py
@@ -2,9 +2,7 @@ from typing import Optional, Tuple
 
 import torch
 from torch import Tensor
-
-from torch_geometric.utils import add_self_loops, scatter, to_undirected
-
+from torch_geometric.utils import to_undirected, add_self_loops
 
 def get_mesh_laplacian(
     pos: Tensor,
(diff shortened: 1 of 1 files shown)


torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -52,58 +52,66 @@ def get_mesh_laplacian(
     assert pos.size(1) == 3 and face.size(0) == 3
 
     num_nodes = pos.shape[0]
+    batch_size = pos.shape[0] // 3
```
rusty1s, line 55: Can you clarify this for me? Why is `batch_size` the number of nodes divided by 3?

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -52,58 +52,66 @@ def get_mesh_laplacian(
...
-    # Face to edge:
-    cot_index = torch.cat([face[:2], face[1:], face[::2]], dim=1)
+    cot_index = torch.cat([face[:, :2], face[:, 1:], face[:, ::2]], dim=1)
     cot_index, cot_weight = to_undirected(cot_index, cot_weight)
 
-    # Compute the diagonal part:
-    cot_deg = scatter(cot_weight, cot_index[0], 0, num_nodes, reduce='sum')
+    cot_deg = torch.zeros((num_nodes * batch_size), dtype=cot_weight.dtype,
+                          device=cot_weight.device)
+    scatter.add_(cot_weight, cot_index[0], out=cot_deg,
```
rusty1s, line 76: This looks wrong to me.

torch_geometric/utils/get_mesh_laplacian.py
```diff
@@ -52,58 +52,66 @@ def get_mesh_laplacian(
...
+                               device=area_weight.device)
+        scatter.add_(area_weight, area_index[0], out=area_deg,
+                     dim_size=num_nodes * batch_size, reduce='sum')
 
         if normalization == 'sym':
             area_deg_inv_sqrt = area_deg.pow_(-0.5)
             area_deg_inv_sqrt[area_deg_inv_sqrt == float('inf')] = 0.0
-            edge_weight = (area_deg_inv_sqrt[edge_index[0]] * edge_weight *
-                           area_deg_inv_sqrt[edge_index[1]])
+            edge_weight.mul_(area_deg_inv_sqrt[edge_index[0]])
```
rusty1s, line 110: I know you can make an argument for in-place ops. For readability, I prefer the out-of-place commands, which might also support backprob to the input. If you desperately need the in-place ops here without running out of memory, we can change it, but I prefer not to.

PR #7701, opened 2023-07-06
Mini-Batch support in `GCNNorm`

mzamini92:
@rusty1s I was thinking of improving the GCN. the code snippet I provided supports both batch processing and sparse operations. It handles both Data and Batch objects, allowing you to apply the transformation to single graphs as well as batches of graphs. what do you think? Does it seem correct to you?

### DIFF
--- a/torch_geometric/transforms/gcn_norm.py
+++ b/torch_geometric/transforms/gcn_norm.py
@@ -1,7 +1,10 @@
-import torch_geometric
-from torch_geometric.data import Data
+from typing import Union
+
+import torch
+from torch_geometric.data import Batch, Data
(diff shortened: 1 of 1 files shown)


rusty1s (commented):
I don't see any issue with the current `GCNNorm` transformation. This should already support both `Data` and `Batch`.

mzamini92:
@rusty1s but this seems more efficient to me. still not sure.

rusty1s:
I don't see how this would be more efficient. Can you clarify? For `edge_index` code path, nothing changes, and for `SparseTensor` code path, you fallback to the less efficient `edge_index` code path via conversion. I am closing this PR, but feel free to re-open if you have any strong opinions.

PR #7703, opened 2023-07-06
Modify `dtype` in `ogbn-products` GraphSAGE example

puririshi98:


rusty1s (approved):
Not totally sure why this is an issue, but shouldn't hurt.

PR #7710, opened 2023-07-07
Add logging to `LocalGraphStore` tests

mzamini92:
The logging module is used to create a logger that writes log messages to both a file `(test.log)` and the console. Within each test function, `logger.info()` is used to log information about the assertions being made.

### DIFF
--- a/test/distributed/test_local_graph_store.py
+++ b/test/distributed/test_local_graph_store.py
@@ -1,18 +1,16 @@
-import torch
 import logging
+
+import torch
+
 from torch_geometric.distributed import LocalGraphStore
 from torch_geometric.testing import get_random_edge_index
 
 
 def setup_logging():
     logging.basicConfig(
-        level=logging.INFO,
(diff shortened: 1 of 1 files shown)


rusty1s (commented):
Can you clarify this PR for me? Why is it necessary to log anything in tests?

mzamini92:
I believe the log messages can serve as documentation for the test results. They record important assertions and their outcomes. By keeping the log messages, we have a clear record of the expected behavior and can easily refer to the log later to understand the results of each test case. This documentation helps ensure that the tests continue to pass as the codebase evolves. While logging is not strictly necessary for the code, it adds transparency, traceability, and aids in the development, maintenance, and debugging of the test suite.
certainly, it is not necessary but it would be a good feature I believe.

mananshah99:
The assertions should be enough to define expected behavior; we can add additional comments if necessary to clarify their use.

PR #7717, opened 2023-07-10
`GraphMaskExplainer` code updation

fork123aniket:
This PR is related to the changes discussed in #7271. This PR removes all redundant loss functions and unessential module calling from the `GraphMask` implementation.


rusty1s (commented):
Something looks to be off with the Git diff here. Do you know why?

fork123aniket:
Hmm, don't know what exactly does cause this `Git diff` issue. Actually, I just made two changes: removing loss functions (Lines `271`, `275`, and `285`) and altering module calling (Line `23`). These are pretty much the changes I've made. After making these changes, I just raised this PR the same way I have always been practicing opening the pull requests. Besides, does this issue mean this PR doesn't show the `code differences` between the current code (`torch_geometric.contrib.explain.GraphMaskExplainer`) and the code associated with this PR??

Kh4L:
@rusty1s @fork123aniket 

Looks like the carriage return character was added ( ^M character) before the EOL ($), as we can see in vim (with `:set list` and  `:e ++ff=unix`):

before
```
import math$
from typing import List, Optional, Tuple, Union$
$
import numpy as np$
import torch$
import torch.nn.functional as F$
from torch import Tensor$
from torch.nn import LayerNorm, Linear, Parameter, ReLU, Sequential$
from tqdm import tqdm$
$
from torch_geometric.explain import Explanation$
from torch_geometric.explain.algorithm import ExplainerAlgorithm$
```

after
```
import math^M$
from typing import List, Optional, Tuple, Union^M$
^M$
import numpy as np^M$
import torch^M$
import torch.nn.functional as F^M$
from torch import Tensor^M$
from torch.nn import LayerNorm, Linear, Parameter, ReLU^M$
from tqdm import tqdm^M$
^M$
from torch_geometric.explain import Explanation^M$
from torch_geometric.explain.algorithm import ExplainerAlgorithm^M$
from torch_geometric.explain.config import MaskType, ModelMode, ModelTaskLevel^M$
^M$
```

fork123aniket:
@Kh4L, Thanks so much for pointing this error out!! 

1. Do you think recommitting the `graphmask_explainer.py` file to this PR might help resolve this automatic addition of the `carriage return character` (`^M`) error?? 

2. JFYI, since the code I committed to this PR doesn't include this character (`^M`), I am interested to know why and how this `^M` is automatically added to the `graphmask_explainer.py` code file. Do you have any clue about this??

akihironitta:
I just added `.gitattributes` to your PR, hoping that your Git (on Windows I assume?) will make sure Python files will always have `LF` line endings:

> `text eol=lf` Git will always convert line endings to `LF` on checkout. You should use this for files that must keep LF endings, even on Windows.

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings

fork123aniket:
Thanks so much, guys (@Kh4L, @akihironitta). Really appreciate your help!!! @rusty1s, please let me know if something is still left to address that could prevent this PR from being merged, I'll do the needful then... Thanks!!!

torch_geometric/contrib/explain/graphmask_explainer.py
```diff
@@ -5,23 +5,18 @@
...
-)
+from torch_geometric.explain.config import MaskType, ModelMode, ModelTaskLevel
 from torch_geometric.nn import MessagePassing
 
 
 def explain_message(self, out: Tensor, x_i: Tensor, x_j: Tensor) -> Tensor:
-    norm = Sequential(LayerNorm(out.size(-1)).to(out.device), ReLU())
-    basis_messages = norm(out)
+    basis_messages = torch.nn.functional.layer_norm(
+        out, [size for size in out.size()]).to(out.device).relu()
```
akihironitta, line 19: Is it on purpose that you changed `normalized_shape`?
https://pytorch.org/docs/2.0/generated/torch.nn.LayerNorm.html#torch.nn.LayerNorm

akihironitta (commented):
Other than the below comment, LGTM! 🚀

torch_geometric/contrib/explain/graphmask_explainer.py
fork123aniket: Yes. Actually, we (@rusty1s and I) didn't want to make unessential module calling (`ReLU` and `LayerNorm`) at every `explain_message` function call, so we did the above to make them inline, replaced `ReLU()` to `.relu()` and `torch.nn.LayerNorm` to `torch.nn.functional.layernorm`. More details on this can be found in #7271.

To properly put `torch.nn.functional.layernorm` to work, I just extracted the list of integers from the tensor (`out`) shape itself and pass this list to `normalized_shape` argument.

akihironitta: Thank you for providing the context :) I was just pointing out that `normalized_shape` to the layer norm is no longer the same, meaning that it'll gives a different result:
```diff
- normalized_shape=out.size(-1)
+ normalized_shape=[size for size in out.size()]
```

Should `normalized_shape` still be the following, no?
```
normalized_shape=out.size(-1)
```

fork123aniket: I tried setting the `normalized_shape` to `out.size(-1)` and it got me the following:

```
TypeError: layer_norm(): argument 'normalized_shape' (position 2) must be tuple of ints, not int
```

This evidently indicates the argument must either be set to the `Tuple` or `List` of `integers`. Now to make the code more interactive and sensible, I did the following and committed the change:

```
basis_messages = torch.nn.functional.layer_norm(out, out.size()).to(out.device).relu()
```

In essence, since `out.size()` itself outputs a `Tuple` of `integers`, I think this could be the best and essential way to code the `LayerNorm`. What do you think??

torch_geometric/contrib/explain/graphmask_explainer.py
```diff
@@ -5,23 +5,18 @@
...
-)
+from torch_geometric.explain.config import MaskType, ModelMode, ModelTaskLevel
 from torch_geometric.nn import MessagePassing
 
 
 def explain_message(self, out: Tensor, x_i: Tensor, x_j: Tensor) -> Tensor:
-    norm = Sequential(LayerNorm(out.size(-1)).to(out.device), ReLU())
-    basis_messages = norm(out)
+    basis_messages = torch.nn.functional.layer_norm(out, out.size()).to(
+        out.device).relu()
```
rusty1s, line 19:
```suggestion
    basis_messages = F.layer_norm(out, (out.size(-1), )).relu()
```

rusty1s (approved):

torch_geometric/contrib/explain/graphmask_explainer.py
rusty1s: I think it needs to be `(out.size(-1), )`.

PR #7725, opened 2023-07-11
Measure time per epoch in some examples

Kh4L:


Kh4L:
@rusty1s @puririshi98

rusty1s (commented):
Thanks for the PR. My only concern here is that we are also measuring GPU start up time. As such, I would prefer reporting per epoch times or something like median over average.

Kh4L:
> Thanks for the PR. My only concern here is that we are also measuring GPU start up time. As such, I would prefer reporting per epoch times or something like median over average.

You are right median does make more sense here, changed to median.

akihironitta (approved):

rusty1s (approved):

PR #7744, opened 2023-07-14
`NetworkX` heterogeneous graphs can be converted to `HeteroData` using `from_hetero_networkx`

### LINKED ISSUE #7340: Convert heterogeneous NetworkX graph to HeteroData without normalizing features across node types

I am working on converting a heterogeneous NetworkX graph to `HeteroData`. Here is an example that I would like to see working:

```python
nodes_types = torch.tensor([1, 2, 3])
nx_graph = nx.MultiGraph()
nx_graph.add_node(0, type=1, foo=4.97, bar=torch.tensor([9, 8]))
nx_graph.add_node(1, type=2, foo=4.97)
nx_graph.add_node(2, type=3)
pyg.utils.from_networkx(nx_graph, group_node_attrs=['type', 'foo', 'bar'])
```
which currently throws `raise ValueError('Not all nodes contain the same attributes')`.

I know that my graph is heterogeneous, but to convert from NetworkX to HeteroData, I need to manually normalize all of the node attributes. I would like to not have to write my own conversion layer.

### Alternatives

N/A

### Additional context

N/A

SimonPop:
In order to answer the following issue: https://github.com/pyg-team/pytorch_geometric/issues/7340 this change proposes the implementation of a `from_hetero_networkx` method in the `convert` module.

Most of the unit tests have been copied from the `from_networkx` version and adapted. 

Some implementation questions can be discussed such as:
- Should the `edge_type_attribute` parameter accept being set to None? (currently, it does, and in that case, it defines the  edge type as 'to')
- Should  `edge_type_attribute` values be tuples `(node_a_type, type, node_b_type)`? Should it also/only accept a single value `type` and infer `node_a_type` and `node_b_type`? 
- When specified for a non-directed graph, what should happen with non-symmetrical `edge_type_attribute` which is bound to be incorrect in at least one of the directions (e.g. if an edge defines its type as `('A', 'to', 'B')` and the graph is not directed, the B -> A edge will have an incorrect label). Currently, it raises an error by checking node type coherency with edge type labels.

### DIFF
--- a/test/utils/test_convert.py
+++ b/test/utils/test_convert.py
@@ -942,7 +942,7 @@ def test_from_hetero_networkx_graph_node_selection():
     import networkx as nx
 
     G = nx.DiGraph()
-    G.add_node(0, type="A")
+    G.add_node(0, type="A", x=0)
     G.add_node("node_1", type="B", y=1)
     G.add_node(2, type="B", x=1)
 
@@ -950,8 +950,7 @@ def test_from_hetero_networkx_graph_node_selection():
                                 edge_type_attribute=None,
                                 graph_attrs=['x', 'y'], nodes=[0, "node_1"])
 
-    assert len(data['A'].type) == 1
-    assert len(data['B'].type) == 1
+    assert len(data['A'].x) == 1
     assert len(data['B'].y) == 1
 
 
--- a/test/utils/test_convert.py
+++ b/test/utils/test_convert.py
@@ -658,7 +658,10 @@ def test_from_dgl_hetero_graph():
 
 @withPackage('networkx')
 def test_from_hetero_networkx():
-    edge_index = get_random_edge_index(5, 10, 20, coalesce=True)
+    author_paper_edge_index = get_random_edge_index(5, 10, 20, coalesce=True)
+    paper_author_edge_index = get_random_edge_index(10, 5, 10, coalesce=True)
+    author_instit_edge_index = torch.tensor([[0, 1], [2, 2]])
+    author_instit_edge_index_bis = torch.tensor([[0, 1], [1, 0]])
     graph_x = [torch.tensor([0, 1, 2])]
 
     data = HeteroData()
@@ -666,8 +669,22 @@ def test_from_hetero_networkx():
     data['graph_x'] = graph_x
     data['author'].x = torch.arange(5)
     data['paper'].x = torch.arange(10)
-    data['author', 'paper'].edge_index = edge_index
-    data['author', 'paper'].edge_attr = torch.arange(edge_index.size(1))
+    data['institution'].x = torch.arange(3)
+    data['author', 'paper'].edge_index = author_paper_edge_index
+    data['paper', 'author'].edge_index = paper_author_edge_index
+    data['author', 'affiliated_with',
+         'institution'].edge_index = author_instit_edge_index
+    data['author', 'affiliated_with_bis',
+         'institution'].edge_index = author_instit_edge_index_bis
+    data['author',
+         'paper'].edge_attr = torch.arange(author_paper_edge_index.size(1))
+    data['paper',
+         'author'].edge_attr = torch.arange(paper_author_edge_index.size(1))
+    data['author', 'affiliated_with', 'institution'].edge_attr = torch.arange(
+        author_instit_edge_index.size(1))
+    data['author', 'affiliated_with_bis',
+         'institution'].edge_attr = torch.arange(
+             author_instit_edge_index_bis.size(1))
 
     G = to_networkx(data, node_attrs=['x'], edge_attrs=['edge_attr'],
                     graph_attrs=['global_id', 'graph_x'])
@@ -678,10 +695,18 @@ def test_from_hetero_networkx():
     assert data['graph_x'].tolist() == [x.tolist() for x in graph_x]
     assert data['author'].x.tolist() == torch.arange(5).tolist()
     assert data['paper'].x.tolist() == torch.arange(10).tolist()
-    assert data[('author', 'to',
-                 'paper')].edge_index.tolist() == edge_index.tolist()
+    assert data[(
+        'author', 'to',
+        'paper')].edge_index.tolist() == author_paper_edge_index.tolist()
+    assert data[(
+        'paper', 'to',
+        'author')].edge_index.tolist() == paper_author_edge_index.tolist()
+    assert data[('author', 'affiliated_with', 'institution'
+                 )].edge_index.tolist() == author_instit_edge_index.tolist()
+    t = data[('author', 'affiliated_with_bis', 'institution')]
+    assert t.edge_index.tolist() == author_instit_edge_index_bis.tolist()
     assert data[('author', 'to', 'paper')].edge_attr.tolist() == torch.arange(
-        edge_index.size(1)).tolist()
+        author_paper_edge_index.size(1)).tolist()
 
 
 @withPackage('networkx')
@@ -803,7 +828,9 @@ def test_from_hetero_networkx_raise_missing_node_type_attribute():
     G.add_node(1)
     G.add_edge(0, 1)
 
-    with pytest.raises(KeyError) as _:
+    with pytest.raises(
+            KeyError,
+            match=r'Given node_type_attribute: .* missing from node .*') as _:
         from_hetero_networkx(G, node_type_attribute="type",
                              edge_type_attribute=None)
 
@@ -817,7 +844,9 @@ def test_from_hetero_networkx_raise_missing_edge_type_attribute():
     G.add_node(1, type='A')
     G.add_edge(0, 1)
 
-    with pytest.raises(KeyError) as _:
+    with pytest.raises(
+            KeyError,
+            match=r'Given edge_type_attribute: .* missing from edge .*') as _:
         from_hetero_networkx(G, node_type_attribute="type",
                              edge_type_attribute="type")
 
@@ -833,7 +862,9 @@ def test_from_hetero_networkx_raise_different_edge_attribute():
     G.add_edge(0, 1, a=1)
     G.add_edge(0, 2, b=2)
 
-    with pytest.raises(ValueError) as _:
+    with pytest.raises(
+            ValueError,
+            match='Not all edges contain the same attributes.') as _:
         from_hetero_networkx(G, node_type_attribute="type",
                              edge_type_attribute=None)
 
@@ -846,7 +877,9 @@ def test_from_hetero_networkx_raise_different_node_attribute():
     G.add_node(0, type='A', a=1)
     G.add_node(1, type='A', b=1)
 
-    with pytest.raises(ValueError) as _:
+    with pytest.raises(
+            ValueError,
+            match='Not all nodes contain the same attributes.') as _:
         from_hetero_networkx(G, node_type_attribute="type",
                              edge_type_attribute=None)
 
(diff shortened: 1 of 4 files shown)


SimonPop:
Hello,

I have been trying to understand why so many indirect coverage changes occur with this commit. 

I am not super familiar with codecov, and it is not obvious for me to see how my change could impact files in completely different modules.

If anyone has experience with that, I would be glad to know more about it so that I can correct this. 

Thanks :)

JosephDenman:
@SimonPop I just used your implementation in my application and it worked like a charm. The only improvements I suggest are to:

1.  tighten up the spelling/grammer in the documentation.
2. add a `str` conversion like so:
 ```
 node_type = str(node_data[node_type_attribute])
 ```
 to accomodate non-string node types.

SimonPop:
@JosephDenman 

Thanks a lot for your review. 

I have tried to update the documentation accordingly. It should be clearer now and better formatted.

I also took into account the non-string node types, as recommended.

torch_geometric/utils/convert.py
```diff
@@ -614,3 +614,178 @@ def from_dgl(
...
+    def get_node_attributes(G, nodes: list,
+                            expected_node_attrs: list = None) -> dict:
+        r"""Collects the attributes of a list of graph nodes in a dictionary.
+
+        Args:
+            G (networkx.Graph or networkx.DiGraph): A networkx graph.
+            nodes (list, optional): The list of nodes whose attributes are to
+                be collected. If set to :obj:`None`, all nodes of the graph
+                will be included. (default: :obj:`None`)
+            node_attrs (list, optional): The list of expected attributes to
```
wsad1, line 688: Lets expose this argument in `from_hetero_netrworkx`

torch_geometric/utils/convert.py
```diff
@@ -614,3 +614,178 @@ def from_dgl(
...
+        edge_list = list(G.edges(data=False))
+        global_edge_index = [edge_list[edge] for edge in group_edges]
+        group_edge_index = [(node_to_group_id[node_a],
+                             node_to_group_id[node_b])
+                            for node_a, node_b in global_edge_index]
+        hetero_data_dict[group_name]["edge_index"] = torch.tensor(
+            group_edge_index, dtype=torch.long).t().contiguous().view(2, -1)
+
+    for key, value in G.graph.items():
+        hetero_data_dict[str(key)] = value
```
wsad1, line 767: Add argument to specify graph attributes.

torch_geometric/utils/convert.py
```diff
@@ -614,3 +614,178 @@ def from_dgl(
...
+        group_name = '__'.join(group)
+        hetero_data_dict[group_name] = get_edge_attributes(
+            G, edge_indexes=group_edges)
+        edge_list = list(G.edges(data=False))
+        global_edge_index = [edge_list[edge] for edge in group_edges]
+        group_edge_index = [(node_to_group_id[node_a],
+                             node_to_group_id[node_b])
+                            for node_a, node_b in global_edge_index]
+        hetero_data_dict[group_name]["edge_index"] = torch.tensor(
+            group_edge_index, dtype=torch.long).t().contiguous().view(2, -1)
```
wsad1, line 764: The edge offsets need to be subtracted. 
In `to_networkx` and offset is added to the edge indicies , see [here](https://github.com/pyg-team/pytorch_geometric/blob/7ac4654db727df73d1443c5102020abe859e5eb6/torch_geometric/utils/convert.py#L167)

test/utils/test_convert.py
```diff
@@ -611,3 +612,231 @@ def test_from_dgl_hetero_graph():
...
+@withPackage('networkx')
+def test_from_hetero_networkx_works_with_non_string_types():
+    import networkx as nx
+
+    G = nx.Graph()
+    G.add_node(0, type=1, x=1)
+    G.add_node(1, type=False, x=1)
+
+    data = from_hetero_networkx(G, node_type_attribute="type",
+                                edge_type_attribute=None)
```
wsad1, line 839: Mention in the docs that the `node_type_attribute` will be cast to a string, and raise an error if we can't cast to string.

test/utils/test_convert.py
```diff
@@ -611,3 +612,231 @@ def test_from_dgl_hetero_graph():
...
+
+@withPackage('networkx')
+def test_from_hetero_networkx_raise_different_node_attribute():
+    import networkx as nx
+
+    G = nx.Graph()
+    G.add_node(0, type='A', a=1)
+    G.add_node(1, type='A', b=1)
+
+    with pytest.raises(ValueError) as _:
```
wsad1, line 807:
```suggestion
    with pytest.raises(ValueError, match = r''Not all nodes .* attributes.'):
```
Since `ValueError` is common and could be raised for other reasons, its important to check if the expected value error is raised. So please use `match` here and in other tests.

test/utils/test_convert.py
```diff
@@ -611,3 +612,231 @@ def test_from_dgl_hetero_graph():
...
+    G = nx.Graph()
+    G.add_node(0, type='A')
+    G.add_node(1, type='A')
+    G.add_node(2, type='A')
+    G.add_edge(0, 1, a=1)
+    G.add_edge(0, 2, b=2)
+
+    with pytest.raises(ValueError) as _:
+        from_hetero_networkx(G, node_type_attribute="type",
+                             edge_type_attribute=None)
```
wsad1, line 796:
```suggestion
    with pytest.raises(ValueError):
        from_hetero_networkx(G, node_type_attribute="type",
                             edge_type_attribute=None)
```
use `match` in `pytest.raises`.

test/utils/test_convert.py
```diff
@@ -611,3 +612,231 @@ def test_from_dgl_hetero_graph():
...
+def test_from_hetero_networkx():
+    edge_index = get_random_edge_index(5, 10, 20, coalesce=True)
+    graph_x = [torch.tensor([0, 1, 2])]
+
+    data = HeteroData()
+    data['global_id'] = 0
+    data['graph_x'] = graph_x
+    data['author'].x = torch.arange(5)
+    data['paper'].x = torch.arange(10)
+    data['author', 'paper'].edge_index = edge_index
```
wsad1, line 627: Add another edge type , to make sure the edge_index gets offset correctly.

wsad1 (commented):
Thanks a lot @SimonPop for building this.
Left some comments. Overall things are looking good.

test/utils/test_convert.py
SimonPop: Alright, I will be adding three more cases:
- Multi-edge between the same type of nodes, with a different edge type
- Opposite direction edge type (from paper to author for example)
- New edge type between author and another node type (e.g. institution)

torch_geometric/utils/convert.py
SimonPop: I think this one is already taken care of. The `node_to_group_id` variable stores an index "by type" (with offset subtracted). 

So for 
```python
G.add_node("node_1", type='A')
G.add_node("node_2", type='A')
G.add_node("node_3" type='B')
```

it should store `{"node_1": 0, "node_2": 1, "node_3": 0}`

Is that what you meant?

torch_geometric/utils/convert.py
SimonPop: Should I expose `nodes ` or `node_attrs`?

SimonPop:
Thanks, @wsad1 for your review! 

I either corrected the code directly or asked for precisions in the comments.

torch_geometric/utils/convert.py
SimonPop: I went for the `nodes` in my last commit.

wsad1: `node_attrs`, so that the user can specify which attributes a node has, similar to `from_networkx`.

SimonPop: Sure! But since HeteroData has different groups of nodes, what implementation do you think would be the best?

1. `node_attrs` is a dictionary {node_group --> attributes} where each type of node has its own list of attributes to stack.

```python
node_attrs = {
type_1: ['y', 'z'],
type_2: ['u', 'v'],
type_3: ['u', 'y']
}
``` 

2. `node_attrs` is a single set of attributes. For each group of nodes, we expect all of them to be present.

```python
node_attrs = {'u', 'v', 'w'} # All required for every node group
``` 

3.  `node_attrs` is a single set of attributes. For each group of nodes, we only stack the ones that are available without raising an error if some are missing.

```python
node_attrs = {'u', 'v', 'w'} # To get when possible for every node group
``` 

n°3 is semi-flexible: we can allow some groups of nodes to have different features on the condition that no two groups of different nodes have the same attribute names for different purposes (assuming we want one but not the other).

I think n°1 is probably the best since we can express any situation, but it is also more complex to use for the user.

What do you think?

SimonPop: In my latest commit, I implemented `attrs` to accommodate both node and edge attributes. 
It is a dictionary that maps one group to the attributes to concatenate. It looks like: 

```python
attrs = {
    "A": ["u", "v", "c"],
    "B": ["w", "z", "c"],
    ("A", "to", "A"): ["a"],
    ("B", "to", "B"): ["b"]
}
```

wsad1: Sorry for the later reply, but lets keep it simple and go for solution 2 because
1. It keeps the API similar to [to_netowrkx](https://pytorch-geometric.readthedocs.io/en/latest/modules/utils.html#torch_geometric.utils.to_networkx) where `node_attrs` and `edge_attrs` are `List[str]`, and all nodes need to have those attributes. 
2. I like the added flexibility of 1, but we should only add more flexiblity (and more complexity) if the end user really needs it. I don't think that most users need this flexibility. 

I see that you have already made some changes. Apologies for asking to change it again.

SimonPop: No worries, I knew that it was subject to change.
I'll make the update for solution n°2 then, thanks!

JosephDenman:
I noticed that if I call `from_hetero_networkx(self, node_type_attribute='type', edge_type_attribute='type', group_node_attrs=['attr'])`, for each node type of the resulting `HeteroData` there is an attribute `type` that is just a list of all the node types. I think it makes sense to not include `type` as an attribute in the resulting `HeteroData`.

SimonPop:
> I noticed that if I call `from_hetero_networkx(self, node_type_attribute='type', edge_type_attribute='type', group_node_attrs=['attr'])`, for each node type of the resulting `HeteroData` there is an attribute `type` that is just a list of the `type` attribute for all nodes of that type, so it's a list of `['A', 'A', 'A', 'A']` for a graph with four nodes with attribute `type='A'`. I think it makes sense to not include `type` as an attribute in the resulting `HeteroData`.

Hi @JosephDenman, 

Yes, you're right, these two attributes are far from necessary. I removed them from the `HeteroData` object in this last commit.

SimonPop:
Hi @wsad1, 

Do not hesitate to tell me if you find anything else missing from the PR

akihironitta (commented):
Thanks for working on this! I haven't looked into this deeper yet, but it looks well-tested and well-documented 🚀

boersmamarcel:
I could really use this feature, will this be merged anytime soon?

threedliteguy:
This would be very useful for demos of PyG

SimonPop:
Tried to refresh this branch a little since it still seems to get some attention :)

brainsqueeze:
I would really appreciate this becoming available. Can I help in anyway to unstick this?

PR #7748, opened 2023-07-15
Update antisymmetric_conv.py

mzamini92:
the bias parameter is optional, so it might not be initialized. I couldn't find where we check if the bias parameter is `None` or not. I guess calling `zeros` on `None` might raise an error in some cases.


EdisonLeeeee:
Thank you for spotting. Actually `zeros` is a no-op when the input is `None`. You can check it here:

https://github.com/pyg-team/pytorch_geometric/blob/9c79ce8570df194c15e4e6857c8b86d757dbbcb7/torch_geometric/nn/inits.py#L52

PR #7503, opened 2023-06-04
Update the documentation of `GraphMaskExplainer`

fork123aniket:
This PR is related to the points discussed in #7271. This PR is meant to add documentation to more functions and to add `datatype hints` to their arguments.

PR #7759, opened 2023-07-17
Add `.gitattributes` for consistent line endings in the repo

akihironitta:
No more weird diffs, e.g. https://github.com/pyg-team/pytorch_geometric/pull/7717#issuecomment-1635607258, https://github.com/pyg-team/pytorch_geometric/pull/7503#issuecomment-1575475212.

See also: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings


sugatoray:
Cool solution. 

Curious: doesn't git take care of this automatically? If I am not wrong, LF/CRLF is used only on the user's end while checking out the code and then committed as Unix style. 

What I would like to know is how is it that people are managing to use different line ending characters?

cc: @akihironitta

akihironitta:
> Curious: doesn't git take care of this automatically?

@sugatoray I'm not an expert on this, but afaik, if it wasn't for `.gitignore`, each user's git would allow them to commit code with any line endings.

> What I would like to know is how is it that people are managing to use different line ending characters?

I've assumed any editors respect the line endings of the original, but it seems like some editors don't as seen in the linked PRs :(

sugatoray:
> I've assumed any editors respect the line endings of the original, but it seems like some editors don't as seen in the linked PRs :(

Have you considered using editorconfig? 

- https://editorconfig.org/
- specifications: https://spec.editorconfig.org/

Example:

```editorconfig

# EditorConfig is awesome: https://EditorConfig.org

# top-most EditorConfig file
root = true

# Unix-style newlines with a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true

# Matches multiple files with brace expansion notation
# Set default charset
[*.{js,py}]
charset = utf-8

# 4 space indentation
[*.py]
indent_style = space
indent_size = 4

# Tab indentation (no size specified)
[Makefile]
indent_style = tab

# Indentation override for all JS under lib directory
[lib/**.js]
indent_style = space
indent_size = 2

# Matches the exact files either package.json or .travis.yml
[{package.json,.travis.yml}]
indent_style = space
indent_size = 2

```

akihironitta:
Thanks for the input @sugatoray! I still feel `.gitattributes` is safer as everyone uses git to push code and some users might use editors (such as `notepad.exe`?) that don't read editorconfig.

However, it seems to be widely used, so I guess it can be added, too :) Feel free to send a PR to hear thoughts from others!

sugatoray:
> Thanks for the input @sugatoray! I still feel `.gitattributes` is safer as everyone uses git to push code and some users might use editors (such as `notepad.exe`?) that don't read editorconfig.
> 
> 
> 
> However, it seems to be widely used, so I guess it can be added, too :) Feel free to send a PR to hear thoughts from others!

Yes, of course. I shared it only for the sake of discussion. I was not aware of the `.gitattribute` hack. So, thank you. It was fun learning that option. 🙏

rusty1s (approved):

PR #7767, opened 2023-07-19
Update imbalanced_sampler.py

mzamini92:
Can we compute the class weights once in the compute_class_weights method and take the dataset and input nodes as arguments? The computed class weights are then stored in the class_weights attribute of the ImbalancedSampler instance. When creating an instance of ImbalancedSampler, the cached class weights are used to initialize the WeightedRandomSampler superclass, eliminating the need to compute the class weights every time.

### DIFF
--- a/torch_geometric/loader/imbalanced_sampler.py
+++ b/torch_geometric/loader/imbalanced_sampler.py
@@ -68,13 +68,11 @@ class ImbalancedSampler(torch.utils.data.WeightedRandomSampler):
         num_samples: Optional[int] = None,
     ):
         # Compute class weights once and cache them for reuse
-        self.class_weights = self.compute_class_weights(dataset, input_nodes)
+        class_weights = self.compute_class_weights(dataset, input_nodes)
 
-        return super().__init__(self.class_weights, num_samples,
-                                replacement=True)
+        super().__init__(class_weights, num_samples, replacement=True)
 
-    def compute_class_weights(self, dataset: Union[Dataset, Data, List[Data],
-                                                   Tensor],
+    def compute_class_weights(self, dataset: Union[Dataset, Data, List[Data], Tensor],
                               input_nodes: Optional[Tensor] = None) -> Tensor:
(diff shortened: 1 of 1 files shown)


torch_geometric/loader/imbalanced_sampler.py
```diff
@@ -67,7 +67,15 @@ def __init__(
         input_nodes: Optional[Tensor] = None,
         num_samples: Optional[int] = None,
     ):
+        # Compute class weights once and cache them for reuse
+        self.class_weights = self.compute_class_weights(dataset, input_nodes)
```
akihironitta, line 71: I might be missing something, but the super class already stores the `weights`:
https://github.com/pytorch/pytorch/blob/v2.0.1/torch/utils/data/sampler.py#L197
so I guess we don't need to store them in `class_weights`, no?

mzamini92: I dont know if I correctly understood what you mean but although the super class also stores the weights internally, it requires the weights to be passed as an argument in its constructor. In this case, the ImbalancedSampler computes the class weights separately in the compute_class_weights method and stores them in the class_weights attribute. Then, it passes the class_weights to the super class constructor. This will allow the ImbalancedSampler class to compute the class weights once and cache them for reuse. Isn't it?

akihironitta: Here're what I see in this PR:
1. `ImbalanceSampler` computes weights
2. `ImbalanceSampler` stores the weights in its attribute `self.class_weights`
3. `ImbalanceSampler` passes the weights to `torch.utils.data.WeightedRandomSampler`
4. `torch.utils.data.WeightedRandomSampler` stores the weights to its attribute `self.weights`

I believe there will be duplicate attributes (i.e. `self.weights` and `self.class_weights`) storing the same tensor after the change in this PR.

mzamini92 (commented):
@akihironitta like this?

PR #7768, opened 2023-07-19
Update home.py

mzamini92:
previous code used the `os.path.expanduser()` function twice, once to expand the user's home directory and once to expand the `$PYG_HOME` environment variable. But this code only uses `os.path.expanduser()` once, to expand the `$PYG_HOME` environment variable which is faster.

### DIFF
--- a/torch_geometric/home.py
+++ b/torch_geometric/home.py
@@ -13,7 +13,8 @@ def get_home_dir() -> str:
     If :meth:`set_home_dir` is not called, the path is given by the environment
     variable :obj:`$PYG_HOME` which defaults to :obj:`"~/.cache/pyg"`.
     """
-    return _home_dir if _home_dir is not None else os.path.expanduser(os.path.expandvars("$PYG_HOME") or DEFAULT_CACHE_DIR)
+    return _home_dir if _home_dir is not None else os.path.expanduser(
+        os.path.expandvars("$PYG_HOME") or DEFAULT_CACHE_DIR)
 
 
 def set_home_dir(path: str):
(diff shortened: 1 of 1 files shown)


torch_geometric/home.py
```diff
@@ -14,12 +13,8 @@ def get_home_dir() -> str:
...
     variable :obj:`$PYG_HOME` which defaults to :obj:`"~/.cache/pyg"`.
     """
-    if _home_dir is not None:
-        return _home_dir
-
-    home_dir = os.getenv(ENV_PYG_HOME, DEFAULT_CACHE_DIR)
-    home_dir = osp.expanduser(home_dir)
-    return home_dir
+    return _home_dir if _home_dir is not None else os.path.expanduser(
+        os.path.expandvars("$PYG_HOME") or DEFAULT_CACHE_DIR)
```
rusty1s, line 17: Can you clarify for me where the old code uses `expanduser` twice? I am also not sure if `os.path.expandvars() or DEFAULT_CACHE_DIR` does what we want to do since `expandvars` will not return `None`.

mzamini92: Sure. In 
```
home_dir = os.getenv(ENV_PYG_HOME, DEFAULT_CACHE_DIR)
home_dir = osp.expanduser(home_dir)
```
I believe, it uses `os.path.expanduser()` twice. The first time, `os.path.expanduser()` is used to expand the user's home directory, which is stored in the `ENV_PYG_HOME` environment variable. The second time, `os.path.expanduser()` is used to expand the `DEFAULT_CACHE_DIR` string, which could contain a tilde character.
But about your second part of question, you are right. Using `os.path.expandvars()` with the or operator and `DEFAULT_CACHE_DIR` ensures that the function returns `DEFAULT_CACHE_DIR` when the environment variable `PYG_HOME` is not set or is empty. I am closing this PR.

Issue #7783, opened 2023-07-21
knn_graph example not working


dyutimoy:
### 🐛 Describe the bug

import torch
from torch_geometric.nn import knn_graph

x = torch.tensor([[-1.0, -1.0], [-1.0, 1.0], [1.0, -1.0], [1.0, 1.0]])
batch = torch.tensor([0, 0, 0, 0])
edge_index = knn_graph(x, k=2, batch=batch, loop=False)

Error message -----
 File "lib/python3.8/site-packages/torch_geometric/nn/pool/__init__.py", line 130, in knn_graph
   return torch_cluster.knn_graph(x, k, batch, loop, flow, cosine,
AttributeError: 'NoneType' object has no attribute 'knn_graph

### Environment

* PyG version: 2.3.1
* PyTorch version: 2.0.1+cu117
* OS: Ubuntu 20.04
* Python version: 3.8
* CUDA/cuDNN version: 12
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`):

EdisonLeeeee:
You are missing the `torch_cluster` dependency.

dyutimoy:
Thanks

PR #7785, opened 2023-07-21
Added a detailed warning if `torch_cluster` is not installed

EdisonLeeeee:
Gives users a more detailed warning message if `torch_cluster` is not installed when using a related API, see also #7783

### DIFF
--- a/torch_geometric/nn/pool/__init__.py
+++ b/torch_geometric/nn/pool/__init__.py
@@ -21,7 +21,12 @@ from .approx_knn import approx_knn, approx_knn_graph
 try:
     import torch_cluster
 except ImportError:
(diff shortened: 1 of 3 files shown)


torch_geometric/nn/pool/__init__.py
```diff
@@ -21,7 +21,12 @@
 try:
     import torch_cluster
 except ImportError:
-    torch_cluster = None
+
+    class PackageWarning:
+        def __getattr__(self, key):
```
rusty1s, line 26:
```suggestion
        def __getattr__(self, key: str):
```

rusty1s (approved):
I like it, thanks :)

PR #7790, opened 2023-07-24
Update graph_gps.py

JanineCHEN:
After a test run of the script, I suspect the use of `Embedding` here should be `Linear` instead.

### DIFF
--- a/examples/graph_gps.py
+++ b/examples/graph_gps.py
@@ -41,10 +41,10 @@ class GPS(torch.nn.Module):
                  attn_type: str, attn_kwargs: Dict[str, Any]):
         super().__init__()
 
-        self.node_emb = Embedding(28, channels - pe_dim)
+        self.node_emb = Linear(28, channels - pe_dim)
         self.pe_lin = Linear(20, pe_dim)
         self.pe_norm = BatchNorm1d(20)
-        self.edge_emb = Embedding(4, channels)
+        self.edge_emb = Linear(4, channels)
 
         self.convs = ModuleList()
         for _ in range(num_layers):


rusty1s:
Thanks. I think this is correct since both `x` and `edge_attr` are discrete values/indices, so the usage of `Embedding` here is preferred over `Linear`.

JanineCHEN:
> 
Hi, thanks a lot for your feedback.

Please kindly correct me if I am mistaken. Still, I understand that the `self.node_emb` and `self.edge_emb` functions here work as a pre-processing function to map node and edge features from their corresponding input dimension into the arbitrarily defined hidden dimension, `Linear` will work in case the input node and edge features are either integer or float values. For example, my customized dataset has float value node features, so `Embedding` triggers issues on my end, but `Linear` works perfectly fine.

rusty1s:
Please feel free to modify the model on your end if you are working with continuous values. I am also not sure how `Linear` works for you on discrete values. For example,
```python
lin = torch.nn.Linear(16, 32)
lin(torch.arange(16))
```
fails for me.

JanineCHEN:
> Please feel free to modify the model on your end if you are working with continuous values. I am also not sure how `Linear` works for you on discrete values. For example,
> 
> ```python
> lin = torch.nn.Linear(16, 32)
> lin(torch.arange(16))
> ```
> 
> fails for me.

Hi @rusty1s , thanks a lot for your feedback again. There might be a need to specify the required input feature format (Long or Float) as well as the expected output dimension. I understand that the `self.node_emb` and `self.edge_emb` functions are to map node and edge features from their corresponding input dimension into the arbitrarily defined hidden dimension, namely NxH -> NxH' (N is the number of nodes, H is the original dimension of node feature, H' is the embedded feature dimension). In such case, the behaviour of `Linear` might suit this requirement.
test

rusty1s:
Sorry, I am not sure I understand. In this case `lin(torch.arange(16).float())` does not return an output of shape `[16, 32]`, and as such is not correct. What do I miss?

JanineCHEN:
> Sorry, I am not sure I understand. In this case `lin(torch.arange(16).float())` does not return an output of shape `[16, 32]`, and as such is not correct. What do I miss?

Sorry for the confusion. I mean, suppose I have `data.x` (with dimension num_nodes * in_channel) as input, I would expect the output to keep the same dimension, namely `num_nodes * **out_channel**`, instead of `num_nodes * **in_channel * out_channel**`.

While using `Linear` only changes the number of channels while keeping the original dimension (in_channel -> out_channel), `Embedding` uplifts the input with an extra dimension (in_channel -> in_channel * out_channel, also according to [the Pytorch doc](https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html)), I'm not sure if this is the expected correct behaviour of `self.node_emb` in the original scripts. My apologies if I understand it wrongly.

akihironitta:
@JanineCHEN I think `Embedding` is preferred over `Linear` in this case because, for `node_emb`, the indices `0` and `1` are as "close" to each other as the indices `0` and `27` are (, meaning there's no meaning in the index value itself).

> I mean, suppose I have `data.x` (with dimension `num_nodes * in_channel`) as input, I would expect the output to keep the same dimension, namely `num_nodes * out_channel`, instead of `num_nodes * in_channel * out_channel`.

If I understand it correctly, the size of `data.x` is:
```
# each element is an index of its node type
[num_nodes, 1]
```
and the expected output size is:
```
# each element is an embedding of its node type
[num_nodes, channels - pe_dim]
```

I'm not too sure if I fully understand your point, but I hope this helps!

JanineCHEN:
> @JanineCHEN I think `Embedding` is preferred over `Linear` in this case because, for `node_emb`, the indices `0` and `1` are as "close" to each other as the indices `0` and `27` are (, meaning there's no meaning in the index value itself).
> 
> > I mean, suppose I have `data.x` (with dimension `num_nodes * in_channel`) as input, I would expect the output to keep the same dimension, namely `num_nodes * out_channel`, instead of `num_nodes * in_channel * out_channel`.
> 
> If I understand it correctly, the size of `data.x` is:
> 
> ```
> # each element is an index of its node type
> [num_nodes, 1]
> ```
> 
> and the expected output size is:
> 
> ```
> # each element is an embedding of its node type
> [num_nodes, channels - pe_dim]
> ```
> 
> I'm not too sure if I fully understand your point, but I hope this helps!

Hi @akihironitta , thanks a lot for your detailed explanation. I now understand the scenario you raised where the input `data.x` is indeed indices (labels or classes of nodes) and `in_channel` will always be 1 in this case; and in this case `Embedding` does fit the scenario.

My concern might be more related to the case when `data.x` is one-hot encoded labels (e.g., `0,1,0`, `1,0,0`), or features in float value (e.g. `11.000000, 15.887014, 37.780000`, `16.000000, 21.395072, 63.350000` in the `ENZYMES` dataset), in these cases, `Linear` might be more suitable. But this might be out of the scope of this particular example script using the `ZINC` dataset.

rusty1s:
Yes, in this case you would have to use a `Linear` layer for the first transformation, but we expect users to make this model change on their own depending on how their data look like. As such, I am closing this PR but feel free to re-open if you still have doubts.

PR #7800, opened 2023-07-25
Save memory in `TwoHop` transform

mzamini92:
Isn't it better to use coalesce operation in-place to save memory and avoid creating a new Data object? Also, can't we instead of resizing multiple times, we create a new tensor edge_attr2 with the correct size right from the beginning?

### DIFF
--- a/torch_geometric/transforms/two_hop.py
+++ b/torch_geometric/transforms/two_hop.py
@@ -1,14 +1,8 @@
 import torch
-
 from torch_geometric.data import Data
 from torch_geometric.data.datapipes import functional_transform
 from torch_geometric.transforms import BaseTransform
-from torch_geometric.utils import (
-    coalesce,
-    remove_self_loops,
-    to_edge_index,
-    to_torch_csr_tensor,
-)
+from torch_geometric.utils import coalesce, to_torch_csr_tensor, to_undirected
 
 
 @functional_transform('two_hop')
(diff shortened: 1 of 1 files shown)


torch_geometric/transforms/two_hop.py
```diff
@@ -19,18 +14,29 @@ def forward(self, data: Data) -> Data:
...
+            edge_index = to_undirected(edge_index, N)
+
+        # Convert to torch_csr_tensor and perform matrix multiplication
         adj = to_torch_csr_tensor(edge_index, size=(N, N))
-        edge_index2, _ = to_edge_index(adj @ adj)
-        edge_index2, _ = remove_self_loops(edge_index2)
+        edge_index2 = (adj @ adj).coalesce()
+
+        # Remove self-loops and duplicate edges
+        edge_index2 = edge_index2.remove_self_loops()
```
rusty1s, line 25: Thank you. Can you clarify to me why this speeds up things or is faster? I don't think there even exists a `remove_self_loops` function as part of PyTorch CSR tensors.

mzamini92: It first converts the edge_index tensor to a torch_csr_tensor, which is a more efficient representation for sparse matrices. Then, it performs matrix multiplication on the adj tensor, which gives the two-hop neighbors of each node. The resulting edge_index2 tensor is then filtered to remove self-loops and duplicate edges. Finally, the edge_index and edge_index2 tensors are combined and coalesced to create a new edge_index tensor. The reason why this code speeds things up is because it takes advantage of the efficient sparse matrix multiplication operations in PyTorch. Additionally, the coalesce function can further optimize the edge_index tensor by merging consecutive edges into a single edge. There is a remove_self_loops function as part of PyTorch CSR tensors. It can be used to remove self-loops from a torch_csr_tensor. am I wrong?

PR #7807, opened 2023-07-27
Add tests for temporal sampling in `LinkNeighborLoader`

akihironitta:


test/loader/test_link_neighbor_loader.py
```diff
@@ -204,6 +204,34 @@ def test_link_neighbor_loader_edge_label():
...
+        num_neighbors=[-1],
+        time_attr="edge_time",
+        edge_label_time=data.edge_time,
+        batch_size=batch_size,
+        shuffle=True,
+    )
+
+    for sample in loader:
+        assert sample.edge_label_index.size() == (2, batch_size)
+        assert sample.edge_label_time.size() == (batch_size, )
```
akihironitta, line 228: Before this PR, the size of `edge_label_time` was `(2*batch_size, )`.

test/loader/test_link_neighbor_loader.py
```diff
@@ -204,6 +204,33 @@ def test_link_neighbor_loader_edge_label():
...
+        assert sample.edge_label.size() == (batch_size, )
+        # TODO(akihironitta): Investigate the failure
+        # assert torch.all(sample.edge_time <= sample.edge_label_time)
```
akihironitta, line 231: Could someone also confirm this assertion is correct before I dig into this?

rusty1s: Yes, will do.

rusty1s (approved):
Thanks. I fixed the assert that was commented out. The issue was that `time` needs to be a node-level vector (for now), and that the endpoints of an edge needs to exist in time.

PR #7814, opened 2023-07-30
Let `coalesce()` and `sort_edge_index()` accept `edge_index` as a tuple of tensors

kaidic:
This PR allows `coalesce()` and `sort_edge_index()` to accept `edge_index` as a tensor or a tuple of tensors.

### DIFF
--- a/test/utils/test_coalesce.py
+++ b/test/utils/test_coalesce.py
@@ -26,7 +26,7 @@ def test_coalesce():
     assert out[1][0].tolist() == [[4], [3], [2], [6]]
     assert out[1][1].tolist() == [4, 3, 2, 6]
 
-    out = coalesce((edge_index[0,:], edge_index[1,:]))
+    out = coalesce((edge_index[0, :], edge_index[1, :]))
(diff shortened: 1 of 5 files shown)


EdisonLeeeee:
Thank you for the PR. However, I believe that users should ensure that the input format is correct (i.e., in PyG format), and we do not need to consider the possibility of diverse inputs.

Otherwise, we would have also to do more refactoring beyond the `coalesce` function.

rusty1s:
Hey @EdisonLeeeee, we have a use-case where we want to use PyG utilities but have data in `(row, col)` format and would like to avoid the (unnecessary) conversion to a contiguous tensor. That‘s why I suggested to support either case here. I am happy to create an issue.

I do agree that it might make our input arguments quite polluted, so for now I would see this as an internal feature that does not necessarily need to be exposed via type hints. Does that work for you?

EdisonLeeeee:
Oh, didn't know that before. Thank you for the clarification. That makes sense :)

rusty1s (approved):

PR #7823, opened 2023-07-31
Measure test coverage of exceptions

akihironitta:
Part of #6528.

IMO, exceptions are also part of the public API so we should measure the test coverage over them, but feel free to close this PR if you think otherwise ;)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,6 +161,7 @@ omit = [
 exclude_lines = [
     "pragma: no cover",
     "pass",
+    "raise NotImplementedError",
     "register_parameter",
     "warn",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -161,8 +161,6 @@ omit = [
 exclude_lines = [
     "pragma: no cover",
     "pass",
-    "raise",
-    "except",
```
rusty1s, line 165:
```suggestion
    "raise NotImplementedError",

rusty1s (approved):

rusty1s (approved):
I agree (except for `raise NotImplementedEror`s)

PR #7827, opened 2023-08-01
Added `HalfHop` graph upsampling augmentation

EdisonLeeeee:
This PR adds the `HalfHop` graph upsampling augmentation from the ICML 2023 paper [Half-Hop: A graph upsampling approach for slowing down message passing](https://openreview.net/forum?id=lXczFIwQkv).

According to the authors, `HalfHop` works by adding artificial slow nodes between neighbors to slow down message propagation, see below.

image

Reference: 
+ https://github.com/nerdslab/halfhop

cc: @mazabou


torch_geometric/transforms/half_hop.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError(
+                f'Ratio of half-hopped edges has to be between 0 and 1 '
+                f'(got {p}')
+        self.p = p
+        self.alpha = alpha
+
+    def forward(self, data: Data) -> Data:
+
+        assert data.edge_weight is None
+        assert data.edge_attr is None
```
wsad1, line 45:
```suggestion
        if data.edge_weight is None or data.edge_attr is None:
            raise ValueError("'HalfHop augmentation is not support if "
                                         "'data' has 'edge_weight' or 'edge_attr'")
```

wsad1: Mention in the docs that `data` with `edge_weight` or `edge_attr` is not supported.

torch_geometric/transforms/half_hop.py
```diff
@@ -0,0 +1,91 @@
...
+class HalfHop(BaseTransform):
+    r"""Graph upsampling augmentation from the
+    `"Half-Hop: A graph upsampling approach for slowing down
+    message passing" <https://openreview.net/forum?id=lXczFIwQkv>`_ paper.
+    The graph is augmented by adding artificial slow nodes between neighbors
+    to slow down message propagation. (functional name: :obj:`half_hop`).
+
+    Args:
+        alpha (float, optional): The interpolation factor for
+            the slow nodes. (default: :obj:`0.5`)
```
wsad1, line 18:
```suggestion
        alpha (float, optional): The interpolation factor
            used to compute slow node features
            :math:`x = \alpha*x_src + (1-\alpha)*x_dst` (default: :obj:`0.5`)
```

test/transforms/test_half_hop.py
```diff
@@ -0,0 +1,46 @@
+import torch
+
+from torch_geometric.data import Data
+from torch_geometric.transforms import HalfHop
+
+
+def test_half_hops():
+    edge_index = torch.tensor([[0, 1, 1, 2], [1, 0, 2, 1]])
```
wsad1, line 8: Can we add self loops here to make sure they remain intact.

wsad1 (commented):
@EdisonLeeeee the changes look good. Thanks for the work.

test/transforms/test_half_hop.py
EdisonLeeeee: Done. Thank you for your review!

RobinFrcd:
We need to take our time in this crazy world, so nice to finally see the slowdown of messages propagation in action!

Well done @mazabou @EdisonLeeeee :clinking_glasses:

PR #7830, opened 2023-08-02
Fix bug in `get_degree_histogram`

DamianSzwichtenberg:
This PR fixes the following error:
`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, xpu:0 and cpu!`
`deg_histogram` now inherits the device type from `edge_index`.

### DIFF
--- a/torch_geometric/nn/conv/pna_conv.py
+++ b/torch_geometric/nn/conv/pna_conv.py
@@ -196,7 +196,8 @@ class PNAConv(MessagePassing):
     def get_degree_histogram(loader: DataLoader) -> Tensor:
         r"""Returns the degree histogram to be used as input for the :obj:`deg`
         argument in :class:`PNAConv`."""
-        deg_histogram = torch.zeros(1, dtype=torch.long)
+        device = next(iter(loader)).edge_index.device
+        deg_histogram = torch.zeros(1, dtype=torch.long, device=device)
         for data in loader:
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/conv/pna_conv.py
```diff
@@ -196,7 +196,8 @@ def __repr__(self):
     def get_degree_histogram(loader: DataLoader) -> Tensor:
         r"""Returns the degree histogram to be used as input for the :obj:`deg`
         argument in :class:`PNAConv`."""
-        deg_histogram = torch.zeros(1, dtype=torch.long)
+        device = next(iter(loader)).edge_index.device
```
EdisonLeeeee, line 199: Can we obtain the device without iterating through the dataloader? This seems to introduce additional costs.

DamianSzwichtenberg: I tried different approaches, but this seems to be the most reliable, as we often lack some members in the loader.
As for performance, there is some cost, but when compared to the whole `get_degree_histogram` function it is a really small fraction. As an alternative, I can propose passing `device` as an argument. WDYT?

rusty1s (approved):

torch_geometric/nn/conv/pna_conv.py
EdisonLeeeee: Looks pretty good to me. Thank you @DamianSzwichtenberg and @rusty1s

Issue #7641, opened 2023-06-26
Sparse Transformer Implementation


RexYing:
### 🚀 The feature, motivation and pitch

Many existing sparse transformers have shown promise in efficient encoding of long sequences (text, images, time series ...). 
There have been many examples, including:
https://arxiv.org/abs/2007.14062
https://arxiv.org/abs/2004.05150
https://arxiv.org/abs/2009.14794
https://arxiv.org/abs/2210.11794

Sparse transformers can be formulated as a sparse attention graph pattern. Instead of the full quadratic self-attention between every pair of tokens, sparse transformers only compute attention.
Therefore, it may benefit from a message passing implementation similar to GNNs (such as GAT), to achieve high memory and runtime efficiency. 

To start with, we will create a general API framework for implementing sparse transformers in PyG, through a few popular and state-of-the-art sparse transformer architectures.
We will further integrate it with HuggingFace to allow the implementation to be able to run on long sequence of text / image / time series benchmarks, and potentially plug in to language models.

In future, such experimentation can be further extended sparse neural networks that are popular, such as [LoRA ](https://arxiv.org/abs/2106.09685) and [Butterfly ](https://arxiv.org/abs/2112.00029) networks.

PR #7840, opened 2023-08-03
Sparse Transformer Implementation

GaotangLi:
This PR addresses issue #7641.

The new commits in the sparse_transformer branch implement a versatile API framework for sparse transformers using efficient message-passing mechanisms in PyG. The framework's cornerstone is the computation of `sparse_messages`, which significantly accelerates attention computation. This framework can seamlessly integrate with the popular PyTorch Lightning package, and we provide a comprehensive example within this codebase.

However, to maintain code organization, we might need to relocate the current directory within the pytorch-geometric folder. Furthermore, we can consider extending the documentation to provide in-depth guidance. We can discuss these aspects further.


rusty1s (commented):
Thanks for the PR. As discussed
* lets move the module implementation to `torch_geometric/nn/transformer`
* add a corresponding test in `test/nn/transformer`
* Add the example in a follow-up PR :)

Also, it would be great if we can type hint all/most of the provided modules/methods.

motivationss:
The latest code change is submitted to PR #8043.

PR #7858, opened 2023-08-08
Add weighted sampling support in `random_walk`

Ali-A-Kiaei:
@rusty1s I have recently submitted a pull request for the node2vec algorithm, which implements a weighted random walk. I am eagerly anticipating the merging of this code into the main branch. 😊

### DIFF
--- a/torch_geometric/nn/models/node2vec.py
+++ b/torch_geometric/nn/models/node2vec.py
@@ -13,44 +13,53 @@ from torch_geometric.utils.sparse import index2ptr, ptr2index
 ####################################################################################### { define a weighted random walk that care the data.attr
 device = 'cuda' if torch.cuda.is_available() else 'cpu'
 
+
 def random_walk_weighted(rowptr, col, weight, batch, walk_length, p, q):
-    row = ptr2index(rowptr) 
+    row = ptr2index(rowptr)
     row_repeat = row.repeat(len(batch), 1)
-    col_repeat = col.repeat(len(batch), 1)    
+    col_repeat = col.repeat(len(batch), 1)
     pos_mats = batch.unsqueeze(dim=0)
     loc_mats = batch.unsqueeze(dim=0)
-    weights_tmp = weight.repeat(len(batch),1).clone().detach()
+    weights_tmp = weight.repeat(len(batch), 1).clone().detach()
     indices = batch.clone().detach()
     for i in range(walk_length):
         starts_with = rowptr[indices]
-        ends_with = rowptr[indices+1]
+        ends_with = rowptr[indices + 1]
 
         weights = (weights_tmp).clone().detach()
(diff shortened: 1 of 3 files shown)


torch_geometric/nn/models/node2vec.py
```diff
@@ -8,7 +8,59 @@
...
 from torch_geometric.utils import sort_edge_index
 from torch_geometric.utils.num_nodes import maybe_num_nodes
-from torch_geometric.utils.sparse import index2ptr
+from torch_geometric.utils.sparse import index2ptr, ptr2index
+
+####################################################################################### { define a weighted random walk that care the data.attr
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
+
+
+def random_walk_weighted(rowptr, col, weight, batch, walk_length, p, q):
```
wsad1, line 17: @rusty1s do we have support for weighted walk in `pyg-lib`? For some reason I rmember seeing it in some roadmap.

wsad1 (commented):
@Ali-A-Kiaei Thanks a lot for adding this. 
One key thing here is that we'd want to implement `random_walk_weighted` in `pyg-lib` using cpp, like we have for `random_walk`. I am checking with @rusty1s to see if there is some plan to implement this already.

Ali-A-Kiaei:
Thank you very much for your feedback.
Actually I have used your nice original random walk cpp in the weighted
random walk, by putting the probability of each existing edge equal to its
weight, in a batch. Therefore, when the user leave the *fastmode* to be
True ( as default), we don't have time consumption for transfering from
random walk to weighted random walk.
On the other hand, if user set *fastmode* to False, the scratch weighted
random walk in python is implemented.

Best regards.

torch_geometric/nn/models/node2vec.py
Ali-A-Kiaei: @wsad1 @rusty1s 
Thank you for your feedback. 
I have incorporated your original random walk C++ code into the weighted random walk by computing the probabilities of existing edges using their weights in each batch. 
By default, the **_fastmode_** is set to True, which means that the weighted walk has the same time complexity as the random walk. 

However, if the user sets **_fastmode_** to False, the following weighted random walk in Python is implemented. 
`def random_walk_weighted(rowptr, col, weight, batch, walk_length, p, q)`
This allows the user to customize the code of the weighted walk, but it results in slower execution time.

` model = Node2Vec(
        data.edge_index,
        embedding_dim=128,
        walk_length=20,
        context_size=10,
        edge_attr=data.edge_attr,  # Add data attribute to the input
        fast_mode=True,  # Fast Estimate or Exact Computation
        walks_per_node=10,
        num_negative_samples=1,
        p=1,
        q=1,
        sparse=True,
    ).to(device)`

lachlancahill:
Wondering if there is any update on this? Edge weights would be an incredible addition to Node2Vec.

NavodPeiris:
Yo Yo weighted graphs are important so can someone try to merge this!!

AkhilaVangara:
Echoing the comments, eagerly waiting for someone to merge this!!

PR #7860, opened 2023-08-08
Added an `ogbn-papers100M` example

puririshi98:
single gpu papers100m on 8xh100:   0.00027184496005873887 s/iter

### DIFF
--- a/examples/ogbn_papers_100m.py
+++ b/examples/ogbn_papers_100m.py
@@ -1,6 +1,5 @@
 import argparse
 import os
-import os.path as osp
 import time
 
 import torch
@@ -8,10 +7,8 @@ import torch.nn.functional as F
 from ogb.nodeproppred import PygNodePropPredDataset
 from torchmetrics import Accuracy
 
-import torch_geometric
 from torch_geometric.loader import NeighborLoader
 from torch_geometric.nn import GCNConv
-from torch_geometric.utils import to_undirected
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--hidden_channels', type=int, default=64)


examples/ogbn_papers_100m.py
```diff
@@ -0,0 +1,110 @@
...
+
+if torch.cuda.is_available():
+    device = torch.device('cuda')
+else:
+    device = torch.device('cpu')
+
+dataset = PygNodePropPredDataset(name='ogbn-papers100M')
+split_idx = dataset.get_idx_split()
+data = dataset[0]
+data.y = data.y.reshape(-1).to(torch.long)
```
rusty1s, line 33: I am pretty certain this explains your index issues. Can you drop this here (since `NaN` cannot be converted to `long`), and just do the conversion inside each mini-batch?

examples/ogbn_papers_100m.py
```diff
@@ -0,0 +1,113 @@
...
+                             batch_size=batch_size, num_workers=pyg_num_work())
+test_loader = NeighborLoader(data, num_neighbors=[args.fan_out, args.fan_out],
+                             input_nodes=split_idx['test'],
+                             batch_size=batch_size, num_workers=pyg_num_work())
+eval_steps = 100
+acc = Accuracy(task="multiclass", num_classes=dataset.num_classes).to(device)
+for epoch in range(args.epochs):
+    for i, batch in enumerate(train_loader):
+        if i >= 10:
+            start = time.time()
```
akihironitta, line 82: nit: do we need this?
```suggestion
```

akihironitta (approved):

examples/ogbn_papers_100m.py
puririshi98: yes, when timing we usually need to allow for some warmup iterations to get accurate average iteration times

akihironitta: I was thinking that because this is not a microbenchmark, it'd still be reasonable to measure the overall time incl. warmups, but I'm fine with either way :)

PR #7867, opened 2023-08-10
Add smiles string to `QM9` dataset

hatemhelal:
The smiles string is useful for many molecular GNNs that try and learn the mapping between the SMILES representation and molecular properties.  This change relies on the dataset being pre-processed with RDKit.


torch_geometric/datasets/qm9.py
```diff
@@ -288,9 +288,11 @@ def process(self):
 
             y = target[i].unsqueeze(0)
             name = mol.GetProp('_Name')
+            smiles = Chem.MolToSmiles(mol, isomericSmiles=True)
 
             data = Data(x=x, z=z, pos=pos, edge_index=edge_index,
-                        edge_attr=edge_attr, y=y, name=name, idx=i)
+                        smiles=smiles, edge_attr=edge_attr, y=y, name=name,
+                        idx=i)
```
rusty1s, line 295:
```suggestion
                        idx=i,)
```

rusty1s (approved):

hatemhelal:
@rusty1s thanks for the speedy review as ever!

Would it be possible to also update the pre-processed dataset used as a fallback when RDKit is not installed?

Happy to help with that if I can.

rusty1s:
> Would it be possible to also update the pre-processed dataset used as a fallback when RDKit is not installed?

Yes, I can take care of that.

PR #7888, opened 2023-08-16
Enable sorted recommendations in `LightGCN` model

daviddavo:
The main motivation behind this is to be able to calculate the precision at 5, 10, 20 without making multiple calls to `model.recommend`


rusty1s:
Thanks for the PR. Since `topk` returns sorted outputs by default, I think we are unblocked for this already. WDYT?

daviddavo:
Sorry, I must have misread the [documentation](https://pytorch.org/docs/stable/generated/torch.topk.html), I thought the default option was `False`. I'll change it and push again. Then it should be ready to merge.

daviddavo:
LGTM

rusty1s (approved):

PR #7894, opened 2023-08-16
[Documentation] Multi-GPU training in Vanilla PyTorch Tutorial

puririshi98:
addresses: https://github.com/pyg-team/pytorch_geometric/issues/7893


rusty1s:
Thanks @puririshi98. Can we do the tutorial on node/graph classification? I think introducing the LP model here is too complex.

puririshi98:
> Thanks @puririshi98. Can we do the tutorial on node/graph classification? I think introducing the LP model here is too complex.

oh yup, just saw this, will update accordingly

puririshi98:
@rusty1s ready for review

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+------------------
+
+To start, we can take a look at the `distributed sampling <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/multi_gpu/distributed_sampling.py>`__ example from :pyg:`PyG`.
+This example shows how to use train a :class:`~torch_geometric.nn.models.GraphSAGE` GNN model on the :class:`~torch_geometric.datasets.Reddit` dataset.
+This example uses the :class:`~torch_geometric.loader.NeighborLoader` with :class:`torch.nn.parallel.DistributedDataParallel` to scale-up training across all available GPU's on your machine.
+
+Defining our GNN
+~~~~~~~~~~~~~~~~
+
+.. code-block:: python
```
rusty1s, line 18: Can we use the `GraphSAGE` model here from `torch_geometric.nn.models` directly? I think the focus of this tutorial shouldn't be on how to build your model.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+                if i < len(self.convs) - 1:
+                    x = x.relu_()
+                xs.append(x.cpu())
+                pbar.update(1)
+            x_all = torch.cat(xs, dim=0)
+
+        pbar.close()
+        return x_all
+
+Defining our Spawnable Runner
```
rusty1s, line 68: I think this section here needs a bit more explanation about how distributed training works in general. What is `rank`, what is `world_size`, etc?

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+       data = dataset[0]
+       data = data.to(rank, 'x', 'y')
+
+Notice we move to features and labels to GPU for faster feature fetching.
+
+Now we split training indices into :obj:`world_size` many chunks for each GPU:
+
+.. code-block:: python
+
+       train_idx = data.train_mask.nonzero(as_tuple=False).view(-1)
```
rusty1s, line 89: I think it is important to mention that `run` gets executed for each node, and so we generate a `NeighborLoader` instance from a reduced set of training indices on each node.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+
+.. code-block:: python
+
+       if rank == 0:
+           subgraph_loader = NeighborLoader(copy.copy(data), num_neighbors=[-1],
+                                            shuffle=False, **kwargs)
+           # No need to maintain these features during evaluation:
+           del subgraph_loader.data.x, subgraph_loader.data.y
+           # Add global node index information:
+           subgraph_loader.data.node_id = torch.arange(data.num_nodes)
```
rusty1s, line 107: I would drop this. Let's just define `val_loader` as `NeighborLoader(data, input_nodes=val_idx)`.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+
+Now that we have our data loaders defined initialize our model and wrap it in DistributedDataParallel
```
rusty1s, line 109: Some clarification of what that does.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+After each training epoch, we evaluate and report accuracies:
+
+.. code-block:: python
+
+        dist.barrier()
+
+        if rank == 0:
+            print(f'Epoch: {epoch:02d}, Loss: {loss:.4f}')
+
+        if rank == 0 and epoch % 5 == 0:  # We evaluate on a single GPU for now
```
rusty1s, line 142: I think it is important to mention here why we only perform this on `rank == 0`.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+Putting it all together, we spawn our runners for each GPU:
+
+.. code-block:: python
+
+   if __name__ == '__main__':
+       dataset = Reddit('../../data/Reddit')
+
+       world_size = torch.cuda.device_count()
+       print('Let\'s use', world_size, 'GPUs!')
+       mp.spawn(run, args=(world_size, dataset), nprocs=world_size, join=True)
```
rusty1s, line 166: I think this needs to be moved to the top where you define the `run()` method.

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+
+      dist.destroy_process_group()
+
+
+Putting it all together, we spawn our runners for each GPU:
+
+.. code-block:: python
+
+   if __name__ == '__main__':
+       dataset = Reddit('../../data/Reddit')
```
rusty1s, line 162: Some note on why we initialize `dataset` here and not in `run` (due to shared memory).

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -0,0 +1,323 @@
...
+
+       world_size = torch.cuda.device_count()
+       print('Let\'s use', world_size, 'GPUs!')
+       mp.spawn(run, args=(world_size, dataset), nprocs=world_size, join=True)
+
+Large Scale
```
rusty1s, line 168: What is the motivation for this section? It feels like we are just repeating ourselves, but swap out the dataset. Would prefer to drop and make the first part more detailed.

docs/source/tutorial/multi_gpu_vanilla.rst
puririshi98: apologies, i'm not sure what you mean by "this", could you help clarify for me?

puririshi98:
not sure why but currently getting Val Acc = 0

docs/source/tutorial/multi_gpu_vanilla.rst
rusty1s: I think we should start the tutorial with this boilerplate:
```python
def run(rank, world_size, dataset):
    ...
    
 if __name__ == '__main__':
     dataset = Reddit('../../data/Reddit')
     
     world_size = torch.cuda.device_count()
     print('Let\'s use', world_size, 'GPUs!')
     mp.spawn(run, args=(world_size, dataset), nprocs=world_size, join=True)
```
rather than introducing ` if __name__ == '__main__':` logic at the end. I think this makes the overall flow and understanding much clearer.

puririshi98: ohh sounds good, i get it now :). agreed thats much better, will do

PR #7896, opened 2023-08-17
Improve inference loop on GPU devices

DamianSzwichtenberg:
In layer-wise inference loop, we perform computations as shown on the following pseudocode:
```
for layer in layers
    for batch in loader
        do inference per layer
```
In models that have more than one layer, we can benefit from caching batches during the
first walk through the data. This PR introduces `CachedLoader`, which transfers
batches to a pointed device and caches them. Additionally, an auxiliary function was provided,
`make_batches_cacheable`, which decorates `BasicGNN` instance with a custom inference
loop.

Selected performance results (gained on Intel PVC):
```
Speedup:
gcn[2L]+Reddit: 1.53x
gcn[3L]+Reddit: 1.69x
sage[2L]+Reddit: 1.55x
sage[3L]+Reddit: 2.02x
gcn[2L]+ogbn-products: 1.72x
gcn[3L]+ogbn-products: 2.11x
sage[2L]+ogbn-products: 1.83x
sage[3L]+ogbn-products: 2.44x
```
Caching mechanism did not have a significant impact on models with a single layer.

Drawbacks:
- User should be aware that caching mechanism requires additional device memory to be allocated.
In experiments, approximately 1GB was needed for the `Reddit` dataset.


torch_geometric/loader/cache.py
```diff
@@ -0,0 +1,55 @@
...
+        self.device = device
+        self.cache = []
+
+    def cache_batch(self, batch: Any) -> BatchType:
+        n_id = batch.n_id.to(self.device)
+        if hasattr(batch, 'adj_t'):
+            edge_index = batch.adj_t.to(self.device)
+        else:
+            edge_index = batch.edge_index.to(self.device)
+        batch = (n_id, edge_index, batch.batch_size)
```
rusty1s, line 33: Wondering why we don't cache the whole data object? Can we make this loader more general by just doing this?

torch_geometric/nn/models/basic_gnn.py
```diff
@@ -435,6 +435,49 @@ def __repr__(self) -> str:
...
+    inference loop, making it work with
+    :class:`torch_geometric.loader.CachedLoader`.
+    """
+    @torch.no_grad()
+    def inference_with_cache(
+        self,
+        loader: NeighborLoader,
+        device: Optional[Union[str, torch.device]] = None,
+        embedding_device: Union[str, torch.device] = 'cpu',
+        progress_bar: bool = False,
```
rusty1s, line 449: Wondering why we need this new function here. Can't we simply add a `cache` option to the `inference` call and do `CachedLoader(loader)` if set to `True`?

DamianSzwichtenberg: With caching the whole batch ([above comment](https://github.com/pyg-team/pytorch_geometric/pull/7896#discussion_r1297118891)), indeed your suggestion is a better option.

torch_geometric/loader/cache.py
DamianSzwichtenberg: Done.

torch_geometric/nn/models/basic_gnn.py
DamianSzwichtenberg: Done.

torch_geometric/loader/cache.py
DamianSzwichtenberg: @rusty1s Update, I have measured peak device memory usage:
```
Reddit:
Before: 8.36GB
Cache only n_id and edge_index: 9.4GB
Cache whole batch: 24.82GB

ogbn-products:
Before: 23.9GB
Cache only n_id and edge_index: 24.54GB
Cache whole batch: 62GB
```
As you can see, when caching the whole batch the peak memory usage jumps almost 3 times. No idea why memory usage is so high. Anyway, I have returned to the original implementation.

torch_geometric/nn/models/basic_gnn.py
DamianSzwichtenberg: Update, Please see [this comment](https://github.com/pyg-team/pytorch_geometric/pull/7896#discussion_r1297570176).

DamianSzwichtenberg:
@rusty1s Any new comments here?

rusty1s:
Will merge ASAP.

torch_geometric/loader/cache.py
```diff
@@ -0,0 +1,55 @@
...
+        device (torch.device): The device to load the data to.
+    """
+    def __init__(self, loader: DataLoader, device: torch.device):
+        self.n_batches = len(loader)
+        self.iter = iter(loader)
+        self.cache_filled = False
+        self.device = device
+        self.cache = []
+
+    def cache_batch(self, batch: Any) -> BatchType:
```
akihironitta, line 27: I'm curious about your thoughts around the interface.

As a user who has a different hardware requirement and dataset, I may want to modify which attributes to cache. I wonder what a clean interface looks like to allow this.

Do you think it's feasible to always return a data object from `__call__`? Currently, I don't think it's very intuitive for users because `next(iter(CachedLoader(loader)))` is a tuple of some of attributes while `next(iter(loader))` is a data object.

test/loader/test_cache.py
```diff
@@ -0,0 +1,29 @@
+import torch
+
+from torch_geometric.data import Data
+from torch_geometric.loader import CachedLoader, NeighborLoader
+from torch_geometric.testing import withPackage
+
+
+@withPackage('pyg_lib')
+def test_cached_loader():
```
akihironitta, line 9: Just for reminder, we should also have a test case for `make_batches_cacheable` once finalised.

torch_geometric/loader/cache.py
```diff
@@ -0,0 +1,55 @@
...
+from torch.utils.data import DataLoader
+
+BatchType = List[Tuple[Tensor, Tensor, int]]
+
+
+class CachedLoader:
+    r"""A caching loader intended to use in layer-wise inference
+    performed on GPU. During the first walk through the data
+    stored in :class:`torch.utils.data.DataLoader` batches are
+    transferred to GPU memory and cached.
```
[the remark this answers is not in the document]
DamianSzwichtenberg: Sure. Done.

torch_geometric/loader/cache.py
DamianSzwichtenberg: Thanks for raising really good points.

> As a user who has a different hardware requirement and dataset, I may want to modify which attributes to cache. I wonder what a clean interface looks like to allow this.

I changed the design a bit, now user can register attributes hooks via `register_attr_hook`. It requires a callable object that accepts batch and returns attribute.

> Do you think it's feasible to always return a data object from __call__? Currently, I don't think it's very intuitive for users because next(iter(CachedLoader(loader))) is a tuple of some of attributes while next(iter(loader)) is a data object.

Unfortunately, I don't think it is possible, it will require caching batches, which will move us back to a problem with increased memory usage (described [here](https://github.com/pyg-team/pytorch_geometric/pull/7896#discussion_r1297570176)). I have modified the documentation a bit, hopefully, it will clear some of user confusion.

test/loader/test_cache.py
DamianSzwichtenberg: Tests of `make_batches_cacheable` are located in `test_basic_gnn.py` file. 😉

DamianSzwichtenberg:
@akihironitta Could you PTAL?

PR #7905, opened 2023-08-19
Add `BreastInvasiveCarcinoma` dataset

Favourj-bit:
This is a dataset that was generated by integrating the breast cancer (BRCA TCGA) dataset from the cBioPortal (cbioportal.org) and a biological network for node connections from Pathway Commons (www.pathwaycommons.org). The dataset contains the gene features of each patient and the overall survival time (in months) of each patient, which are the labels.


torch_geometric/datasets/brca_tgca.py
```diff
@@ -0,0 +1,107 @@
...
+)
+
+
+class BrcaTcga(InMemoryDataset):
+    r"""The breast cancer (BRCA TCGA) dataset from `cBioPortal
+    <https://www.cbioportal.org>`_ and the biological network for node
+    connections from `Pathway Commons <https://www.pathwaycommons.org>`_.
+    The dataset contains the gene features of each patient in graph_features
+    and the overall survival time (in months) of each patient,
+    which are the labels.
```
akihironitta, line 22: Could you check the docstring change as I've just removed some redundant sentences? Also, it'd be ncie if you could describe what nodes and edges represent to help new users understand this dataset. (example: https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.datasets.KarateClub.html#torch_geometric.datasets.KarateClub)

akihironitta (commented):
Thank you for adding this dataset! I've just directly made some changes to your PR, but please have a look and feel free to revert them if necessary :)

Favourj-bit:
Hi @akihironitta 
Thank you for your review. I noticed that you removed the link that directs to the repository where it shows how the data was created and modelling with the data. I think this is an important information that should be included, and during my discussion with @rusty1s , he mentioned i could include this in the docstring.
I'll be adding this back, but apart from that, I'm okay with the other changes

rusty1s (approved):

PR #7907, opened 2023-08-20
Fix broken URLs in `HGBDataset`

frinkleko:
Old links are not available. According to changes in its [repo](https://github.com/THUDM/HGB), dataset can be found in [here](https://drive.google.com/drive/folders/10-pf2ADCjq_kpJKFHHLHxr_czNNCJ3aX) now.


rusty1s (approved):
thank you :)

PR #7921, opened 2023-08-23
Added `papers100m` multi-GPU example

puririshi98:
will use this in my tutorial: https://github.com/pyg-team/pytorch_geometric/pull/7894/

Average Training Iteration Time: 0.01475715453494085 s/iter


examples/multi_gpu/papers100m_multigpu.py
```diff
@@ -0,0 +1,136 @@
...
+
+from torch_geometric.loader import NeighborLoader
+from torch_geometric.nn import GCNConv
+
+
+def pyg_num_work():
+    num_work = None
+    if hasattr(os, "sched_getaffinity"):
+        try:
+            num_work = len(os.sched_getaffinity(0)) / 2
```
akihironitta, line 21: n00b question: I think I didn't comment this in the previous PR, but why was this divided by 2?

puririshi98: @rusty1s  gave me this hueristic a while back when I was setting up our GNN training infra inside NVIDIA and ive used it ever since. My hunch is that this is the most workers you can use at once without hurting perf for other processes due to busy cpu

examples/multi_gpu/papers100m_multigpu.py
```diff
@@ -0,0 +1,137 @@
...
```
akihironitta, line 21: This might open too many processes, leading to slowdown, as there will be `(pyg_num_work()+1)*torch.cuda.device_count()` processes on a single machine, but users are expected to adjust this anyway, so it may be fine as is.

akihironitta (approved):
Although I haven't run this example as I don't have access to a multi-gpu machine, it looks good to me. Shall we also follow the coding style in #7954 for consistency?

examples/multi_gpu/papers100m_multigpu.py
puririshi98: yeah it might need tuning, it seemed relatively performant for my set up

Issue #7795, opened 2023-07-24
Introduce PyG lifetime benchmarks


akihironitta:
### 🚀 The feature, motivation and pitch

Introduce PyG lifetime benchmarks with a dashboard 🚀 (using https://github.com/airspeed-velocity/asv)

## Motivation

There are quite a number of benchmarks in PyG already:
- micro-benchmarks in `test/` (See `git grep benchmark test/`)
- macro-benchmarks in `benchmark/`

However, since the results are currently not recorded, we cannot tell whether the performance of a feature has improved or regressed over time.

## TODO

See https://github.com/pyg-team/pytorch_geometric/tree/akihironitta/bench for progress for now.

**v0**
_Introduction of a minimal set of benchmarks on CPU/GPU_
- [x] Run benchmarks manually
- [x] Run benchmarks in a CI/CD pipeline on CPU (GitHub hosted runner)
- [ ] Run benchmarks in a CI/CD pipeline on a GPU (🔴 Blocked by availability of a GPU instance)
- [x] Add micro-benchmarks (e.g. small fwd/bwd operations)

**v1**
_More expensive benchmarks, more accelerators_
- [ ] Add macro-benchmarks (e.g. training and inference)
- [ ] Support other devices (mps, xpu, rocm, ...)

## List of benchmarks and metrics

- Model benchmarks:
  - peak memory usage
  - seconds per fwd/bwd
  - seconds per training iteration
  - final target metrics (loss/precision/rocauc/etc.) after a certain number of iterations
- Operation benchmarks:
  - peak memory usage
  - seconds per fwd/bwd
- ... (open to suggestions!)

### Additional context

- PyG benchmark results will be available at https://pyg-team.github.io/pytorch_geometric/
- PyTorch Lightning results are available at https://www.akihironitta.com/lightning-benchmarks-private/
  <img width="411" alt="image" src="https://github.com/pyg-team/pytorch_geometric/assets/20610905/8292b7eb-7b01-48f6-95ce-8c175af9d3b2">

I'll modify this issue description as I make progress. Also, please feel free to modify this :)

rusty1s:
@kiriti-kumo Can you help me with setting up GPU runners for nightly PyG runs?

akihironitta:
[...]

PR #7925, opened 2023-08-24
Add possibility to run training benchmarks on XPU device

DamianSzwichtenberg:
Add possibility to run training  benchmarks on XPU device.
Exemplary CMD:
`python training_benchmark.py --device xpu --datasets Reddit --models gcn --num-layers 2 --num-hidden-channels 128 --batch-sizes 8192`


akihironitta (approved):
LGTM! We could potentially consider adding xpu to benchmark PyG on via #7795 if we have access to an instance with a xpu device.

PR #7929, opened 2023-08-24
Propose & Integrate CF-Explainer

KintsugiArtist:
With @RexYing and @Cather-Chen, we propose an integration based on https://arxiv.org/abs/2102.03322 to create Counterfactual Explainability in PyG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added counterfactual explainability support in PyG in PyG ([#7929](https://github.com/pyg-team/pytorch_geometric/pull/7929))
 - Added `utils.ppr` for personalized PageRank computation ([#7917](https://github.com/pyg-team/pytorch_geometric/pull/7917))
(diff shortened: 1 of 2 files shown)


kano5266:
Hi there,

Thanks for the fantastic implementation of CF-GNNExplainer! It’s really well written and easy to follow. I did notice a few things that differ slightly from the original implementation or the paper, and I thought I’d point them out:

1. Indicator in line 142: The `loss_pred` indicator should be based on `y_hat` with a _binarized_ `edge_mask`, not `y_hat` with a _fractional_ `edge_mask`. You can see this in the [Your code](https://github.com/pyg-team/pytorch_geometric/blob/08d8f14548c2bb26b45f131645d773d9f27a194b/torch_geometric/explain/algorithm/cf_explainer.py#L142) and the [original code](https://github.com/a-lucic/cf-gnnexplainer/blob/cce5b2dc0dc95979a2083d95f87142678c12a6c0/src/cf_explanation/cf_explainer.py#L96C5-L96C77).

2. Input `edge_index`: The `edge_index` input should be the `l-hop` neighborhood around the node to be explained (the computation graph), rather than the full edge_index. This is mentioned in section 3.1 of the paper.

3. Naming: The correct name should be CF-GNNExplainer, not CF-Explainer. CF-GNNExplainer is just one of several counterfactual explainers, like RCExplainer and CLEAR.

Thanks again for your hard work on this!

PR #7966, opened 2023-09-01
Draft: adding `EdgeBankPredictor` from TGBN

puririshi98:
status: torch implementation is drastically slower than numpy.

> on 8x A100 machine:
>   bs=200
>     mine: 3.85it/s
>     numpy: 20.97it/s
> 
> 
>   bs 2048:
>     numpy:   4.41it/s
>     mine:   2.93s/it
> 
>   bs 8192:
>     mine:   11.40s/it
>     numpy 1.2 it/s
> 
> 
> 
> on single A100 machine bs200:
>   numpy: 20.00it/s
>   original (https://github.com/pyg-team/pytorch_geometric/pull/7966/commits/58d859e66fe1f7274787281adfb55a15aa92fe0d)  src/dst looping torch on CPU: 12.44it/s
>   original src/dst looping torch on CUDA: 6.55it/s
>   new edge_index no loop torch on CUDA: 4.22it/s

TODO: a FakeTemporalDataset to use for the unit test for test_edge_bank


test/nn/test_edgebank.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+# ==================
+# ==================
+# ==================
+@withPackage('tgb')
+def test_edge_bank_pred():
+    from tgb.linkproppred.dataset import LinkPropPredDataset
+    from tgb.linkproppred.evaluate import Evaluator
+    from tgb.utils.utils import save_results, set_random_seed
```
rusty1s, line 26: Can we test here instead on a small synthetic dataset instead? I don't think there is a need to test this model on a large real world use-case within unit tests.

torch_geometric/nn/models/edge_bank_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+Reference:
+    - https://github.com/fpour/DGB/tree/main
+"""
+
+import warnings
+
+import torch
+
+
+class EdgeBankPredictor(object):
```
rusty1s, line 14: `torch.nn.Module`.

torch_geometric/nn/models/edge_bank_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(
+        self,
+        src: torch.tensor,
+        dst: torch.tensor,
+        ts: torch.tensor,
```
rusty1s, line 19: I think it is kind of bad practice to initialize a model with input data. I would prefer if users need to call `update` explicitly after initializing.

torch_geometric/nn/models/edge_bank_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+        pred = torch.zeros(len(query_src))
+        idx = 0
+        for src, dst in zip(query_src, query_dst):
+            if (src, dst) in self.memory:
+                if (self.memory_mode == 'fixed_time_window'):
+                    if (self.memory[(src, dst)] >= self.prev_t):
+                        pred[idx] = self.pos_prob
+                else:
+                    pred[idx] = self.pos_prob
+            idx += 1
```
rusty1s, line 159: Can we get rid of the for-loop here and compute this in PyTorch directly?

Issue #252, opened 2019-04-30
Batched Data usage problem.


RuihongQiu:
## ❓ Questions & Help

Hello. I came across a strange bug when manipulating the batched Data. I can print the `data` and assign it to a variable. But when I print the attribute of it, it failed. Though some other datas don't have this bug. Could you help me?

    In[3]: data
    Out[3]: Batch(batch=[176], edge_attr=[137], edge_index=[2, 137], in_degree_inv=[137], 
    out_degree_inv=[137], sequence=[282], sequence_len=[100], x=[176, 1], y=[100])
    In[4]: x = data.x
    In[5]: print(x)
    Traceback (most recent call last):
      File "/home/test/pt-1.0/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line
    3296, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
      File "<ipython-input-5-fc17d851ef81>", line 1, in <module>
        print(x)
      File "/home/test/pt-1.0/lib/python3.6/site-packages/torch/tensor.py", line 66, in __repr__
        return torch._tensor_str._str(self)
      File "/home/test/pt-1.0/lib/python3.6/site-packages/torch/_tensor_str.py", line 277, in _str
        tensor_str = _tensor_str(self, indent)
      File "/home/test/pt-1.0/lib/python3.6/site-packages/torch/_tensor_str.py", line 195, in 
    _tensor_str
        formatter = _Formatter(get_summarized_data(self) if summarize else self)
      File "/home/test/pt-1.0/lib/python3.6/site-packages/torch/_tensor_str.py", line 80, in __init__
        value_str = '{}'.format(value)
      File "/home/test/pt-1.0/lib/python3.6/site-packages/torch/tensor.py", line 378, in __format__
        return self.item().__format__(format_spec)
    RuntimeError: CUDA error: device-side assert triggered

rusty1s:
To me this does not look like a PyG bug, but it is hard to say what may be the cause of this error. You can print `data` because it only requests `data.x.size()`. I'm pretty sure that `print(x.size())` does not throw an error either, but accessing the real data does.

RuihongQiu:
Yeah! You are right. `x.size()` is also valid. So it is a data generation bug? Do you have any idea about how to debug? I have tested every single graph when generating the `InMemoryDataset`?

rusty1s:
[...]

Issue #246, opened 2019-04-29
ValueError: too many values to unpack (expected 2)


youjinChung:
I was using pytorch-geometric that I installed from github repo.
and got this error while running autoencoder example code with my custom dataset.

```
Traceback (most recent call last):
  File "json2edgelist.py", line 72, in <module>
    data = model.split_edges(data)
  File "/home/paperspace/miniconda2/envs/thesis/lib/python3.6/site-packages/torch_geometric-1.1.3-py3.6.egg/torch_geometric/nn/models/autoencoder.py", line 114, in split_edges
    row, col = data.edge_index
ValueError: too many values to unpack (expected 2)
```
I read that this is due to the confliction between pypi packages from former issues, I guess I got this issue using the master branch version.

youjinChung:
Actually, I just checked again.
It worked well as example code without editing. but I just got this error when I run with my custom data.

rusty1s:
Related https://github.com/rusty1s/pytorch_geometric/issues/221

youjinChung:
Thank you so much!

PR #7974, opened 2023-09-04
Add `DistNeighborSampler`

kgajdamo:

`DistNeighborSampler` leverages the `NeighborSampler` class from `pytorch_geometric` and the `neighbor_sample` function from `pyg-lib`. However, due to the fact that in case of distributed training it is required to synchronise the results between machines after each layer, the part of the code responsible for sampling was implemented in python.

Added suport for the following sampling methods:
- node, edge, negative, disjoint, temporal

**TODOs:**

- [x] finish hetero part
- [x] subgraph sampling

**This PR should be merged together with other distributed PRs:**
pyg-lib: [#246](https://github.com/pyg-team/pyg-lib/pull/246), [#252](https://github.com/pyg-team/pyg-lib/pull/252)
GraphStore\FeatureStore: https://github.com/pyg-team/pytorch_geometric/pull/8083
DistLoaders:
1.  https://github.com/pyg-team/pytorch_geometric/pull/8079
2.  https://github.com/pyg-team/pytorch_geometric/pull/8080
3.  https://github.com/pyg-team/pytorch_geometric/pull/8085


JakubPietrakIntel (approved):
all issues were addressed!

torch_geometric/distributed/event_loop.py
```diff
@@ -0,0 +1,97 @@
+import asyncio
+import logging
+from threading import BoundedSemaphore, Thread
+from typing import Callable, Optional
+
+import torch
+
+# This code is taken from alibaba graphlearn-for-pytorch repository:
```
kgajdamo, line 8: Hi @rusty1s FYI. Sorry, I should have added it earlier.

rusty1s (approved):
Overall, this is a massive PR and it is very hard to review/merge it. I took a pass and made the following changes
* `remove_duplicates` shouldn't operate on numpy arrays
* Adjusted `NodeDict` and `BatchDict` to be correctly defined data classes
* I dropped support for `sample_from_edges` since I don't like how `asyncio` is used within `NeighborSampler` - I think we need a different solution for this.
* I really like the changes to support homgeoneous `FeatureStore` and `GraphStore` implementations, this is very cool :)
* Ideally, I feel it would be good to move `sample_one_hop` to `DistributedNeighborSampler` but this isn't blocking.

While re-factoring, I probably added a bunch of breaking changes (sorry for that), but since there is no test case I cannot test my changes.

PR #7996, opened 2023-09-07
Replace typ annotation  List[str] with str for *args

apfelsinecode:
Following [PEP 484: _Type Hints/Arbitrary argument lists and default argument values_](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values) the type annotation for argument list `*args` is annotated with the type of each element (`str`) instead of the type `List` itself.

This affects the class `Data` and its attributes `Storage` and `View`.

Previously, IDEs like PyCharm would mark `data.to('cuda', 'x', 'edge_index')` as incorrect, because it expects every element of the list `args` to be another list.

### DIFF
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -265,43 +265,43 @@ class BaseData:
         r"""Returns :obj:`True` if graph edges are directed."""
         return not self.is_undirected()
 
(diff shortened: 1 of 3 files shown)


rusty1s (approved):
Thank you, didn't know :)

PR #8001, opened 2023-09-08
ensure that last batch doesn't have any nan labels (causing nan losses) in Papers100m

puririshi98:


puririshi98:
doesn't seem to happen in latest version

PR #8002, opened 2023-09-08
Update remote_backend_utils.py

mzamini92:
can't we optimize the `_internal_num_nodes` function by directly accessing the attribute values and using list comprehensions to extract the sizes. I believe this can avoid iterating through the entire list of edge attributes multiple times. This should make the function more efficient and concise.


rusty1s (commented):
Thanks. I don't see yet why we are currently iterating over the list of edge attributes multiple times. I also don't think your implementation is correct. Here, we are interested in the number of nodes per node type, while your implementation seems to sum them up alltogether.

rusty1s:
Closing this as I don't see why this function needs changes. Feel free to re-open if you disagree.

PR #8019, opened 2023-09-11
Make `DimeNet` pickleable

bkmi:
This is an extremely simple and small change to make instantiated versions of `SphericalBasisLayer` pickleable.

The issue is that the lambda is a local function and therefore not pickleable.

I made the function into a "private" static method within the class. It is unlikely to cause name collisions with subclasses.

I have not run `pytest`; however, I am using the code in my own project.


rusty1s (approved):
Thank you :)

PR #8024, opened 2023-09-13
Added the option to pass keyword arguments to the underlying normalization layers within `BasicGNN` and `MLP`

erikhuck:


rusty1s (approved):
Thanks. I modified this a bit and confirmed that TorchScript tests pass with this. For your circular dependency issues, let's revisit this in a separate PR.

erikhuck:
@rusty1s the purpose of norm_kwargs was so a user could pass in whatever arguments they need to in the norm. Now they will only be able to pass in batch and batch_size whereas other normalization layers take other arguments. For example, MessageNorm takes a p and a msg argument in its forward method. I think the main issue was passing in the batch but someone in the future may require more flexibility.

PR #8029, opened 2023-09-14
Add distributed training sage e2e example on homogeneous graphs [5/6]

ZhengHongming888:

You can run this example as command below in 2 nodes (2 partitions) - 

This homo e2e example will base on ogbn-product dataset with 2 partitions which is generate by our partition example script in the same folder. The default number of sampler workers is 2 and concurrency for mp.queue is 2.. And you can also setup other number to give you more gain on throughput.

node 0:
python dist_train_sage_for_homo.py --dataset_root_dir=your partition folder --num_nodes=2 --node_rank=0 --num_training_procs=1 --master_addr= master ip

node 1:
python dist_train_sage_for_homo.py --dataset_root_dir=your partition folder --num_nodes=2 --node_rank=1 --num_training_procs=1 --master_addr= master ip

The training will output the progress bar as the status like
Epoch 00:  27%|████████████████████████████████▊                                                                                        | 26624/98307 [   ]

After training/test epochs there will generate 3 result files -

1) dist_train_sage_for_homo.txt   -   general training information /arguments
2) dist_train_sage_for_homo_rank0.txt  - results for training loss, accuracy, epoch time on rank0 node
3) dist_train_sage_for_homo_rank1.txt  - results for training loss, accuracy, epoch time on rank1 node

You also can reference from our e2e README for more distributed example cases and will add more like hetero, edge_sampler, etc and their running command. Welcome for any comments!  

Thanks.


examples/distributed/pyg/dist_train_sage_for_homo.py
```diff
@@ -0,0 +1,348 @@
...
+    xs = []
+    y_true = []
+    for i, batch in enumerate(test_loader):
+        if i == 0:
+            device = batch.x.device
+        x = model(batch.x, batch.edge_index)[:batch.batch_size]
+        xs.append(x.cpu())
+        y_true.append(batch.y[:batch.batch_size].cpu())
+        del batch
+        if i == len(test_loader) - 1:
```
JakubPietrakIntel, line 33: I suggest that we remove this barrier wait here and instead in the enable `persistent_workers=True` in the `DistNeighborLoader` when `num_workers > 0`. This should allow for the RPC to stay alive in worker subprocesses and allow for processing the remote requests if one of the iterator exits sooner than the other (when async processing is on).
I tested this on my system, didn't get any RPC connection issues.
Would you please add such a line to the `dist_neighbor_loader.py`?
`NodeLoader.__init__(
            self,
             ...
            persistent_workers=True if kwargs['num_workers'] > 0 else False,
            )`

examples/distributed/pyg/dist_train_sage_for_homo.py
```diff
@@ -0,0 +1,348 @@
...
+            if i == 0:
+                pbar.set_description(f'Epoch {epoch:02d}')
+            optimizer.zero_grad()
+            out = model(
+                batch.x,
+                batch.edge_index)[:batch.batch_size].log_softmax(dim=-1)
+            loss = F.nll_loss(out, batch.y[:batch.batch_size])
+            loss.backward()
+            optimizer.step()
+            if i == len(train_loader) - 1:
```
JakubPietrakIntel, line 168: Same as above

JakubPietrakIntel (changes requested):
@ZhengHongming888 I left a request to remove barrier waits and enable persistent workers instead.

examples/distributed/pyg/dist_train_sage_for_homo.py
JakubPietrakIntel: The connectivity issue should be fixed by https://github.com/pyg-team/pytorch_geometric/pull/8605, no need to use `persistent_workers`

JakubPietrakIntel:
https://github.com/pyg-team/pytorch_geometric/pull/8713 adds an updated version of e2e examples for dist PyG, so I'm closing this PR in agreement with @ZhengHongming888

PR #8032, opened 2023-09-14
Add example employing both PVC tiles

DamianSzwichtenberg:
Distributed GAT training, targeting XPU devices. PVC has 2 tiles, each reports itself as a separate device. DDP approach allows us to employ both tiles.

Additional requirements:
- IPEX (`intel_extension_for_pytorch`)
- oneCCL (`oneccl_bindings_for_pytorch`)

We need to import both these modules, as they extend torch module with XPU/oneCCL related functionality.

Run with:
`mpirun -np 2 python distributed_sampling_xpu.py`


DamianSzwichtenberg:
@pyg-team/intel-team Please take a look.

kgajdamo (approved):

PR #8033, opened 2023-09-14
Adds the option to use batch-level normalization layers in `AttentionalAggregation`/`MLPAggregation`

erikhuck:


torch_geometric/nn/aggr/attention.py
```diff
@@ -47,11 +47,13 @@ def reset_parameters(self):
 
     def forward(self, x: Tensor, index: Optional[Tensor] = None,
                 ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
-                dim: int = -2) -> Tensor:
+                dim: int = -2, batch: Optional[Tensor] = None,
```
erikhuck: @rusty1s This allows the `batch` and `batch_size` arguments to be passed to the underlying MLP modules within the `AttentionalAggregation`. Since we provide the arguments to the `MLP` class itself, we must provide that same capability to the modules that use the `MLP`.

torch_geometric/nn/aggr/attention.py
```diff
@@ -45,13 +45,21 @@ def reset_parameters(self):
...
-                ptr: Optional[Tensor] = None, dim_size: Optional[int] = None,
-                dim: int = -2) -> Tensor:
+    def forward(
+        self,
+        x: Tensor,
+        index: Optional[Tensor] = None,
+        ptr: Optional[Tensor] = None,
+        dim_size: Optional[int] = None,
+        dim: int = -2,
+        batch: Optional[Tensor] = None,
```
erikhuck, line 55: @rusty1s This allows the batch and batch_size arguments to be passed to the underlying MLP modules within the AttentionalAggregation. Since we provide the arguments to the MLP class itself, we must provide that same capability to the modules that use the MLP.

torch_geometric/nn/aggr/attention.py
```diff
@@ -45,13 +45,21 @@ def reset_parameters(self):
...
+        batch_size: Optional[int] = None,
+    ) -> Tensor:
 
         self.assert_two_dimensional_input(x, dim)
-        gate = self.gate_nn(x)
-        x = self.nn(x) if self.nn is not None else x
+        gate = self.gate_nn(x, batch=batch, batch_size=batch_size)
```
erikhuck, line 60: @rusty1s I'm not sure if `self.gate_nn` and `self.nn` are typed as `Module` in the documentation because they're supposed to be anything other than `MLP` or because you needed to avoid the circular import error. The papers discussing attentional aggregation describe H gate (gate_nn) and H theta (nn) as MLPs.  To avoid a type error, we'd need to clarify in the documentation that the expectation is for them to be `MLP`s despite the more generic `Module` label.

erikhuck:
@rusty1s I wanted to open this PR in draft for initial discussion. If we can agree on a design, then I can make these same changes to the MLP aggregation as well.

torch_geometric/nn/aggr/attention.py
rusty1s: I don't think we need these new arguments here. Actually, `batch` is already given through `index` here, and `batch_size` is given through `dim_size`. You should be able to just pass these instead.

torch_geometric/nn/aggr/attention.py
rusty1s: Yes, it doesn't necessarily need to be a `torch_geometric.nn.MLP`. I changed this accordingly.

torch_geometric/nn/aggr/attention.py
rusty1s: Fixed.

PR #8038, opened 2023-09-15
Weighted sampling in `NeighborLoader` and `LinkNeighborLoader`

rusty1s:


ShuaiWang97:
Thanks for updating this function! I want to ask are these weights can be learnable as part of the model? optimized by the loss function?

PR #8048, opened 2023-09-18
Add `edge_attr` support for `ResGatedGraphConv`

royvelich:


torch_geometric/nn/conv/res_gated_graph_conv.py
```diff
@@ -116,5 +119,9 @@ def forward(self, x: Union[Tensor, PairTensor], edge_index: Adj) -> Tensor:
 
         return out
 
-    def message(self, k_i: Tensor, q_j: Tensor, v_j: Tensor) -> Tensor:
-        return self.act(k_i + q_j) * v_j
+    def message(self, x_i: Tensor, x_j: Tensor, edge_attr: Tensor) -> Tensor:
```
royvelich, line 122: I think that I forgot to check here whether `edge_attr` is `None`, and act accordingly... will fix it.

royvelich: Resolved

rusty1s (approved):

PR #8071, opened 2023-09-22
Multi-Node-Multi-GPU Tutorial

puririshi98:
ready for review


Kh4L (approved):

puririshi98:
@akihironitta @rusty1s lmk if anything else needed to merge

puririshi98:
@akihironitta @rusty1s lmk if anything else needed to merge

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added a tutorial for multi-node-multi-GPU training with pure PyTorch ([#8071](https://github.com/pyg-team/pytorch_geometric/pull/8071)
```
akihironitta, line 10:
```suggestion
- Added a tutorial for multi-node multi-GPU training with pure PyTorch ([#8071](https://github.com/pyg-team/pytorch_geometric/pull/8071))
```

docs/source/tutorial/multi_gpu_vanilla.rst
```diff
@@ -1,4 +1,4 @@
-Multi-GPU Training in Pure PyTorch
+Multi-GPU GNN Training
 ==================================
```
akihironitta, line 2:
```suggestion
Multi-GPU GNN Training
======================
```

docs/source/tutorial/multi_node_multi_gpu_tutorial.rst
```diff
@@ -0,0 +1,252 @@
...
+        ...
+
+
+    def get_local_process_group():
+        ...
+
+
+    def run(device, data, world_size, model, epochs, batch_size, fan_out,
+                  split_idx, num_classes):
+        ...
```
akihironitta, line 36: We don't need these func definitions right?

docs/source/tutorial/multi_node_multi_gpu_tutorial.rst
```diff
@@ -0,0 +1,252 @@
...
+            'test':rand_id[:int(.8 * num_nodes):],
+        }
+
+        model = GCN(dataset.num_features, args.hidden_channels, 2,
+                    dataset.num_classes)
+        run(device, data, nprocs, model, args.epochs, args.batch_size,
+                  args.fan_out, split_idx, dataset.num_classes)
+
+
+Similarly to the warm up example, we define a :meth:`run` function. However, in this case we are using torch distributed with NVIDIA NCCL backend, instead of relying on :class:`~torch.multiprocessing`. Because we are running on multiple nodes, we want to set up a local process group for each node, and use :obj:`args.ngpu_per_node` GPUs per node. We then select the the CUDA device that will be used by each process within each process group. The next steps are fairly basic :pyg:`PyG` and :pytorch:`PyTorch` usage. We load our (synthetic) dataset and then set up our 60/20/20 train/val/test split. Next, we define our :class:`~torch_geometric.nn.models.GCN` model and finally call our :meth:`run` function.
```
akihironitta, line 81:
```suggestion
Similarly to the single-node multi-gpu example, we define a :meth:`run` function. However, in this case we are using torch distributed with NVIDIA NCCL backend, instead of relying on :class:`~torch.multiprocessing`. Because we are running on multiple nodes, we want to set up a local process group for each node, and use :obj:`args.ngpu_per_node` GPUs per node. We then select the the CUDA device that will be used by each process within each process group. The next steps are fairly basic :pyg:`PyG` and :pytorch:`PyTorch` usage. We load our (synthetic) dataset and then set up our 60/20/20 train/val/test split. Next, we define our :class:`~torch_geometric.nn.models.GCN` model and finally call our :meth:`run` function.
```

Also, I think we should breifly explain or add a link to some documentation about process group as I believe the target readers are someone who doesn't have much experience in multi-node multi-gpu training.

docs/source/tutorial/multi_node_multi_gpu_tutorial.rst
puririshi98: these are definitely needed functions to set up the process groups for multinode

akihironitta: I was asking because I see the definitions right below this section.

puririshi98: oh i see what you mean. ill leave it up to @rusty1s to decide, i chose to show these as a skeleton to give an idea of the key base functions needed for multi-node but i can understand your reasoning as well.

rusty1s (approved):

flxmr:
Is it only me or is this a very weird style to use Slurm? I got a student being very excited about this "official" tutorial for pyg and it confused him (and me to a lesser degree) a lot.

Specifically:
1) your SLURM-cluster  seems to not have their GPUs managed by GRES? That's a highly unusual setup for anything a random person might find in their slurm setup?
2) you need environment variables set  (e.g. `RANK`). Probably your container does this? 
3) why would you tie the programming example to the very weird way of using Slurm (an example should use `sbatch` I'd say)?
4) the `get_num_workers function` is also very weird... https://github.com/pyg-team/pytorch_geometric/blob/3e55a4c263f04ed6676618226f9a0aaf406d99b9/examples/multi_gpu/multinode_multigpu_papers100m_gcn.py#L46 You get the number of CPUs in your slurm-task and then divide that by 2x the `world_size` (aka the total number of *tasks* you have on multiple (or one) nodes, refreshing on the process groups now). This means that without any slurm settings of your installation this is 0 (`--cpus-per-task < 2*world_size)? What's the point, especially for an example?

Maybe @puririshi98 can help me out on this points and I can let my student prep a more generic and slurmy version.

rusty1s:
Thanks for the feedback. I agree that the tutorial is missing out on some information in this aspect. It would be great to get your student and @puririshi98 to update the confusing parts.

puririshi98:
@flxmr thank you for your concerns. I will reply back soon w/ more details after discussing internally at NVIDIA.

flxmr:
Yes, we will prepare something the upcoming week. Our own cluster doesn't have any container-integration, but I found an university-available array of DGX-systems which have, so we can also integrate the container instructions.

puririshi98:
I will make a quick PR for some cleanups:
1. I will remove get_local process group and create_local_process_gropu functions for simplicity and use int(os.environ['LOCAL_RANK'])) for simplicity. These functions were being used in an example for more complicated multi-node setups.
2. I will remove the line that sets env variable NVSHMEM_SYMMETRIC_SIZE, again this was for different usage. this is for using nvshmem for feature store
3. update the suggested slurm command to:
 srun -l -N<num_nodes> --ntasks-per-node=<ngpu_per_node> \
    --container-name=cont  --container-image=<image_url> --container-mounts=<data-directory>/ogb-papers100m/:/workspace/dataset \
    python3 path_to_script.py --ngpu_per_node <> 
4. I'll remove --ngpu-per-node as well since I will access env by LOCAL_RANK.

puririshi98:
i would recommend making your further improvements on this after my PR has landed ill work on it today/tmrw

puririshi98:
https://github.com/pyg-team/pytorch_geometric/pull/8292
here it is

puririshi98:
@flxmr, curious what additional changes would you like to see after my PR gets merged?

PR #8074, opened 2023-09-25
Update graph.py

mzamini92:
This modification is more efficient because it only subtracts the minimum weight and divides by the maximum weight once. This is because the edge_weight.min() and edge_weight.max() terms are the same in both steps, so we can calculate them once and then reuse them.


rusty1s:
I would say this implementation is less efficient because you compute `edge_weight.min()` twice. Closing this as I don't see how one could make this more efficient. Please feel free to re-open if you have a strong concern.

PR #8075, opened 2023-09-25
Added a random permutation transformation to the LCM 2nd minimum example

ArchieGertsman:
In the LCM paper, it is suggested to randomly permute the data as an augmentation technique for better generalization. I have included this as a dataset transform in the 2nd minimum example.


rusty1s (approved):
Sorry, missed this part :(

PR #8080, opened 2023-09-27
Add `DistributedNeighborLoader` [3/6]

JakubPietrakIntel:
**[2/3] Distributed Loaders PRs**
This PR includes`DistributedNeighborLoader` used for processing node sampler output in distributed training setup.

1.  https://github.com/pyg-team/pytorch_geometric/pull/8079
2.  https://github.com/pyg-team/pytorch_geometric/pull/8080
3.  https://github.com/pyg-team/pytorch_geometric/pull/8085

Other PRs related to this module:
DistSampler: https://github.com/pyg-team/pytorch_geometric/pull/7974
GraphStore\FeatureStore: https://github.com/pyg-team/pytorch_geometric/pull/8083

PR #8083, opened 2023-09-27
Update `GraphStore` and `FeatureStore` [1/6]

kgajdamo:

Please be aware that this PR should be merged before Loaders package! - @JakubPietrakIntel 
Loaders:
1.  https://github.com/pyg-team/pytorch_geometric/pull/8079
2.  https://github.com/pyg-team/pytorch_geometric/pull/8080
3.  https://github.com/pyg-team/pytorch_geometric/pull/8085

Other PRs related to this module:
DistSampler: https://github.com/pyg-team/pytorch_geometric/pull/7974

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Update `GraphStore` and `FeatureStore` to support distributed training ([#8083](https://github.com/pyg-team/pytorch_geometric/pull/8083))
 - Disallow the usage of `add_self_loops=True` in `GCNConv(normalize=False)` ([#8210](https://github.com/pyg-team/pytorch_geometric/pull/8210))
 - Disable device asserts during `torch_geometric.compile` ([#8220](https://github.com/pyg-team/pytorch_geometric/pull/8220))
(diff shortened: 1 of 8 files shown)


JakubPietrakIntel:
@rusty1s @ZhengHongming888 
I added [ac49add](https://github.com/pyg-team/pytorch_geometric/pull/8083/commits/ac49add09e2fe95b9771b5f9a4a97527d22d9ce8) which enforces sorting by destination node in LGS `edge_index` (in other words sort by col, or `edge_index[1]`) to match CSC format used for sampling. This is condition ensures that input data won't be permuted again at neighbor sampler init.
We do that to make sure that correct edge_ids are returned at sampler's output and in `filter_fn` in `NodeLoader` i.e. [L161](https://github.com/pyg-team/pytorch_geometric/blob/ac49add09e2fe95b9771b5f9a4a97527d22d9ce8/torch_geometric/loader/node_loader.py#L161) the expression is evaluated correctly with `perm = None`.
We have considered alternative solutions and decided that this one will be the best regarding performance, at a cost of adding a constraint on incoming data. We can't have a global perm stored in LGS, as METIS is not performing a simple split, so the ordering would be lost. The other solution would be to use local values per partition in `edge_id`, hence increasing complexity and adding yet another re-indexing step. Furthermore, it would require RPC requests for perm values located in remote partitions, because the sampler may return edge_ids located in remote partition.

rusty1s (approved):

PR #8096, opened 2023-09-29
Added `interval` argument to `BaseTransform`

mzamini92:
Following to [PR5733](https://github.com/pyg-team/pytorch_geometric/pull/7533)  I've Added the interval argument to the BaseTransform class, and any transform that inherits from BaseTransform will now have access to the interval attribute in its forward method. @rusty1s


rusty1s:
Thanks. I don't understand why this is necessary. Not all transforms need `interval ` to be defined. What problem does this solve?

PR #8098, opened 2023-09-30
Add `interval` arugment to `Compose` transform

mzamini92:
Following to [PR5733](https://github.com/pyg-team/pytorch_geometric/pull/7533)  I've Added the interval argument. @rusty1s With these modifications, both the Compose and ComposeFilters classes accept an interval argument, and it will be passed to the transformed data when composing transforms or applying filters.


rusty1s:
Thanks. I don't understand why this is necessary. What problem does this solve?

PR #8127, opened 2023-10-04
Fix weight type for `segment_matmul`

Jokeren:
`weight` should be a float tensor as `x`, the previous code initializes `weight` as an int tensor

### DIFF
--- a/torch_geometric/typing.py
+++ b/torch_geometric/typing.py
@@ -29,7 +29,7 @@ try:
         try:
             x = torch.randn(3, 4, device='cuda')
             ptr = torch.tensor([0, 2, 3], device='cuda')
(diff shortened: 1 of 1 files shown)


rusty1s (approved):

torch_geometric/typing.py
```diff
@@ -29,7 +29,7 @@
         try:
             x = torch.randn(3, 4, device='cuda')
             ptr = torch.tensor([0, 2, 3], device='cuda')
-            weight = torch.tensor([2, 4, 4], device='cuda')
+            weight = torch.randn((2, 4, 4), device='cuda')
```
rusty1s, line 32:
```suggestion
            weight = torch.randn(2, 4, 4, device='cuda')
```

PR #8135, opened 2023-10-05
Draft: CUDF Data integration

puririshi98:
still brainstorming:
it is unclear to me how we would store Data attributes as CUDF data frames in an effective way. The issue is foresee is that a super majority of functions in PyG (dataloaders and GNNs and most utils as well) are designed for torch tensors. If we store Data[key] = value by converting value to CUDF frame, we would inevitably have to convert back to Tensor to use the Data with any part of PyG. To avoid converting back to Tensor from CUDF frame, we would have to reimplement the majority of PyG to work on it and this doesnt seem reasonable given how well Torch works already

### DIFF
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -504,7 +504,10 @@ class Data(BaseData, FeatureStore, GraphStore):
     # __getitem__, __setitem__, and __delitem__ so, for example, we
     # can accept key: Union[str, TensorAttr] in __getitem__.
     def __getitem__(self, key: str) -> Any:
-        return self._store[key]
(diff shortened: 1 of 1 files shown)


rusty1s:
@puririshi98 I think you are overcomplicating what I had in mind. I was expecting something similar to https://github.com/pyg-team/pytorch_geometric/pull/8110/files. In the end, it would be good to store `pandas.DataFrame` and `cudf.DataFrame` also as part of `Data`, and we would expect that `num_nodes` can correctly read the data frames and return the number of rows from it, and that `DataLoader()` would stack them in the first dimension.

Issue #8200, opened 2023-10-16
GATconv and GATv2conv 


SZiesche:
### 📚 Describe the documentation issue

In the documentation there are several subtleties hidden, that make it hard to understand, what is really going on without looking at the actual implementation. These are:
- it is not right away apparent that the layer works on bipartite graphs. The initial formulas rather indicate, that this is not the case.
- the stacking operator || is not well known and could be easily circumvented by writing out the additions in GATconv
- in GATv2conv the operator is actually used very sloppy already in the paper and the docs would also just be easier to read if || was just replaced by the proper addition.
- in GATv2conv `share_weights` is by default `False` however the formulas indicate, that it should be `True`. 

### Suggest a potential alternative/fix

I suggest to explicitly name the different `\Theta`s used in the different layers by an index `s` and `t` (like it is already done in the docs when talking about input sizes). Do the same for `a` in the `GATconv` layer. Finally remove the unclear `||` operator and just write out the few additions.

I could create a PR if you give me some rights to push a branch. Otherwise you could replace the class docstrings by 

`gat_conv.py`
```
    r"""The graph attentional operator from the `"Graph Attention Networks"
    <https://arxiv.org/abs/1710.10903>`_ paper

    .. math::
        \mathbf{x}^{\prime}_i = \alpha_{i,i}\mathbf{\Theta}_{s}\mathbf{x}_{i} +
        \sum_{j \in \mathcal{N}(i)} \alpha_{i,j}\mathbf{\Theta}_{t}\mathbf{x}_{j},

    where the attention coefficients :math:`\alpha_{i,j}` are computed as

    .. math::
        \alpha_{i,j} =
        \frac{
        \exp\left(\mathrm{LeakyReLU}\left(
        \mathbf{a}^{\top}_{s} \mathbf{\Theta}_{s}\mathbf{x}_i
        + \mathbf{a}^{\top}_{t} \mathbf{\Theta}_{t}\mathbf{x}_j
        \right)\right)}
        {\sum_{k \in \mathcal{N}(i) \cup \{ i \}}
        \exp\left(\mathrm{LeakyReLU}\left(
        \mathbf{a}^{\top}_{s} \mathbf{\Theta}_{s}\mathbf{x}_i
        + \mathbf{a}^{\top}_{t}\mathbf{\Theta}_{t}\mathbf{x}_k
        \right)\right)}.

    If the graph has multi-dimensional edge features :math:`\mathbf{e}_{i,j}`,
[...]

PR #8201, opened 2023-10-16
[Documentation] make documentation of `GATConv` and `GATv2Conv` more explicit and accurate

SZiesche:
## What?
I explicitly named the different `\Theta`s used in the different layers by an index s and t (like it is already done in the docs when talking about input sizes). I did the same for a in the GATconv layer. Finally removed the unclear `||` operator and just wrote out the few additions.

## Why?
To solve the issues discussed in https://github.com/pyg-team/pytorch_geometric/issues/8200

## Testing?
No tests needed, as only documentation was changed.


EdisonLeeeee (approved):

SZiesche:
ok great, should we make the two failing checks pass? I had a look at the changelog, but the PR doesn't seem to contain something worth mentioning there. The test for previous pytorch version seems not to be failing because of this PR.

How can the stuff be merged in the end?

SZiesche:
hey guys, I wanted to ask again, who can push the merge button (once this is updated with master) and if we need to make the changelog happy.

EdisonLeeeee:
@rusty1s would take care of it :)

rusty1s (approved):
Thank you, this is indeed much clearer :)

SZiesche:
> Thank you, this is indeed much clearer :)

It was a pleasure. I always like to share such improvements. Documentation is always a difficuilt point, because the author of the code is often hugely biased and has a large knowledge advantage, which makes it pretty hard for him to write the docs.

PR #8207, opened 2023-10-17
[Documentation] remove legacy argument from `HGTConv` docstring

### LINKED ISSUE #8185: unused `group` parameter in `HGTConv` still documented
### 📚 Describe the documentation issue

#7117 replaces `HGTConv` with the implementation done for the faster `FastHGTConv`. In the process of doing so, the `group` parameter has been removed falling back to the default `sum` operation. (@puririshi98: this was intentional, right?). The docs, however, haven't been updated accordingly.

https://github.com/pyg-team/pytorch_geometric/blob/737707c37fc2bd712a2289b683ec14549926ff49/torch_geometric/nn/conv/hgt_conv.py#L40-L43

Remove the unused parameter from the docs.

stadlmax:
closes #8185

Background: #7117 replaces the previous `HGTConv` with an updated implementation but kept the argument `group` in the docstring which isn't used anymore. As only the documentation is touched, no further testing is needed.


rusty1s (approved):
Thanks!

PR #8085, opened 2023-09-27
Add `DistributedLinkNeighborLoader` [4/6]

JakubPietrakIntel:
**[3/3] Distributed Loaders PRs**
This PR includes `DistributedLinkNeighborLoader` used for processing edge sampler output in distributed training setup.

1.  https://github.com/pyg-team/pytorch_geometric/pull/8079
2.  https://github.com/pyg-team/pytorch_geometric/pull/8080
3.  https://github.com/pyg-team/pytorch_geometric/pull/8085

Other PRs related to this module:
DistSampler: https://github.com/pyg-team/pytorch_geometric/pull/7974
GraphStore\FeatureStore: https://github.com/pyg-team/pytorch_geometric/pull/8083

PR #8209, opened 2023-10-17
Update `DistNeighborSampler` for homo graphs [2/6]

kgajdamo:
**Changes made:**
- added support for temporal sampling
- use torch.Tensors instead of numpy arrays
- move _sample_one_hop() from NeighborSampler to DistNeighborSampler
- do not go with disjoint flow in _sample() function - this is not needed because batch is calculated after
- added tests for node sampling and disjoint (works without DistNeighborLoader)
- added tests for node temporal sampling (works without DistNeighborLoader)
- some minor changes like changing variables names etc

This PR is based on the #8083, so both must be combined to pass the tests.

Other distributed PRs:
#8083 
#8080 
#8085

### DIFF
--- a/torch_geometric/sampler/neighbor_sampler.py
+++ b/torch_geometric/sampler/neighbor_sampler.py
@@ -119,10 +119,11 @@ class NeighborSampler(BaseSampler):
             feature_store, graph_store = data
 
             # Obtain graph metadata:
-            node_attrs = feature_store.get_all_tensor_attrs()
-            self.node_types = list(
-                set(attr.group_name for attr in node_attrs
-                    if isinstance(attr.group_name, NodeType)))
+            node_attrs = [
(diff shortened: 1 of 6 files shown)


rusty1s (approved):

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -119,7 +119,10 @@ def __init__(
             feature_store, graph_store = data
 
             # Obtain graph metadata:
-            node_attrs = feature_store.get_all_tensor_attrs()
+            node_attrs = [
```
kgajdamo, line 122: @rusty1s  Thank you for the review :). One comment about these lines: they are common for both homo and hetero. In case of homo the attr.group_name will be None and therefore the condition `isinstance(attr.group_name, NodeType)` will never be met and node_attrs will be always empty. I will open a new PR with the previous version.

Issue #6251, opened 2022-12-16
Missing self_loop for GCNConv when nomalize is false


elharchaoui:
### 🐛 Describe the bug

Helllo,

The vanilla GCNConv does not consider adding self-loop when normalization is False, it can be visible in the implementation code of GCNConv, meanwhile here is an example that illustrates the issue.

I took the same example PyTorch geo introduction 
 [https://pytorch-geometric.readthedocs.io/en/latest/_images/graph.svg](url)

1- case without self-loop False on parameters :
```
import torch
from torch_geometric.data import Data

edge_index = torch.tensor([[0, 1, 1, 2],
                           [1, 0, 2, 1]], dtype=torch.long)

x = torch.tensor([[0], [2], [1]],
                 dtype=torch.float)

data = Data(x=x, edge_index=edge_index)


conv = GCNConv(1, 1, add_self_loops=False,
                      bias=False,
                      normalize=False,
                    aggr='max')

# force weight update for the sake of illustration

conv.lin.bias=torch.nn.Parameter(torch.tensor(0., requires_grad=True))
conv.lin.weight=torch.nn.Parameter(torch.tensor([[1.]], requires_grad=True))

with torch.no_grad():
    x1 = conv(x,edge_index)

```
result :
```
tensor([[2.],
        [1.],
        [2.]])
```
Node 1 didn't get max value 2 because there is no self loop, it's okey

2- case with self-loop True :
```
import torch
from torch_geometric.data import Data

edge_index = torch.tensor([[0, 1, 1, 2],
                           [1, 0, 2, 1]], dtype=torch.long)

x = torch.tensor([[0], [2], [1]],
                 dtype=torch.float)

data = Data(x=x, edge_index=edge_index)


conv = GCNConv(1, 1, add_self_loops=True,
                      bias=False,
                      normalize=False,
                    aggr='max')

# force weight update for the sake of illustration

conv.lin.bias=torch.nn.Parameter(torch.tensor(0., requires_grad=True))
conv.lin.weight=torch.nn.Parameter(torch.tensor([[1.]], requires_grad=True))

with torch.no_grad():
    x1 = conv(x,edge_index)
```
result :
```
tensor([[2.],
        [1.],
        [2.]])
```

Same result as before. I encountered this problem in a more complex problème, my example is simple for the illustration.

If there is something wrong in my understanding of the way of usage please let me know. thank you 

### Environment

* PyG version: 2.2.0
* PyTorch version: 1.13.0
[...]

PR #8210, opened 2023-10-17
Disallow `add_self_loops=True` in `GCNConv` when `normalize=False`

xnuohz:
Fix https://github.com/pyg-team/pytorch_geometric/issues/6251


EdisonLeeeee:
Thank you for the PR, I don't think it is necessary to make these changes, as 

- `normalize` is the core operation in `GCNConv`, and in most cases, it is recommended to set it to `True` when using `GCNConv` as the convolution layer.
- While there are situations where using `GCNConv` without normalization but with self-loops might be desired, it is possible to manually add self-loops during the forward pass or data preprocessing. In such cases, it might be more appropriate to consider using `SAGEConv`.
- To support such cases, adding the functionality of self-loops without normalization to `GCNConv` would require introducing additional code, which could potentially make the implementation more complex and harder to read.

Perhaps we can simply add a warning when `normalize=False` but `add_self_loops=True`, WDYT?

xnuohz:
> Thank you for the PR, I don't think it is necessary to make these changes, as
> 
> * `normalize` is the core operation in `GCNConv`, and in most cases, it is recommended to set it to `True` when using `GCNConv` as the convolution layer.
> * While there are situations where using `GCNConv` without normalization but with self-loops might be desired, it is possible to manually add self-loops during the forward pass or data preprocessing. In such cases, it might be more appropriate to consider using `SAGEConv`.
> * To support such cases, adding the functionality of self-loops without normalization to `GCNConv` would require introducing additional code, which could potentially make the implementation more complex and harder to read.
> 
> Perhaps we can simply add a warning when `normalize=False` but `add_self_loops=True`, WDYT?

Generally agree. There are two solutions I've considered

- Directly forbid `normalize=False`, indeed in most cases we will not set it like this
- Support `normalize=False` to make sure the return is correct, since currently the results are incorrect when `add_self_loops=True`.

And I chose the 2nd one, haha.

rusty1s (commented):
Thank you. This is actually a breaking change which we would need to avoid. As such, can we make `add_self_loops` be of type `Optional[bool] = None`, and set it based on `normalize=False/True` whenever it is not explicitly passed?

xnuohz:
> Thank you. This is actually a breaking change which we would need to avoid. As such, can we make `add_self_loops` be of type `Optional[bool] = None`, and set it based on `normalize=False/True` whenever it is not explicitly passed?

How about `add_self_loops=False` for `normalize=False` and `add_self_loops=True` for `normalize=True` ? If so, then a warning is needed when we explicitly pass `normalize=False` and `add_self_loops=True`.

rusty1s:
I just read @EdisonLeeeee's post, and need to agree with this. Normalization in `GCNConv` code is quite complex and we shouldn't move away from the original implementation too much.

rusty1s (approved):

Issue #7985, opened 2023-09-05
GCNConv: Weighted Self-Loops, Gaussian Initialization 


GageDeZoort:
### 🚀 The feature, motivation and pitch

I'd like to propose the implementation of a new `GCNConv` using recommendations from this paper:

[Principles for Initialization and Architecture Selection in Graph Neural Networks with ReLU Activations
](https://arxiv.org/abs/2306.11668)
The relevant modifications include:

- The linear layer `self.lin` would have a Gaussian initialization scheme with no biases and weights sampled from a zero-mean Gaussian with variance `C_w/self.in_channels`, where `C_w` is set just above 2 by default. 
- In the `gcn_norm` function, allow for self-loops to contribute more in proportion to  weights calculated via `edge_weight = deg_inv_sqrt[row] * edge_weight * deg_inv_sqrt[col]` (even if self loops are already included in `deg_inv_sqrt`). Specifically, if the `edge_weight` values are stored in a weighted adjacency matrix P, message passing would be (1-`t`) * identity + `t` * P for a hyperparameter 0<`t`<1. 

In summary, this would simply generalize the `GCNConv` to include hyperparameters `C_w` and `t`.

rusty1s:
This sounds cool :) We can either add this to the original `GCNConv` layer or in a separate module that inherits from it. What do you think?

GageDeZoort:
Thanks for the response, Matthias :) I think it'd be nice to have `t` and `C_W` as options in the original `GCNConv` layer. I'd be fine with having a separate module, though - this is only a loose preference. Either way, I'm happy to help with any implementation or surrounding documentation!

rusty1s:
Super. Would it be possible for you to provide the initial implementation directly in `GCNConv`, and I can take it from there?

GageDeZoort:
Sure, I'm happy to - will open a PR in the next few days!

PR #8221, opened 2023-10-18
Add Gaussian init and residual aggregations to `GCNConv`

GageDeZoort:
Addresses #7985

### DIFF
--- a/torch_geometric/nn/conv/gcn_conv.py
+++ b/torch_geometric/nn/conv/gcn_conv.py
@@ -27,15 +27,15 @@ from torch_geometric.utils.sparse import set_sparse_value
 
 
 @torch.jit._overload
-def gcn_norm(edge_index, edge_weight, num_nodes, improved, add_self_loops,
-             t, flow, dtype):
+def gcn_norm(edge_index, edge_weight, num_nodes, improved, add_self_loops, t,
+             flow, dtype):
     # type: (Tensor, OptTensor, Optional[int], bool, bool, str, Optional[int]) -> OptPairTensor  # noqa
     pass
 
 
 @torch.jit._overload
-def gcn_norm(edge_index, edge_weight, num_nodes, improved, add_self_loops,
-             t, flow, dtype):
+def gcn_norm(edge_index, edge_weight, num_nodes, improved, add_self_loops, t,
+             flow, dtype):
     # type: (SparseTensor, OptTensor, Optional[int], bool, bool, str, Optional[int]) -> SparseTensor  # noqa
     pass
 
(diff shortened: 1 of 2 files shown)


torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -60,7 +60,8 @@ def gcn_norm(edge_index, edge_weight=None, num_nodes=None, improved=False,
         deg_inv_sqrt.masked_fill_(deg_inv_sqrt == float('inf'), 0.)
         adj_t = torch_sparse.mul(adj_t, deg_inv_sqrt.view(-1, 1))
         adj_t = torch_sparse.mul(adj_t, deg_inv_sqrt.view(1, -1))
-
+        adj_t = (1 - t) * torch.eye(
```
EdisonLeeeee, line 63: IMO, it is better to place the residual part in the forward of `GCNConv` or a new module outside of `gcn_norm`.

torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -171,8 +177,10 @@ def __init__(
         improved: bool = False,
         cached: bool = False,
         add_self_loops: bool = True,
+        t: float = 1.0,
         normalize: bool = True,
-        bias: bool = True,
+        bias: bool = False,
+        weight_initializer: str = "normal",
```
EdisonLeeeee, line 183:
```suggestion
        weight_initializer: str = "glorot",
```
Keep it as `glorot` for compatibility.

torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -171,8 +177,10 @@ def __init__(
         improved: bool = False,
         cached: bool = False,
         add_self_loops: bool = True,
+        t: float = 1.0,
         normalize: bool = True,
-        bias: bool = True,
+        bias: bool = False,
```
EdisonLeeeee, line 182:
```suggestion
        bias: bool = True,
```
Keep it as`True` for compatibility.

torch_geometric/nn/conv/gcn_conv.py
GageDeZoort: In this case, would it be possible to add a note that "normal" is an option to the documentation? Maybe with a pointer to our paper, highlighting the importance of this init?

torch_geometric/nn/conv/gcn_conv.py
GageDeZoort: I'm open to it - but how would you suggest to implement this? The goal is to interpolate between the message passing operator (P) and the identity (I) like (1-t)I + tP, so it almost has to be done at this stage given the current implementation of `gcn_norm`.

GageDeZoort: @EdisonLeeeee bumping this question, curious what you had in mind!

EdisonLeeeee: Sorry for the delayed reply. The residual part can be performed after the (aggregated) output of `GCNConv`, i.e., `out = (1-t)*x + t*out`, WDYT?

PR #8222, opened 2023-10-19
Fix URL typo in `PCMQ4Mv2` dataset

pbielak:
In the `download()` method is a typo: `self.url_2d` should be named `self.url`


EdisonLeeeee (approved):
Thank you!

PR #8223, opened 2023-10-19
Allow for device agnostic testing

joao-alex-cunha:
This PR changes the test suite to allow running the tests on any hardware.

The change is rather small and targets the tests that run both on the cpu and gpu. For these tests the decorator is changed to also read from a couple of environment variables that can specify the device and optionally the backend.
Using environment variables has the benefit of avoiding further changes to the code while also providing some degree of future proofing. If in the future new hardware vendors want to run the PyG test suite on their hardware they simply can use the environment variables.

Usage:
`TORCH_DEVICE=some_new_hw TORCH_BACKEND=torch_new_hw pytest test`


rusty1s (approved):

torch_geometric/testing/decorators.py
```diff
@@ -169,6 +169,13 @@ def withCUDA(func: Callable):
     devices = [torch.device('cpu')]
     if torch.cuda.is_available():
         devices.append(torch.device('cuda:0'))
+    import importlib
+    import os
+    device = os.getenv("TORCH_DEVICE")
+    if device:
+        backend = os.getenv("TORCH_BACKEND")
+        importlib.import_module(backend)
+        devices.append(torch.device(device))
```
rusty1s, line 178:
```suggestion
    # Additional devices can be registered through environment variables:
    device = os.getenv('TORCH_DEVICE')
    if device:
        backend = os.getenv('TORCH_BACKEND')
        if backend is None:
            warnings.warn("Please specify the backend via 'TORCH_BACKEND' in"
                          "order to test againt '{device}'")
        else:
            import_module(backend)
            devices.append(torch.device(device))
```

rusty1s (approved):

torch_geometric/testing/decorators.py
```diff
@@ -169,6 +169,16 @@ def withCUDA(func: Callable):
     devices = [torch.device('cpu')]
     if torch.cuda.is_available():
         devices.append(torch.device('cuda:0'))
+    # Additional devices can be registered through environment variables:
```
rusty1s, line 172:
```suggestion

    # Additional devices can be registered through environment variables:
```

Issue #8219, opened 2023-10-18
Sparse tensor support on CPU requires MKL


agitter:
### 😵 Describe the installation problem

I am installing PyG 2.4 on Windows 10 in a cpuonly setting. When running the `examples/graph_gps.py` example code, I encountered a RuntimeError related to MKL:
```
Processing train dataset:   0%|          | 0/10000 [00:00<?, ?it/s]C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\utils\sparse.py:264: UserWarning: Sparse CSR tensor support is in beta state. If you miss a functionality in the sparse tensor support, please submit a feature request to https://github.com/pytorch/pytorch/issues. (Triggered internally at  C:\cb\pytorch_1000000000000\work\aten\src\ATen\SparseCsrTensorImpl.cpp:68.)
  adj = torch.sparse_csr_tensor(
Traceback (most recent call last):
  File "C:\xxxxx\Desktop\madison\teaching\bmi826-f21\homework\pyg-yeast-hw\graph_gps.py", line 26, in <module>
    train_dataset = ZINC(path, subset=True, split='train', pre_transform=transform)
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\datasets\zinc.py", line 100, in __init__
    super().__init__(root, transform, pre_transform, pre_filter)
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\data\in_memory_dataset.py", line 76, in __init__
    super().__init__(root, transform, pre_transform, pre_filter, log)
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\data\dataset.py", line 102, in __init__
    self._process()
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\data\dataset.py", line 235, in _process
    self.process()
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\datasets\zinc.py", line 162, in process
    data = self.pre_transform(data)
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\transforms\base_transform.py", line 32, in __call__
    return self.forward(copy.copy(data))
  File "C:\xxxxx\.conda\envs\pyg-2023\lib\site-packages\torch_geometric\transforms\add_positional_encoding.py", line 144, in forward
[...]

PR #8225, opened 2023-10-19
[Transforms] Fix sparse-sparse matrix multiplication support on Windows (Part 2)

EdisonLeeeee:
Address #8219


rusty1s (approved):
Oh, is that a bug in PyTorch?

Issue #4698, opened 2022-05-24
Bug: Pre-trained `DimeNet++` performance on QM9


rusty1s:
### 🐛 Describe the bug

As discussed in https://github.com/pyg-team/pytorch_geometric/pull/4432, loading the pre-trained `DimeNetPlusPlus` model yields incorrect performance on QM9:
```
Target: 00, MAE: 4.85522 ± 3.42503
Target: 01, MAE: 15.25113 ± 6.21239
Target: 02, MAE: 603.06146 ± 484.24606
Target: 03, MAE: 994.61578 ± 703.35645
Target: 05, MAE: 977.33118 ± 332.41879
Target: 06, MAE: 594.56433 ± 317.66690
Target: 07, MAE: 3095.07080 ± 1889.53760
```
cc: @arunppsg

PR #8239, opened 2023-10-21
Fix Pre-trained `DimeNet++` performance on QM9

xnuohz:
Fix https://github.com/pyg-team/pytorch_geometric/issues/4698
```
Target: 00, MAE: 0.02975 ± 0.05869
Target: 01, MAE: 0.04322 ± 0.15740
Target: 02, MAE: 24.43286 ± 35.56775
Target: 03, MAE: 19.42164 ± 31.41084
Target: 05, MAE: 0.28941 ± 0.66366
Target: 06, MAE: 1.21997 ± 2.13234
Target: 07, MAE: 6.15220 ± 13.26018
Target: 08, MAE: 6.20371 ± 12.61268
Target: 09, MAE: 6.48553 ± 14.85413
Target: 10, MAE: 7.41481 ± 14.50145
Target: 11, MAE: 0.02268 ± 0.02604
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed Pre-trained `DimeNet++` performance on QM9 ([#8239](https://github.com/pyg-team/pytorch_geometric/pull/8239))
 - Fixed `GNNExplainer` usage within `AttentiveFP` ([#8216](https://github.com/pyg-team/pytorch_geometric/pull/8216))
 - Fixed `to_networkx(to_undirected=True)` in case the input graph is not undirected ([#8204](https://github.com/pyg-team/pytorch_geometric/pull/8204))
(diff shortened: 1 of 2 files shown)


rusty1s:
Thank you. I would assume this change will break `DimeNet` due to the difference in their implementation?

xnuohz:
Actually, it shouldn't be. I'll continue to fix this.

xnuohz:
@rusty1s Checked the origin code again(See [dimenet](https://github.com/gasteigerjo/dimenet/blob/66a8168ddb3c6e972977c7fb7d2143d6baee1402/dimenet/model/dimenet.py#L83-L94) and [dimenet++](https://github.com/gasteigerjo/dimenet/blob/66a8168ddb3c6e972977c7fb7d2143d6baee1402/dimenet/model/dimenet_pp.py#L92-L103)), and the author indeed used different angle calculations for these 2 models. Seems it's not consistent with the paper, anyway we just align with the author's code.

rusty1s:
Yeah, can we somehow condition that on which DimeNet version we are using?

rusty1s (approved):

Issue #5657, opened 2022-10-12
[Community Sprint] Add missing type hints and TorchScript support 🚀


rusty1s:
### 🚀 The feature, motivation and pitch

**We are kicking off our very first community sprint!**

The community sprint resolves around **adding missing type hints and TorchScript support** for various functions across PyG, aiming to improve and clean-up our core codebase. Each individual contribution is designed to only take **around 30 minutes to two hours** to complete.

The sprint begins **Wednesday October 12th** with a kick off meeting at 8am PST. The community sprint will **last 2 weeks** and we will have another live hangouts when the sprint has completed. If you are interested in helping out, please also join our **[PyG slack](https://data.pyg.org/slack.html) channel `#community-sprint-type-hints`** for more information. **You can assign yourself to the type hints you are planning to work on [here](https://docs.google.com/spreadsheets/d/1QpwREo8_iQ0sfnXCa22i1b1XEM9Udxu7X16Z77eyRg8/edit?usp=sharing).**

# 🚀 Add missing type hints and TorchScript support

Type hints are currently used inconsistently in the `torch-geometric` repository, and it would be nice to make them a complete, consistent thing across all datasets, models and utilities. Adding type hint support in models also helps us to improve our TorchScript coverage across layers and models provided in `nn.*`.

## Example

Take a look at the current implementation of [`contains_isolated_nodes`](https://github.com/pyg-team/pytorch_geometric/blob/56610200da56fc43ae7ba4ed27b207cae6c7613c/torch_geometric/utils/isolated.py#L8):

```python
def contains_isolated_nodes(edge_index, num_nodes=None):
    num_nodes = maybe_num_nodes(edge_index, num_nodes)
    edge_index, _ = remove_self_loops(edge_index)
    return torch.unique(edge_index.view(-1)).numel() < num_nodes
 ```
 
 Adding type hints support to the function signature helps us to better understand its input and output, improving code readability:
 
 ```python
def contains_isolated_nodes(
    edge_index: Tensor,
    num_nodes: Optional[int] = None,
) -> bool:
    ...
```

Importantly, it also lets us use it as part of a TorchScript model. Without it, all arguments that miss type hints are expected to be PyTorch tensors (which is clearly not the case for the `num_nodes` argument). Without it, `torch.jit.script` compilation will fail:

```python
import torch

from torch_geometric.utils import contains_isolated_nodes

contains_isolated_nodes = torch.jit.script(contains_isolated_nodes)

[...]

PR #8254, opened 2023-10-24
Enable typing support

akihironitta:
Addresses https://github.com/pyg-team/pytorch_geometric/discussions/3429#discussioncomment-6755720.
Related to #5657.

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well.

https://peps.python.org/pep-0561/#packaging-type-information

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -66,6 +66,5 @@ jobs:
       - name: Check type hints
         continue-on-error: true
         run: |
-          # https://github.com/python/mypy/issues/10863
-          mkdir -p .mypy_cache
+          mkdir -p .mypy_cache  # https://github.com/python/mypy/issues/10863
           mypy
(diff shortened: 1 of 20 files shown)


akihironitta:
I'm not very familiar with [TorchScript](https://pytorch.org/docs/stable/jit_language_reference_v2.html), but it seems like there's an issue with around annotations for TorchScript:
https://github.com/pyg-team/pytorch_geometric/blob/699120e2584077b0113eb9519e58c5e3dbcc1ec6/torch_geometric/nn/conv/edge_conv.py#L116-L120

```concole
$ mypy
torch_geometric/nn/conv/edge_conv.py:120: error: expected an indented block  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```
https://github.com/pyg-team/pytorch_geometric/actions/runs/7072892589/job/19252222130?pr=8254

---

I would appreciate any help from the PyG community ❤️

PR #8269, opened 2023-10-26
Add alternative edge indices shape to `GCNConv` input description

666even666:
Screenshot 2023-10-26 at 1 05 22 PM
According to the implementation https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/gcn_conv.py#L49, sparse adjacency matrix of shape |V|*|V| can also be used as edge_indices.

### DIFF
--- a/torch_geometric/nn/conv/gcn_conv.py
+++ b/torch_geometric/nn/conv/gcn_conv.py
@@ -159,7 +159,8 @@ class GCNConv(MessagePassing):
     Shapes:
         - **input:**
           node features :math:`(|\mathcal{V}|, F_{in})`,
-          edge indices :math:`(2, |\mathcal{E}|)`,
+          edge indices :math:`(2, |\mathcal{E}|)`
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/conv/gcn_conv.py
```diff
@@ -159,7 +159,8 @@ class GCNConv(MessagePassing):
     Shapes:
         - **input:**
           node features :math:`(|\mathcal{V}|, F_{in})`,
-          edge indices :math:`(2, |\mathcal{E}|)`,
+          edge indices :math:`(2, |\mathcal{E}|)`
+          or :math:`(|\mathcal{V}|, |\mathcal{V}|)` *(sparse)*,
```
rusty1s, line 163:
```suggestion
          or sparse matrix :math:`(|\mathcal{V}|, |\mathcal{V}|)`,
```

rusty1s (approved):

Issue #8272, opened 2023-10-27
`torch_gemoetric.data.Data.subgraph()` fails on graph without edges


QuentinSoubeyranAqemia:
### Description
Calling `torch_geometric.data.Data.subgraph` on a graph with no edge (no `edge_index` attribute) raises an `AttributeError: 'NoneType' object has no attribute 'device'`.

### Expected behavior
The method correclty returns a subgraph of the current graph, with only the `pos`, `x`, `y`, etc... for the selected nodes.

### Minimal Reproducible Example

```python
import torch
from pytorch_geometric import data

graph = data.Data(
    pos=torch.Tensor([[0, 0, 0], [1, 1, 1]]),
    x=torch.Tensor([[0], [1]]),
)
graph.subgraph(torch.LongTensor([0]))  # AttributeError: NoneType has no attribute 'device'
```

### Workaround
Define an empty `edge_index` attribute (`torch.LongTensor([[], []])`) before calling subgraph, and remove it afterward

### Context
The edges are not yet define on my graphs because they are built afterward by code I do not control. After loading data, graphs I produce contains many nodes which I don't need, which I thus filter out with `.subgraph()` (at data-loading time) to produce smaller graph and reduce load.

### Details of the bug

Bug is due to:
1. `torch_geometric.data.Data.edge_index` is an [optional attribute (doc)](https://pytorch-geometric.readthedocs.io/en/latest/get_started/introduction.html#data-handling-of-graphs) that might be missing (e.g. graphs with only nodes, where edges are to be built later)
2. [`torch_geometric.data.Data.subgraph`](https://github.com/pyg-team/pytorch_geometric/blob/3752d94447950e83e9d6e03c1332d0c031a2376c/torch_geometric/data/data.py#L626-L634) unconditionally passes `self.edge_index` to `torch_geometric.utils.subgraph.subgraph()`, even though it might be undefined and thus `None`
3. [`torch.utils.subgraph.subgraph()`](https://github.com/pyg-team/pytorch_geometric/blob/3752d94447950e83e9d6e03c1332d0c031a2376c/torch_geometric/utils/subgraph.py#L42) expects a `torch.Tensor` as the `edge_index` attribute

Not sure which `subgraph()` routines needs to be fixed. I'd lean toward `utils.subgraph.subgraph()` because it makes sense to take the subgraph of an edge-less graph.

### Environment

* PyG version: `2.4.0` (bug still present in repo `master` source code AFAIK)
[...]

PR #8276, opened 2023-10-28
Fix `torch_geometric.utils.subgraph` on graph without edges

xnuohz:
Fix https://github.com/pyg-team/pytorch_geometric/issues/8272


rusty1s:
Oh, it looks like we worked on a fix for this in parallel, really sorry. Here is my PR: https://github.com/pyg-team/pytorch_geometric/pull/8276

PR #8282, opened 2023-10-29
Replace `Union` type hints

rusty1s:


EdisonLeeeee:
@rusty1s Should we also replace something like `Optional[int]` with `int | None`?

rusty1s:
> Should we also replace something like Optional[int] with int | None?

What do you prefer? I actually like `Optional` syntax more (but no strong opinion).

EdisonLeeeee:
> > Should we also replace something like Optional[int] with int | None?
> 
> What do you prefer? I actually like `Optional` syntax more (but no strong opinion).

I prefer using `| None` as it explicitly indicates the optional `None` choices, especially in cases where `None` is not the default value. I don't have a strong opinion either - just a proposal :)

rusty1s:
Looks like we cannot merge this PR here yet anyway due to broken JIT tests.

PR #8288, opened 2023-10-30
Add multi GPU training benchmarks

DamianSzwichtenberg:
Benchmarks currently cover only homogeneous cases. All tested configurations show improvement in loss over epochs.


benchmark/multi_gpu/training/training_benchmark.py
```diff
@@ -0,0 +1,300 @@
...
+
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
+
+    if args.device == 'xpu':
+        model, optimizer = ipex.optimize(model, optimizer=optimizer)
+
+    train = train_hetero if hetero else train_homo
+
+    if rank == 0:
+        maybe_synchronize(args.device)
```
akihironitta, line 201: n00b question: Is it correct that it's synchronised on rank zero only? Should it be on all ranks?

akihironitta (commented):
🚀

benchmark/multi_gpu/training/training_benchmark.py
DamianSzwichtenberg: That's a very good question and a good catch. Indeed I should synchronize on all ranks. Thanks 😉 just fixed that.

JakubPietrakIntel (approved):
LGTM

PR #8297, opened 2023-10-31
Do not silently override configurations in `ConfigStore`; make duplicate node logic more robust

rusty1s:


vid-koci (approved):
Approve to unblock, with two comments:
 - The code doesn't prevent us from registering the same data class multiple times, but it will error out when it's trying to access the config store. Ideally, it'll already error out when registering a data class that's already in the config store.
 - tests missing

PR #8317, opened 2023-11-03
[Refactor] Replace `torch.save` and `torch.load` with `InMemoryDataset.save` and `InMemoryDataset.load`

EdisonLeeeee:
A follow-up PR of  #7250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Disallow the usage of `add_self_loops=True` in `GCNConv(normalize=False)` ([#8210](https://github.com/pyg-team/pytorch_geometric/pull/8210))
 - Disable device asserts during `torch_geometric.compile` ([#8220](https://github.com/pyg-team/pytorch_geometric/pull/8220))
-- Replace `torch.save` and `torch.load` with `InMemoryDataset.save` and `InMemoryDataset.load` ([#8317](https://github.com/pyg-team/pytorch_geometric/pull/8317))
 
 ### Deprecated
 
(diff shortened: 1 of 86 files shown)


torch_geometric/datasets/actor.py
```diff
@@ -50,7 +50,7 @@ class Actor(InMemoryDataset):
     def __init__(self, root: str, transform: Optional[Callable] = None,
                  pre_transform: Optional[Callable] = None):
         super().__init__(root, transform, pre_transform)
-        self.data, self.slices = torch.load(self.processed_paths[0])
+        self.load(self.processed_paths[0], data_cls=Data)
```
akihironitta, line 53: nit: do we need to specify `data_cls`?
```suggestion
        self.load(self.processed_paths[0])
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Disallow the usage of `add_self_loops=True` in `GCNConv(normalize=False)` ([#8210](https://github.com/pyg-team/pytorch_geometric/pull/8210))
 - Disable device asserts during `torch_geometric.compile` ([#8220](https://github.com/pyg-team/pytorch_geometric/pull/8220))
+- Replace `torch.save` and `torch.load` with `InMemoryDataset.save` and `InMemoryDataset.load` ([#8317](https://github.com/pyg-team/pytorch_geometric/pull/8317))
```
akihironitta, line 23: (nit: IMHO we don't need a changelog entry for internal refactor that doesn't impact users, but feel free to ignore my comment :))

```suggestion
```

akihironitta (approved):
Neat! ✨

torch_geometric/datasets/actor.py
EdisonLeeeee: Good catch! Let's drop it.

torch_geometric/datasets/rel_link_pred_dataset.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, root: str, name: str,
         self.name = name
         assert name in ['FB15k-237']
         super().__init__(root, transform, pre_transform)
-        self.data, self.slices = torch.load(self.processed_paths[0])
+        self.load(self.processed_paths[0])
```
rusty1s, line 55: here again.

EdisonLeeeee:
@rusty1s Thank you for the fix!

Issue #8271, opened 2023-10-27
Add NDCG@K for link prediction performance evaluation


songsong0425:
Dear PyG community,

Hi, always thank you for your effort in package maintenance and development.
When I tried to implement the link prediction metrics (i.e., hits@k, ndcg@k), I couldn't find the example code for ndcg@k.
Can you consider adding the example about the ndcg@k for the evaluation metrics?
If it has already existed and I missed it, can anyone tell me where can I find it?

Thank you for reading!

_Originally posted by @songsong0425 in https://github.com/pyg-team/pytorch_geometric/discussions/7792_

SimonPop:
Hi there! 

I would gladly work on that subject :)

I have some questions with regard to the implementation details.

## Location of the implementation

I noticed that the `hits@k` metric was implemented as part of the `test` method of the KGE base class: `torch_geometric.nn.kge.base`. 

What would be the best approach to implement `ndcg@k` ? 

1. Add a method parameter (e.g. `metric`=`ndcg` / `hits`) to the test function to choose which to compute (`hits@k` or `ndcg@k`) and compute it inside this test method.
2. Implement a separate method with a new name only for computing `ndcg@k`(e.g. `test_ndgck`).
3. Compute both in the method (but the signature will change, which is a breaking change `Tuple[float, float]` -> `Tuple[float, float, float, float]`).
4. Only implement it in an example but not in the library itself.
5. Somewhere else? 

## Details of implementation

In its general form, if I understand it correctly, NDCG allows gains from different possible tails for a single head. For example, for a link prediction task, any of the neighbors (tails) of a given node (head) can bring a gain. 

Currently, in the hits@k, we allow a single tail per head. 

Do we want to let the user give a list of different tails (fixed number or flexible)? (again, a possible change in the signature of the method)

```python
def test(
        self,
        head_index: Tensor,
        rel_type: Tensor,
        tail_indexes: List[Tensor],
        batch_size: int,
        k: int = 10,
        log: bool = True,
    )
```

Or do we want to keep the one head - one tail limit? 

```python
def test(
        self,
        head_index: Tensor,
        rel_type: Tensor,
        tail_index: Tensor,
        batch_size: int,
        k: int = 10,
        log: bool = True,
    )
```

Thanks! :)

rusty1s:
I would prefer we implement these as metrics that follow the design of `torchmetrics`, such that we can compute them in a mini-batch fashion during training. This would then be something like
```python
[...]

PR #8326, opened 2023-11-05
`NDCG@k` metric for link-prediction

SimonPop:
Hello!

This branch implements an example of link prediction using the [ndcg@k](https://en.wikipedia.org/wiki/Discounted_cumulative_gain) metric.

This has been discussed in the #8271 issue. 

I have tried to first implement it in a new example as a starting point. It probably should be moved somewhere else, or have a more distinctive example. This example is a simple copy from the `link_prediction.py` example, replacing AUC with NDCG. 

There remain some questions about the signature of the `update` method, as well as how to call the method in the example.

For example, we might want to demonstrate that we can call it during training, adding mini-batches one at a time. 

We might also want to avoid predicting links from another set, that way, we are not considering valid links as undesired.


examples/link_pred_ndgck.py
```diff
@@ -0,0 +1,163 @@
...
+
+if torch.cuda.is_available():
+    device = torch.device('cuda')
+elif hasattr(torch.backends, 'mps') and torch.backends.mps.is_available():
+    device = torch.device('mps')
+else:
+    device = torch.device('cpu')
+
+
+class NdcgK():
```
rusty1s, line 18: Can we move this to `torch_geometric` directly? For example, in `torch_geometric.metrics`?

SimonPop: Yes, of course, I will be doing that.

rusty1s (approved):

PR #8345, opened 2023-11-08
Support for multiple aggregations in `torch.compile`

rusty1s:


torch_geometric/nn/aggr/fused.py
```diff
@@ -282,8 +283,9 @@ def forward(self, x: Tensor, index: Optional[Tensor] = None,
             outs[i] = (pow_sum / count) - (mean * mean)
 
         # Compute `StdAggregation` last:
-        i = self.aggr_index.get('StdAggregation')
-        if i is not None:
+        if 'StdAggregation' in self.aggr_index:
+            i = self.aggr_index['StdAggregation']
```
akihironitta, line 287: Ah! Interesting :) I thought there'd be no fix for this on our side.

PR #8356, opened 2023-11-09
Fix batching issue for `HeteroData` that is converted using `ToSparseTensor` when `torch_sparse` is not installed

### LINKED ISSUE #8355: Incorrect batching with graphs converted by `ToSparseTensor()`
### 🐛 Describe the bug

I've been debugging my code and have run into an issue with my custom dataset - at present, I have it return a `HeteroData` object that has been converted with the `ToSparseTensor()` transform to construct the sparse form (`adj_t`). I then feed this dataset into the torch_geometric custom dataloader (`torch_geometric.loader.DataLoader`).

However, when I get batches from this dataloader, the `adj_t` of the batched graph seems wildly incorrect (that is, it appears to properly extend the sparse adjacency matrix along just one dimension, but not both, leading to incorrect indices). 

Minimally reproducible code snippet:

```
import torch
import torch_geometric

rng = torch.Generator().manual_seed(9)
bsize = 2

h_dataset = torch_geometric.datasets.FakeHeteroDataset(num_graphs=128, 
                                                  num_node_types=4, 
                                                  num_edge_types=16, 
                                                  avg_num_nodes=1024, 
                                                  num_classes=2, 
                                                  transform=torch_geometric.transforms.ToSparseTensor())


train_ds, test_ds = torch.utils.data.random_split(h_dataset, [64, 64], rng)

for i in range(bsize):
    print(train_ds[i])

train_dl = torch_geometric.loader.DataLoader(train_ds, shuffle=False, batch_size=2, drop_last=False)
batched_data = next(iter(train_dl))
print(batched_data)
```

Specifically, you'll see that the printing of the first two graphs in the train_ds (before batching) produces values around 1000 for the number of each node type and the adjacency matrices match appropriately. 

For example, from my environment, the edge data contains: `(v2, e0, v2)={ adj_t=[1237, 1237] }` and `(v2, e0, v2)={ adj_t=[1005, 1005] }` for the two graphs.

After batching, the edge data contains: `(v2, e0, v2)={ adj_t=[2242, 1237] }` - which appears as if the batching worked along the first dimension but not the second.

### Environment

* PyG version: 2.4.0
* PyTorch version: 2.1.0
* OS: Mac OSX 
* Python version: 3.10.13
* CUDA/cuDNN version: N/A
* How you installed PyTorch and PyG (`conda`, `pip`, source): pip
* Any other relevant information (*e.g.*, version of `torch-scatter`): N/A

rachitk:
Fixes #8355.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `GNNExplainer` usage within `AttentiveFP` ([#8216](https://github.com/pyg-team/pytorch_geometric/pull/8216))
 - Fixed `to_networkx(to_undirected=True)` in case the input graph is not undirected ([#8204](https://github.com/pyg-team/pytorch_geometric/pull/8204))
 - Fixed sparse-sparse matrix multiplication support on Windows in `TwoHop` and `AddRandomWalkPE` transformations ([#8197](https://github.com/pyg-team/pytorch_geometric/pull/8197), [#8225](https://github.com/pyg-team/pytorch_geometric/pull/8225))
+- Fixed batching of `HeteroData` converted using `ToSparseTensor()` ([#8356](https://github.com/pyg-team/pytorch_geometric/pull/8356))
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


rusty1s (approved):
Thank you!

PR #8363, opened 2023-11-10
Avoid a graph break in `ModuleDict` and `ParameterDict`

akihironitta:
We have a graph break at `hasattr` call in `ModuleDict.to_internal_key`.

Repro:

```python
import torch
from torch_geometric.nn.module_dict import ModuleDict

edge_type = ("a", "to", "b")

class SomeModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.module_dict = ModuleDict({
            edge_type: torch.nn.Linear(1, 1),
        })

    def forward(self, x):
        # need to convert tuple to string in advance to avoid a graph break
        # due to https://github.com/pytorch/pytorch/issues/111551
        key = ModuleDict.to_internal_key(edge_type)
        x = self.module_dict[key](x)
        return x

from torch._dynamo.utils import CompileProfiler
model = SomeModel()
with CompileProfiler() as prof:
    model = torch.compile(model)
    model(torch.randn(1, 1))
    print(prof.report())
```

### DIFF
--- a/test/nn/test_module_dict.py
+++ b/test/nn/test_module_dict.py
@@ -97,7 +97,6 @@ def test_compile_module_dict(device):
             x = self.module_dict[key](x)
             return x
 
-
     x = torch.randn(1, 1, device=device)
     module = TestModule().to(device)
     explanation = dynamo.explain(module)(x)
--- a/test/nn/test_module_dict.py
+++ b/test/nn/test_module_dict.py
@@ -11,9 +11,9 @@ from torch_geometric.testing import (
 
 
 def test_internal_external_key_conversion():
-    assert ModuleDict.to_internal_key('a.b') == 'a#b'
-    assert ModuleDict.to_internal_key('ab') == 'ab'
-    assert ModuleDict.to_internal_key('a.b.c') == 'a#b#c'
+    assert ModuleDict.to_internal_key('a.b') == '<a#b>'
+    assert ModuleDict.to_internal_key('ab') == '<ab>'
+    assert ModuleDict.to_internal_key('a.b.c') == '<a#b#c>'
     assert ModuleDict.to_internal_key(('a', 'b')) == '<a___b>'
     assert ModuleDict.to_internal_key(('a.b', 'c')) == '<a#b___c>'
     assert ModuleDict.to_internal_key('type') == '<type>'
@@ -32,7 +32,7 @@ def test_dot_syntax_keys():
         'model.sub_model.lin3': torch.nn.Linear(4, 4),
     })
 
-    expected_keys = {'lin1', 'model.lin2', 'model.sub_model.lin3'}
+    expected_keys = {'<lin1>', '<model.lin2>', '<model.sub_model.lin3>'}
     assert set(module_dict.keys()) == expected_keys
     assert set([key for key, _ in module_dict.items()]) == expected_keys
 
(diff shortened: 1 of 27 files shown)


test/nn/test_module_dict.py
```diff
@@ -68,3 +74,36 @@ def test_reserved_keys():
...
+
+        def forward(self, x):
+            key = ModuleDict.to_internal_key(edge_type)
+            x = self.module_dict[key](x)
+            return x
+
+    x = torch.randn(1, 1, device=device)
+    module = TestModule().to(device)
+    explanation = dynamo.explain(module)(x)
+    assert explanation.graph_break_count == 0
```
akihironitta, line 103: Even without the fix, this test case always passes somehow while
```python
import torch
from torch_geometric.nn.module_dict import ModuleDict

edge_type = ("a", "to", "b")

class SomeModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.module_dict = ModuleDict({
            edge_type: torch.nn.Linear(1, 1),
        })

    def forward(self, x):
        key = ModuleDict.to_internal_key(edge_type)
        x = self.module_dict[key](x)
        return x

from torch._dynamo.utils import CompileProfiler
model = SomeModel()
with CompileProfiler() as prof:
    model = torch.compile(model)
    model(torch.randn(1, 1))
    print(prof.report())
```

says there're graph breaks:

```
Torchdynamo Profiler Report
===========================

Graph Breaks
------------
Graph breaks happen when torchdynamo encounters code it can't safely trace.
If you want to find out why breaks are happening, check below for each break reason
You may gain additional insight by passing `fullgraph=True` to torch.compile,
to stop at the first break.

Graph Break Reason                     Count
-----------------------------------  -------
hasattr: UserDefinedClassVariable()        1
hasattr no source                          2

Recompilation
-------------
These subgraphs were recompiled more than once due to guard failures
Guard failures indicate some condition assumed to be static by the tracer changed,
making it unsafe to reuse the compiled program.

No recompilation detected.
```

akihironitta: https://github.com/pytorch/pytorch/issues/113443

torch_geometric/nn/module_dict.py
```diff
@@ -30,9 +30,8 @@ def to_internal_key(cls, key: Key) -> str:
             key = f"<{'___'.join(key)}>"
         assert isinstance(key, str)
 
-        # ModuleDict cannot handle keys that exists as class attributes:
-        if hasattr(cls, key):
-            key = f'<{key}>'
+        # Always wrap key with <> to avoid conflicts with class attributes
+        key = f'<{key}>'
```
akihironitta, line 34: @rusty1s looks like `hasattr` call produces a graph break (https://github.com/pyg-team/pytorch_geometric/pull/8363/files#r1389351664). What do you think would be a more reasonable fix here? The current fix will be a breaking change...

akihironitta:
Converting this to draft as it's currently not a proper fix.

torch_geometric/nn/module_dict.py
rusty1s: I think this fix is okay. An alternative fix would be to pre-compute the class attributes and do `key in CLASS_ATTRIBUTES`. WDYT?

akihironitta: Update this PR to accomondate the alternative fix. Confirmed it also produces no graph break, so this PR is ready for your another review!

rusty1s (approved):

PR #8367, opened 2023-11-13
Fix homo temporal distributed sampling

kgajdamo:
First get node attributes and then extract node_types to prevent node_attrs from being always be empty in case of homo sampling.

### DIFF
--- a/torch_geometric/sampler/neighbor_sampler.py
+++ b/torch_geometric/sampler/neighbor_sampler.py
@@ -119,11 +119,10 @@ class NeighborSampler(BaseSampler):
             feature_store, graph_store = data
 
             # Obtain graph metadata:
-            node_attrs = [
-                attr for attr in feature_store.get_all_tensor_attrs()
-                if isinstance(attr.group_name, NodeType)
-            ]
-            self.node_types = list(set(attr.group_name for attr in node_attrs))
+            node_attrs = feature_store.get_all_tensor_attrs()
+            self.node_types = list(
+                set(attr.group_name for attr in node_attrs
+                    if isinstance(attr.group_name, NodeType)))
 
             edge_attrs = graph_store.get_all_edge_attrs()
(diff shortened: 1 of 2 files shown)


torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -119,11 +119,10 @@ def __init__(
             feature_store, graph_store = data
 
             # Obtain graph metadata:
-            node_attrs = [
-                attr for attr in feature_store.get_all_tensor_attrs()
-                if isinstance(attr.group_name, NodeType)
-            ]
-            self.node_types = list(set(attr.group_name for attr in node_attrs))
+            node_attrs = feature_store.get_all_tensor_attrs()
```
rusty1s, line 122: Can you clarify for me why this fixes temporal sampling?

rusty1s: But `time_attrs` is a subset of `node_attrs` for which group names are a node type. If the `node_attrs` are empty here, I would expected `time_attrs` to be empty as well?

kgajdamo: It is about the condition that checks whether the attr.group_name is of NodeType. In the case of homo graphs, group name will always be None because there are no node types. Therefore, in the case of homo graphs, node_attrs will always be empty.

kgajdamo: It is needed because for some reason get_all_tensor_attrs returns edge_types as well.

kgajdamo: I'm sorry but I need to go. I'll check that tomorrow.

rusty1s: Ok, that makes sense to me.

rusty1s (approved):

PR #8368, opened 2023-11-13
Cleanup multinode papers100m `get_num_workers`

puririshi98:
address comments from discussion at end of https://github.com/pyg-team/pytorch_geometric/pull/8071

### DIFF
--- a/examples/multi_gpu/multinode_multigpu_papers100m_gcn.py
+++ b/examples/multi_gpu/multinode_multigpu_papers100m_gcn.py
@@ -19,15 +19,15 @@ from torch_geometric.loader import NeighborLoader
 from torch_geometric.nn import GCNConv
 
 
-def get_num_workers(world_size: int) -> int:
+def get_num_workers() -> int:
     num_workers = None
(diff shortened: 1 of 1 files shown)


examples/multi_gpu/multinode_multigpu_papers100m_gcn.py
```diff
@@ -19,15 +19,15 @@
 from torch_geometric.nn import GCNConv
 
 
-def get_num_workers(world_size: int) -> int:
+def get_num_workers() -> int:
     num_workers = None
     if hasattr(os, "sched_getaffinity"):
```
akihironitta, line 24: > I also wonder, on which platform with pyg and CUDA `os.sched_getaffinity` is not available and then you have to fallback to `os.cpu_count`...

@flxmr `os.sched_getaffinity` is only available on UNIX systems: https://docs.python.org/3/library/os.html#interface-to-the-scheduler
but since this script is already specialised for a slurm cluster, I think we could also drop it. I'm fine as is though.

akihironitta (approved):

rusty1s (approved):

PR #8369, opened 2023-11-13
Add faster `mask_select` method

toenshoff:
This PR addresses a minor performance issue with the `utils.mask.mask_select` method.

The previous implementation unsqueezed the mask to be broadcastable along all dimensions.
This broadcasted masking is significantly slower in torch than applying a 1-dimensional mask to the first dimension of a tensor.
The new implementation simply transposes the tensor such that the masked dimension comes first and later undoes this.

This mainly affects the methods for computing subgraphs which are needlessly slow on bigger graphs with many features.

Here is a code example to illustrate the difference:
```python
import torch
from timeit import timeit


N = 1000000
E = 10000000

data = Data(
    x=torch.randn((N, 200)),
    edge_index=torch.randint(0, N, (2, E)),
    edge_attr=torch.randn((E, 200), dtype=torch.float16)
)

subset_bool = torch.rand((N,)) < 0.5
time_bool = timeit(lambda: data.subgraph(subset_bool), number=1)
print(f'Time: {time_bool:.4f}s')
```

The prior version yields:
```bash
Time: 13.7826s
```
The updated version yields:
```bash
Time:  0.3541s
```

So more than a 30-fold speedup. We thought this may be helpful to accelerate workflows involving subgraph extraction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Update `utils.mask.mask_select` with faster method ([#8369](https://github.com/pyg-team/pytorch_geometric/pull/8369))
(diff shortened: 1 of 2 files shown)


akihironitta (commented):
LGTM!

rusty1s (approved):
Great catch, thank you!

PR #8375, opened 2023-11-14
Enable homo distributed link sampling

kgajdamo:
This implementation covers distributed homo edge sampling. Hetero is in progress.

The edge_sample() function is almost the same as the one from the `NeighborSampler` class. The difference is that it calls the node_sample() from the distributed package.
Why I decided to copy paste it to the DistSampler?:
- to not interfere with the original edge_sample() function from the `NeighborSampler`
- To make this function async (seamlessly)
- In case of hetero there will be more changes needed

This PR introduces support for edge sampling:
- disjoint, negative, temporal, bidirectional
- with unit tests 
- unit tests for edge-level temporal sampling


torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -164,6 +164,8 @@ def __init__(
         else:  # self.data_type == DataType.remote
             feature_store, graph_store = data
 
+            self.edge_time = None  # TODO(kgajdamo): only for the tests to pass
```
rusty1s, line 167: Can you clarify?

torch_geometric/loader/link_loader.py
```diff
@@ -262,6 +262,7 @@ def filter_fn(
             data.num_sampled_edges = out.num_sampled_edges
 
             data.input_id = out.metadata[0]
+            data.batch_size = out.metadata[0].size(0)
```
rusty1s, line 265: Is this needed? This is not necessarily correct.

rusty1s (approved):

torch_geometric/sampler/neighbor_sampler.py
kgajdamo: This PR was added before I added support for edge-level sampling. I forgot to remove that.

torch_geometric/loader/link_loader.py
kgajdamo: I added this line because the test for the DistLinkNeighborLoader was failing because it expected the batch_size field in the batch. I deleted that line. Thank you for your comments.

PR #8377, opened 2023-11-14
Fix output metadata processing in loaders 'filter_fn()` for multinode

jjpietrak:
The output of `DistributedNeighborSampler._collate_fn()` includes `nfeats, nlabels, efeats` which are our `metadata[-3]`, `metadata[-2]` and `metadata[-1]`. Here: [L717C1-L718C22](https://github.com/pyg-team/pytorch_geometric/blob/88d7986b6d0a6de5895872270d2ff4fc95fae3b7/torch_geometric/distributed/dist_neighbor_sampler.py#L717C1-L718C22)

### DIFF
--- a/torch_geometric/loader/link_loader.py
+++ b/torch_geometric/loader/link_loader.py
@@ -224,23 +224,20 @@ class LinkLoader(torch.utils.data.DataLoader, AffinityMixin):
                     self.link_sampler.edge_permutation)
 
             else:  # Tuple[FeatureStore, GraphStore]
+                edge_index = torch.stack([out.row, out.col])
+                data = Data(edge_index=edge_index)
+
+                # TODO Respect `custom_cls`.
+                # TODO Integrate features.
+
                 # Hack to detect whether we are in a distributed setting.
-                if self.node_sampler.__class__.__name__ == "DistNeighborSampler":
-                    # metadata entries are populated in `DistributedNeighborSampler._collate_fn()`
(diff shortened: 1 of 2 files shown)


torch_geometric/loader/link_loader.py
```diff
@@ -224,11 +224,21 @@ def filter_fn(
...
             else:  # Tuple[FeatureStore, GraphStore]
-                # TODO Respect `custom_cls`.
-                # TODO Integrate features.
                 edge_index = torch.stack([out.row, out.col])
                 data = Data(edge_index=edge_index)
 
+                # TODO Respect `custom_cls`.
+                # TODO Integrate features.
+
+                # Hack to detect whether we are in a distributed setting.
```
rusty1s, line 233: Thanks. I misunderstood the implementation, sorry :)

From a PyG interface perspective, I personally think it is quite confusing that `SamplerOutput` also returns the features, but I see that it makes sense to fetch the features during sampling. However, this does not really seem to generalize to more feature attributes (e.g., `time`), so let me know if there exists plans to extend this to more general cases.

rusty1s (approved):

torch_geometric/loader/link_loader.py
JakubPietrakIntel: No worries :) thanks for correcting that quickly. You're right, we will need revise our Sampler output to be more customizable. Remark noted!

PR #8379, opened 2023-11-14
Add filesystem abstraction using `fsspec`

asherbondy:
**Changes made:**
- Add fs_utils.py for API that offers 1:1 replacement for current filesystem usage.
- Update usages of open, glob, makedirs, normpath to use new API.
- Update test/datasets/* to use in-memory filesystem, simplifying test and testing fsspec API.


asherbondy:
Hey folks, Just in case it is not clear from the PR title. I would like to make sure the proposed API is agreed upon based on the presented cases before attempting to carry the fs abstraction throughout the rest of the code base.

torch_geometric/data/fs_utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+import fsspec
+import torch
+from fsspec.core import url_to_fs
+
+DEFAULT_CACHE_PATH = '/tmp/pyg_simplecache'
+
+
+def get_fs(path: str) -> fsspec.AbstractFileSystem:
+    return url_to_fs(path)[0]
```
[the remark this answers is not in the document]
asherbondy: Added some documentation.

torch_geometric/data/fs_utils.py
```diff
@@ -0,0 +1,137 @@
...
+def fs_cp(path1: str, path2: str, kwargs1: Optional[dict] = None,
+          kwargs2: Optional[dict] = None, extract: bool = False,
+          cache_path: Optional[str] = DEFAULT_CACHE_PATH,
+          blocksize: int = 2 * 22):
+    # Initialize kwargs for source/destination.
+    kwargs1 = kwargs1 or {}
+    kwargs2 = kwargs2 or {}
+
+    # Cache result if the protocol is not local and we have a cache folder.
+    if get_fs(path1).protocol != 'file' and cache_path:
```
asherbondy, line 80: @rusty1s I'm pointing out 3 things that fs_cp is doing that might be surprising to help you focus on this function. Here we automatically add a simplecache stage to the path so that results of a remote source path will first go to a local filesystem cache location so that subsequent fetches will succeed without making remote requests, e.g., http, gs, s3, etc.

torch_geometric/data/fs_utils.py
```diff
@@ -0,0 +1,137 @@
...
+        path1 = f'tar://**::{path1}'
+    elif extract and path1.endswith('.zip'):
+        path1 = f'zip://**::{path1}'
+    else:
+        name = osp.basename(path1)
+        if extract and name.endswith('.gz') or name.endswith('.bz2'):
+            name = osp.splitext(name)[0]
+        path2 = osp.join(path2, name)
+
+    if '*' in path1:
```
asherbondy, line 101: @rusty1s Finally, 3rd thing, is that we need to handle multi-file copy differently than single file due to some quirks of the fsspec API.

rusty1s (approved):
Thanks. I think this is a great start. I moved the files to `torch_geometric.io` and suggest usage of
```python
from torch_geometric.io import fs
```
I think this is a bit cleaner.

The `cp` function for my taste seems a bit overloaded. Is it common practice to fuse downloading and extraction here? I think this is a bit error-prone. Made some updates but broke `memory://` support in `conftest`, since `fs.ls` doesn't return the protocol :(

PR #8381, opened 2023-11-15
Fix for new `pyg_lib.neighbor_sample` argument order

rusty1s:


test/distributed/test_dist_neighbor_loader.py
```diff
@@ -172,6 +172,7 @@ def dist_neighbor_loader_hetero(
 @pytest.mark.parametrize('num_parts', [2])
 @pytest.mark.parametrize('num_workers', [0])
 @pytest.mark.parametrize('async_sampling', [True])
+@pytest.mark.skip(reason="Breaks due to METIS timeout")
```
kgajdamo, line 175: It's probably because of the new sampler argument - edge_time. If we set this to "None" at the moment in [here](https://github.com/pyg-team/pytorch_geometric/blob/31b386b442f4800b35e963a90286d3486421bf7d/torch_geometric/distributed/dist_neighbor_sampler.py#L622) then test should pass.

rusty1s: Good point.

PR #8382, opened 2023-11-15
Add support for heterogeneous workloads in multi XPU benchmarks

DamianSzwichtenberg:
Fix for the following error:
```
RuntimeError: Modules with uninitialized parameters can't be used with DistributedDataParallel. Run a dummy forward pass to correctly initialize the modules.
```

HeteroGAT requires additional treatment, as the `lin_dst` from each of its convolutions is not initialized for `paper-cites-paper` relation (for more details, look at [this code](https://github.com/pyg-team/pytorch_geometric/blob/HEAD/torch_geometric/nn/conv/gat_conv.py#L233)).


JakubPietrakIntel (approved):
The fix seems valid, but I wonder if there's a way to include this on the GNN-layer level.
@rusty1s don't you think that we should still initialize 
`x_dst = self.lin_dst(x_dst).view(-1, H, C)`
[L235C13-L235C59](https://github.com/pyg-team/pytorch_geometric/blob/7357a34433c70a5d210f518e91b764095854463c/torch_geometric/nn/conv/gat_conv.py#L235C13-L235C59) when there's a self-loop on `x_src` ?

rusty1s:
Discussed with @DamianSzwichtenberg. We should update `GATConv` here, yes.

DamianSzwichtenberg:
> Discussed with @DamianSzwichtenberg. We should update `GATConv` here, yes.

A possible solution is available in #8397.

PR #8386, opened 2023-11-15
Add cuda to multigpu (xpu) bench

puririshi98:


DamianSzwichtenberg (approved):
I have pushed changes that separate CUDA and XPU starting points. This approach fixes all XPU-related errors + makes things a bit clearer (I've also tested CUDA path, though only using 1 GPU).

puririshi98:
thanks!, works for multigpu too

PR #8397, opened 2023-11-17
Fix issue with uninitialized module in `GAT+hetero+DDP` scenario

DamianSzwichtenberg:
This PR fixes the error described in #8382.


rusty1s (approved):
Hey, I pushed an alternative fix. I think this one should resolve your issues without any `is_unitialized` handling.

DamianSzwichtenberg:
> Hey, I pushed an alternative fix. I think this one should resolve your issues without any `is_unitialized` handling.

Yep, I confirm, it works. Thanks 😉

Issue #8226, opened 2023-10-19
Manipulation tools for Batches


mova:
### 🚀 The feature, motivation and pitch

Currently,  Batches should not manipulated once created.
Yet, I would be useful to be able 
- create batches from an index (`init_batch`)
- add features for the nodes/edges/graph (`add_?_attr`)
- concatenate and slice batches (`from_batch_list` and TBD)
without turning the batch into a list and batching the list again.
I have implemented these functions here:
https://github.com/DeGeSim/caloutils/blob/main/caloutils/utils/batch.py
Tests 
https://github.com/DeGeSim/caloutils/blob/main/tests/test_batch.py
Do you see any chance of integrating this? 
What would need to happen before that? 
Cheers!

rusty1s:
Thanks for the issue. Can you clarify what `init_batch` and `add_?_attr` would do? More efficient concatenating and slicing batches is a cool, and I am happy to include them if you wanna give it a shot.

mova:
`init_batch` would provide an "empty" batch object, just containing `batch.batch`.
With `add_node_attr` we can then add the `batch.x` later, with `batch._slice_dict` and `batch._inc_dict`  updated accordingly.
Here an example from the tests:
```
    batch_size = 3
    node_range = (1, 10)
    nodes_v = torch.randint(*node_range, (batch_size,)).to(device)
    x_list = [torch.rand(n, 3).to(device) for n in nodes_v]
    graph_attr_list = torch.rand(batch_size).to(device)

    batch_list = [Data(x=x, ga=ga) for x, ga in zip(x_list, graph_attr_list)]
    batch_truth = Batch.from_data_list(batch_list)
    batch_idx = batch_truth.batch
    batch = init_batch(batch_idx)
    add_node_attr(batch, "x", torch.vstack(x_list))
    add_graph_attr(batch, "ga", graph_attr_list)
    compare(batch_truth, batch)
```

mova:
Currently `init_batch` takes the mapping from the points to the graph as an input. 
I'd like to change that to a vector containing the number of points per graph.
Ideally `init_batch` would become the new constructor or a classmethod like `Batch.from_num_points`.
The `add_?_method` I would like to turn into methods of the batch class, replacing the `batch` argument with`self`. 
What I'm currently wondering is how to make that work with `Batch` objects derived from `HeteroData`.

rusty1s:
[...]

Issue #5083, opened 2022-07-28
Add `from_data` builder method to `Batch` that accepts batched data.


ethanabrooks:
### 🚀 The feature, motivation and pitch

PyG currently offers only one method for creating `Batch` objects, namely `from_data_list`. This method takes a list of `BaseData` objects as input. However, there are settings, such as Reinforcement Learning, where it is commonplace to process data that is already batched, e.g. using `stable_baselines3` [SubprocVecEnv](https://stable-baselines.readthedocs.io/en/master/_modules/stable_baselines/common/vec_env/subproc_vec_env.html#SubprocVecEnv). In these settings, it becomes necessary to unbatch the data and subsequently re-batch it using `from_data_list`. It would be helpful if there was a method that accepted already batched data in order to create `Batch` objects.

### Alternatives

Currently our best alternatives are un-batching and re-batching, as described above, or overriding the methods in `SubprocVecEnv`.

### Additional context

I am willing to work on a Pull Request for this.

rusty1s:
Thanks for creating this issue. Can you clarify how it may look like to already batch "batched data"? Won't there be an inherent limitation of only being able to batch equally-sized data?

ethanabrooks:
Hi Rusty. Yes our expectation is that the batched data would be padded to a maximum size with a designated padding index.

rusty1s:
The current `collate` and `separate` functions are quite complex since they also allow for recursive collating and supports different data formats such as `Data` and `HeteroData`. You can probably achieve something more simpler than this in your case, e.g. (untested):
```python
x = ... [batch_size, num_nodes, num_features]
adj = [batch_size, num_nodes, num_nodes]

def collate(x, adj):
    x = x.view(batch_size * num_nodes, num_features)
    
    edge_index = adj.nonzero()[:1]
    inc = torch.arange(0, num_nodes * batch_size, batch_size).view(-1, 1)
    edge_index += inc
    edge_index = edge_index.t()

    return x, edge_index
````

ethanabrooks:
Hi @rusty1s a member of my team came up with a solution to this based on what you wrote. We will be sharing shortly. Thanks for your help!

yikai5518:
For `x`, we reshaped with the padding as the padded nodes will not induce message passing since they are disjoint from the rest of the graphs. For `edge_index`, we used the `inc` tensor you suggested, and we removed the padded edges with boolean indexing. A similar approach was used for `edge_attr`. 

[...]

PR #8414, opened 2023-11-21
Manipulation tools for Batches

mova:
Hey,
this is the initial PR  for 
https://github.com/pyg-team/pytorch_geometric/issues/8226
I needs some work, I'd like to get the ball rolling, maybe we could have a separate branch for this? 

Apparently something similar has been implemented in https://github.com/pyg-team/pytorch_geometric/issues/5083
@yikai5518 @ethanabrooks

Cheers

### DIFF
--- a/torch_geometric/data/batch.py
+++ b/torch_geometric/data/batch.py
@@ -1,10 +1,12 @@
 import inspect
+from collections import defaultdict
 from collections.abc import Sequence
-from typing import Any, List, Optional, Type, Union
+from typing import Any, List, Literal, Optional, Tuple, Type, Union
 
 import numpy as np
 import torch
-from torch import Tensor
+from torch import LongTensor, Tensor
+from torch.nn.functional import pad
 from typing_extensions import Self
 
 from torch_geometric.data.collate import collate
@@ -80,6 +82,24 @@ class Batch(metaclass=DynamicInheritance):
     tensors of the same attribute should be concatenated together.
     """
     @classmethod
+    def from_empty(cls, num_nodes: Union[Tensor, Sequence]) -> Self:
+        if isinstance(num_nodes, Sequence):
+            num_nodes = torch.tensor(num_nodes, dtype=torch.long)
+        if not num_nodes.dtype == torch.long:
+            raise Exception('`num_nodes` dtype must be torch.long')
+        if not num_nodes.dim() == 1:
+            raise Exception('`num_nodes` must have one dimension')
+        if (num_nodes < 0).any():
+            raise Exception('`num_nodes` must be positive')
+
+        batch = Batch()
+        batch.batch, batch.ptr = cls._batch_ptr_from_num_nodes(cls, num_nodes)
+        batch._num_graphs = int(batch.batch.max() + 1)
+        batch._slice_dict = defaultdict(dict)
+        batch._inc_dict = defaultdict(dict)
+        return batch
+
+    @classmethod
     def from_data_list(
         cls,
         data_list: List[BaseData],
@@ -89,7 +109,7 @@ class Batch(metaclass=DynamicInheritance):
         r"""Constructs a :class:`~torch_geometric.data.Batch` object from a
         list of :class:`~torch_geometric.data.Data` or
         :class:`~torch_geometric.data.HeteroData` objects.
-        The assignment vector :obj:`batch` is created on the fly.
+        The assignment vector :obj:`batch` is adjusted on the fly.
         In addition, creates assignment vectors for each key in
         :obj:`follow_batch`.
         Will exclude any keys given in :obj:`exclude_keys`.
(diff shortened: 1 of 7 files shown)


mova:
Hey,
I'm still trying to make a more of the test to run, but I guess this would be ready for some feedback.
Cheers

rusty1s:
Any reason to close this? :(

mova:
Alright, I guess this is ready for some review

mova:
Hey,
is there anything I can do to move this forward?

torch_geometric/data/batch.py
```diff
@@ -78,10 +81,29 @@ class Batch(metaclass=DynamicInheritance):
     Furthermore, :meth:`~Data.__cat_dim__` defines in which dimension graph
     tensors of the same attribute should be concatenated together.
     """
+    @classmethod
+    def from_batch_index(cls, batch_idx: Tensor) -> Self:
```
rusty1s, line 85: What is the purpose of this method? This doesn't necessarily look particular useful to me.

torch_geometric/data/batch.py
```diff
@@ -105,6 +127,48 @@ def from_data_list(cls, data_list: List[BaseData],
...
+        batches: List[Self],
+        follow_batch: Optional[List[str]] = None,
+        exclude_keys: Optional[List[str]] = None,
+    ) -> Self:
+        r"""Same as :meth:`~Batch.from_data_list```,
+        but for concatenating existing batches.
+        Constructs a :class:`~torch_geometric.data.Batch` object from a
+        list of :class:`~torch_geometric.data.Batch` objects.
+        The assignment vector :obj:`batch` is created on the fly.
+        In addition, creates assignment vectors for each key in
```
rusty1s, line 142:
```suggestion
        The assignment vector :obj:`batch` is adjusted on the fly.
        In addition, creates assignment vectors for each key in
```

torch_geometric/data/batch.py
```diff
@@ -105,6 +127,48 @@ def from_data_list(cls, data_list: List[BaseData],
 
         return batch
 
+    @classmethod
+    def from_batch_list(
+        cls,
+        batches: List[Self],
+        follow_batch: Optional[List[str]] = None,
+        exclude_keys: Optional[List[str]] = None,
```
rusty1s, line 135: Do we need these here?

torch_geometric/data/batch.py
```diff
@@ -128,16 +192,12 @@ def get_example(self, idx: int) -> BaseData:
...
-    def index_select(self, idx: IndexType) -> List[BaseData]:
-        r"""Creates a subset of :class:`~torch_geometric.data.Data` or
-        :class:`~torch_geometric.data.HeteroData` objects from specified
-        indices :obj:`idx`.
-        Indices :obj:`idx` can be a slicing object, *e.g.*, :obj:`[2:5]`, a
-        list, a tuple, or a :obj:`torch.Tensor` or :obj:`np.ndarray` of type
-        long or bool.
-        The :class:`~torch_geometric.data.Batch` object must have been created
-        via :meth:`from_data_list` in order to be able to reconstruct the
-        initial objects.
```
rusty1s, line 140: This still seems to be the case here - or am I missing something?

torch_geometric/data/batch.py
```diff
@@ -210,3 +305,98 @@ def __len__(self) -> int:
     def __reduce__(self):
         state = self.__dict__.copy()
         return DynamicInheritanceGetter(), self.__class__.__bases__, state
+
+    def add_node_attr(self, attrname: str, attr: Tensor) -> None:
```
rusty1s, line 309: Can we use `setattr` functionality for this?

torch_geometric/data/batch.py
```diff
@@ -210,3 +305,98 @@ def __len__(self) -> int:
...
+        assert attr.device == self.batch.device
+
+        self[attrname] = attr
+        self._slice_dict[attrname] = torch.arange(self.num_graphs + 1,
+                                                  dtype=torch.long)
+
+        self._inc_dict[attrname] = torch.zeros(self.num_graphs,
+                                               dtype=torch.long)
+
+    def set_edge_index(self, edge_index: Tensor,
```
rusty1s, line 342: This interface is a bit unclear to me. Wouldn't it make more sense to just expect `list[Tensor]` here? Also, what's the use-case for this?

torch_geometric/data/batch.py
mova: My bad, that must have been a mistake.

torch_geometric/data/batch.py
mova: So the issues is, that we do not know if the attribute is a graph, node or edge attribute and thus cannot construct the `_slice_dict` and `_inc_dict`.
I could merge the `add_node_attr` / `add_graph_attr` / `set_edge_attr` into one function with switch if that helps?

torch_geometric/data/batch.py
mova: I guess it should be the new EdgeIndex Type.
I would like to be able to predict edges, update the graph and run it through a GNN again without losing the ability to slice the batch.

mova:
Hey, sorry this took so long, but I've finished my thesis now and hope I can carry this over the finish line :)

torch_geometric/data/batch.py
mova: I removed them

torch_geometric/data/batch.py
mova: I merged `add_node_attr` / `add_graph_attr` / `set_edge_attr` into `set_attr`.

torch_geometric/data/batch.py
mova: I implemented both ways now. Passing the tensors directly should only really matter for very large batches.
I can also remove it and only implement the `List[Tensor]` method. 
What do you prefer?

mova:
Hey @rusty1s ,  
I have updated most of the code, any comments?
The Linting/mypy and macos errors seem not to be caused by me, right?

mova:
@wsad1 @mananshah99 Could you review this?

mova:
@rusty1s @wsad1 @mananshah99 Is there any way to still release this or should I close it?

PR #8419, opened 2023-11-22
Shadow GNN using Neighborloader

pmpalang:
This is an implementation (example) for using torch.loader.NeighborLoader with the argument subgraph='induced' for mimicking the behavior of shadow GNN sampler. 

Thanks,
Poovaiah


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added example for shadow GNN using NeighborLoader iwth argument subgraph='induced' ([#8419](https://github.com/pyg-team/pytorch_geometric/pull/8419))
```
EdisonLeeeee, line 10:
```suggestion
- Added example for shadow GNN using `NeighborLoader` with argument `subgraph='induced'` ([#8419](https://github.com/pyg-team/pytorch_geometric/pull/8419))
```

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+                             input_nodes=data.test_mask,
+                             subgraph_type='induced', **kwargs)
+
+
+class GNN(torch.nn.Module):
+    def __init__(self, in_channels, hidden_channels, out_channels):
+        super().__init__()
+        self.conv1 = SAGEConv(in_channels, hidden_channels)
+        self.conv2 = SAGEConv(hidden_channels, hidden_channels)
+        self.conv3 = SAGEConv(hidden_channels, hidden_channels)
```
EdisonLeeeee, line 36: Just a bit confused - why do we use three convolutional layers when only two-hop neighbors are sampled?

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
+
+
+def train():
+    model.train()
+    total_loss = total_examples = 0
+    for data in tqdm(train_loader, desc="Train"):
+        data = data.to(device)
+        optimizer.zero_grad()
+        # print(data.input_id.size())
```
EdisonLeeeee, line 66:
```suggestion
```

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+        x = self.conv1(x, edge_index).relu()
+        x = F.dropout(x, p=0.3)
+        x = self.conv2(x, edge_index).relu()
+        x = F.dropout(x, p=0.3, training=self.training)
+        x = self.conv3(x, edge_index).relu()
+        x = F.dropout(x, p=0.3, training=self.training)
+
+        # We merge both central node embeddings and subgraph embeddings:
+
+        # x = torch.cat([x[root_n_id], global_mean_pool(x, batch)], dim=-1)
```
EdisonLeeeee, line 49: Is that accidentally commented out?

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+@torch.no_grad()
+def test(loader, desc='Eval'):
+    model.eval()
+    total_correct = total_examples = 0
+    for data in tqdm(loader, desc=desc):
+        data = data.to(device)
+        out = model(data.x, data.edge_index, data.batch, data.input_id)
+        total_correct += int(
+            (out.argmax(dim=-1) == data.y).sum()) / len(data.y)
+        total_examples += 1
```
EdisonLeeeee, line 85:
```suggestion
        total_examples += len(data.y)
```

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+
+@torch.no_grad()
+def test(loader, desc='Eval'):
+    model.eval()
+    total_correct = total_examples = 0
+    for data in tqdm(loader, desc=desc):
+        data = data.to(device)
+        out = model(data.x, data.edge_index, data.batch, data.input_id)
+        total_correct += int(
+            (out.argmax(dim=-1) == data.y).sum()) / len(data.y)
```
EdisonLeeeee, line 84:
```suggestion
            (out.argmax(dim=-1) == data.y).sum())
```

examples/shadow_neighborloader.py
```diff
@@ -0,0 +1,94 @@
...
+
+def train():
+    model.train()
+    total_loss = total_examples = 0
+    for data in tqdm(train_loader, desc="Train"):
+        data = data.to(device)
+        optimizer.zero_grad()
+        # print(data.input_id.size())
+        out = model(data.x, data.edge_index, data.batch, data.input_id)
+        loss = F.cross_entropy(out, data.y)
```
EdisonLeeeee, line 68: Shouldn't we do `F.cross_entropy(out[:data.batch_size], data.y[:data.batch_size])`?

examples/shadow_neighborloader.py
pmpalang: First, the sampling depth and model depth are independent parameters. Meaning, one can have a subgraph sampled using 2 hops and such a subgraph can be used on a model with 3 hops. The same node might appear on multiple levels as a result.

Second, the core feature of ShadowGNN is that is can sample shallow sub-graphs and enable deeper GNNS.

examples/shadow_neighborloader.py
pmpalang: Thanks, this line is not needed. I shall delete the commented line.

examples/shadow_neighborloader.py
pmpalang: Yes, I think you are right. I'll update the code. Thanks

PR #8428, opened 2023-11-23
Enable edge based temporal sampling in `torch_geometric.distributed`

kgajdamo:
This PR enables edge based temporal distributed training for node and link sampling.

**Comment about the edge temporal data definition:**
In the case of distributed training, it is necessary to create a separate vector for each partition that will store the time information of the edges included in the partition. (I mention this just to point out that this works differently than with node-based temporal sampling, where we can have one vector common to each partition because we operate on node ids.)
Why:
Each partition has its own unique edge_index in COO format, which is later converted to a matrix in CSR/CSC format in the neighbor sampler. Therefore, we do not have information about the global edge IDs when sampling and we would not be able to find the correct time information for a specific edge. Therefore, this information must be local.

**Changes made:**
- added support for edge_time argument
- `seed_time` needs to be specified (requirement for edge level temporal sampling)
- added unit tests
- unit tests for link sampler are in [this PR #8375](https://github.com/pyg-team/pytorch_geometric/pull/8375)

### DIFF
--- a/torch_geometric/distributed/dist_neighbor_sampler.py
+++ b/torch_geometric/distributed/dist_neighbor_sampler.py
@@ -183,15 +183,14 @@ class DistNeighborSampler:
         src_batch_dict = None
 
         seed = inputs.node.to(self.device)
-        seed_time = None
+        seed_time: Optional[Tensor] = None
         if self.time_attr is not None:
             if inputs.time is not None:
                 seed_time = inputs.time.to(self.device)
+            elif self.node_time is not None:
+                seed_time = self.node_time[seed]
             else:
-                if self.time_attr == 'time':
-                    seed_time = self.node_time[seed]
-                else:  # self.time_attr == 'edge_time'
-                    raise ValueError("Seed time needs to be specified.")
+                raise ValueError("Seed time needs to be specified")
 
         src_batch = torch.arange(batch_size) if self.disjoint else None
         metadata = (seed, seed_time)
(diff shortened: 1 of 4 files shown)


rusty1s (approved):

torch_geometric/distributed/dist_neighbor_sampler.py
```diff
@@ -620,12 +619,12 @@ def _sample_one_hop(
             True,  # csc
             self.replace,
             self.subgraph_type != SubgraphType.induced,
-            self.disjoint and self.time_attr is not None,
+            self.disjoint or self.time_attr is not None,
```
kgajdamo, line 622: Hi @rusty1s, thank you for the review. One comment from me about this change: during distributed training, layer-by-layer sampling loop is located in Python and not in the sample() function from pyg-lib, therefore, in the next layers, information about nodes belonging to the subgraph is lost in the c++ code and the batch calculated in the C++ is not important. Because of that the whole pipeline was created such that the sample() function must work in disjoint mode only during temporal sampling. If there is only a disjoint, the sample() function does not have to use this option because the batch is counted later anyway based on the sampled nodes with duplicates info and the one from C++ is not used anyway. So this condition should stay as it was.

rusty1s: Thanks for the clarification. I can change that back. Why do we then expose `disjoint` as an argument to `DistNeighborSampler`?

rusty1s: https://github.com/pyg-team/pytorch_geometric/pull/8482

kgajdamo: This is what the algorithm looks like:

1. Subgraph ids are assigned to the seed nodes at the beginning before layer-by-layer sampling starts in the node_sample() function in the DistNeighborSampler:
`src_batch = torch.arange(batch_size) if self.disjoint else None`

2. src_batch is being passed to the sample_one_hop() function where all machines starts sampling.
3. When all machines finished sampling, their results must be merged - it is being done in the merge_sampler_outputs().
4. In the merge_sampler_outputs() function, based on the sampled nodes with duplicates and information about how many neighbors each node has sampled, subgraph ids are assigned to the newly sampled nodes and out_batch is being returned.
5. Then, in the next layer, we assign the newly created batch (of the newly sampled nodes) to the src_batch variable and go back to the bullet 2.

rusty1s: Got it, thanks for clarifying :)

PR #8454, opened 2023-11-27
Merge `TemporalData` with `Data`/`HeteroData` [1/n]

DamianSzwichtenberg:
This PR attempts to merge TemporalData into Data.
Currently, only edge events are supported. For `(src, dst)` event definition `edge_index` is reused as its definition is aligned with our needs. For `msg`, `edge_attr` is reused. An additional attribute is `time`, which represents an event's occurrence time.
New methods include the following:
- `is_sorted_by_time()`: returns `True` if data is sorted according to `time`, `False` otherwise.
- `sort_by_time()`: sorts event data according to `time`.
- `snapshot(start_time, end_time)`: returns event data that occurred in range `<start_time, end_time>`.
- `up_to(time)`: returns event data that occurred before or at a given `time`.
- `concat(data)`: returns new data that concatenates the object on which the method was called and given `data`. Can be used to add new events.

Related issue: #3230


torch_geometric/data/data.py
```diff
@@ -84,6 +85,10 @@ def node_stores(self) -> List[NodeStorage]:
     def edge_stores(self) -> List[EdgeStorage]:
         raise NotImplementedError
 
+    @property
```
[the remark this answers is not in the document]
DamianSzwichtenberg: Hi @rusty1s, in my initial design, I thought that it made sense to put event data into a separate abstraction. After rethinking, I went for a simple design, as described in the roadmap. The code is now updated, please take a look.

torch_geometric/data/storage.py
```diff
@@ -734,6 +734,65 @@ def is_edge_attr(self, key: str) -> bool:
...
+        if 'time' in self:
+            self.time, perm = torch.sort(self.time, stable=True)
+            if self.is_edge_attr('time'):
+                keys = self.edge_attrs()
+                keys.remove('time')
+                self._index_select(keys, self._dims_for_attrs(keys), perm)
+            # TODO: cover case where 'time' is a node attribute
+        return self
+
+    def snapshot(self, start_time: Union[float, int],
```
EdisonLeeeee, line 779: Hi @DamianSzwichtenberg, I think the `snapshot` function should accept the `start`-th and `end`-th time steps, as people usually don't care about the exact time values. For example, one can call `snapshot` to obtain its graph snapshot from 0th to 2rd time step: `snapshot(0, 2)`

DamianSzwichtenberg: Hi @EdisonLeeeee, thanks for the feedback. As `TemporalData` was representing CTDGs, I assumed that the same applies here - hence user will not operate on `start`-th and `end`-th steps, but on concrete `time` values. In my understanding `snapshot` may be used to generate a DTDGs. Please correct me if I'm wrong.

rusty1s (approved):

PR #8472, opened 2023-11-28
replace sklearn heuristic with rudimentary timing cache

puririshi98:
@stadlmax made this branch. I took a look and i think this would be a good update


puririshi98:
needs rebase

PR #8483, opened 2023-11-29
Update `bug-report.yml`

akihironitta:
Adds commands to collect environment information to include in the bug report issue, assuming the PyTorch team won't add any malicious code to `collect_env.py`.


rusty1s (approved):
We would be missing `pyg-lib` version in the script, but that's okay I guess.

Issue #284, opened 2019-05-08
Learning rate / epochs search for benchmark at the paper.


LeeJunHyun:
## ❓ Questions & Help

Hi, at first, I really appreciate this wonderful work!
This makes it easier for many researchers to conduct their experiments.

I just want to ask you about the benchmark performance for graph classification in your paper.

Is this reported after hyper parameter search such as learning rate, the number of epochs, learning rate decay, and the things that we can adjust with `argparse` in your code?

Thanks,

rusty1s:
The `argparse` hyperparameters are fixed and there is no tuning performed on them. If you want to tune them, you should add them as lists to [this line](https://github.com/rusty1s/pytorch_geometric/blob/master/benchmark/kernel/main.py#L54).

LeeJunHyun:
So, if I am doing a learning rate search, there's room for performance change?

Because I want to report baseline performance with this report for homogeneous training scenario.

rusty1s:
I do not know what you mean with "room for performance change".
If you simply want to tune the learning rate, do something like this:
```python
lrs = [0.1, 0.01, 0.001, ...]
```
and replace line 54 with:
```
for lr in lrs:
```

```

**Comment by LeeJunHyun** [comment]
yes, and I also should change line 63 like this:

```python
lr=lr
```

Thanks.
Your answer was helpful for me.

 "room for performance change" means that
I should tune baselines again If I want to report my model with learning rate search.
Because baselines performance may be different with your paper.
(The reported performance may not be best performance)

rusty1s:
Ah okay, I see what you mean. I guess you need to do that for a fair comparison :(

LeeJunHyun:
Thanks, I will do that!

Actually, this repo is great contribution for GNN research community!

rusty1s:
Thank you :)

PR #8503, opened 2023-12-01
Update `DistNeighborSampler` for hetero

kgajdamo:
The purpose of this PR is to improve distributed hetero sampling algorithm.
**IMPORTANT INFO**: This PR is complementary with [#284](https://github.com/pyg-team/pyg-lib/pull/284) from pyg-lib. The pyg-lib one needs to be merged for this one to work properly.

**Description:** (sorry if too long)
Distributed hetero neighbor sampling is a procedure analogous to homo sampling, but more complicated due to the presence of different types of nodes and edges.
Sampling in distributed training imitates the `hetero_neighbor_sample()` function in pyg-lib. Therefore, the mechanism of action and the nomenclature of variables are similar.
Due to the fact that in distributed training, after sampling each layer, it is necessary to synchronize the results between machines, the loop iterating through the layers was implemented in Python.

The main two loops iterate sequentially: over layers and edge types. Inside the loop, the `sample_one_hop()` function is called, which performs sampling for one layer.
The input to the `sample_one_hop()` function is data of a specific type, so its execution is almost identical to homo.
The sample_one_hop() function, depending on whether the input nodes are located on a given partition or a remote one, performs sampling or sends an RPC request to the remote machine to do so. The `dist_neighbor_sample()`->`neighbor_sample()` function is used for sampling. Nodes are sampled with duplicates so that they can later be used to construct local to global node mappings.
When all machines have finished sampling, their outputs are merged and synchronized in the same way as for homo.
Then the results return to the `node_sample()` function where they are written to the output dictionaries and the src nodes for the next layer are calculated.
After going through all the layers, the global node indices are finally mapped to the local ones in the `hetero_dist_relabel()` function.

Information about some of the variables used in a node_sample() function:
`node_dict` - class storing information about nodes. It has three fields: `src`, `with_dupl`, `out`, which are described in more detail in the distributed/utils.py file.
`batch_dict` - class used when sampling with the disjoint option. It stores information about the affiliation of nodes to subgraphs. Just like `node_dict`, it has three fields: `src`, `with_dupl`, `out`.
`sampled_nbrs_per_node_dict` - a dictionary that stores information about the number of sampled neighbors by each src node. To facilitate subsequent operations, for each edge type is additionally divided into layers.
`num_sampled_nodes_dict`, `num_sampled_edges_dict` - needed for HGAM to work.


JakubPietrakIntel (approved):
Thanks Kinga, looks amazing. 
Depending if https://github.com/pyg-team/pytorch_geometric/pull/8605 comes in before that, we might do some adjustments in `test_dist_neighbor_loader.py` to account for a different RPC shutdown method or I'll make corrections in the other PR.

rusty1s (approved):

PR #8524, opened 2023-12-04
Add relabel node functionality to `dropout_node`

will-leeson:
Currently, the dropout_node function removes any edge containing a dropped node from the edge_index. It also produces an node mask to remove nodes encodings from x. However, the edge_mask doesn't have the option to relabel nodes accordingly. If node n is dropped and we use the node mask, node n+1 is now the nth node in x, but the edge set still references it as n+1.

This functionality can be added by feeding the relabel node flag directly to the subgraph call in relabel_node


EdisonLeeeee (approved):
Thank you for the PR!

rusty1s (approved):

PR #8540, opened 2023-12-05
Add `MultithreadMixin` that enables multithreading in `NodeLoader` and `LinkLoader` and extend CPU affinity with default multithreading

JakubPietrakIntel:
The `MultithreadMixin` fixes a known issue with PyG Dataloaders being constrained by base Pytorch Dataloader class to use a sinlge thread in each sampler's subprocess. The implementation simply modifies the `worker_init_fn` to execute `toch.multiprocessing.set_num_threads(N)`, however for these changes to take place a user needs to change the start method from `fork` to `spawn` prior to script execution, i.e.:
```
if __name__ == '__main__':
    torch.set_num_threads(...)
    torch.multiprocessing.set_start_method('spawn')
    run()
```
More information on spawning subprocesses in pytorch [here](https://pytorch.org/docs/stable/multiprocessing.html#spawning-subprocesses).
The drawback of switching to `spawn` is that the OS will take more time to instantiate a process (up to 20x is quoted by [some sources](https://superfastpython.com/fork-faster-than-spawn/#:~:text=The%20fork%20method%20is%20only,faster%20than%20spawning%20a%20process)). In a long run, this overhead seems to be negligible and is outweighed by the gain of achieving higher throughput with multiple threads in sampler processes.
The measurements performed for the isolated data loading times revealed that we're seeing approx**. 2x data loading time decrease for the hetero datasets and 2.8x decrease for the homo datasets**. What is also important, the best time can be achieved at lower number of workers, which reduced memory requirements. Full results are presented here, as mean times per each config per `num_workers`: [TH_mean.xlsx](https://github.com/pyg-team/pytorch_geometric/files/13564345/TH_mean.xlsx)
).
HETERO
HOMO

The results were obtained by measuring end-to-end data loading time for 10 randomly generated homo/hetero datasets with avg. num. nodes 10**6 and avg. node degree 3 on a dual-socket 2x40 core CPU Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz.
Legend:
- _base_: baseline dataloader with `fork`
- _TH-n_: dataloader with `spawn` and `enable_multithreading()` where _n_ denotes the number of thread in each sampler
- _AFF-fork_: baseline dataloader with `fork` and `enable_cpu_affinity()` 
- _AFF-spawn_: dataloader with `spawn` and modified `enable_cpu_affinity()` 

Observations:
1.  The best loading times are usually achieved at relatively low number of threads, while using too many threads for sampling leads to perf. degradation (due to core oversubscription), see TH-40, where 40 is a number of all available cores.
2. The results revealed that in the isolated case the best data loading times are achieved with multithreading enabled and avoiding CPU affinizations. 
3. Using modified CPU affinity with multithreading may lead to perf. degradation for hetero datasets. This is due to inherited parallelization in `neighbor_kernel.cpp` in pyg-lib that parallelizes hetero sampling based on number of edge types, assigning a single thread to process input nodes of the common edge type. In effect hetero neighbor kernel doesn't utilize multithreading fully. This case will be considered in the future to see if further improvements are possible.

Side note:
I'm also adding an additional `MemMixin` class with `enable_memory_log()` that I've found useful for mapping the amount of memory used by each worker. Spawning processes doesn't pre-allocate memory as forking does, so changing to `spawn` may add an additional memory advantage.


test/loader/test_neighbor_loader.py
```diff
@@ -662,27 +662,28 @@ def test_memmap_neighbor_loader(tmp_path):
...
-                            num_workers=1)
-
+                            num_workers=2)
     out = []
     with loader.enable_cpu_affinity(loader_cores):
         iterator = loader._get_iterator()
         workers = iterator._workers
         for worker in workers:
-            sleep(1)  # Gives time for worker to initialize.
+            sleep(2)  # Gives time for worker to initialize.
```
mananshah99, line 675: What changed here? Was this flaky?

test/loader/test_neighbor_loader.py
```diff
@@ -662,27 +662,28 @@ def test_memmap_neighbor_loader(tmp_path):
...
-        if not loader_cores:
-            assert out == [0]
+            out.append(stdout.split(':')[1].strip())
+        if loader_cores:
+            out == ['[1]', '[2]']
         else:
-            assert out == loader_cores
+            n, m = out[0].split('-')
+            assert int(n) == 0
+            assert int(out[1].split('-')[0]) == int(m) + 1
```
mananshah99, line 686: Can we add a brief comment about what we are asserting here?

torch_geometric/loader/mixin.py
```diff
@@ -71,6 +74,86 @@ def __call__(self, worker_id):
             self.func(worker_id)
 
 
+class MemMixin:
+    r"""A context manager to enable logging of dataloder workers
+    memory consumption.
+    """
+    def mem_init_fn(self, worker_id):
+        logging.debug(
+            f"woker {worker_id} process is using {round((psutil.Process(                os.getpid()).memory_info().rss / 1024 ** 2),2)} MB of memory"
```
mananshah99, line 83:
```suggestion
            f"Worker {worker_id} process is using {round((psutil.Process(                os.getpid()).memory_info().rss / 1024 ** 2),2)} MB of memory"
```

mananshah99, line 83: Also log PID?

torch_geometric/loader/mixin.py
```diff
@@ -71,6 +74,86 @@ def __call__(self, worker_id):
...
+        self.worker_init_fn_old = WorkerInitWrapper(self.worker_init_fn)
+        try:
+            self.worker_init_fn = self.mem_init_fn
+            yield
+        finally:
+            self.worker_init_fn = self.worker_init_fn_old
+
+
+class MultithreadMixin:
+    r"""A context manager to enable multithreading in dataloder workers.
```
mananshah99, line 98:
```suggestion
    r"""A context manager to enable multithreading in dataloader workers.
```

torch_geometric/loader/mixin.py
```diff
@@ -71,6 +74,86 @@ def __call__(self, worker_id):
...
+
+    @contextmanager
+    def enable_multithreading(
+            self, worker_threads: Optional[int] = torch.get_num_threads() / 2):
+        r"""Enables multithreading in worker subprocess.
+
+        Args:
+            worker_threads ([int], optional): Number of cores for sampling.
+            By defualt it uses half of all avaialbe CPU cores
+            (default: : `torch.get_num_threads() / 2`)
```
mananshah99, line 116: Should the default heuristic also be based on `num_workers`, to avoid oversubscription?

torch_geometric/loader/mixin.py
```diff
@@ -71,6 +74,86 @@ def __call__(self, worker_id):
...
+        """
+        self.worker_threads = worker_threads
+        nthreads_old = (torch.get_num_threads()
+                        )  # 1 is torch Dataloader default value
+        self.worker_init_fn_old = WorkerInitWrapper(self.worker_init_fn)
+
+        if not self.num_workers > 0:
+            raise ValueError(
+                f"'enable_multithread_sampling' should be used with at least one "
+                f"worker (got {self.num_workers})")
```
mananshah99, line 136: Why?

torch_geometric/loader/mixin.py
```diff
@@ -71,6 +74,86 @@ def __call__(self, worker_id):
...
+            with loader.enable_multithreading(10):
+                for batch in loader:
+                    pass
+        if if __name__ == '__main__':
+            torch.set_start_method('spawn')
+            run()
+        """
+        self.worker_threads = worker_threads
+        nthreads_old = (torch.get_num_threads()
+                        )  # 1 is torch Dataloader default value
```
mananshah99, line 130: Why do we need this here, if sets are only happening in subprocesses?

mananshah99 (commented):
Thanks! This looks pretty cool. A couple notes:

> however for these changes to take place a user needs to change the start method from fork to spawn prior to script execution

Why is this the case, if we set `num_threads` after the `fork`/`exec`?

It would also be worth adding some more information in the documentation about the subtleties in the `'fork'` vs `'spawn'` methods as they relate to GNN training. In particular, while PyTorch already handles sharing of tensors (via shmem + some IPC mechanism), this does not generally hold true for other Python objects. While `'fork'` supports Linux CoW, `'spawn'` does not work this way, so users may see significant memory blowups if, for example, they are using a feature store / graph store combination that stores state not in Tensor format.

torch_geometric/loader/mixin.py
```diff
@@ -29,20 +31,21 @@ def get_numa_nodes_cores() -> Dict[str, Any]:
     nodes = {}
     try:
         for node_path in numa_node_paths:
-            numa_node_id = int(osp.basename(node_path)[4:])
+            numa_node_id = int(os.path.basename(node_path)[4:])
```
kgajdamo, line 34:
```suggestion
            numa_node_id = int(osp.basename(node_path)[4:])
```

https://github.com/pyg-team/pytorch_geometric/pull/8435

test/loader/test_neighbor_loader.py
JakubPietrakIntel: https://github.com/pyg-team/pytorch_geometric/blob/7d41e0f18e92854982b77cf856d56550b02568b4/test/loader/test_mixin.py#L37C1-L41C30

torch_geometric/loader/mixin.py
JakubPietrakIntel: You're right, I've changed it to: 
```
self.worker_threads = int(torch.get_num_threads() / self.num_workers)
```

JakubPietrakIntel:
Thank you for the review @mananshah99. To answer your first question
> Why is this the case, if we set `num_threads` after the `fork`/`exec`?

According to my best knowledge forking a process with multitple threads is currently not possible/supported (some, but not great explanation can be found here: [python multiprocessing docs](https://docs.python.org/3/library/multiprocessing.html#:~:text=The%20parent%20process%20uses%20os.fork()%20to%20fork%20the%20Python%20interpreter.%20The%20child%20process%2C%20when%20it%20begins%2C%20is%20effectively%20identical%20to%20the%20parent%20process.%20All%20resources%20of%20the%20parent%20are%20inherited%20by%20the%20child%20process.%20Note%20that%20safely%20forking%20a%20multithreaded%20process%20is%20problematic.)
We can test that by printing the output of `at::get_num_threads()` in `neighbor_kernel.cpp` in pyglib. When the start method is set to `fork`, calling `toch.get_num_threads()` essentially doesn't change anything and the print will return 1. With the method changed to `spawn` the print returns set number of threads as expected.

> It would also be worth adding some more information in the documentation about the subtleties in the `'fork'` vs `'spawn'` methods as they relate to GNN training. In particular, while PyTorch already handles sharing of tensors (via shmem + some IPC mechanism), this does not generally hold true for other Python objects. While `'fork'` supports Linux CoW, `'spawn'` does not work this way, so users may see significant memory blowups if, for example, they are using a feature store / graph store combination that stores state not in Tensor format.

This is a great suggestion and I agree that switching to `spawn` is a rather fundamental change. I'm also not aware of all possible implications so thanks for mentioning that GS/FS case. Do you think that I should add some extra notes it in docs to the mixin function? 

One thing I've noticed is that sampler processes build with `fork` pre-allocate much more memory than these build with 'spawn' at init, i.e. for the same dataset with `shuffle=False` and exact same loader configs
Spawn:
`NeighborLoader() worker 0 process is using 372.07 MB of memory`
Fork:
`NeighborLoader() worker 0 process is using 3152.64 MB of memory`
Might be something worth investigating.

torch_geometric/loader/mixin.py
JakubPietrakIntel: done

rusty1s (approved):

PR #8546, opened 2023-12-05
Example of a recommender system

pmpalang:
Hi Everyone,

I'm adding a recommender system example with the following salient features

1. Dataset MovieLens – a heterogenous use case
2. Demonstrates the use of edge based temporal sampling
3. Visualization
  t-SNE based visualization (--visualize_emb)
4. Uses torch_geometric.nn.pool.MIPSKNNIndex for getting recommedations
5. Integration of the LinkPred metrics -- precision@k and ndcg@k

Thanks,
Poovaiah


examples/hetero/recommender_system.py
```diff
@@ -0,0 +1,434 @@
...
+
+test_dataloader = LinkNeighborLoader(
+    data=test_data,
+    num_neighbors=[-1, -1, -1],
+    neg_sampling_ratio=1,
+    edge_label_index=(('user', 'rates', 'movie'),
+                      test_data[('user', 'rates', 'movie')].edge_index),
+    batch_size=BATCH_SIZE,
+    shuffle=True,
+)
```
rusty1s, line 132: As discussed, let's drop these and just call retrieval metrics.

examples/hetero/recommender_system.py
```diff
@@ -0,0 +1,434 @@
...
+if args.use_weighted_loss:
+    weight = torch.bincount(data['user', 'movie'].edge_label)
+    weight = weight.max() / weight
+else:
+    weight = None
+
+
+def weighted_mse_loss(pred, target, weight=None):
+    weight = 1. if weight is None else weight[target].to(pred.dtype)
+    return (weight * (pred - target.to(pred.dtype)).pow(2)).mean()
```
rusty1s, line 145: Can be dropped since we only want to do link prediction, not rating prediction.

examples/hetero/recommender_system.py
```diff
@@ -0,0 +1,434 @@
...
+        z = self.lin1(z).relu()
+        z = self.lin2(z)
+        return z.view(-1)
+        # return z
+
+
+class Model(torch.nn.Module):
+    def __init__(self, num_users, num_movies, hidden_channels):
+        super().__init__()
+        self.user_emb = Embedding(num_users, hidden_channels, device=device)
```
rusty1s, line 186: If you are using an `Embedding` here, then there is no need to add a one-hot matrix for user features above.

pmpalang: Agreed. I have removed the one hot encoding. Thanks

examples/hetero/recommender_system.py
```diff
@@ -0,0 +1,434 @@
...
+        row, col = edge_label_index
+        u_to_i = self.user_to_item(z_dict['user'][row])
+        i_to_u = self.item_to_user(z_dict['movie'][col])
+
+        z = torch.cat([u_to_i, i_to_u], dim=-1)
+
+        z = self.lin1(z).relu()
+        z = self.lin2(z)
+        return z.view(-1)
+        # return z
```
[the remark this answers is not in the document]
pmpalang: Done. Thanks

examples/hetero/recommender_system.py
pmpalang: Agreed. Removed this portion -- I'm using F.mse_loss() now. Thanks

examples/hetero/recommender_system.py
```diff
@@ -0,0 +1,400 @@
...
+    # Get the recommendations for user; only those
+    # that the user is yet to watch
+    real_recs = {}
+    for user in range(train_data['user'].num_nodes):
+        try:
+            real_recs[user] = list(
+                OrderedSet(knn_index[user].tolist()) -
+                OrderedSet(user_pos_items[user]))[:k]
+        except KeyError:
+            real_recs[user] = knn_index[user].tolist()[:k]
```
rusty1s, line 233: I think this example looks mostly good, thank you :) The only remaining issue I have is in this code-block. I think I would prefer if we can do `mipsknn.search` in mini-batches over users, and then also compute the metrics in a mini-batch fashion. Let me know if you need help on that.

pmpalang: Implemented in the latest commit. Thanks

rusty1s (approved):

PR #8551, opened 2023-12-06
Add `channel_lists` to `__init__` of the `Basic_GNN`

### LINKED ISSUE #8330: Add channel_list to `BasicGNN`
### 🛠 Proposed Refactor

Today, all pre-defined GNN models (`GCN`, `GAT`, etc.) can only have a constant hidden size. The base class for these models,  which is [`BasicGNN`](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/basic_gnn.py), disallows the use of varying hidden channels. However, there are many use cases when one would want to instantiate a model with varying layer sizes.  For instance, [BGRL](https://github.com/nerdslab/bgrl) method utilizes GAT or GCN architectures with decreasing hidden size in consecutive layers, e.g., `[512, 256]`. Also, it is common to experiment with different layer configurations, and the current implementation of GNN models in PyG forces us to write this class from scratch each time.

I propose to refactor the implementation of `BasicGNN` and make it similar to [`MLP`](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.models.MLP.html), i.e.:
* add `channel_list: Optional[list[int]]` parameter to the `__init__` method, 
* keep `input_channels`,  `hidden_channels`, and `num_layers` parameters as an alternative to  `channels_list` (as in `MLP`)

However, unlike in `MLP`,  there is `JumpingKnowledge`  module in `BasicGNN`, which requires constant hidden size. To fix this problem, one could prohibit using `channel_list` with varying hidden sizes together with `JumpingKnowledge` by raising an exception. 

Let me know if you agree to that change or have some suggestions. I can prepare PR with the proposed changes.

brovatten:
Fixes #8330.

My first PR here so let me know if I missed something. If you want new unit tests using the new parameter, please provide some guidance.


rusty1s:
I think this looks good. My only concern is that this is a breaking change since the order of arguments changed. Wondering if there is any way to prevent this :(

brovatten:
> I think this looks good. My only concern is that this is a breaking change since the order of arguments changed. Wondering if there is any way to prevent this :(

Okay I understand. Would it be better to provide `channel_list` as a optional argument and keep the others as positional in that case?

brovatten:
I think it should be fixed now. I added some unit tests as well.  

PS: Should I have git hard reset my branch after the rework?

PR #8569, opened 2023-12-07
Fixed `edge_index` shape when is `SparseTensor` type for `hgt_conv`

jykr:
When the edges with type `SparseTensor` was passed into `hgt_conv`, the edges were converted into the wrong size of `SparseTensor` as its initialization weren't provided of the `sparse_sizes`. Fixed the issue by providing the correct # of source & target nodes.


torch_geometric/utils/hetero.py
```diff
@@ -119,11 +122,11 @@ def construct_bipartite_edge_index(
...
         edge_index = SparseTensor(
-            row=edge_index[1],
-            col=edge_index[0],
+            row=edge_index[0],
+            col=edge_index[1],
             value=edge_attr,
-        )
-
+            sparse_sizes=n_nodes,
+        ).t()
```
akihironitta, line 130: I'm not too familiar with `torch_sparse.SparseTensor`, but do we want to avoid calling `t()` to preserve contiguity of the tensor? @rusty1s

akihironitta (commented):
@jykr Could we have a test case?

torch_geometric/utils/hetero.py
jykr: (Has been a while so this may already be resolved but) Changed the number of edges in `test_hgt_conv.py`, which leads to the following error in master but passes with the edit.
```
====================================== short test summary info ======================================
FAILED test/nn/conv/test_hgt_conv.py::test_hgt_conv_same_dimensions - ValueError: Encountered tensor with size 10 in dimension 0, but expected size 6.
```

jykr:
This issue has been resolved by #8805 so I'm closing it!

PR #8591, opened 2023-12-11
Collect memory statistics in distributed XPU example

DamianSzwichtenberg:


torch_geometric/profile/utils.py
```diff
@@ -153,7 +153,7 @@ def get_gpu_memory_from_ipex(
     max_allocated = byte_to_megabyte(max_allocated, digits)
     max_reserved = byte_to_megabyte(max_reserved, digits)
     max_active = byte_to_megabyte(max_active, digits)
-    ipex.xpu.reset_peak_memory_stats()
+    ipex.xpu.reset_peak_memory_stats(device)
```
akihironitta, line 156: It looks like we don't test this much even before this PR, but it could be nice to add a quick test case maybe 😄

DamianSzwichtenberg: I think `reset_peak_memory_stats()` is well-tested in IPEX. If we would like to test `get_gpu_memory_from_ipex` per device, it would require CI environment with at least two XPU devices - I'm not sure it's doable.

examples/multi_gpu/distributed_sampling_xpu.py
```diff
@@ -103,6 +104,17 @@ def inference(
         return x_all
 
 
+@profileit('xpu')
```
rusty1s, line 107: Is this desired? I would assume that profiling comes with an overhead.

examples/multi_gpu/distributed_sampling_xpu.py
```diff
@@ -127,20 +139,23 @@ def run(rank: int, world_size: int, dataset: PygNodePropPredDataset):
...
-            optimizer.zero_grad()
-            out = model(batch.x,
-                        batch.edge_index.to(device))[:batch.batch_size]
-            loss = F.cross_entropy(out, batch.y[:batch.batch_size].squeeze())
-            loss.backward()
-            optimizer.step()
+            batch = batch.to(device)
+            loss, stats = train_step(model, optimizer, batch.x,
+                                     batch.edge_index, batch.y,
+                                     batch.batch_size)
```
rusty1s, line 148:
```suggestion
            loss, stats = train_step(model, optimizer, batch)
```

rusty1s (approved):

examples/multi_gpu/distributed_sampling_xpu.py
DamianSzwichtenberg: Yes, XPU profiling is lightweight, it brings `< 2%` overhead.

examples/multi_gpu/distributed_sampling_xpu.py
DamianSzwichtenberg: This cannot be done, as in the current design `profileit` decorator detects device id based on input tensors.

PR #8592, opened 2023-12-11
Use `find_unused_parameters` only for heterogeneous cases

DamianSzwichtenberg:


akihironitta (commented):
n00b q: Why do we enable it only for hetero cases?

DamianSzwichtenberg:
> n00b q: Why do we enable it only for hetero cases?

`to_hetero` transformation sometimes generates a model that has not used parameters, which in the DDP model ends with the following error:
```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss.
```
It does not apply to homogeneous cases, and because `find_unused_parameters` may introduce overhead to forward pass we do not want to enable it in that scenario.

akihironitta:
> `to_hetero` transformation sometimes generates a model that has not used parameters

(Feel free to ignore me 😄 but just out of curiousity) Do you know when that'll be the case? I wonder if it's a bug or expected behaviour.

DamianSzwichtenberg:
> > `to_hetero` transformation sometimes generates a model that has not used parameters
> 
> (Feel free to ignore me 😄 but just out of curiousity) Do you know when that'll be the case? I wonder if it's a bug or expected behaviour.

I know that it happens in rgcn, but unfortunately, I didn't manage to track the core issue.

rusty1s:
The core issue is that `to_hetero` will replicate GNN layers for every edge type present in your data. However, some of these edge types will not contribute to the final node representation (e.g., in case the number of message passing hops it too small or there does not exist a path from the node type to the seed node type).

PR #8602, opened 2023-12-12
Implement annealed Langevin dynamics sampling for graph generative models

Loquats:
Implement annealed Langevin dynamics sampling, as defined in Algorithm 1 in https://arxiv.org/abs/2003.00638. 
In a subsequent PR, we will implement the EDP-GNN model defined in https://arxiv.org/abs/2003.00638. The sampling procedure here will be used with the EDP-GNN model to generate graphs. 

Also see https://github.com/pyg-team/pytorch_geometric/pull/8608 and https://github.com/pyg-team/pytorch_geometric/pull/8347


Loquats:
FYI @666even666 @atimashov

PR #8608, opened 2023-12-12
CS224W: Adding multi channel and Edgewise Dense Prediction (EDP) convolutional layers

atimashov:
Me and my partners intend to implement Permutation Invariant Graph Generation via Score-Based Generative Model-
ing. (https://arxiv.org/abs/2003.00638)

1) Multi-channel convolutional layer is similar to GINConv, we just assume that we want ot process a few different adjacency matrices for given node embeddings in parallel. Besides, we have weights on each edge.
2. EDP is the main component of this paper. It returns node embeddings (from 1) and edge predictions.

pyg


torch_geometric/nn/conv/edp_conv.py
```diff
@@ -0,0 +1,183 @@
...
+        kwargs.setdefault('aggr', 'add')
+        super().__init__(**kwargs)
+        self.nn = nn
+        self.initial_eps = eps
+        if train_eps:
+            self.eps = torch.nn.Parameter(torch.empty(1))
+        else:
+            self.register_buffer('eps', torch.empty(1))
+
+    def message(self, x_j, edge_weight):
```
Loquats, line 64: @atimashov should we add docstrings for all the methods here?

atimashov: @Loquats I checked a few implementations of layers, folks don't add. But theoretically it might be beneficial.

PR #8615, opened 2023-12-13
HeteroLinear/SEGMM: switch from heuristic to timing-cache

puririshi98:
needed to rebase https://github.com/pyg-team/pytorch_geometric/pull/8472


akihironitta (commented):
What does this PR do on a high level?

torch_geometric/nn/dense/linear.py
```diff
@@ -258,64 +259,110 @@ def reset_parameters(self):
...
+            if (type_vec[1:] < type_vec[:-1]).any():
+                type_vec, perm = index_sort(type_vec, self.num_types)
+                x = x[perm]
+
+        type_vec_ptr = index2ptr(type_vec, self.num_types)
+
+        if torch_geometric.backend.use_segment_matmul is None:
+            use_segment_matmul = False
+
+            # TODO check cses of compiling and scripting properly
```
akihironitta, line 330: ~noob question: What does "cses" stand for?~

```suggestion
            # TODO check cases of compiling and scripting properly
```

torch_geometric/nn/dense/linear.py
```diff
@@ -258,64 +259,110 @@ def reset_parameters(self):
...
+            subset_out = x[off_start:off_end] @ self.weight[i]
+            # The data type may have changed with mixed precision:
+            out[off_start:off_end] = subset_out.to(out.dtype)
+
+        return out
+
+    @torch.jit.unused
+    def _update_timing_cache(self, x: Tensor, type_vec_ptr: Tensor,
+                             num_rows: int) -> bool:
+        measure_iter = 3
```
akihironitta, line 282: nit: I'm in favor of putting this as a constant outside the class so that one can easily experiment this.

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Update `GraphStore` and `FeatureStore` to support distributed training ([#8083](https://github.com/pyg-team/pytorch_geometric/pull/8083))
 - Disallow the usage of `add_self_loops=True` in `GCNConv(normalize=False)` ([#8210](https://github.com/pyg-team/pytorch_geometric/pull/8210))
 - Disable device asserts during `torch_geometric.compile` ([#8220](https://github.com/pyg-team/pytorch_geometric/pull/8220))
+- switch to selecting `use_segment_matmul` in `HeteroLinear` based on a rudimentary timing-cache instead of relying on heuristic ([#8615](https://github.com/pyg-team/pytorch_geometric/pull/8615))
```
akihironitta, line 57: I don't think timing-cache is clear to users. Maybe something like this?

```suggestion
- Changed `use_segment_matmul` in `HeteroLinear`, and it now benchmarks both with and without the flag and uses a faster one in subsequent calls ([#8615](https://github.com/pyg-team/pytorch_geometric/pull/8615))
```

akihironitta (commented):
- Is this something similar to https://pytorch.org/docs/stable/backends.html#torch.backends.cudnn.benchmark in a sense that it first benchmarks all algos and then select the fastest one? Shall we make this as an opt-in feature just like `cudnn.benchmark`?
- I'm not sure how unlikely it is, but is the measurament robust enough for each rank to always pick up the same algorithm in a distributed setting?

torch_geometric/nn/dense/linear.py
puririshi98: agreed

puririshi98:
> Is this something similar to https://pytorch.org/docs/stable/backends.html#torch.backends.cudnn.benchmark in a sense that it first benchmarks all algos and then select the fastest one? Shall we make this as an opt-in feature just like cudnn.benchmark?

i think in our case there are only two algorithms to check and in many cases the speed difference is drastic(in either direction). we are currently selecting it based on sklearn hueristics i trained from benchmarks on a single A100 machine on a single software stack from about half a year ago. this is far less robust on the majority of cards besides A100. i think the new solution is much better. I think instead of making the hueristic opt-in seperately, we can make it a part of the existing `torch_geometric.backend.use_segment_matmul

> I'm not sure how unlikely it is, but is the measurament robust enough for each rank to always pick up the same algorithm in a distributed setting?

i think this is a very low likelihood and we run a similarly low risk with the current heuristic when using dynamic shapes on each rank

puririshi98:
@rusty1s any thoughts on the above discussion?

rusty1s:
Yes, I agree that this solution is much better than the heuristic-based one. I think we would need to avoid measuring warmup times though, and likely adjust `MEASURE_ITER` based on whether we are in a test or in an actual example.

puririshi98:
`
Yes, I agree that this solution is much better than the heuristic-based one. I think we would need to avoid measuring warmup times though, and likely adjust MEASURE_ITER based on whether we are in a test or in an actual example.

`

that sounds like a good idea to me. do you have a suggested method of implementing such a system?

rusty1s:
Currently, we use `'pytest' in sys.modules` to detect whether we are testing.

puririshi98:
okay thanks @rusty1s, do you have a desired value for MEASURE_ITER when we are testing? how about for when we aren't

rusty1s:
I think for testing `MEASURE_ITER=1` would be sufficient. The best value for scripts needs to be recommended by you :)

akihironitta (approved):
LGTM! (but I'd suggest waiting for another review from Matthias)

Also, I think we can follow up in a separate PR, but we should update `RGCNConv` right?

puririshi98:
`but we should update RGCNConv right?`
i agree, i can add this to my todo list and will follownup w/ a future PR

puririshi98:
@rusty1s anything else needed to merge?

rusty1s:
Please let me take a look tomorrow :) I'll try to get to it. If you are feeling blocked, please go ahead and merge.

rusty1s (approved):
That's nice :)

PR #8623, opened 2023-12-15
CS224W: PanRep: Graph neural networks for extracting universal node embeddings in heterogeneous graphs

Joey-Ji:
Our team (for CS224W) implemented the PanRep (https://arxiv.org/abs/2007.10445), a graph neural network for extracting universal node embeddings in heterogeneous graphs.
1. This commit includes four decoders and one RGCN encoder (as described in the paper)


rachitk:
The primary reason why all of the checks are failing is because this imports `dgl` and makes extensive use of `dgl` functions, which is not currently a PyG dependency. While some of the `dgl` functions seem useful, I do think most of them can be refactored out or replaced with PyG alternatives to avoid adding such a bulky dependency. 

Some additional first-pass comments:
- The input graph is passed as an argument to initialize the model and is stored as an instance attribute. This could probably be refactored to instead pass the graph metadata when initializing the model and its respective layers and instead pass the graph during the forward pass (which would also allow one to operate on a different graph without reinitializing the model). 
- The decoder models are initialized using a normal Python dictionary; however, this will likely lead to issues with registering their parameters and they should instead be initialized inside of a `ModuleDict`.

Joey-Ji:
> The primary reason why all of the checks are failing is because this imports `dgl` and makes extensive use of `dgl` functions, which is not currently a PyG dependency. While some of the `dgl` functions seem useful, I do think most of them can be refactored out or replaced with PyG alternatives to avoid adding such a bulky dependency.
> 
> Some additional first-pass comments:
> 
> * The input graph is passed as an argument to initialize the model and is stored as an instance attribute. This could probably be refactored to instead pass the graph metadata when initializing the model and its respective layers and instead pass the graph during the forward pass (which would also allow one to operate on a different graph without reinitializing the model).
> * The decoder models are initialized using a normal Python dictionary; however, this will likely lead to issues with registering their parameters and they should instead be initialized inside of a `ModuleDict`.

Thanks for the comment! Working on fixing the issues now.

PR #8651, opened 2023-12-21
Added the `GeoGNN` model

### LINKED ISSUE #8626: Request to Implement GeoGNN

I'd like to implement the paper [Geometry-enhanced molecular representation learning for property prediction](https://www.nature.com/articles/s42256-021-00438-4) by Fang _et al_.

In short, they propose a novel architecture called GeoGNN which encodes additional spatial information about molecular geometries by modeling both atom to bond and bond to bond angle relations. They do so by creating two graphs, one for each relation type as shown below

<p align="center">
Screenshot 2023-12-15 at 3 11 28 PM
</p>

The message passing for graph $G$ (the atom-bond graph) looks like

```math
\begin{aligned}
    \mathbf{a^{(k)}_{u}} &= AGG^{(k)}_G\left( \{ (\mathbf{h^{(k-1)}_u}, \mathbf{h^{(k-1)}_v}, \mathbf{h^{(k-1)}_{uv}} : v \in \mathit{N}(u) \}\right)\\
    \mathbf{h^{(k)}_{u}} &= COMBINE^{(k)}_G (\mathbf{h^{(k-1)}_{u}}, \mathbf{a^{(k)}_{u}})
\end{aligned}
```

and for graph $H$ (the bond-bond angle graph) it looks like

```math
\begin{aligned}
    \mathbf{a^{(k)}_{uv}} &= AGG^{(k)}_H\left( \{ (\mathbf{h^{(k-1)}_{uv}}, \mathbf{h^{(k-1)}_{uw}},  \mathbf{x_{wuv}}) : w \in \mathit{N}(u) \} \right. \notag \\ 
    & \qquad \qquad \qquad \left. \cup \{ (\mathbf{h^{(k-1)}_{uv}}, \mathbf{h^{(k-1)}_{vw}},  \mathbf{x_{uvw}}) : w \in \mathit{N}(v) \}\right)\\
    \mathbf{h^{(k)}_{uv}} &= COMBINE^{(k)}_H (\mathbf{h^{(k-1)}_{uv}}, \mathbf{a^{(k)}_{uv}})\\
\end{aligned}
```

They show it does quite well on the MoleculeNet dataset. Would it be possible to contribute it to PyG?

Screenshot 2023-12-15 at 4 00 38 PM

### Alternatives

There's a repo [here](https://github.com/PaddlePaddle/PaddleHelix/blob/dev/pahelix/model_zoo/gem_model.py) written in `paddle`.

### Additional context

_No response_

jasona445:
Fixes #8626. Added the modified GIN in `geognn_conv.py` and two-graph GeoGNN in `geognn.py` from [Geometry-enhanced molecular representation learning for property prediction](https://www.nature.com/articles/s42256-021-00438-4). 

 Not sure if this is the right place for everything - please let me know if I can move things around / add tests or examples.


torch_geometric/contrib/nn/models/geognn.py
```diff
@@ -0,0 +1,76 @@
+import torch
+
+from torch_geometric.contrib.nn.conv.geognn_conv import GeoGNNConv
+from torch_geometric.nn import global_mean_pool
+
+
+class GeoGNN(torch.nn.Module):
+    """Modified version of GeoGNN from the
```
rusty1s, line 8: Can you shed some light here about its modifications and what that means?

torch_geometric/contrib/nn/models/geognn.py
```diff
@@ -0,0 +1,76 @@
...
+    <https://www.nature.com/articles/s42256-021-00438-4>`_.
+
+    Args:
+        embedding_size (int): The size of the node embeddings.
+        layer_num (int): The number of GNN layers.
+        dropout (float, optional): Dropout rate for regularization.
+            Defaults to 0.2.
+    """
+    def __init__(self, embedding_size, layer_num, dropout=0.2):
+        super(GeoGNN, self).__init__()
```
rusty1s, line 20:
```suggestion
        super().__init__()
```

jasona445:
Modifications are with respect to the `PaddlePaddle` implementation, but I think follow the main implementation described by the paper. In the PaddlePaddle model, [`forward`](https://github.com/PaddlePaddle/PaddleHelix/blob/bc48e6be56a0c39fe30437bd744406eb30f961e5/pahelix/model_zoo/gem_model.py#L127) expects two arguments, `atom_bond_graph` and `bond_angle_graph`.

`atom_bond_graph` looks like

```
{
  "class": "Graph",
  "num_nodes": 49534,
  "edges_shape": [144176,2],
  "node_feat": [
    {"name": "atomic_num", "shape": [49534], "dtype": "paddle.int64"}
        …
    {"name": "hybridization", "shape": [49534], "dtype": "paddle.int64"}
  ],
  "edge_feat": [
    {"name": "bond_dir", "shape": [144176], "dtype": "paddle.int64"}, 
        …
    {"name": "bond_length", "shape": [144176], "dtype": "paddle.float32"}
  ]
}
```

And `bond_angle_graph` look like this, where only one edge feature is present. The number of nodes is the same as the number of edges in `atom_bond_graph`, as here the bonds are the vertices and the bond-angles are the edges.

```
{
  "class": "Graph",
  "num_nodes": 144176,
  "edges_shape": [412564,2],
  "node_feat": [],
  "edge_feat": [
    {"name": "bond_angle",  "shape": [412564], "dtype": "paddle.float32”}
  ]
}
```

Inside the `forward` pass, features for both graphs are embedded using `nn.Embedding` layers for the discrete features, and RBF + `nn.Linear` for the continuous features. A single embedding for each node is generated by sum pooling the embeddings of each of its features. The same is done for the edges.

In this PR, I’ve pulled the embedding process out of the `GeoGNN` class to allow for embeddings to be done separately, and the model can be run via

```
model = GeoGNN(embedding_size, layer_num, dropout)

ab_graph = torch_geometric.data.Data(
    x=torch.randn(10, embedding_size),
    edge_index=torch.tensor([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
                                [1, 2, 3, 4, 5, 6, 7, 8, 9,
                                0]]), batch=torch.tensor([0] * 10))

ba_graph = torch_geometric.data.Data(
    x=torch.randn(10, embedding_size),
    edge_index=torch.tensor([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
                                [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]]),
    edge_attr=torch.randn(10,
                            embedding_size), batch=torch.tensor([0] * 10))

ab_graph_repr, ba_graph_repr, node_repr, edge_repr = model(ab_graph, ba_graph)
```

Here, the `torch.randn(10, embedding_size)` used for the node/edge features are assumed to be preprocessed embeddings. Atom embeddings are specified in `ab_graph.x` and bond and angle embeddings are specified in `ba_graph.x` and `ba_graph.edge_attr` respectively. My hope was that pulling the embedding out from the model class would make it more modular and similar to other `torch_geometric.nn.model`’s.

The larger change comes as some what of a consequence of pulling out the embedding stage, as the way the PaddlePaddle implementation is formed, the bond embeddings fed into each bond-angle [`GeoGNNBlock`](https://github.com/PaddlePaddle/PaddleHelix/blob/bc48e6be56a0c39fe30437bd744406eb30f961e5/pahelix/model_zoo/gem_model.py#L32-L58) layer are calculated *de novo* from the original features instead of from the previous’ layers outputs. That is, the encodings for each layer are calculated like the figure below.

Screenshot 2023-12-23 at 2 45 47 PM

The atom embedding weights correspond to the `nn.Embedding` params learned for each of the atom features. This is only done once, as the outputs of the atom-bond GIN (AB-GIN) are used as the atom embeddings for subsequent layers. In contrast, there are multiple sets of bond embedding weights, as [the input to each bond-angle layer](https://github.com/PaddlePaddle/PaddleHelix/blob/bc48e6be56a0c39fe30437bd744406eb30f961e5/pahelix/model_zoo/gem_model.py#L146-L149) is calculated by applying its corresponding `nn.Embedding` and `nn.Linear` params to the original inputs, instead of just using the outputs of the previous BA-GIN layer. The bond-angles are omitted for clarity, but they are [embedded just like the bonds are](https://github.com/PaddlePaddle/PaddleHelix/blob/bc48e6be56a0c39fe30437bd744406eb30f961e5/pahelix/model_zoo/gem_model.py#L145).

This implementation deviated slightly from my understanding of the original message passing, which looks like this. The authors note that the AGGREGATE and COMBINE are those from the original GIN paper.
Screenshot 2023-12-22 at 4 08 30 PM
Screenshot 2023-12-22 at 4 08 39 PM

To my understanding, this corresponds to a scheme that looks more like this, where the previous BA-GIN layer’s outputs are used as inputs to the next.

Screenshot 2023-12-23 at 2 58 31 PM

This is the version that I ended up implementing, as it seemed to match with what the message passing that the paper suggested in equations (3) and (4), and allowed the prepocessing to be left to the user. I’m still not completely sure which is better, and could attempt to implement the other version as well. I might also attempt to benchmark both. Let me know what you think.

Takaogahara:
Hey!

I had a chance to take a look at the proposed model, and I must say it's pretty interesting!
Do you know if there have been any updates on the merge?

Looking forward to hearing more about it.

PR #8718, opened 2024-01-04
Handle node-level and edge-level temporal information when generating partitions

kgajdamo:
**Description**

**Temporal data definition:**
Time data can be added to the FeatureStore in two ways:
- it can be obtained directly from the partition using `LocalFeatureStore.from_partition()` function
or
- you can add them yourself using the `put_tensor()` function on the `LocalFeatureStore` object.

- Node-level temporal data: each partition must have the same time vector, which is global and its size is equal to the number of nodes in the whole graph.
Why:
We operate on global node ids.

- Edge-level temporal data: each partition has its own time vector, which is local for a given partition and its size is equal to the number of edges in the given partition (part_data.edge_index.size(1)).
Why:
Each partition has its own unique edge_index in COO format, which is later converted to a CSR/CSC matrix in the neighbor sampler. So we do not have information about the global edge IDs when sampling and we would not be able to find the correct time information for a specific edge. Therefore, this information must be local.

**How to distinguish node-level or edge-level temporal data:**
- `time_attr`='time' for node-level temporal sampling.
- `time_attr`='edge_time' for edge-level temporal sampling. It is different from a single machine case when both edge time and node time have `time_attr`='time'. It is handled this way because of the lack of the node_store/edge_store in the feature store, so at the moment we determine whether to use node-level or edge-level temporal sampling based on the attribute name.

**Where temporal data is saved:**
- `time` has been added to the node features -> node_feats.pt
- `edge_time` has been added to the edge features -> edge_feats.pt


torch_geometric/distributed/partition.py
```diff
@@ -167,9 +175,9 @@ def generate_partition(self):
                         data.edge_index[:, global_eid],
                         torch.stack((global_row, global_col), dim=0),
                     )
-                    local_eid = global_eid - edge_offset[edge_type]
```
JakubPietrakIntel, line 170: I think it's more clear if we keep the original naming because `global_eid` becomes `local_eid` within edge group.

torch_geometric/distributed/partition.py
```diff
@@ -167,9 +175,9 @@ def generate_partition(self):
...
+                    global_eid = global_eid - edge_offset[edge_type]
                     assert torch.equal(
```
JakubPietrakIntel, line 179: `local_eid = global_eid - edge_offset[edge_type]
`

torch_geometric/distributed/partition.py
```diff
@@ -167,9 +175,9 @@ def generate_partition(self):
...
-                        self.data[edge_type].edge_index[:, local_eid],
```
JakubPietrakIntel, line 172: `self.data[edge_type].edge_index[:, local_eid],`

JakubPietrakIntel (commented):
looks great, thanks :)

torch_geometric/distributed/partition.py
kgajdamo: These are still global edge ids but belonging to a specific partition. If we would keep the original naming we will write to edge_feats:
{
'global_id': local_eid,
...
}
which I found a bit misleading. Therefore, I suggest changing the name to, for example: partition_eids? What do You think?

kgajdamo: @JakubPietrakIntel After some thought, I changed the name to "offseted_eid", because from what I noticed, the introduction of offset is the reason for changing the name to local_eid.

torch_geometric/distributed/partition.py
```diff
@@ -85,6 +85,32 @@ def __init__(
...
+        if 'edge_time' in self.data:
+            return True
+        if 'time' in self.data:
+            if self.is_hetero:
+                for store in self.data.edge_stores:
+                    if 'time' in store:
+                        return True
+            else:
+                if self.data.is_edge_attr('time'):
+                    return True
```
rusty1s, line 111: Can't we use `is_edge_attr` both for hetero/homo case?

kgajdamo: I am receiving _KeyError: 'time'_ when I check if the 'time' is an edge attribute.

kgajdamo: It was solved in the same way in the `NeighborSampler.init()`: [link](https://github.com/pyg-team/pytorch_geometric/blob/32e3872721b9475c760d54a2c22ca8365f957017/torch_geometric/sampler/neighbor_sampler.py#L110C15-L110C15)

torch_geometric/distributed/partition.py
```diff
@@ -267,17 +310,24 @@ def generate_partition(self):
...
-                    torch.save(
-                        {
-                            'global_id': edge_id,
-                            'feats': dict(edge_attr=part_data.edge_attr[perm]),
-                        }, osp.join(path, 'edge_feats.pt'))
+                    efeat = {
+                        'global_id': edge_id,
+                        'feats': dict(edge_attr=part_data.edge_attr[perm]),
+                    }
+                    if is_edge_level_time:
```
rusty1s: Is it expected that we only add `edge_time` in case there exists edge features?

rusty1s (approved):

torch_geometric/distributed/partition.py
kgajdamo: While writing an e2e script that uses the MovieLens dataset (specifically: MovieLens(path, model_name='all-MiniLM-L6-v2') I noticed that there were no edge attributes there. So I had to move edge_time outside the condition. Does it make sense?

rusty1s: Yes, but here it is inside the condition, right? Maybe I am misunderstanding.

kgajdamo: You are right. Here is the fix: [#8815](https://github.com/pyg-team/pytorch_geometric/pull/8815)

PR #8722, opened 2024-01-05
Enable distributed link hetero sampling

kgajdamo:
This PR enables distributed edge sampling for heterogeneous graphs.

**Added:**
- Distributed edge heterogeneous sampling.
- Distributed edge heterogeneous node-level and edge-level temporal sampling.
- `DistEdgeHeteroSamplerInput` class, which serves as an input data to the `node_sample` function when for a given input edge there are different source and target node types.
- unit tests

**Comments:**
- In the case when a given input edge has distinct source and destination node types it is necessary to handle the data of each of these types separately, so it is slightly different from the situation when we have only one input node type.
- This PR depends on: [#8718](https://github.com/pyg-team/pytorch_geometric/pull/8718)


denadai2:
thanks for doing this!!

rusty1s (approved):

PR #8739, opened 2024-01-08
Merge `TemporalData` with `Data`/`HeteroData` [2/n]

DamianSzwichtenberg:
This PR attempts to merge `TemporalData` into `HeteroData`.
Currently, one can perform time-related modification/query using a specific node or edge type:
```Python
# bad: usage based on `BaseStorage` methods
for e in hetero_data.edge_types:
    hetero_data[e].up_to(10)
res = all([hetero_data[e].is_sorted_by_time() for e in hetero_data.edge_types])
```
This is possible because `BaseStorage` implements those functionalities. For better usability, however,
these functions should be exposed in the `HeteroData` API directly. Such an approach would also allow us to
perform operations on many node or edge types at once:
```Python
# good: usage based on `HeteroData` API
out = hetero_data.up_to(hetero_data.edge_types, 10)
res = hetero_data.is_sorted_by_time(hetero_data.edge_types)
```

Related: #3230, #8454

### DIFF
--- a/torch_geometric/data/hetero_data.py
+++ b/torch_geometric/data/hetero_data.py
@@ -24,6 +24,7 @@ from torch_geometric.typing import (
     QueryType,
     SparseTensor,
     TensorFrame,
+    TimeType,
     torch_frame,
 )
 from torch_geometric.utils import (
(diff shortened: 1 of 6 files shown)


torch_geometric/data/hetero_data.py
```diff
@@ -778,6 +779,74 @@ def edge_type_subgraph(self, edge_types: List[EdgeType]) -> Self:
...
+    def concat(self, hetero_data: Self) -> Self:
+        r"""Concatenates :obj:`self` with another :obj:`hetero_data` object.
+        All values needs to have matching shapes at non-concat dimensions.
+        """
+        out = copy.copy(self)
+        for store, other_store in zip(out.stores, hetero_data.stores):
+            store.concat(other_store)
+        return out
+
+    def is_sorted_by_time(
```
rusty1s, line 791: Isn't the base implementation sufficient?

torch_geometric/data/hetero_data.py
```diff
@@ -778,6 +779,74 @@ def edge_type_subgraph(self, edge_types: List[EdgeType]) -> Self:
...
+                                        List[NodeType]]) -> Self:
+        r"""Sorts data associated with :obj:`time` according to :obj:`time` for
+        all :obj:`types` (either edge types or node types).
+        """
+        if not types:
+            return self
+
+        out = copy.copy(self)
+        # TODO: should we limit `out` to be a [edge|node]_type_subgraph?
+        for t in [self._to_canonical(t) for t in types]:
```
rusty1s, line 817: no need to call `_to_canonical` here, right?

torch_geometric/data/hetero_data.py
DamianSzwichtenberg: Well, base implementation is sufficient for all methods, if we want to **always** apply them to all node/edge types. The API provided here is more flexible.

DamianSzwichtenberg: @rusty1s, any more comments here? 😉

torch_geometric/data/hetero_data.py
```diff
@@ -778,6 +779,71 @@ def edge_type_subgraph(self, edge_types: List[EdgeType]) -> Self:
...
+        """
+        if not types:
+            return self
+
+        out = copy.copy(self)
+        for t in types:
+            out[t].sort_by_time()
+        return out
+
+    def snapshot(self, types: Union[List[EdgeType], List[NodeType]],
```
wsad1, line 820: I have a question about `snapshot` which is relevant to this PR and #8454. 
Assume nodes n1 and n2 with time equals 1 and 4 respectively, and an edge e1 (between n1 and n2 ) with time = 5. 
On calling `snapshot(start_time = 4, end_time = 5)` what would the output look like, from the code it looks like the returned graph will have node n2 and edge e1, but e1 doens't have a source node.

torch_geometric/data/hetero_data.py
```diff
@@ -778,6 +779,71 @@ def edge_type_subgraph(self, edge_types: List[EdgeType]) -> Self:
...
+        """
+        if not types:
+            return self
+
+        out = copy.copy(self)
+        for t in types:
+            out[t].snapshot(start_time, end_time)
+        return out
+
+    def up_to(self, types: Union[List[EdgeType], List[NodeType]],
```
wsad1, line 834: For both `up_to` and `snapshot` , do we reindex edges. Like if we remove the 0th node from the graph, then does the 1st node become the 0th node in the edge index?

torch_geometric/data/hetero_data.py
```diff
@@ -778,6 +779,71 @@ def edge_type_subgraph(self, edge_types: List[EdgeType]) -> Self:
...
+            out[t].sort_by_time()
+        return out
+
+    def snapshot(self, types: Union[List[EdgeType], List[NodeType]],
+                 start_time: TimeType, end_time: TimeType) -> Self:
+        r"""Returns a snapshot restricted by :obj:`start_time` and
+        :obj:`end_time` for all :obj:`types` (either edge types or node
+        types).
+        """
+        if not types:
```
wsad1, line 826: To simpliy the interface for the user, if `types=None` we could assume the user wants to check for all `types` in the graph.

torch_geometric/data/hetero_data.py
DamianSzwichtenberg: In the current design, `time` cannot be an edge and node attribute at the same time, so this scenario is impossible. As these functionalities are moved directly from `TemporalData` I would consider underlying data rather as a list of events.

torch_geometric/data/hetero_data.py
DamianSzwichtenberg: We don't. But that's a good point. When I integrated `TemporalData` into `Data` I kept the assumption that we represent events, so `time` as a node attribute is currently not well defined.

torch_geometric/data/hetero_data.py
wsad1: Is there some part of the code that explicitly prevents this?
Does this also apply to hetero graphs?

DamianSzwichtenberg: 2x Yes, for example at [BaseStorage](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/data/storage.py#L360) and [NeighborSampler](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/sampler/neighbor_sampler.py#L75).

PR #8743, opened 2024-01-09
Upgrade visualize_graph for explain module

Tonniia:
For current `visualize_graph`, the node colors are default 'white', which is not as intuitive as in GNNExplainer: target node is red, different node class with different node color.
We add three optional params:
- node_label: node class label
- color_dict: node color for each node class
- target_node: target node to explain for node-level explanation
- draw_node_idx: if or not draw node index. For some cases, it is not necessary, and when nodes number is large, the visualization will be messy

We add lines at two examples `examples/explain/gnn_explainer.py` and `examples/explain/gnn_explainer_ba_shapes.py`

I am not sure about how to add optional params. Currently we add them at **kwargs.


torch_geometric/visualization/graph.py
```diff
@@ -127,10 +134,26 @@ def _visualize_graph_via_networkx(
             ),
         )
 
+    node_color = ['white'] * len(g.nodes)
+    if node_label != None:
```
wsad1, line 138: So if `node_label` is `None` won't all nodes be white?

wsad1 (commented):
Great, thanks for improving the visualization. 
Could you please post some images of how things looked before and after your change.
I've left some comments.

PR #8752, opened 2024-01-12
Return an empty tensor from `utils.group_argsort` if its input tensor is empty

akihironitta:


torch_geometric/utils/_scatter.py
```diff
@@ -243,11 +243,20 @@ def group_argsort(
             (default: :obj:`False`)
         stable (bool, optional): Controls the relative order of equivalent
             elements. (default: :obj:`False`)
+
+    Example:
+        >>> src = torch.tensor([0, 0, 1, 1, 1, 1, 2, 2, 2])
+        >>> index = torch.tensor([0, 1, 5, 4, 3, 2, 6, 7, 8])
```
rusty1s: I assume you need to swap `src` and `index` here.

rusty1s:
```suggestion
        >>> src = torch.tensor([0, 1, 5, 4, 3, 2, 6, 7, 8])
        >>> index = torch.tensor([0, 0, 1, 1, 1, 1, 2, 2, 2])
```

rusty1s (approved):

torch_geometric/utils/_scatter.py
akihironitta: ah yes! PTAL at 563c007 :)

PR #8795, opened 2024-01-19
add `examples/rel_bench.py`

puririshi98:


puririshi98:
> @puririshi98 Thanks for the contribution. Is this something beyond just copy-and-paste of the example code from relbench? We are still working on relbench scripts, so it's better to keep the example in one place.

nope, just that. i was just adding it to make it a part of our CI for PyG. i can go the alternative route of just testing the script in side relbench

PR #8802, opened 2024-01-21
feat(ci/labeler): bump `actions/labeler` and add `skip-changelog` label

SauravMaheshkar:
This PR aims to make the CI process better for contributors by adding the ability to auto-assign the `skip-changelog` label by naming the branch `skip*`. With the [new version of `actions/labeler` we can specify a list of regex expressions](https://github.com/actions/labeler?tab=readme-ov-file#match-object) to match with for the head branch.

This will prevent failing CI in case where updating the `CHANGELOG` isn't considered necessary. 

Perhaps this should also be reflected in the docs somewhere, most likely the contributing guide. More than happy to update the docs as well.

Request for Review: @rusty1s

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,54 +1,54 @@
 documentation:
-- changed-files:
-  - any-glob-to-any-file: "docs/**/*"
+  - changed-files:
+      - any-glob-to-any-file: "docs/**/*"
 
 example:
-- changed-files:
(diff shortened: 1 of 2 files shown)


SauravMaheshkar:
`yamllint` seems to pass locally:

```bash
pre-commit run yamllint --files .github/workflows/labeler.yml
Lint yaml................................................................Passed
```

PR #8815, opened 2024-01-24
[`Distributed`] Partition `MovieLens` dataset

kgajdamo:
Changes made:

- added MovieLens dataset to the partitioning script (partition_graph.py)
- edge time can be defined independently of edge_attrs

### DIFF
--- a/examples/distributed/pyg/partition_graph.py
+++ b/examples/distributed/pyg/partition_graph.py
@@ -6,7 +6,7 @@ import torch
 from ogb.nodeproppred import PygNodePropPredDataset
 
 import torch_geometric.transforms as T
-from torch_geometric.datasets import OGB_MAG, Reddit
+from torch_geometric.datasets import OGB_MAG, MovieLens, Reddit
 from torch_geometric.distributed import Partitioner
 from torch_geometric.utils import mask_to_index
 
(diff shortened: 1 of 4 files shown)


examples/distributed/pyg/partition_graph.py
```diff
@@ -106,17 +129,32 @@ def get_idx_split(dataset, dataset_name, split_data):
...
 
+    elif dataset_name == 'MovieLens':
+        edge_type = ('user', 'rates', 'movie')
+        train_data, val_data, test_data = T.RandomLinkSplit(
+            num_val=0.1,
+            num_test=0.1,
+            neg_sampling_ratio=0.0,
+            edge_types=[edge_type],
+            rev_edge_types=[('movie', 'rev_rates', 'user')],
+        )(dataset[0])
```
rusty1s, line 140: Looks like we drop the adjustment of message passing here computed in `train_data`, `val_data` and `test_data`. This is not necessarily a blocker (since I don't necessarily see a good way to fix this for now), but something you should be aware of. Currently, you would like information.

rusty1s (approved):

examples/distributed/pyg/partition_graph.py
kgajdamo: Sorry, I didn't get that. What do you mean by "we drop the adjustment of message passing here computed in train_data, val_data and test_data"?

rusty1s: The `edge_index` is different across different splits for link prediction tasks (in order to not leak information).

kgajdamo: I see. What about the solution used in the temporal_link_pred example?: [temporal_link_pred.py #L27](https://github.com/pyg-team/pytorch_geometric/blob/787598715c4f0f89d82ab310372c58ed0cc61308/examples/hetero/temporal_link_pred.py#L27) Maybe we can use that one instead?

rusty1s: Yeah, using temporal sampling is usually a good option to resolve this.

PR #8818, opened 2024-01-25
Rectify the default attention mode "node" of the `nn.HypergraphConv` layer

A-LOST-WAPITI:
Greetings, esteemed individuals.

While acquainting myself with the utilization of the nn.HypergraphConv layer through the application of mock inputs, I have encountered an issue with the default attention mode, namely "node," which disrupts the functionality of the forward function.

Upon examining the source code, I have identified a potential problem with the num_nodes parameter within the softmax function at line 177. From my perspective, when attempting to compute attention scores for nodes within a hyperedge, the number of hyperedges should be considered as num_nodes.

Therefore, I have modified the value of num_nodes to correspond to the pre-calculated quantity of num_edges. As a result, the issue has been resolved, and the functionality has been restored harmoniously.

Warmest regards.


rusty1s (approved):
Yes, you are absolutely right. Thanks for catching :)

A-LOST-WAPITI:
You're most welcome.

I appreciate your prompt response in addressing this issue, and it is also my aspiration to enhance the excellence of pyg.

PR #8823, opened 2024-01-25
Draft:Improvements to multinode papers100m default hyperparams and adding eval on all ranks

puririshi98:
will test before and after this PR before its ready to merge

### DIFF
--- a/examples/multi_gpu/papers100m_gcn_multinode.py
+++ b/examples/multi_gpu/papers100m_gcn_multinode.py
@@ -1,6 +1,6 @@
 """Multi-node multi-GPU example on ogbn-papers100m.
 
-To run:
+Example way to run using srun:
 srun -l -N<num_nodes> --ntasks-per-node=<ngpu_per_node> \
 --container-name=cont --container-image=<image_url> \
 --container-mounts=/ogb-papers100m/:/workspace/dataset
@@ -16,7 +16,8 @@ from ogb.nodeproppred import PygNodePropPredDataset
 from torch.nn.parallel import DistributedDataParallel
 
 from torch_geometric.loader import NeighborLoader
-from torch_geometric.nn import GCNConv
+from torch_geometric.nn.models import GCN
+from torchmetrics import Accuracy
 
 
 def get_num_workers() -> int:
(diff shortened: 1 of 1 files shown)


puririshi98:
@akihironitta @rusty1s anything needed to get this merged?

examples/multi_gpu/papers100m_gcn_multinode.py
```diff
@@ -54,97 +41,128 @@ def run(world_size, data, split_idx, model):
...
-                out = model(batch.x, batch.edge_index)[:batch.batch_size]
-            pred = out.argmax(dim=-1)
-            y = batch.y[:batch.batch_size].view(-1).to(torch.long)
-
-            total_correct += int((pred == y).sum())
-            total_examples += y.size(0)
-        print(f"Test Acc: {total_correct / total_examples:.4f}")
+                out = model(batch.x, batch.edge_index)[:batch_size]
+            acc_sum += acc(out[:batch_size].softmax(dim=-1),
+                           batch.y[:batch_size])
```
rusty1s, line 118: unclear why we use `torchmetrics` here when we don't actually use it. Would be in favor of dropping the dependency.

examples/multi_gpu/papers100m_gcn_multinode.py
```diff
@@ -54,97 +41,128 @@ def run(world_size, data, split_idx, model):
...
+        acc_sum = torch.tensor(float(acc_sum), dtype=torch.float32,
+                               device=device)
+        dist.all_reduce(acc_sum, op=dist.ReduceOp.SUM)
+        num_batches = torch.tensor(float(i + 1), dtype=torch.float32,
```
rusty1s, line 122: This would be only correct with `drop_last=True`.

examples/multi_gpu/papers100m_gcn_multinode.py
```diff
@@ -54,97 +41,128 @@ def run(world_size, data, split_idx, model):
...
+        if rank == 0:
+            print(
+                f"Validation Accuracy: {acc_sum/(num_batches) * 100.0:.4f}%", )
+            sec_per_iter = (time.time() - start) / (num_batches - warmup_steps)
+            print(f"Avg Inference Iteration Time: {sec_per_iter:.6f} s/iter")
+    dist.barrier()
+
+    model.eval()
+    acc_sum = 0.0
+    for i, batch in enumerate(test_loader):
```
rusty1s, line 135: Can't we out-source to its own function to avoid the duplication between val/test?

puririshi98: yes

puririshi98:
@rusty1s lmk if anything else needed to merge

rusty1s (approved):

PR #8841, opened 2024-01-31
ARGO: an easy-to-use runtime to improve GNN training performance on multi-core processors 

jasonlin316:
## Overview
The GNN training performance on multi-core processors is limited as the current design cannot scale well. We propose a runtime system named ARGO that can improve the scalability of GNN training on multi-core processors. On a CPU platform where the original program can only scale to 16 cores (meaning that no performance improvement is achieved if more than 16 cores are applied), ARGO can further scale the design up to 64 cores, achieving up to 5x speedup compared to the original design without ARGO.


examples/argo.py
```diff
@@ -0,0 +1,76 @@
...
+                 space=[(2, 8), (1, 4), (1, 32)], random_state=1):
+        self.n_search = n_search # number of configuration searches the auto-tuner will conduct
+        self.epoch = epoch # number of epochs of training
+        self.batch_size = batch_size # size of the mini-batch
+        self.space = space # range of the search space, [number of processes, number of samplers, number of trainers]
+        self.random_state = random_state # number of random initializations before searching
+        self.acq_func = 'EI' # acqusition function of the auto-tuner
+        self.counter = [0]
+
+    def core_binder(self, num_cpu_proc, n_samp, n_train, rank):
```
wsad1, line 19: Add type hints.

examples/argo.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def auto_tuning(self, train, args):
+        ep = 1
+        result = gp_minimize(lambda x: self.mp_engine(x, train, args, ep),
+                             dimensions=self.space, n_calls=self.n_search,
+                             random_state=self.random_state,
+                             acq_func=self.acq_func)
+        return result
+
+    def mp_engine(self, x, train, args, ep):  # Multi-Process Engine
```
wsad1, line 39: Type hints here too.

wsad1 (commented):
This is great thanks for adding!
Need your help with a few things to merge this PR
1. Add a link to the paper and also give a short explanation of how ARGO speeds things up.
2. Could you share some benchmarks on pyg models. Showing how much ARGO speeds up py models on high core cpus?
3. Add type hints to all functions and also a short doc for each function.

examples/argo.py
jasonlin316: Thanks for your suggestions!
I added type hints to each function and provided a short doc.
The description of ARGO is available on top of the file now.
For benchmarking results, they are available in the paper (Fig. 10 and 11).
Please let me know if you need anything else from me :)

examples/argo.py
```diff
@@ -0,0 +1,230 @@
...
+        epoch=200,
+        batch_size=4096,
+        space=[(2, 8), (1, 4), (1, 32)],
+        random_state=1,
+    ):
+        """
+        Initialization
+
+        Parameters
+        ----------
```
akihironitta, line 34: Can we follow the same docstring style as the PyG codebase?
```suggestion
        Args:
```

examples/argo.py
```diff
@@ -0,0 +1,230 @@
...
+        batch_size: int
+            Size of the mini-batch
+
+        space: list[Tuple(int,int)]
+            Range of the search space; [range of processes, range of samplers for each process, range of trainers for each process]
+
+        random_state: int
+            Number of random initializations before searching
+
+        acq_function: str
```
akihironitta, line 50: Looks like `acq_function` isn't used. Mind updating the docsting or the code?

akihironitta (commented):
Looks interesting! Do you think we could have an end-to-end example?

examples/argo.py
jasonlin316: Hmm...I can change the term to "Args." But I think the PyG doc uses "PARAMETERS"? 
For example, the doc here: https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html

examples/argo.py
jasonlin316: Sure thing.

jasonlin316:
> Looks interesting! Do you think we could have an end-to-end example?

Yes. I do have end-to-end example: https://github.com/jasonlin316/ARGO/tree/main/PyG
Please see [flickr_example_ARGO.py](https://github.com/jasonlin316/ARGO/blob/main/PyG/flickr_example_ARGO.py) which enables ARGO on the [flickr_example.py](https://github.com/jasonlin316/ARGO/blob/main/PyG/flickr_example.py).

examples/argo.py
akihironitta: The built docs show "PARAMETERS", but in the codebase, we follow this docstring style afaik: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

akihironitta (commented):
I feel like we should have a complete example because only looking at this class may not be clear enough for most PyG users to adapt it to their use cases IMHO.

examples/argo.py
jasonlin316: Understood. I have updated the document accordingly.

jasonlin316:
> I feel like we should have a complete example because only looking at this class may not be clear enough for most PyG users to adapt it to their use cases IMHO.

Yes, I agree having a complete example is important for the users.
@rusty1s Do you want to comment on this? I recall the plan is to add ARGO to the PyG package, and then add an example. Should we finalize this PR first, and then create another PR for the end-to-end example?

rusty1s:
@jasonlin316 Can you give me write access to your fork?

examples/argo.py
```diff
@@ -0,0 +1,252 @@
...
+ARGO is a runtime system that offers scalable performance by overlapping the computation and communication during GNN training.
+With ARGO enabled, we are able to scale over 64 cores, allowing ARGO to speedup GNN training (in terms of epoch time) by up to 5.06x and 4.54x on a Xeon 8380H and a Xeon 6430L, respectively.
+--------------------------------------------
+Paper Link: https://arxiv.org/abs/2402.03671
+"""
+
+import time
+from typing import Callable, Tuple
+
+import dgl.multiprocessing as dmp
```
rusty1s, line 14: Is this needed? We don't want to have a `dgl` dependency in PyG. Can't we use Python/PyTorch for this?

rusty1s:
@jasonlin316 I moved to `torch_geometric` package directly, moved the doc-string and adjusted input arguments. If you have time, can you follow up with adjusting docs/arguments in the remainder? In addition, how hard would it be to test `ARGO` in `test/nn/models/test_argo.py`?

examples/argo.py
XGZepto: replace with torch.multiprocessing works in my use case

PR #8842, opened 2024-01-31
`graphgym`: A correction in documentation and added information on how to install requirements

rraadd88:
1. In the [documentation of `graphgym`](https://pytorch-geometric.readthedocs.io/en/latest/advanced/graphgym.html), `configs/example_node.yaml` should be `configs/pyg/example_node.yaml`, based on the current location of that file in the repository.
2. The documentation lacked information on how to install the requirements for `graphgym`. So I added a `graphgym/configure.py` file which extracts `graphgym`'s requirements from the `pyproject.toml`. Accordingly, following line is added  in the documentation:
```
python configure.py;pip install -r requirements.txt
```
I hope this helps.


rraadd88 (commented):
`example_node.yaml` is located in `pyg` folder.

rusty1s (approved):
Thanks!

rraadd88 (commented):
About `pip install torch-geometric[graphgym]`, I wonder if running this command could replace the installed version of the package. I suppose one should have the same version of the package installed and the (git-)cloned to use `graphgym`. But if this is not the case, users could lose their installed version of `pyg`.

I had suggested separating the requirements from toml and installing them, for this reason. By installing only the requirements, users would not lose their installed version of `pyg`.

rusty1s:
IMO, it will not override the existing version of the package. This is definitely the recommended way to install the missing dependencies. I personally would like to avoid adding a script that parses the toml and creates a requirement file from it.

PR #8873, opened 2024-02-06
Intel/rect hgam homo - draft for HGAM based on rectangular adj matrix in homogeneous case (CSR only)

andreazanetti:
Sharing this code just to discuss the idea. 
**Not to be merged.**

As per the title, this is a draft of the rectangular adj matrix HGAM implementation -for homogeneous data, CSR flow- which computes only the representations needed to proceed with the next step in the computation of the GNN.

In real settings the only files that should be changed are:
- [sage_conv.py](https://github.com/pyg-team/pytorch_geometric/compare/master...andreazanetti:pytorch_geometric:intel/rect_hgam_homo?expand=1#diff-9e954450b69fcce15f59d201d40bc959f14a3d6133f2fa9147d2f64e816082b0)  
- [torch_geometric/utils/_trim_to_layer.py](https://github.com/pyg-team/pytorch_geometric/compare/master...andreazanetti:pytorch_geometric:intel/rect_hgam_homo?expand=1#diff-49458186c476522ae2cf5cecb6cec81fe59536cf2aa66c51c2d273470166b8c9)

Other changes that are appear in the code included are connected to the data collection for analysis, or remained there from the initial testing and should be ignored.

**Overall idea for rect adj HGAM**
The overall idea is about identifying 2 sets of nodes at each layer. 
S1: The set of nodes which are necessary for the computation of the sought nodes representation at that layer (our "inputs")
S2: The set of nodes which we need to compute a new representation for (our "outputs")

S2 is a subset of S1 in general (at least assuming that the current representation of each node is used to compute its next layer representation)

Therefore S1 populates the columns of adj_t, whereas S2 populates the rows of adj_t (edge_index)
In the HGAM current implementation in PyG, at each layer we identify S1 and the representations for all the nodes in S1 are computed, since we use a square adj matrix, which gets trimmed as we proceed through the layers, moving from S1 at a layer to the S1 of the next layer. That is why in the current implementation HGAM [does not act](https://github.com/pyg-team/pytorch_geometric/blob/64a12680315dbb5bcd00c4e09cf836b88373d3b3/torch_geometric/utils/_trim_to_layer.py#L87) `if layer  == 0`, because all the nodes in the batch graph are in S1, at the first layer.
Although that approach is not optimal, it is much better than computing representations for all nodes in the graph batch at all layers.

Optimality is reached when at each layer we compute only the representations for nodes from the S2 set, for that layer (which will become S1 for the next layer).
That is what we are doing here in the case of homo, sage_conv, CSR.
Hope it makes sense.

Thanks in advance for your comments.


andreazanetti:
Closing to open a new PR that contains hetero case as well (still for CSR only).

PR #8874, opened 2024-02-06
Change method of partitioning MovieLens

kgajdamo:
**Changes made:**
- Perform link-level split instead of using T.RandomLinkSplit(...)
- partition{X}.pt consists of a dictionary with the following data: edge_label_index, edge_label, edge_label_time.
- added  support to sort edge_index with respect to node time or edge time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added PyTorch 2.2 support ([#8857](https://github.com/pyg-team/pyg-lib/pull/8857))
 - Added fallback code path for `segment` in case `torch-scatter` is not installed ([#8852](https://github.com/pyg-team/pytorch_geometric/pull/8852))
 - Added support for custom node labels in `visualize_graph()` ([#8816](https://github.com/pyg-team/pytorch_geometric/pull/8816))
-- Added support for graph partitioning for temporal data in `torch_geometric.distributed` ([#8718](https://github.com/pyg-team/pytorch_geometric/pull/8718), [#8815](https://github.com/pyg-team/pytorch_geometric/pull/8815))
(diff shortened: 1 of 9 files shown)


rusty1s (commented):
Let me know once this is ready for review :)

kgajdamo:
> Let me know once this is ready for review :)

Sure, I am still getting this error and I can't find the reason :|

rusty1s:
This would mean that local neighborhoods are not sorted according to time.

kgajdamo:
> This would mean that local neighborhoods are not sorted according to time.

Yes I know, this error comes from pyg-lib code. But I wonder why it happens, since I used a similar approach as in the [temporal_link_pred.py](https://github.com/pyg-team/pytorch_geometric/blob/master/examples/hetero/temporal_link_pred.py) example and no additional sorting was used there so the time data from the dataset should be valid.

rusty1s:
And all local graph store instances respect the constraint from pyg-lib?

kgajdamo:
> And all local graph store instances respect the constraint from pyg-lib?

When I check manually, the time values are actually not sorted. The most probable thing to me was that there is an error during partitioning, for example that the data was not permuted correctly. However, looking at the dataset and the corresponding edge_index and time values, the dependencies between them are preserved after the partitioning process.

kgajdamo:
@rusty1s I see what's missing. There is an additional sort in here: [utils.py#L26](https://github.com/pyg-team/pytorch_geometric/blob/98b5e7f4fb5c2bdeda68f30d193d83bc1a443041/torch_geometric/sampler/utils.py#L26). We don't use this function during partitioning. I'll try to fix this as soon as possible.

kgajdamo:
@rusty1s  It’s ready for review. Sorry for the delay. I’m glad you noticed that edge_label_time initialisation. Now I see that the values I provided were so small that it probably discarded the entire neighborhood before checking whether the time data was sorted. I added `lexsort([edge_time, global_col])` with respect to edge time and now it works fine.

rusty1s (approved):

PR #8888, opened 2024-02-08
Make `ModuleDict` and `ParameterDict` available at `torch_geometric.nn`-level

akihironitta:
Mainly for docs. Open to any other suggestions!

Built docs: https://pytorch-geometric--8888.org.readthedocs.build/en/8888/modules/nn.html

### DIFF
--- a/docs/source/modules/nn.rst
+++ b/docs/source/modules/nn.rst
@@ -1,10 +1,20 @@
 torch_geometric.nn
 ==================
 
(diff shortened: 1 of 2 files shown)


akihironitta:
Ok, looking at the source having no docstrings, I guess these classes might be meant to be private. I just wanted a doc page to link to for our release note, but closing this as I think it's unnecessary.

PR #8898, opened 2024-02-12
added modifed version of PMLP with edge attributes

chaous:


chaous:
I used this architecture for my project and adding edge features increased accuracy of PMLP

PR #8935, opened 2024-02-19
Make `batch` optional in the global pooling layers

JeS24:
Currently, `batch=None` has to be mandatorily passed to the global pooling layers in `torch_geometric/nn/pool/glob.py`. This PR makes it optional, as the type annotations assume.

I can update the [tests](https://github.com/pyg-team/pytorch_geometric/blob/master/test/nn/pool/test_glob.py) too, if needed.

### DIFF
--- a/torch_geometric/nn/pool/glob.py
+++ b/torch_geometric/nn/pool/glob.py
@@ -5,7 +5,7 @@ from torch import Tensor
 from torch_geometric.utils import scatter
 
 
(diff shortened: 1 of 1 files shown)


rusty1s:
Hey, we had a similar PR a while ago: https://github.com/pyg-team/pytorch_geometric/pull/6857.

In the end, I would like to avoid adding `None` here as the default as it increases the risk to misuse this function.

JeS24:
Ah, I see. Makes sense. I am closing this PR.

PR #8978, opened 2024-02-27
Improved distributed tests

arlesniak:
This PR fixes distributed tests in cases when an Exception raised in subprocess is not correctly manifested to pytest, finally reported as PASSED, causing false positive result.
Changes made: 

- `assert_run_mproc` test utility function for decorating and running Processes
- `MPCaptOutput` context utility class, capturing std output and stderr


JakubPietrakIntel (dismissed):
LGTM, we might need to keep an eye on dist tests after this is merged

test/distributed/test_dist_link_neighbor_sampler.py
```diff
@@ -445,21 +445,11 @@ def test_dist_link_neighbor_sampler(disjoint):
...
-        target=dist_link_neighbor_sampler,
-        args=(world_size, 1, port, disjoint),
-    )
-
-    w0.start()
-    w1.start()
-    w0.join()
-    w1.join()
+    procs = (mp_context.Process(target=dist_link_neighbor_sampler,
+                                args=(0, port, disjoint)),
```
kgajdamo, line 449: missing world_size=2 at the beginning of the args? Same for other cases.

arlesniak: world size is calculated in `dist_util.py:59` from number of elements of procs parameter:
`world_size = len(procs)`

kgajdamo: So we can remove it from the function arguments?

arlesniak: `world_size` is used i.e. in the above code in `dist_link_neighbor_sampler` function, so it's still needed. However calculation of it later in the assert_run_proc makes it easiy to pass any number of processes without the need for remembering to pass it as the first parameter for the original target function

JakubPietrakIntel:
`Notice:  A new release of pip is available: 23.0.1 -> 24.0
Notice:  To update, run: pip install --upgrade pip
torch_geometric/testing/dist_utils.py:25: error: "TextIO" has no attribute "getvalue"  [attr-defined]
torch_geometric/testing/dist_utils.py:28: error: "TextIO" has no attribute "getvalue"  [attr-defined]
torch_geometric/testing/dist_utils.py:64: error: "Process" has no attribute "_target"  [attr-defined]
torch_geometric/testing/dist_utils.py:65: error: "Process" has no attribute "_target"  [attr-defined]
torch_geometric/testing/dist_utils.py:66: error: "Process" has no attribute "_args"  [attr-defined]`

@arlesniak we need to look into thos CI issues first

test/distributed/test_dist_neighbor_sampler.py
```diff
@@ -396,21 +401,10 @@ def test_dist_neighbor_sampler(disjoint):
...
-    w0 = mp_context.Process(
-        target=dist_neighbor_sampler,
-        args=(world_size, 0, port, disjoint),
-    )
-
-    w1 = mp_context.Process(
-        target=dist_neighbor_sampler,
-        args=(world_size, 1, port, disjoint),
-    )
+    procs = (ProcArgs(target=dist_neighbor_sampler, args=(0, port, disjoint)),
```
JakubPietrakIntel, line 404: is there a reason for using tuple and list interchangeably?

JakubPietrakIntel (approved):

test/distributed/test_dist_neighbor_sampler.py
arlesniak: no special reason, `assert_run_mproc`  accepts Collection as the parameter `pargs: Collection[ProcArgs],` dist_utils.py:59

kgajdamo (approved):
Looks good to me

rusty1s (approved):
Amazing :)

PR #8979, opened 2024-02-27
Add transformation with a chronological split

DamianSzwichtenberg:
This PR adds the `ChronologicalSplit` transformation, which splits the data into training, validation, and test parts, satisfying the following relationship:
```
max(train_time) < min(val_time) && max(val_time) < min(test_time)
```

Related issues: #3230 #8454

### DIFF
--- a/torch_geometric/data/data.py
+++ b/torch_geometric/data/data.py
@@ -292,7 +292,7 @@ class BaseData:
         end_time: Union[float, int],
     ) -> Self:
         r"""Returns a snapshot of :obj:`data` to only hold events that occurred
-        in period :obj:`[start_time, end_time]`.
+        in period :obj:`<start_time, end_time)`.
         """
         out = copy.copy(self)
         for store in out.stores:
@@ -301,7 +301,7 @@ class BaseData:
 
     def up_to(self, end_time: Union[float, int]) -> Self:
         r"""Returns a snapshot of :obj:`data` to only hold events that occurred
-        up to :obj:`end_time` (inclusive of :obj:`edge_time`).
+        before :obj:`end_time`.
         """
         out = copy.copy(self)
         for store in out.stores:
--- a/torch_geometric/transforms/chronological_split.py
+++ b/torch_geometric/transforms/chronological_split.py
@@ -1,12 +1,12 @@
 import copy
-from itertools import chain
-from typing import Optional, Tuple, Union
+from typing import List, Tuple, Union
 
 import numpy as np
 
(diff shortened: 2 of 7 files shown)


torch_geometric/data/data.py
```diff
@@ -308,6 +308,25 @@ def up_to(self, end_time: Union[float, int]) -> Self:
             store.up_to(end_time)
         return out
 
+    def train_val_test_split_time(
```
rusty1s, line 311: Any objection to introduce a `TemporalSplit` as a `transform`?

DamianSzwichtenberg: No objections :wink: Code updated.

torch_geometric/transforms/chronological_split.py
```diff
@@ -0,0 +1,81 @@
...
+            return self._forward_hetero_data(data)
+
+    def _forward_data(self, data: Data) -> Tuple[Data, Data, Data]:
+        assert hasattr(data, self.key)
+
+        val_time, test_time = np.quantile(
+            data[self.key].numpy(),
+            [1 - self.val_ratio - self.test_ratio, 1 - self.test_ratio])
+        train_data = data.snapshot(0, val_time)
+        val_data = data.snapshot(val_time, test_time)
```
wsad1, line 58: Snapshot return values in the period `[start, end]`. Given that both `train_data` and `val_data` can have data at `val_time`.

torch_geometric/transforms/chronological_split.py
```diff
@@ -0,0 +1,81 @@
...
+        if isinstance(data, Data):
+            return self._forward_data(data)
+        else:  # HeteroData
+            return self._forward_hetero_data(data)
+
+    def _forward_data(self, data: Data) -> Tuple[Data, Data, Data]:
+        assert hasattr(data, self.key)
+
+        val_time, test_time = np.quantile(
+            data[self.key].numpy(),
```
wsad1, line 55:
```suggestion
        val_time, test_time = torch.quantile(
            data[self.key],
```
Nit: Can we use torch.quantile.

torch_geometric/transforms/chronological_split.py
DamianSzwichtenberg: This PR changes that behavior, take a look at [this line](https://github.com/pyg-team/pytorch_geometric/pull/8979/files#diff-46e4e6f655447088ed5da6e3517b0417723c19498cf31d0c28c781bea7e7d551R295). After re-thinking this I thought that right-excusive range is more natural.

Issue #8997, opened 2024-03-01
Behavior of `EdgeIndex.get_csr` for directed networks with zero-outdegree nodes


IngoScholtes:
### 🐛 Describe the bug

Hi

I have a question about the intended behavior of `torch_geometric.EdgeIndex.get_csr`. Specifically, for directed graphs where some nodes have zero out-degree, the `row_ptr` returned by `torch_geometric.EdgeIndex.get_csr` does not contain repeated index entries for all-zero rows. This implies that the dense representation of the CSR matrix misses all rows that only contain zero elements. Moreover, it has consequences when using CSR-based indexing to access elements in rows that are zero (which raises an IndexError in pyG, but yields the correct result in scipy). 

As this behavior deviates from the CSR format used in `scipy.sparse` (see below), I wonder whether this is intended? 

Minimal example to reproduce the issue (and comparison to scipy's CSR format):

```
import torch
from torch_geometric import EdgeIndex

# graph where nodes 1 and 2 have zero out-degree
e = EdgeIndex(torch.tensor([[0,0,0],[0,1,2]]), sort_order='row')
((row_ptr, col), perm) = e.get_csr()
print(row_ptr)
print(e.to_dense())
```

Output: 

```
tensor([0, 3])
tensor([[1., 1., 1.]])
```

I would have expected the same output as for the CSR matrix class in `scipy.sparse`

```
from scipy.sparse import csr_matrix
import numpy as np

m = csr_matrix((3, 3), dtype=np.int32)
m[0,0] = 1
m[0,1] = 1
m[0,2] = 1
print(m.indptr)
print(m.todense())
```

Output: 

```
[0 3 3 3]
[[1 1 1]
 [0 0 0]
 [0 0 0]]
```

Note: This behavior is due to the use of `get_num_rows()` in the call to function `torch._convert_indices_from_coo_to_csr` [here](https://github.com/pyg-team/pytorch_geometric/blob/9b660ac6ca882604d1ae521912d20ded1d180ecf/torch_geometric/edge_index.py#L632), which does not account for zero rows: 

```
torch._convert_indices_from_coo_to_csr(
            self[dim],
            self.get_num_rows(),
            out_int32=self.dtype != torch.int64,
        )
```

Replacing it by the following, would make it work for directed networks with nodes with zero out-degree (but the issue then remains for networks with completely isolated nodes).

```
torch._convert_indices_from_coo_to_csr(
            self[dim],
[...]

PR #8998, opened 2024-03-01
partial fix for issue #8997

IngoScholtes:
Possible (partial) fix for issue #8997, which would retain scipy's CSR semantics at least for directed graphs that include nodes with zero out-degree.

### DIFF
--- a/torch_geometric/edge_index.py
+++ b/torch_geometric/edge_index.py
@@ -629,7 +629,7 @@ class EdgeIndex(Tensor):
         else:
             rowptr = self._T_indptr = torch._convert_indices_from_coo_to_csr(
                 row,
-                self.get_num_rows(),
+                max(self.get_num_rows(), self.get_num_cols()),
                 out_int32=self.dtype != torch.int64,
             )
 


rusty1s:
Is this "fix" still needed based on my answer in https://github.com/pyg-team/pytorch_geometric/issues/8997? I would like to avoid this fix here since it would assume quadratic adjacency matrices.

IngoScholtes:
> Is this "fix" still needed based on my answer in #8997? I would like to avoid this fix here since it would assume quadratic adjacency matrices.

Thanks for the quick response! Yes, we can use that to work around the issue.

Our issue is that we need to do CSR-based indexing in instances of `EdgeIndex` that are created and passed by the user. We thus now check whether sparse_size() corresponds to a quadratic matrix with the correct numnber of nodes and issue an error otherwise. We could alternatively resize the sparse matrix. 

I can follow that you wouldn't like to impose quadratic matrices. We were just wondering whether it is intended that way.

Thanks!

rusty1s:
You can now also use `edge_index.sparse_resize_()` to make your matrix quadratic :)

PR #9005, opened 2024-03-02
Fixed `NeighborLoader` with remote envs

denadai2:
NeighborLoader is not sampling all seed nodes in the graph. Particularly, it discards all isolated nodes. This is a fix for that 

ref: https://github.com/pyg-team/pytorch_geometric/discussions/8990


rusty1s:
Closing this as it seemed to be resolved already.

Issue #8501, opened 2023-12-01
[Extended Roadmap] Enhanced Hypergraph Support in PyG


mehular0ra:
## Objective
Proposing enhancements to the existing roadmap for hypergraph support in PyG (refer to issue [#7312](https://github.com/pyg-team/pytorch_geometric/issues/7312)).

## Enhancements Proposed

1. **Enhancing HyperGraphData Class:**
    - Develop static functions for converting existing PyG graph datasets into hypergraph datasets.
    - Benefit: Improved benchmarking capabilities and seamless data format transitions.

2. **Symmetric Normalization in HyperGraphConv:**
    - Introduce Symmetric Normalization alongside Row Normalization.
    - Rationale: Align with hypergraph research advancements. References: ["Hypergraph Neural Networks"](https://arxiv.org/pdf/1809.09401.pdf) and ["Deep Hypergraph Learning"](https://arxiv.org/pdf/1901.08150.pdf).

3. **Integration of New Hypergraph Models:**
    - Add new models that utilize both row and symmetric normalization, as discussed in issue [#7120](https://github.com/pyg-team/pytorch_geometric/issues/7120).
    - Goal: Expand hypergraph applications in PyG.

4. **Creating an Example Utilizing New Modules:**
    - Develop a comprehensive example that showcases the use of the newly added hypergraph functionalities and models.
    - Purpose: To provide a practical, hands-on demonstration of the enhancements for users.

### Call for Collaboration
Inviting input and collaboration from the PyG community and core contributors, particularly @rusty1s and @wsad1.

### Acknowledgments
Expressing gratitude to the PyG team and community for their support and valuable discussions with @wsad1 .

colltoaction:
Hi everyone, I'm Martín and I'm a developer working in the [HIF](https://github.com/pszufe/HIF-standard) project, an effort to share hypergraphs across many libraries. It is an incidence-based encoding that is developed by members of [HypergraphX](https://github.com/HGX-Team/hypergraphx), [HyperNetX](https://github.com/pnnl/HyperNetX) and [XGI](https://github.com/xgi-org/xgi) projects for higher-order networks.

I hope you would consider taking a look at our repository where we provide the specification of a JSON format and tutorials for interoperating with all our libraries. This integration would be incredible beneficial for all! For starters we would have the ability to share datasets, but with time all Hypergraph libraries should benefit from standardizing this data structure across the whole Scientific Python ecosystem.

[...]

Issue #7312, opened 2023-05-06
[Roadmap] Better support for hypergraphs


wsad1:
### 🚀 The feature, motivation and pitch

Following this [discussion](https://github.com/pyg-team/pytorch_geometric/discussions/7120). The goal is to improve support for hypergraphs in pyg via:

- [x]  Add `HyperGraphData` object that inherits from `BaseData` and implements `__inc__`, `__cat_dim__`, `num_nodes`, `subgraph` etc for hyper graphs. The `hyperedge_index` will be of shape `[2, num_hyper_edges*num_nodes_per_hyper_edge]`, see [here](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HypergraphConv.html#torch_geometric.nn.conv.HypergraphConv) for an example.[#7611]
- [x] Test that `DataLoader` works with `HyperGraphData`. [#7611]
- [x] Add a hyper graph dataset to `Datasets`
- [ ] Add hyper graph gnns like [HGNN](https://arxiv.org/abs/1809.09401), 
- [ ] Add an example that uses all of the above new modules.

CodeTal:
@wsad1 Hello sir, I am a new open-source contributor and PyG user. I happened to see this issue and found that the hypergraph is a really interesting notion. Could I work on this issue?

CodeTal:
@rusty1s OK. I will try to do something these days.

CodeTal:
@rusty1s Thank you for letting me know!

CodeTal:
Hello, I am trying to implement the HyperGraphData class first. I have gone through some relevant codes and formulated some initial ideas. I think most functions can be inherited from the Data class, and here are something that we need to change:

1. Re-name some parameters. For example, edge_index to hyperedge_index, adj to incidence
2. Re-implement some functions. 
- Prior issues (#7365）mentioned that the \_\_inc\_\_ function should be refactored. Based on my understanding of how the tensors are collated during batching, the \_\_inc\_\_ of HyperGraphData should still return the num_nodes for the first row of hyperedge_index but return the num_hyperedges for the second row of hyperedge_index. 
- The \_\_cat_dim\_\_ function could remain the same as the concatenation method doesn't change. 
- The subgraph function should be changed because the expression of edge_index has changed.
- Other functions could remain the same as the original Data class.

I am new to this project, so I still have many confusions. If I am wrong, could you please provide some feedback when you are free? Thanks a lot!

[...]

PR #9090, opened 2024-03-21
feat: add `CornellTemporalHyperGraphDatasets`

SauravMaheshkar:
Reference: #8501 #7312 

Request for Review: @rusty1s @wsad1

This PR aims to add HyperGraph datasets consisting of timestamped simplices where each simplex is a set of nodes. Released with the paper [Simplicial Closure and higher-order link prediction](https://huggingface.co/papers/1802.06916).

Notes:
> "[In compliance with PEP 8 and PEP 257, Ruff prefers double quotes for triple quoted strings and docstrings even when using quote-style = "single".](https://docs.astral.sh/ruff/settings/#format_quote-style)". 

Thus even with pre-commit passing there are changes which convert `'` to `"`. Should probably look into setting the `quote-style` property to `"single"`, if single quotes are desired. That way it can be enforced via pre-commit as well.

---
_**Update 26th March 2024**: There are multiple datasets which follow a similar processing and loading workflow. Therefore, I have refactored to make a single `CornellTemporalHyperGraphDataset` class. I'm still encountering errors with the test on windows, would appreciate some help on this_


SauravMaheshkar:
I can't figure out this path error on windows, would appreciate some help on this.

SauravMaheshkar:
@rusty1s I ended up refactoring to allow for all datasets to use the same class

SauravMaheshkar:
Gentle ping: @rusty1s

rusty1s (approved):
Cool :)

Issue #9107, opened 2024-03-26
Add edge_attr reduction argument to torch_geometric.nn.pool.pool.pool_edge


MatthieuMelennec:
### 🚀 The feature, motivation and pitch

I am working on my own graph pooling scheme based on cluster vectors assigning nodes to the cluster they get pooled into. To pool my edges and edge_attributes, I am using `pool_edge` from `torch_geometric/nn/pool/pool.py`. However, I would want to be able to choose the reduce operation that will be used in `coalesce`. As of now, it is hardcoded to the default value `'sum'`.

### Alternatives

The only solution I see is to rewrite the `pool_edge` function identically and adding an argument `reduce` that will be passed to `coalesce`.

rusty1s:
That's a good request. Do you mind sending a PR for that?

PR #9116, opened 2024-03-27
Add `reduce` argument to `nn.pool.pool.pool_edge`

MatthieuMelennec:
To specify edge_attr reduction scheme (issue #9107)

### DIFF
--- a/torch_geometric/nn/pool/pool.py
+++ b/torch_geometric/nn/pool/pool.py
@@ -5,8 +5,12 @@ import torch
 from torch_geometric.utils import coalesce, remove_self_loops, scatter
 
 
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/pool/pool.py
```diff
@@ -5,12 +5,14 @@
 from torch_geometric.utils import coalesce, remove_self_loops, scatter
 
 
-def pool_edge(cluster, edge_index, edge_attr: Optional[torch.Tensor] = None):
+def pool_edge(cluster, edge_index, edge_attr: Optional[torch.Tensor] = None,
+              reduce: Optional[str] = "add"):
```
rusty1s, line 9:
```suggestion
              reduce: Optional[str] = 'sum', ):
```

rusty1s (approved):

PR #9154, opened 2024-04-04
G-retriever (GNN+LLM) example w/ demo

puririshi98:
GNN reduces hallucinations by: ~62.0%
Note: hallucinations detected by regex hence the ~

old PR: https://github.com/pyg-team/pytorch_geometric/pull/8984 (closed)

PR #9167, opened 2024-04-08
Add G-retriever (GNN+LLM) example

puririshi98:
1. https://github.com/pyg-team/pytorch_geometric/pull/9462
2. https://github.com/pyg-team/pytorch_geometric/pull/9480
3. https://github.com/pyg-team/pytorch_geometric/pull/9481
4. **->** https://github.com/pyg-team/pytorch_geometric/pull/9167

---

repro:
Latest NVIDIA PyG container
+
`git config --global credential.helper store; huggingface-cli login; cd /opt/pyg; pip uninstall -y torch-geometric; rm -rf pytorch_geometric; git clone -b gnn-llm-model-integration https://github.com/pyg-team/pytorch_geometric.git; cd /opt/pyg/pytorch_geometric; pip install .; pip install peft datasets transformers pcst_fast sentencepiece; python3 examples/llm_plus_gnn/g_retriever.py`

old PR: https://github.com/pyg-team/pytorch_geometric/pull/9154/

note: pure cpu is 220x slower than pure GPU using a single Grace Hopper (for llama-7b)

info:
tried gemma, performs worse in all train/val/test metrics. most likely needs some tuning, will leave this as future work as part of the community sprint to try many LLM and GNN combos and tune them. Therefore keeping the default llama2

the new gemma-v2 is also much worse than llama2


puririshi98:
@Kh4L reviews welcome

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+        if llm_name == "llama2":
+            self.printable_llm_name = "LLAMA2"
+            self.huggingface_str = llama2_str_name
+        elif llm_name == "gemma":
+            self.printable_llm_name = "GEMMA"
+            self.huggingface_str = gemma_str_name
+        else:
+            self.printable_llm_name = llm_name
+            self.huggingface_str = llm_name
+        self.mem_needed = 85 * num_params / 7
```
Kh4L, line 76: Can you add a comment explaining why 85 and 7 here

Kh4L (approved):

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -0,0 +1,359 @@
...
+        if not WITH_DATASETS:
+            missing_str_list.append('datasets')
+            missing_imports = True
+        if not WITH_PANDAS:
+            missing_str_list.append('pandas')
+            missing_imports = True
+        if missing_imports:
+            missing_str = ' '.join(missing_str_list)
+            error_out = f"`pip install {missing_str}` to use this dataset."
+            raise ImportError(error_out)
```
rusty1s, line 269: We can just import here on the fly. IMO no need to have this complex logic.

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -0,0 +1,359 @@
...
+        self.graph_type = "Knowledge Graph"
+        self.model_name = "sbert"
+        self.device = torch.device(
+            "cuda" if torch.cuda.is_available() else "cpu")
+        super().__init__(root, None, None, force_reload=force_reload)
+        self.load(self.processed_paths[0])
+
+    @property
+    def raw_file_names(self) -> List[str]:
+        return []
```
rusty1s, line 281: Is there a good reason this is empty?

torch_geometric/nn/models/__init__.py
```diff
@@ -28,50 +28,19 @@
...
-    'ViSNet',
+    'MLP', 'GCN', 'GraphSAGE', 'GIN', 'GAT', 'PNA', 'EdgeCNN',
+    'JumpingKnowledge', 'MetaLayer', 'Node2Vec', 'DeepGraphInfomax',
+    'InnerProductDecoder', 'GAE', 'VGAE', 'ARGA', 'ARGVA', 'SignedGCN',
+    'RENet', 'GraphUNet', 'SchNet', 'DimeNet', 'DimeNetPlusPlus',
+    'to_captum_model', 'to_captum_input', 'captum_output_to_dicts',
+    'MetaPath2Vec', 'DeepGCNLayer', 'TGNMemory', 'LabelPropagation',
+    'CorrectAndSmooth', 'AttentiveFP', 'RECT_L', 'LINKX', 'LightGCN',
+    'MaskLabel', 'GroupAddRev', 'GNNFF', 'PMLP', 'NeuralFingerprint', 'ViSNet',
+    'GNN_LLM'
```
rusty1s, line 45:
```suggestion
    'GNN_LLM', 
```

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+
+    for i in range(gpus_2_use_4_llm):
+        max_mem_dict[i] = str(avail_mem_dict[i]) + "GiB"
+    kwargs["max_memory"] = max_mem_dict
+    kwargs["device_map"] = "auto"
+    return kwargs
+
+
+class LLM(nn.Module):
+    def __init__(self, llm_name: str = "llama2", llm_dtype=torch.bfloat16,
```
rusty1s, line 64:
```suggestion
    def __init__(self, model_name: str = "llama2", dtype=torch.bfloat16,
```

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+                 num_params: int = 7):
```
rusty1s, line 65: What does this mean?

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+            break
+
+    for i in range(gpus_2_use_4_llm):
+        max_mem_dict[i] = str(avail_mem_dict[i]) + "GiB"
+    kwargs["max_memory"] = max_mem_dict
+    kwargs["device_map"] = "auto"
+    return kwargs
+
+
+class LLM(nn.Module):
```
rusty1s, line 63: Doc-string would be great.

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+                                                       use_fast=False)
+        self.tokenizer.pad_token_id = pad_token_id
+        self.tokenizer.padding_side = padding_side
+        self.llm = AutoModelForCausalLM.from_pretrained(
+            self.huggingface_str, torch_dtype=self.llm_dtype,
+            low_cpu_mem_usage=True, **kwargs)
+        self.llm_device = self.llm.device
+        self.word_embedding = self.llm.model.get_input_embeddings()
+
+    def encode_inputs(self, samples: Batch):
```
rusty1s, line 91: Can we drop the `Batch` dependency here and clearly define the expected function header?

torch_geometric/nn/models/gnn_llm.py
```diff
@@ -0,0 +1,449 @@
...
+
+        return {
+            'pred': pred,
+            'label': samples['label'],
+            'question': samples['question'],
+            'desc': samples['desc'],
+        }
+
+
+class GNN_LLM(nn.Module):
```
rusty1s, line 213: Shouldn't we call it `GRetriever` in this case?

torch_geometric/nn/models/gnn_llm.py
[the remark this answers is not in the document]
puririshi98: yes will do

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -0,0 +1,359 @@
...
+        if isinstance(index, torch.Tensor):
+            index = index.item()
+        batch_data = dict()
+        for key in self.data.keys():
+            if self.data[key] is not None:
+                batch_data[key] = self.data[key][index]
+        return batch_data
+
+
+class Sentence_Transformer(torch.nn.Module):
```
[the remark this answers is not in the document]
puririshi98: i agree ill work on this

torch_geometric/nn/models/gnn_llm.py
puririshi98: i will describe with docstring

torch_geometric/nn/models/gnn_llm.py
puririshi98: done

torch_geometric/datasets/web_qsp_dataset.py
puririshi98: the reason I have this complicated logic is that these imports are used throughout this file, not just in this class. so the names won't be recognized in other scopes.
Example:
```
>>> def f():
...     import gc
... 
>>> gc.collect()
  File "<stdin>", line 1, in <module>
NameError: name 'gc' is not defined
>>> f()
>>> gc.collect()
  File "<stdin>", line 1, in <module>
NameError: name 'gc' is not defined
```

torch_geometric/datasets/web_qsp_dataset.py
puririshi98: i left it blank since the raw files are handled by HuggingFace datasets:
```
    @property
    def raw_file_names(self) -> List[str]:
        return []

    @property
    def processed_file_names(self) -> List[str]:
        return ["list_of_graphs.pt", "pre_filter.pt", "pre_transform.pt"]

    def download(self) -> None:
        dataset = datasets.load_dataset("rmanluo/RoG-webqsp")
        self.raw_dataset = datasets.concatenate_datasets(
            [dataset["train"], dataset["validation"], dataset["test"]])
        self.split_idxs = {
            "train":
            torch.arange(len(dataset["train"])),
            "val":
            torch.arange(len(dataset["validation"])) + len(dataset["train"]),
            "test":
            torch.arange(len(dataset["test"])) + len(dataset["train"]) +
            len(dataset["validation"])
        }
```

i am open to suggestions

torch_geometric/nn/models/gnn_llm.py
puririshi98: yes

puririshi98:
@akihironitta @rusty1s addressed reviews, plz lmk if anything else needed to merge.

akihironitta (commented):
Sorry for the extreme delay :( Will have a look during this weekend (long weekend in UK) :)

puririshi98:
@akihironitta anything else needed to merge?

puririshi98:
@rusty1s @akihironitta anything else needed to merge?

puririshi98:
@rusty1s @akihironitta happy to make any required changes to get this merged finally

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -0,0 +1,251 @@
...
+        desc = textual_nodes.to_csv(index=False) + "\n" + textual_edges.to_csv(
+            index=False, columns=["src", "edge_attr", "dst"])
+        graph = Data(x=graph.x, edge_index=graph.edge_index,
+                     edge_attr=graph.edge_attr, num_nodes=graph.num_nodes)
+        return graph, desc
+
+    root = -1  # unrooted
+    num_clusters = 1
+    pruning = "gw"
+    verbosity_level = 0
```
akihironitta, line 53:
```suggestion
```

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -0,0 +1,251 @@
...
+
+        topk_e_values, _ = torch.topk(e_prizes.unique(), topk_e, largest=True)
+        e_prizes[e_prizes < topk_e_values[-1]] = 0.0
+        last_topk_e_value = topk_e
+        for k in range(topk_e):
+            indices = e_prizes == topk_e_values[k]
+            value = min((topk_e - k) / sum(indices), last_topk_e_value - c)
+            e_prizes[indices] = value
+            last_topk_e_value = value
+        # cost_e = max(min(cost_e, e_prizes.max().item()-c), 0)
```
akihironitta, line 76: unused?
```suggestion
```

akihironitta (commented):
Sorry for the delay. I feel like we should probably split this PR into (at least four) smaller PRs like #9350 so that we don't miss any important changes during review:

1. Add `LLM`
2. Add `GRetriever`
3. Add a new dataset `WebQSPDataset`
4. Add an example `g_retriever.py`

I think smaller PRs are usually easier to review :)

torch_geometric/datasets/web_qsp_dataset.py
puririshi98: its not clear to me why we need to remove this?  these are used later in the function
```
    vertices, edges = pcst_fast(edges, prizes, costs, root, num_clusters,
                                pruning, verbosity_level)
```

akihironitta: nit: it just looked unnecessary to me to assign literals to local variables, do nothing on the variables and feed them to `pcst_fast` :D

puririshi98: oh i see what you mean, yes i can move these into the function definition for cleanliness, i misunderstood the request. will do

puririshi98: okay i actually looked into this. the pcst-fast function is defined as so:
```
TypeError: pcst_fast(): incompatible function arguments. The following argument types are supported:
    1. (arg0: numpy.ndarray[numpy.int64], arg1: numpy.ndarray[numpy.float64], arg2: numpy.ndarray[numpy.float64], arg3: int, arg4: int, arg5: str, arg6: int) -> tuple[numpy.ndarray[numpy.int32], numpy.ndarray[numpy.int32]]

Invoked with: array([[   0,    1],
       [   2,    0],
       [   3,    4],
       ...,
       [1722, 1567],
       [ 551, 1723],
       [1723,   15]]), array([0.        , 0.        , 0.        , ..., 0.16666669, 0.15000004,
       0.33333337], dtype=float32), array([0.5, 0.5, 0.5, ..., 0. , 0. , 0. ]); kwargs: root=-1, num_clusters=1, pruning='gw', verbosity_level=0
```

therefore something like
```
    vertices, edges = pcst_fast(edges, prizes, costs, root=-1, num_clusters=1,
                                pruning="gw", verbosity_level=0)
```
fails because theres no kwarg support. and so the cleanest way to pass these values while making it clear what they mean is the original way. will leave as is

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
+"""This example implements G-retriever using PyG.
+Original Paper: https://arxiv.org/abs/2402.07630
+“G-Retriever significantly reduces hallucinations
+by 54% compared to the [LLM] baseline“.
+Requirements on top of basic PyG:
+`pip install datasets transformers pcst_fast sentencepiece tqdm accelerate`.
```
zechengz, line 6: Do we use `pcst_fast` here?

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
...
+def compute_accuracy(eval_output) -> float:
+    df = pd.concat([pd.DataFrame(d) for d in eval_output])
+    all_hit = []
+    all_precision = []
+    all_recall = []
+    all_f1 = []
+
+    for pred, label in zip(df.pred.tolist(), df.label.tolist()):
+        try:
+            pred = pred.split('[/s]')[0].strip().split('|')
```
zechengz, line 48: Why we need to split by `"[/s]"`?

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
...
+from tqdm import tqdm
+
+from torch_geometric import seed_everything
+from torch_geometric.datasets import WebQSPDataset
+from torch_geometric.loader import DataLoader
+from torch_geometric.nn.models import GRetriever
+from torch_geometric.nn.nlp import LLM
+
+
+def detect_hallucinate(pred, label):
```
zechengz, line 27: Can you add typing to these functions and maybe a short description on the definition of the hallucination here?

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
...
+        return model.inference(batch.question, batch.x, batch.edge_index,
+                               batch.batch, batch.ptr, batch.edge_attr,
+                               batch.desc)
+
+
+def train(since, num_epochs, hidden_channels, num_gnn_layers, batch_size,
+          eval_batch_size, lr, loss_fn, inference_fn, model=None, dataset=None,
+          checkpointing=False, tiny_llama=False):
+    def adjust_learning_rate(param_group, LR, epoch):
+        # Decay the learning rate with half-cycle cosine after warmup
```
zechengz, line 123: Any reason or reference for this LR adjustments?

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
...
+                              drop_last=True, pin_memory=True, shuffle=True)
+    val_loader = DataLoader(val_dataset, batch_size=eval_batch_size,
+                            drop_last=False, pin_memory=True, shuffle=False)
+    test_loader = DataLoader(test_dataset, batch_size=eval_batch_size,
+                             drop_last=False, pin_memory=True, shuffle=False)
+
+    # Step 2: Build Model
+    if model is None:
+        gc.collect()
+        if tiny_llama:
```
zechengz, line 156: Can we define an argument in argparse such as `model`. We can check if it's tinyllama and set corresponding arguments here.

examples/llm_plus_gnn/g_retriever.py
```diff
@@ -0,0 +1,472 @@
...
+    # save model
+    print("Saving Model...")
+    save_params_dict(model, model_save_name + ".pt")
+    print("Saving eval output for downstream demo...")
+    torch.save(eval_output, model_save_name + "_eval_outs.pt")
+    print("Done!")
+    return prep_time, dataset, eval_output
+
+
+def minimal_demo(gnn_llm_eval_outs, dataset, lr, epochs, batch_size,
```
zechengz, line 265: Maybe we can put this to a separate file? We can create a folder named with `g-retriever` and have files such as train.py, utils.py and compare.py. WDYT?

examples/llm_plus_gnn/g_retriever.py
puririshi98: its used in the dataset webqsp

examples/llm_plus_gnn/g_retriever.py
puririshi98: yes

examples/llm_plus_gnn/g_retriever.py
puririshi98: original authors of g-retriever paper from meta use this

examples/llm_plus_gnn/g_retriever.py
puririshi98: in general i prefer standalone examples since they give users of our container a smoother out of the box experience. open to other options if @rusty1s feels strongly about it

examples/llm_plus_gnn/g_retriever.py
puririshi98: sure ooto till aug 27, can circle back then if still needed

examples/llm_plus_gnn/g_retriever.py
puririshi98: the issue with doing something like that is there are too many knobs such mlp out_dim and num_llm_params which must vary with the LLM used. We would need a user to set correctly through argparser for each  model. This code serves as an example for which they can look into the api and adjust the dimensions of their model as needed for custom workflows.

rusty1s (approved):

PR #9215, opened 2024-04-19
Support `EdgeIndex` and `torch.compile()` in PyTorch 2.3

rusty1s:


torch_geometric/edge_index.py
```diff
@@ -1364,239 +1408,209 @@ def flip(
...
+#     other: Union[int, Tensor, EdgeIndex],
+#     *,
+#     alpha: int = 1,
+# ) -> Tensor:
+#     aten.add_.Tensor(
+#         input._data,
+#         other._data if isinstance(other, EdgeIndex) else other,
+#         alpha=alpha,
+#     )
+#     return input
```
akihironitta, line 1554: could be nit: it this still relevant?

akihironitta (commented):
:tada:

torch_geometric/edge_index.py
rusty1s: commented out on purpose, and brought back in #9239

PR #9254, opened 2024-04-29
Sparse version of `nn.dense.DMoNPooling`

xiaohan2012:
# What and why

- `nn.dense.DMoNPooling` requires the input adjacency matrix to be dense, which does not scale to large graphs.
- This PR adds the "sparse" version of `DMoNPooling`


carloabate:
Hey! We developed `tgp` (`torch-geometric-pool`), a graph pooling library for GNNs that operates on top of PyG. In particular, we implemented `DMoNPooling`, including a sparse adjacency path: in `tgp`, when operating with dense poolers, you can set `batched=False` to operate on sparse adjacency inputs (without dense padding/densification) and compute the auxiliary losses in that sparse mode.

Repo: https://github.com/tgp-team/torch-geometric-pool  
Docs (DMoNPooling): https://torch-geometric-pool.readthedocs.io/en/latest/api/poolers.html#tgp.poolers.DMoNPooling

PR #9335, opened 2024-05-19
Added optional `attr` argument to `BaseData.snapshot` method

### LINKED ISSUE #9333: Add optional `attr` argument to `BaseData.snapshot` method
### 🛠 Proposed Refactor

The `snapshot` method of a PyG `dataset` currently requires that the temporal attribute data is stored as `dataset.time`.

The method supports [both node- and edge-level]() temporal attributes. However, some datasets may include both node- and edge-level temporal data under distinct keys (e.g., `time` and `edge_time`), and there is currently no way to specify it.

Moreover, the result of `snapshot` when passing a node-level or edge-level data is different; the first returns a `dataset` object with filtered node data only, while the latter returns filtered edge data only, therefore fulfilling different use cases.

In order to avoid having to modify the `dataset.time` object on e.g. datasets with both attributes, PyG should IMHO ideally support receiving an `attr` argument that allows to specify the name of the attribute key in which the data is stored.

`...`

nelsonaloysio:
Fixes #9333.

### DIFF
--- a/torch_geometric/data/storage.py
+++ b/torch_geometric/data/storage.py
@@ -370,18 +370,20 @@ class BaseStorage(MutableMapping):
         self,
         start_time: Union[float, int],
         end_time: Union[float, int],
+        attr: str = 'time',
     ) -> Self:
-        if 'time' in self:
-            mask = (self.time >= start_time) & (self.time <= end_time)
+        if attr in self:
+            mask = getattr(self, attr)
+            mask = (mask >= start_time) & (mask <= end_time)
 
-            if self.is_node_attr('time'):
+            if self.is_node_attr(attr):
                 keys = self.node_attrs()
-            elif self.is_edge_attr('time'):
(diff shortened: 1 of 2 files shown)


torch_geometric/data/storage.py
```diff
@@ -370,18 +370,20 @@ def snapshot(
         self,
         start_time: Union[float, int],
         end_time: Union[float, int],
+        attr: str = 'time',
     ) -> Self:
-        if 'time' in self:
-            mask = (self.time >= start_time) & (self.time <= end_time)
+        if attr in self:
+            mask = getattr(self, attr)
+            mask = (mask >= start_time) & (mask <= end_time)
```
rusty1s, line 377:
```suggestion
            time = self[attr]
            mask = (time >= start_time) & (time <= end_time)
```

rusty1s (approved):

PR #9341, opened 2024-05-21
Convert molecular features to floating-point values

aehrlich1:
Change node feature data type from long to float.

### DIFF
--- a/torch_geometric/utils/smiles.py
+++ b/torch_geometric/utils/smiles.py
@@ -119,7 +119,7 @@ def from_smiles(smiles: str, with_hydrogen: bool = False,
         row.append(x_map['is_in_ring'].index(atom.IsInRing()))
         xs.append(row)
 
-    x = torch.tensor(xs, dtype=torch.long).view(-1, 9)
+    x = torch.tensor(xs, dtype=torch.float).view(-1, 9)
 
     edge_indices, edge_attrs = [], []
     for bond in mol.GetBonds():  # type: ignore


rusty1s:
Thanks. This doesn't look correct to me. These are categorical features which are expected to be processed via `torch.nn.EmbeddingBag`.

aehrlich1:
The issue arises when trying to pass the data object through a GCN that uses F.linear.

Here a small example where the problem comes up:

`import torch
import torch.nn.functional as F
from torch_geometric.utils import from_smiles

smiles = 'COc1cccc(c2cccc(F)c2C(=O)[O-])c1'
data = from_smiles(smiles)

# data.x = data.x.float()
print(data.x.dtype, data.edge_attr.dtype, data.edge_index.dtype)
class GCN(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.conv1 = GCNConv(9, 128)
        self.conv2 = GCNConv(128, 2)

    def forward(self, data):
        x, edge_index = data.x, data.edge_index

        x = self.conv1(x, edge_index)

        return F.log_softmax(x, dim=1)

model = GCN()
model(data)
`

This will throw an error "RuntimeError: Found dtype Long but expected Float".

rusty1s:
Yes, this is expected. The correct usage is:
```python
self.emb = torch.nn.EmbeddingBag(100, 64)

x = self.emb(data.x)
x = conv(x, edge_index)

aehrlich1:
Ahhh I see. I must have overlooked this in the documentation. My apologies for opening the PR. It can be closed.

PR #9394, opened 2024-06-05
 [FutureWarning] Fixing the warning that appears when calling the constructor of the `LogisticRegression` class

drivanov:
The following warning, which appeared in three tests, has been fixed:
```
/usr/local/lib/python3.10/dist-packages/sklearn/linear_model/_logistic.py:1247: FutureWarning: 'multi_class' was 
deprecated in version 1.5 and will be removed in 1.7. From then on, it will always use 'multinomial'. Leave it to 
its default value to avoid this warning.
```


rusty1s (approved):
Nice :)

PR #9397, opened 2024-06-06
Make train/val/test splits disjoint sets in RandomLinkSplit

mk-61:
It seems train split is erroneously used instead of validation split, and also I don't understand what `train_val_edges` variable is used for. So this PR is an attempt to fix this.
But some other tests are failing, so consider this as WIP, for now...


mk-61:
Right, makes sense, thanks for clarification!

PR #9402, opened 2024-06-06
Add pure SSG with no lin

Kh4L:
Recent work in link prediction tasks (e.g., https://arxiv.org/abs/2007.00970) has shown the need for a pure GCN convolution, without applying any linearity to the output. This PR modifies the existing `SGConv` to support GCN convolution.


Kh4L:
Closing after offline convo w @rusty1s : https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.LGConv.html exists

PR #9445, opened 2024-06-22
Fix minor issues with `ogbn_papers_100m` example

devanshamin:
### Issues
- Missing `num_workers`. 
- `epoch` passed to the `train` function that doesn't accept any arguments. 
### Fix
- Renamed `workers` to `num_workers`.
- Removed the `epoch` argument.


akihironitta (approved):
Nice catch! Congrats on your first contribution to PyG! 🎉

PR #9453, opened 2024-06-24
Update NVIDIA docker readme

puririshi98:
cuda 10.1 is pretty old, we are well into CUDA 12 now. the containers we provide are the latest and greatest of the whole software stack

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -10,7 +10,7 @@ The creation of [our dockerfile](https://github.com/pyg-team/pytorch_geometric/b
 1. `$ docker build -t "custom image name"`
 1. `$ docker run --rm -it --init --runtime=nvidia --ipc=host --network=host --volume=$PWD:/app -e NVIDIA_VISIBLE_DEVICES=0 "custom image name" /bin/bash`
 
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -1,6 +1,9 @@
 # Docker on NVIDIA GPU
 
-You can run PyG with CUDA 10.1 inside a docker image.
+The recommended way to use Docker for NVIDIA hardware is described [here](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pyg).
+
+You can also run PyG with CUDA 10.1 inside a docker image. This method is deprecated and we highly recommend the above mentioned official NVIDIA docker containers instead.
```
akihironitta, line 5: I guess 10.1 can retire :D 
```suggestion
You can also run PyG with CUDA 12.1 inside a docker image. This method is deprecated and we highly recommend the above mentioned official NVIDIA docker containers instead.
```

akihironitta (approved):
Nice!

Would be also nice to update the line below
```diff
- If you encounter any problems, please feel free to contact <liangshengwen@ict.ac.cn>.
+ If you encounter any problems, please feel free to create a GitHub issue.
```

Issue #6891, opened 2023-03-10
global_add_pool() missing default batch argument


qihuazhong:
### 🐛 Describe the bug

According to the doc (https://pytorch-geometric.readthedocs.io/en/stable/modules/nn.html#torch_geometric.nn.pool.global_add_pool) and the code. The `batch` argument should be optional. However, the default None is not provided in the function signature and would raise an error if not provided.

Should be an easy fix by chaning 
```
def global_add_pool(x: Tensor, batch: Optional[Tensor],
                    size: Optional[int] = None) -> Tensor:
```
to 
```
def global_add_pool(x: Tensor, batch: Optional[Tensor] = None,
                    size: Optional[int] = None) -> Tensor:
```

### Environment

* PyG version:
* PyTorch version:
* OS:
* Python version:
* CUDA/cuDNN version:
* How you installed PyTorch and PyG (`conda`, `pip`, source):
* Any other relevant information (*e.g.*, version of `torch-scatter`):

rusty1s:
Yes, it is optional (meaning you can pass in `None`) but it does not have a default argument (which is intended).

eurunuela:
> Yes, it is optional (meaning you can pass in `None`) but it does not have a default argument (which is intended).

It is actually a required argument. If you try to call `global_add_pool(x)` it will raise an error saying you didn't pass the `batch` argument. It should be given a default value or the docs should be updated to say that it is a required argument.

PR #9454, opened 2024-06-25
Fix issue #6891

cshjin:
Fix issue #6891

Add a default value "None" to `batch` in the global pooling functions, including

* `global_add_pool`
* `global_mean_pool`
* `global_max_pool`


rusty1s:
I would like to avoid this since in 99% of the cases, `batch` is what you need to input. I am afraid making this argument having a default, this leads to numerous mis-uses downstream.

PR #9456, opened 2024-06-25
[ISSUE 9455] Data preprocess fix in multi XPU training

zhouyu5:
This is a follow-up PR for issue #9455 .

Say we want to lauch training on two XPUs:

- Before fixing
Data will be downloaded twice and process twice, and causing conflict, see below:
image

- After fixing
Data will be downloaded and processed only once on main process, see below:
image

### DIFF
--- a/benchmark/multi_gpu/training/training_benchmark_xpu.py
+++ b/benchmark/multi_gpu/training/training_benchmark_xpu.py
@@ -50,7 +50,8 @@ if __name__ == '__main__':
         f"Dataset {args.dataset} isn't supported."
 
     # if the dataset is not present, it will be downloaded only by process rank=0
-    # other process will use the dataset cache from rank=0, and will not re-download and process it
+    # other process will use the dataset cache from rank=0, 
+    # and will not re-download and process it
     if rank == 0:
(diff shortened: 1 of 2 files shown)


benchmark/multi_gpu/training/training_benchmark_xpu.py
```diff
@@ -48,6 +48,11 @@ def custom_optimizer(model: Any, optimizer: Any) -> Tuple[Any, Any]:
 
     assert args.dataset in supported_sets.keys(), \
         f"Dataset {args.dataset} isn't supported."
-    data, num_classes = get_dataset(args.dataset, args.root)
+
+    if rank == 0:
```
DamianSzwichtenberg, line 52: nit: Please add a comment explaining why we need this, e.g. "if the dataset is not present, it will be downloaded only by process rank=0".

DamianSzwichtenberg (approved):

benchmark/multi_gpu/training/training_benchmark_xpu.py
DamianSzwichtenberg: LGTM

PR #9461, opened 2024-06-26
Fix `ConfigMixin.from_config` when config is of type `dict`

akihironitta:


akihironitta:
No need for a changelog entry as `ConfigMixin` hasn't been included in any release yet.

test/test_config_mixin.py
```diff
@@ -68,3 +68,36 @@ def test_config_mixin() -> None:
...
+
+    cfg = asdict(cfg)
+
+    model = Module.from_config(cfg)
+    assert isinstance(model, Module)
+    assert model.x == 0
+    # TODO: Support recursively restoring objects when cfg is dict.
+    # assert isinstance(model.data, Dataclass)
+    # assert model.data.x == 1
+    # assert model.data.y == 2
```
akihironitta, line 80: Haven't taken a deeper look, but ideally we'd want to support recursively restoring objects.

rusty1s (approved):

PR #9471, opened 2024-06-30
Add retailhero and movielens25 bipartite networks with causal parameters

geopanag:
I would like to add two new bipartite networks in PyG.
One is based on the RetailHero (https://ods.ai/competitions/x5-retailhero-uplift-modeling/data) (users-buy-products) with causal outcomes and interventions from a marketing campaign,  and the other is based on MovieLens25 (https://grouplens.org/datasets/movielens/25m/) (movies-ratedby-users) with observational causal parameters .

I use them to test a GNN for causal outcome prediction in https://hal.science/hal-04601553/document ( ECMLPKDD 2024 ) and I believe they re useful for this line of research.

### DIFF
--- a/test/datasets/test_movielens25.py
+++ b/test/datasets/test_movielens25.py
@@ -1,0 +1,16 @@
+from torch_geometric.testing import onlyFullTest, onlyOnline
+
+
+@onlyOnline
+@onlyFullTest
+def test_movielens25(get_dataset):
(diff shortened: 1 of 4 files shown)


geopanag:
Since there 's been some time can I ask, I am not sure if I should make any corrections e.g. for the changelog and typehints, or I should first wait for the CR. Are these checks blocking the CR?

PR #9462, opened 2024-06-26
Integrate `LLM` class

puririshi98:
breaking https://github.com/pyg-team/pytorch_geometric/pull/9167/ into subPRs

PR #9480, opened 2024-07-02
Add `nn.models.GRetriever`

puririshi98:
1. https://github.com/pyg-team/pytorch_geometric/pull/9462
2. **->** https://github.com/pyg-team/pytorch_geometric/pull/9480
3. https://github.com/pyg-team/pytorch_geometric/pull/9481
4. https://github.com/pyg-team/pytorch_geometric/pull/9167

---

breaking https://github.com/pyg-team/pytorch_geometric/pull/9167/ down further, focusing on G-retriever model this time


torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,328 @@
...
+
+        # encode graphs
+        graph_embeds = self.encode_graphs(node_feat, edge_index, edge_attr,
+                                          batch)
+        graph_embeds = self.projector(graph_embeds)
+        batch_inputs_embeds = []
+        batch_attention_mask = []
+        batch_label_input_ids = []
+        num_nodes_per_graph = ptr[1:] - ptr[:-1]
+        for i in range(batch_size):
```
rusty1s, line 181: There seems to be a lot of duplication here and in `LLM`. Why was it done this way? Is there a way to de-duplicate the amount of code here?

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,248 @@
...
+        trainable_params = 0
+        all_param = 0
+        for _, param in self.named_parameters():
+            num_params = param.numel()
+
+            all_param += num_params
+            if param.requires_grad:
+                trainable_params += num_params
+
+        return trainable_params, all_param
```
akihironitta, line 248: I think this method is too general to put on this model:
```suggestion
```

test/nn/models/test_g_retriever.py
```diff
@@ -0,0 +1,41 @@
...
+    batch.ptr = torch.tensor([0, batch.num_nodes]).to(torch.int64)
+    # simple cycle graph
+    batch.edge_index = torch.tensor(
+        [list(range(10)), list(range(1, 10)) + [0]])
+
+    model = GRetriever(
+        llm_to_use="TinyLlama/TinyLlama-1.1B-Chat-v0.1",
+        num_llm_params=1,  # 1 Billion
+        mlp_out_dim=2048,
+    )
```
akihironitta, line 31: What would be nice is to mock the HF code that `GRetriever` is calling so that we don't have to download anything for testing.

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,248 @@
...
+        try:
+            self.graph_encoder = gnn_to_use(
+                in_channels=gnn_in_channels,
+                out_channels=gnn_out_channels,
+                hidden_channels=gnn_hidden_channels,
+                num_layers=num_gnn_layers,
+                heads=num_gnn_heads,
+                norm='batch_norm',
+            ).to(self.llm_device)
+        except:  # noqa
```
akihironitta, line 104: I'm not in favor of simply adding `noqa`, but at least it should have a specific rule code to ignore, e.g., E501 for too long line length.

torch_geometric/nn/nlp/llm.py
```diff
@@ -63,21 +62,12 @@ def __init__(
...
 
-        if model_name == 'llama2-7b':
-            pretty_model_name = 'LLAMA2'
-            model_name = 'meta-llama/Llama-2-7b-chat-hf'
-        elif model_name == 'gemma':
-            pretty_model_name = 'GEMMA'
-            model_name = 'google/gemma-7b'
-        else:
-            pretty_model_name = model_name
-
```
akihironitta, line 74: I removed these aliases as they're not transparent to users.

akihironitta (commented):
Sorry for the extreme delay. Wish I could leave a thorough review here :( but I pushed a commit that updates docs, test, model selection logic. PTAL!

akihironitta:
I wonder from next time if it would make sense to start adding models to `torch_geometric.contrib.nn` first and then promote them to `torch_geometric.nn` when we're confident about the code around them.

puririshi98:
> I wonder from next time if it would make sense to start adding models to `torch_geometric.contrib.nn` first and then promote them to `torch_geometric.nn` when we're confident about the code around them.

from NVIDIAs pov we are already including this in our containers each release so I'd rather save the effort of moving these too pyg.contrib and wait until all of your needs are met before merging these prs. I will address these reviews as soon as i can, and will gladly address any follow up reviews

test/nn/models/test_g_retriever.py
puririshi98: the problem is that huggingface code under the hood changes often, it will be hard to mantain a mirror of that implementation just for testing purposes. in general thats why we chose to use huggingface as the LLM instead of implementing our own, open source LLM engineering progresses far quicker than we can keep up with for GNN+LLM, which is a smaller community. On top of this, the tiny LLM used in testing is not likely to be the same LLM used in production settings where this code will be ultimately used. the testing for huggingface this way still tests the underlying transformers backbone they use.

puririshi98:
@rusty1s @akihironitta anything else needed to merge the 3 remaining PRs?

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,236 @@
...
+        batch: torch.Tensor,
+        ptr: torch.Tensor,
+        label: List[str],
+        edge_attr: Optional[torch.Tensor] = None,
+        additional_text_context: Optional[List[str]] = None,
+    ):
+        r"""Forward pass.
+
+        Args:
+            question (List[str]): The questions/prompts.
```
zechengz, line 146: Does this means that passing a list of questions / prompts and they will use the same subgraph?

torch_geometric/nn/nlp/llm.py
```diff
@@ -117,6 +107,102 @@ def _encode_inputs(
         return (batch_size, questions, context, eos_user_tokens, bos_embeds,
                 pad_embeds)
 
+    def _label_input_ids(self, i, label, eos_tokens):
```
zechengz, line 110: Can we also add typing to the arguments in these functions? Thanks.

torch_geometric/nn/models/g_retriever.py
puririshi98: no each q/prompt is matched to a subgraph, they are just combined by the pyg dataloader collate step

torch_geometric/nn/nlp/llm.py
puririshi98: sure, im ooto until tues aug 27, but can circle back then if still needed

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.word_embedding = self.llm_to_use.word_embedding
+
+    def encode_graphs(self, node_feat, edge_index, edge_attr, batch):
+        x = node_feat.to(self.llm_device)
+        edge_index = edge_index.long().to(self.llm_device)
+        edge_attr = edge_attr.to(self.llm_device)
+        n_embeds = self.graph_encoder(x, edge_index.long(), edge_attr)
+        batch = batch.to(self.llm_device)
+        g_embeds = scatter(n_embeds, batch, dim=0, reduce='mean')
```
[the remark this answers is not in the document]
puririshi98: i think in the future we can support customizeable pooling but maybe we can do this as a follow up PR?

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,236 @@
...
+                lora_dropout=lora_dropout,
+                bias="none",
+                task_type="CAUSAL_LM",
+            )
+            self.llm_generator = get_peft_model(self.llm_generator, config)
+        self.llm_device = self.llm_to_use.llm_device
+        self.tokenizer = self.llm_to_use.tokenizer
+        try:
+            self.graph_encoder = gnn_to_use(
+                in_channels=gnn_in_channels,
```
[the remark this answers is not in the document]
puririshi98: open to whatever @rusty1s thinks is best

torch_geometric/nn/models/g_retriever.py
```diff
@@ -0,0 +1,236 @@
...
+        num_gnn_heads: int = 4,
+        mlp_hidden_dim: int = 2048,
+        mlp_out_dim: int = 4096,
+    ) -> None:
+        super().__init__()
+        self.llm_to_use = LLM(llm_to_use, num_llm_params, llm_dtype)
+        self.llm_generator = self.llm_to_use.llm
+        self.llm_dtype = llm_dtype
+        if llm_use_lora:
+            from peft import (
```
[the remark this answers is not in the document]
puririshi98: open to whatever @rusty1s thinks is best here as well

rusty1s (approved):

torch_geometric/nn/models/g_retriever.py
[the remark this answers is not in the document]
rusty1s: Moved `LLM` and `GNN` to input args.

rusty1s:
Very clean now, thanks for the updates :)

torch_geometric/nn/models/g_retriever.py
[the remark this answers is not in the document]
rusty1s: We can follow-up with that.

torch_geometric/nn/models/g_retriever.py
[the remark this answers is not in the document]
rusty1s: We can follow-up with that.

PR #9487, opened 2024-07-04
Adding `PatchTransformerAggregation`

andyhuang-kumo:
Adding patch transformer architecture from dynamic graph learning literature as a PyG aggregation module. 
Mostly based on "Simplifying Temporal Heterogeneous Network for Continuous-Time Link Prediction"
- [x] added patch transformer aggregation module in `torch_geometric/nn.aggr.patch_transformer.py`
- [x] added all related util functions in `pt_utils.py`
- [x] passed precommit hooks
- [x] added `test_patch_transformer_aggr.py` in `test/nn/aggr/ , test passed 
- [x] add documentation for helper functions
- [ ] integrate into workflow
- [x] test on dataset

Notes:

- `PatchTransformerAggregation` requires edge features(`edge_feats`) and edge timestamps(`edge_ts`)
- `PatchTransformerAggregation` primarily focus on aggregating edge information for each node based on their past interactions, the embedding of edges are fed into the patch transformer blocks which learns interaction between these edges and by pooling over multiple patches, the representation of each node is generated. Lastly, this representation is merged with the input node features (`x`) to produce an output.
- `PatchTransformerAggregation` do not run `self.to_dense_batch()` at the moment thus do not reshape the input `x` tensor.
- the output dimension is set via the `out_dim` argument


torch_geometric/nn/aggr/patch_transformer.py
```diff
@@ -0,0 +1,159 @@
...
+class PatchTransformerAggregation(Aggregation):
+    r"""Performs  Heterogeneous Link Encoder embedding
+    and Semantic Patches Fusion
+    from `"Simplifying Temporal Heterogeneous Network for
+    Continuous-Time Link Prediction
+    "<https://dl.acm.org/doi/pdf/10.1145/3583780.3615059>`_paper.
+
+    Args:
+        channels (int): Size of each edge feature
+        use to specify the time embedding dimension of edge timestamps
```
zechengz, line 24: nit:
```suggestion
            use to specify the time embedding dimension of edge timestamps.
```

torch_geometric/nn/aggr/patch_transformer.py
```diff
@@ -0,0 +1,159 @@
...
+        hidden_dim (int, optional): hidden dimension
```
zechengz, line 25: nit: we usually use `hidden_channels` in pyg. Also capitalize the first char and adding the dot at last.
```suggestion
        hidden_channels (int, optional): Hidden channels.
```

torch_geometric/nn/aggr/pt_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def reset_parameters(self):
+        self.transformer_encoder.reset_parameters()
+
+    def forward(self, x, mask: Optional[Tensor] = None):
+        x = x + self.transformer_encoder(x, x, mask, mask)
+        return x
+
+
+class PositionalEncoding1D(torch.nn.Module):
```
zechengz, line 73: Any difference between this and the https://github.com/pyg-team/pytorch_geometric/blob/fbafbc4fc9181e8759ec1f39d9618992793b5fe1/torch_geometric/nn/encoding.py#L7 Seems that the output tensor is 3D instead of 2D?
Can we use existing one and repeat to 3D?

torch_geometric/nn/aggr/patch_transformer.py
```diff
@@ -0,0 +1,159 @@
...
+        x_feat = self.pad_projector(x_feat)
+
+        # x_feat = [batch_size, max_edge, edge_feat_size]
+        x_feat = x_feat + self.pe(x_feat)
+
+        # x_feat = [batch_size, max_edge, edge_feat_size]
+        for atten_block in self.atten_blocks:
+            x_feat = atten_block(x_feat)
+
+        x_feat = self.layernorm(x_feat)
```
zechengz, line 132: image
Do we need this `layernorm`, in the paper seems they just have pooling and FC

torch_geometric/nn/aggr/patch_transformer.py
```diff
@@ -0,0 +1,159 @@
...
+                                        time_dim=self.time_dim)
+        self.layernorm = torch.nn.LayerNorm(self.hidden_dim)
+        self.atten_blocks = torch.nn.ModuleList([
+            TransformerBlock(dims=self.hidden_dim, heads=self.heads,
+                             dropout=self.dropout)
+            for _ in range(num_encoder_blocks)
+        ])
+
+        # input is node feature and transformer node embed
+        self.mlp_head = torch.nn.Linear(self.hidden_dim * (len(self.aggr)),
```
zechengz, line 81: As this is just one layer, maybe something like self.lin or self.fc is a better name?

torch_geometric/nn/aggr/patch_transformer.py
andyhuang-kumo: updated in recent PR

torch_geometric/nn/aggr/patch_transformer.py
andyhuang-kumo: updated in recent commit

torch_geometric/nn/aggr/patch_transformer.py
andyhuang-kumo: This was in the original implementation as well and seem to produce slightly better performance, let's keep it for now

torch_geometric/nn/aggr/patch_transformer.py
andyhuang-kumo: addressed in recent commit

torch_geometric/nn/aggr/pt_utils.py
andyhuang-kumo: I tried both and seem to get better performance with the `PositionalEncoding1D`, though they are very close

torch_geometric/nn/aggr/pt_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, dims):
```
zechengz: Add typing and modify the docstring for this class.

zechengz (approved):
In general LGTM. Maybe @rusty1s can help to make another review when available.

rusty1s (approved):

PR #9499, opened 2024-07-09
Fix `visualize_feature_importance` in heterogenous graphs with single features

nilserranestle:
When calling visualize_feature_importance for a HeteroExplanation object, an error occurs if one of the Heterogenous graph's node types has one single feature. The error message also had a bug, as "node_mask_dict" has no attribute "size()". 

image

This is a simple fix for both bugs.


zechengz (approved):

akihironitta:
> Do I have to update CHANGELOG.md for this simple bug fix?

Mind adding a test case and updating the changelog? It'll be useful for PyG users to see what are fixed even the fix itself is simple. Let us know if you need help!

rusty1s (approved):

PR #9515, opened 2024-07-17
Added the `residual` option for `GATConv` and `GATv2Conv`.

Linnore:
As discussed here [https://github.com/pyg-team/pytorch_geometric/discussions/7555#discussioncomment-9973267](https://github.com/pyg-team/pytorch_geometric/discussions/7555#discussioncomment-9973267), the GAT paper mentioned using skip-connection to reach the reported metrics on PPI dataset. This PR adds the option of `residual` to enable skip-connection. 

Specifically, we consider the following for adding residual:
 1. Define the `actual_out_channels` = `heads` * `out_channels` if concat else `out_channels`
 2. If `in_channels` == `actual_out_channels`, then residual is just `x` before the convolution. Otherwise, residual is a linear projection of the input `x` onto the dimension of `actual_out_channels`.

For bipartite case where the input is (x_src, x_dst), the above consideration will be applied for x_dst.


Linnore:
An example `gat_ppi.py` is added under `examples/gat_ppi.py`. Running the example with argument `--residual` and `--no-residual` can show that the implementation of `residual` is effected. 

Note that for the case `in_channels` == `actual_out_channels` described above, there is no fully connected layers added as in `examples/ppi.py` to get the residual; instead, an identity mapping is used. Though having slower convergence speed, this is a reasonable design and is the same design as DGL's `GATConv` just for reference.

I am not sure whether the `residual` is correctly implemented for the bipartite input though. It's probably problematic when the bipartite input is (x_src, x_dst=None). In `GATv2Conv`, such input should be asserted and should not be allowed, but `GATConv` seems not having the same checking. 

Feel free to comment!

torch_geometric/nn/conv/gat_conv.py
```diff
@@ -176,6 +181,21 @@ def __init__(
             self.lin_edge = None
             self.register_parameter('att_edge', None)
 
+        if residual and concat:
+            if in_channels != out_channels * heads:
+                self.lin_res = Linear(in_channels, out_channels * heads,
```
zechengz, line 186: QQ does GAT paper uses this linear transformation if the `in_channels` is different from `out_channels`?

torch_geometric/nn/conv/gat_conv.py
```diff
@@ -195,6 +215,10 @@ def reset_parameters(self):
             self.lin_dst.reset_parameters()
         if self.lin_edge is not None:
             self.lin_edge.reset_parameters()
+        if isinstance(self.lin_res, Linear):
+            self.lin_res.reset_parameters()
+            if self.lin_res.bias is not None:
```
zechengz, line 220: I think `zeros` also works if self.lin_res.bias is None

torch_geometric/nn/conv/gat_conv.py
Linnore: No need for linear transformation if the dimension matches.

torch_geometric/nn/conv/gat_conv.py
Linnore: You are right, we can save it here.

torch_geometric/nn/conv/gatv2_conv.py
Linnore, line 1: Fail to pass the test for type hints if removing these lines. So still check `self.lin_res.bias is not None` here.

torch_geometric/nn/conv/gatv2_conv.py
```diff
@@ -218,8 +218,7 @@ def reset_parameters(self):
             self.lin_edge.reset_parameters()
         if isinstance(self.lin_res, Linear):
             self.lin_res.reset_parameters()
-            if self.lin_res.bias is not None:
```
Linnore: Fail to pass the test for type hints if removing these lines. So still check self.lin_res.bias is not None here.

Linnore:
The Linting check fails due to `torch_geometric/config_mixin.py:85: error: "Type[DataclassInstance]" has no attribute "_target_"  [attr-defined]`. It is not related to this PR.

torch_geometric/nn/conv/gat_conv.py
```diff
@@ -176,6 +181,21 @@ def __init__(
             self.lin_edge = None
             self.register_parameter('att_edge', None)
 
+        if residual and concat:
```
zechengz, line 184: Maybe we can make this cleaner. For example, we can define something like
```python
if concat:
    out_channels = out_channels * heads
if residual and in_channels != out_channels:
    self.lin_res = Linear(in_channels, out_channels, ...)
elif residual:
    self.lin_res = Identity()
else:
    self.register_parameter('lin_res', None)
```

Linnore: I think in GATConv, out_channels means out channels per head. Let out_channels = out_channels * heads will require a refactor of other usages of out_channels in GATConv.

I can help adjust this if it's necessary.

zechengz: You can use a different name

Linnore: Sure. 

Since the `out_channels` behaves the same for `residual` and `bias`, I also use the new variable `out_channels_total` to simplify the codes for `bias`. 
No other usage of `heads * out_channels` should be replaced by this new variable now.

zechengz (approved):
In general LGTM. Maybe @rusty1s can take a look

Linnore (commented):
Question about adding linear transformation when `in_channels == total_out_channels`

torch_geometric/nn/conv/gatv2_conv.py
```diff
@@ -186,23 +185,22 @@ def __init__(
...
-                                      bias=bias, weight_initializer='glorot')
-            else:
-                self.lin_res = Identity()
+            self.res = Linear(
+                in_channels
+                if isinstance(in_channels, int) else in_channels[1],
+                total_out_channels,
+                bias=False,
+                weight_initializer='glorot',
+            )
```
Linnore: In this way, a linear transformation will also be applied when `in_channels == total_out_channels`. It is consistent with the original `example/ppi.py` but over-complicates the idea of skip-connection.

Maybe it's better to directly add the residual without linear transformation following the GAT authors' official implementation  here:
> https://github.com/PetarV-/GAT/blob/5af87e7fce2b90ae1cbd621cd58059036a3c7436/utils/layers.py#L27-L32

torch_geometric/nn/conv/gat_conv.py
```diff
@@ -176,10 +180,22 @@ def __init__(
...
+        if residual:
+            self.res = Linear(
+                in_channels
+                if isinstance(in_channels, int) else in_channels[1],
+                total_out_channels,
+                bias=False,
+                weight_initializer='glorot',
+            )
+        else:
+            self.register_parameter('res', None)
```
Linnore, line 195: In this way, a linear transformation will also be applied when `in_channels == total_out_channels`. It is consistent with the original `example/ppi.py` but over-complicates the idea of skip-connection.

Maybe it's better to directly add the residual without linear transformation following the GAT authors' official implementation  here:
> https://github.com/PetarV-/GAT/blob/5af87e7fce2b90ae1cbd621cd58059036a3c7436/utils/layers.py#L27-L32

PR #9562, opened 2024-08-02
Add `RemoveSelfLoops` transformation

### LINKED ISSUE #9561: Request for `RemoveSelfLoops` Transformation in PyG

I was exploring the BA2MotifDataset and noticed that it contains 40 graphs with self-loops. I wanted to remove these self-loops using a transformation like the following:
```python
from torch_geometric.datasets import BA2MotifDataset

transform = T.Compose([T.ToDevice(device=device), T.RemoveSelfLoops()])
dataset = BA2MotifDataset(root="./", transform=transform)
```

However, I found that while there are `AddSelfLoops` and `AddRemainingSelfLoops` transformations available, there is no `RemoveSelfLoops` transformation.

Since `remove_self_loops` is already implemented in `torch_geometric.utils`, it seems straightforward to add a `RemoveSelfLoops` transformation. This could be achieved by replacing the `add_remaining_self_loops`  in [add_remaining_self_loops.py](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/transforms/add_remaining_self_loops.py) with `remove_self_loops`.

Would it be possible to add this transformation to PyG? I believe it would be a useful addition for many users dealing with datasets containing self-loops.

Thank you for considering this request!

Best regards

### Alternatives

_No response_

### Additional context

_No response_

kano5266:
Fix  #9561


rusty1s (approved):
Super :)

PR #9569, opened 2024-08-03
Add OPF solution objective function value as a graph level property

viktor-ktorvi:
As discussed with @mzgubic 

This PR has to do with the recently added OPFDataset. It reads the OPF solution objective function value from the json file, and adds it as a graph level property. This value is useful in creating optimality metrics that illustrate how far an ML predictions objective function is from the one achieved by a solver.


viktor-ktorvi:
Should I make changes to the CHANGELOG myself?

mzgubic (approved):
LGTM

rusty1s (approved):

Issue #8415, opened 2023-11-21
Scatter_reduce onnx export error


sevenineone:
I'm trying to export gnn model with onnx and get following error:

```bash
OnnxExporterError: ONNX does not support include_self=False for scatter_reduce
```

Here is example to reproduce this error:

```python
import torch
from torch import nn
from torch_geometric.nn import aggr, TransformerConv
from torch_geometric.nn.norm import GraphNorm


class GNN(nn.Module):
    def __init__(self):
        super(GNN, self).__init__()

        self.gnn = TransformerConv(10, 10, 1, dropout=0.1).jittable()         
        self.normalization = GraphNorm(in_channels=10)
        self.act = nn.GELU()
        self.aggr = aggr.AttentionalAggregation(gate_nn=nn.Linear(10, 1), nn=nn.Linear(10, 10)) #doesn't compile

    def forward(self, x, edge_index, batch_ptr):
        hidden = x
        hidden = self.gnn(hidden, edge_index)
        hidden = self.normalization(hidden)
        hidden = self.act(hidden)

        graph_rep = self.aggr(hidden, ptr=batch_ptr)

        return graph_rep

    
device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
model = GNN().to(device)

x = torch.randn(100, 10).to(device)
edge_index = torch.randint(100, size=(2, 20)).to(device)
batch_ptr = torch.tensor([0, 100]).to(device)
res = model(x, edge_index, batch_ptr)
print(res)

torch.onnx.export(model, (x, edge_index, batch_ptr), 'model.onnx',
                  input_names=('x', 'edge_index', 'batch_ptr'), opset_version=18)
```

### Environment

* PyG version: 2.4.0 (master)
* PyTorch version: 2.1.0
* OS: ubuntu 22
* Python version: 3.10
* CUDA/cuDNN version: 12.3
* How you installed PyTorch and PyG (`conda`, `pip`, source): source

rusty1s:
There is sadly not much we can do here besides waiting that `torch.scatter_reduce` is supported by ONNX :(

sevenineone:
`torch.scatter_reduce` is supported by ONNX. ` include_self=False ` parameter is unsuported. Changing this to `True` will have almost no effect on the logic, but will make conversion to onnx possible

[...]

PR #9587, opened 2024-08-12
Support ONNX export for `scatter` with min/max reduction

lee-clement-oxb:
Attempt to address issue https://github.com/pyg-team/pytorch_geometric/issues/8415 and unblock ONNX export. 

As I understand it, the issue is that when we're applying min/max reductions with `scatter_reduce_` to populate a `self` tensor of zeroes or an empty tensor, we need to set `include_self=False` to prevent the initial fill value from influencing the min/max results. For that reason simply setting `include_self=True` in these cases is not the right solution since it would change the behaviour of the reduction.

A simple behaviour-preserving solution is to initialize the `self` tensor such that it doesn't influence the min/max result when `include_self=True`. This can be done using the dtype min/max values provided by `torch.finfo` or `torch.iinfo` to initialize the elements of `self` that will be populated by `scatter_reduce_`. Unfortunately they aren't currently scriptable (see https://github.com/pytorch/pytorch/issues/25661), but it's possible to get them in a roundabout way by casting `torch.inf` (which also works for int types).

There was a nuance where certain test cases expected zeros in the `self` tensor for entries that were not involved in the scatter-reduce index. We can preserve that behaviour by scattering the dtype min/max value into a zero tensor along the index so that only the relevant entries are initialized that way.

Another less efficient option would be to calculate the max/min value of the `src` tensor itself and use that as the fill value.

### DIFF
--- a/torch_geometric/utils/_scatter.py
+++ b/torch_geometric/utils/_scatter.py
@@ -97,9 +97,9 @@ if torch_geometric.typing.WITH_PT112:  # pragma: no cover
                                   f" package, but it was not found")
 
                 res = src.new_zeros(size)
+                # `scatter_reduce_` with `include_self=False` is not currently
+                # supported by onnx
                 if src.numel() > 0:
-                    # `scatter_reduce_` with `include_self=False` is not
-                    # currently supported by onnx
                     fill_value = src.max() if "min" in reduce else src.min()
                     [res.select(dim, i).fill_(fill_value) for i in index]
                 index = broadcast(index, src, dim)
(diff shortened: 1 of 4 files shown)


torch_geometric/utils/_scatter.py
```diff
@@ -96,10 +96,17 @@ def scatter(
                                   f"can be accelerated via the 'torch-scatter'"
                                   f" package, but it was not found")
 
+                res = src.new_zeros(size)
+                if src.numel() > 0:
+                    # `scatter_reduce_` with `include_self=False` is not
+                    # currently supported by onnx
+                    fill_value = (src.detach().max().item() if "min" in reduce
+                                  else src.detach().min().item())
+                    [res.select(dim, i).fill_(fill_value) for i in index]
```
lee-clement-oxb, line 105: Can't use `index_select` for this as it returns a copy not a view.

lee-clement-oxb:
ONNX export works with these changes and all the tests except for 4 torch.dynamo related tests are passing - the failures are due to graph breaks. I couldn't find a way to fix those ones, perhaps someone else can.

rusty1s (approved):
Nice :)

Issue #9457, opened 2024-06-25
`LineDigraph` transformation


Flunzmas:
### 🚀 The feature, motivation and pitch

Recent scientific work [1] leveraging GNNs for routing in computer networks has used a line digraph representation of the computer network as input for their GNN policies. Informally, the line digraph is formed by taking the original graph's edges as new nodes, and drawing an edge between those new nodes that, as edges in the original graph, form a directed path of length two. I.e., `E' = {(u, v), (w, x) | (u, v) in E; (w, x) in E; v = w}`.

However, AFAIK there is no existing PyG implementation that provides this conversion, as the `LineGraph` transformation is defined differently and produces a different output. Bernárdez et al., in [their code](https://github.com/BNN-UPC/MAGNNETO-TE/blob/2491465f8d5b279f6102c5b33614b2e09ba31b0c/environment/environment.py#L289), have done implemented the conversion manually on a `networkX.DiGraph` object, and while this works fine I suspect a PyG transformation implementation working on a `Data` object can provide the functionality more efficiently.

---

[1]: Bernárdez et al. "Magnneto: A graph neural network-based multi-agent system for traffic engineering." _IEEE Transactions on Cognitive Communications and Networking 9.2 (2023)_

### Alternatives

As mentioned above, the existing `LineGraph` transformation does something slightly different and is therefore not an alternative. Also, while it's possible to manually construct the graph as Bernárdez et al. did in [their code](https://github.com/BNN-UPC/MAGNNETO-TE/blob/2491465f8d5b279f6102c5b33614b2e09ba31b0c/environment/environment.py#L289), I assume we can do it quicker with tensor ops in PyG.

### Additional context

Besides, is it possible to apply the transformations of PyG to `Batch` objects?

rusty1s:
You can send a PR with your initial implementation. I can try to optimize it in a follow-up.

petern48:
Hi @Flunzmas , first-timer here. I optimized the for loop out of your code. I also ran some small test cases to ensure it produces the same result as your original code, and it was already 10x faster using some small graphs. Let me know if there's anything else I can do to help. I would be happy to integrate this into the codebase and send a PR, but wanted to leave the option to you, in case you prefer to do it yourself.

```
def to_line_digraph_optim(data: Data) -> Data:  # self
    assert data.edge_index is not None
    assert data.is_directed()
    edge_index, edge_attr = data.edge_index, data.edge_attr
[...]

PR #9592, opened 2024-08-14
Add LineDigraph transformation

petern48:
Addresses #9457 in collaboration with @Flunzmas

Changes:
- Implement an optimized transforms module for LineDiGraph based off of the [implementation](https://github.com/BNN-UPC/MAGNNETO-TE/blob/2491465f8d5b279f6102c5b33614b2e09ba31b0c/environment/environment.py#L289) by Bernárdez et al.
- Add a test for testing the implementation

Motivation from Flunzmas:

Recent scientific work [1] leveraging GNNs for routing in computer networks has used a line digraph representation of the computer network topology as input for their GNN policies. The line digraph is formed by taking the original graph's edges as new nodes, and drawing an edge between those new nodes that, as edges in the original graph, form a directed path of length two: E' = {(u, v), (w, x) | (u, v) in E; (w, x) in E; v = w}.

[1]: Bernárdez et al. "Magnneto: A graph neural network-based multi-agent system for traffic engineering." IEEE Transactions on Cognitive Communications and Networking 9.2 (2023)


test/transforms/test_line_digraph.py
```diff
@@ -0,0 +1,35 @@
...
+    edge_index = torch.tensor([
+        [0, 1, 2, 2, 3],
+        [1, 2, 0, 3, 0],
+    ])
+    data = Data(edge_index=edge_index, num_nodes=4)
+    data = transform(data)
+    assert data.edge_index.tolist() == [[0, 1, 1, 2, 3, 4], [1, 2, 3, 0, 4, 0]]
+    assert data.num_nodes == data.edge_index.max().item() + 1
+
+    # Undirected.
```
Kh4L, line 21: This is not undirected, and based on the implem, it wouldn't work for an undirected graph

torch_geometric/transforms/line_digraph.py
```diff
@@ -0,0 +1,50 @@
...
+        E = data.num_edges
+
+        edge_index, edge_attr = coalesce(edge_index, edge_attr, data.num_nodes)
+        row, col = edge_index
+
+        # Broadcast to create a mask for matching row and col elements
+        mask = row.unsqueeze(0) == col.unsqueeze(1)  # (num_edges, num_edges)
+        # Generate new_row/col indices. 2D tensor of arange
+        new_row = torch.arange(E).unsqueeze(1).expand(-1, mask.size(1))
+        new_col = torch.arange(mask.size(1)).unsqueeze(0).expand(E, -1)
```
Kh4L, line 41:
```suggestion
        new_row = torch.arange(E).unsqueeze(1).expand(-1, E)
        new_col = torch.arange(E).unsqueeze(0).expand(E, -1)
```

torch_geometric/transforms/line_digraph.py
```diff
@@ -0,0 +1,50 @@
...
+        new_row = new_row[mask]
+        new_col = new_col[mask]
+        new_edge_index = torch.stack([new_row, new_col], dim=0)
```
Kh4L, line 44: You can replace this whole logic by:

```suggestion
        new_edge_index = torch.nonzero(mask).T
```

Kh4L (changes requested):

test/transforms/test_line_digraph.py
petern48: Whoops, I started with the existing test_line_graph.py code as a template. I updated the test cases, but must've forgotten to clean up the comments. Thanks.

petern48:
@rusty1s Do I need to do anything to update the docs? I thought it would produce the page automatically, but I don't see it in the built docs.

Flunzmas:
Hi, I am reviving this MR by adding an option to put the node features of the original graph into the line digraph's edge features.

Let us know if there is anything missing to conclude this MR 🙂.

PR #9599, opened 2024-08-16
Fix Profiler String Parsing Bug

zaristei:
Fix a bug in the profiler where the stack trace isn't parsed correctly due to string parsing issues. It looks like this case was already checked for in the unittests but was previously failing.


zaristei:
Closing for now, it looks like the error that was occurring a few months ago may have been fixed in the pytest. Will reopen the PR once I can confirm the bug I was encountering is still reproducable.

PR #9621, opened 2024-08-22
Control factor to limit data download in `OPFDataset`

kaarthiksundar:
Adds additional parameter ``num_groups``, with default value of 20 to control the amount of data being downloaded. This factor is necessary because the larger test cases's size can be > 3 TB.

### DIFF
--- a/torch_geometric/datasets/opf.py
+++ b/torch_geometric/datasets/opf.py
@@ -41,10 +41,10 @@ class OPFDataset(InMemoryDataset):
             If :obj:`"test"`, loads the test dataset. (default: :obj:`"train"`)
         case_name (str, optional): The name of the original pglib-opf case.
             (default: :obj:`"pglib_opf_case14_ieee"`)
-        num_groups (int, optional): The dataset is divided into 20 groups with 
+        num_groups (int, optional): The dataset is divided into 20 groups with
             each group containing 15000 samples. For large systems, this amount
             of data can be overwhelming. The num_groups parameters controls the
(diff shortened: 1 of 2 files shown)


torch_geometric/datasets/opf.py
```diff
@@ -42,9 +42,10 @@ class OPFDataset(InMemoryDataset):
         case_name (str, optional): The name of the original pglib-opf case.
             (default: :obj:`"pglib_opf_case14_ieee"`)
         num_groups (int, optional): The dataset is divided into 20 groups with
-            each group containing 15000 samples. For large systems, this amount
-            of data can be overwhelming. The num_groups parameters controls the
-            amount of data being downloaded. Allowable values are [1, 20].
+            each group containing 15,000 samples.
+            On smaller systems, this amount of data can be overwhelming.
```
kaarthiksundar, line 46: It should be "On large systems" instead of smaller. Large system corresponds to large number of nodes in the graph.

rusty1s: Got it. I was thinking system=machine you are processing that dataset on.

rusty1s (approved):

kaarthiksundar:
@rusty1s I missed one commit in this PR. Without that commit, it introduces a bug. Is there a unit test for these datasets that I can run locally?

rusty1s:
Can you send another PR? We generally only have unit tests for a small fraction of datasets since downloading the datasets is too inefficient.

PR #9627, opened 2024-08-26
Added the `ClusterPooling` layer

thijssnelleman:
Our method that we will present at ECML MLG workshop. Based on [EdgePooling](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/pool/edge_pool.py)


wsad1 (commented):
Nice work. 
Could you add unit tests.

rusty1s (approved):

Issue #9631, opened 2024-08-29
Link pred metric: mean reciprocal rank (MRR)


rhjohnstone:
### 🚀 The feature, motivation and pitch

[MRR](https://en.wikipedia.org/wiki/Mean_reciprocal_rank) is another common metric for recommendations. There are a few uses of it throughout the codebase, but it would be good to have an implementation consistent with the other link prediction metrics.

### Additional context

I am working on a PR to implement this.

PR #9632, opened 2024-08-29
Mean reciprocal rank metric

rhjohnstone:
Addresses #9631. Implements `LinkPredMRR` as a `LinkPredMetric`, with test included.


rusty1s (approved):
Super :)

PR #9649, opened 2024-09-08
Autogen release notes

akihironitta:


rusty1s (commented):
How does it work?

akihironitta:
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

We'll still adjust auto-generated release notes, but it will be useful for categorising entries.

rusty1s:
Yes, part of the reason I am not necessarily eager to use this is since we are no longer able to group features of multiple PRs into a single bullet point. Am I misunderstanding?

akihironitta:
When writing a release note for a release, I read both
1. `CHANGELOG.md` and
2. titles of all commits from its previous release.

Ideally, 1. `CHANGELOG.md` should cover all changes that we want to highlight in the release so that I don't have to read through all commits between the new release and its previous release, but in the past few releases, I had to look back at commit titles to make sure we highlight all items and link pointers to them in the release notes.

This PR was aiming to reduce my manual work reading through 2. titles of all commits from the previous release by letting GitHub show me no commit titles tagged `skip-changelog` and group the rest commit titles by their labels.

akihironitta:
Just for note, adding this donesn't mean we publish auto-generated release note as is. GitHub will just fill an initial release note draft with commit entries according to the rule we define in the file.

PR #9651, opened 2024-09-10
Update metapath2vec.py

yangfa-zhang:
fix bug:
 col = col[rand] if col.numel() > 0 else rand
IndexError: index 2282976 is out of bounds for dimension 0 with size 2282976

### DIFF
--- a/torch_geometric/nn/models/metapath2vec.py
+++ b/torch_geometric/nn/models/metapath2vec.py
@@ -136,19 +136,22 @@ class MetaPath2Vec(torch.nn.Module):
         """
         emb = self.embedding.weight[self.start[node_type]:self.end[node_type]]
         if batch is not None:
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/models/metapath2vec.py
```diff
@@ -257,6 +272,10 @@ def sample(rowptr: Tensor, col: Tensor, rowcount: Tensor, subset: Tensor,
     rand = rand.to(torch.long) + rowptr[subset].view(-1, 1)
     rand = rand.clamp(max=col.numel() - 1)  # If last node is isolated.
 
-    col = col[rand] if col.numel() > 0 else rand
+    if col.numel() > 0:
+        rand = torch.min(rand, torch.tensor(col.numel() - 1))
```
rusty1s, line 276: We already clamp the value in line 273. What does this do here?

PR #9668, opened 2024-09-17
Make `ConfigMixin.from_config` no-op if dataclass resolution is unnecessary

akihironitta:


torch_geometric/config_mixin.py
```diff
@@ -82,7 +82,10 @@ def _recursive_from_config(value: Any) -> Any:
     cls: Any = None
     if is_dataclass(value):
         if getattr(value, '_target_', None):
-            cls = _locate_cls(value._target_)  # type: ignore[attr-defined]
+            try:
+                cls = _locate_cls(value._target_)
```
akihironitta, line 86:
```suggestion
                cls = _locate_cls(value._target_)  # type: ignore[attr-defined]
```

rusty1s (approved):

PR #9672, opened 2024-09-17
rebasing interactive demo for g-retriever

puririshi98:
rebase https://github.com/pyg-team/pytorch_geometric/pull/9417


puririshi98:
realized the new g-retriever example doesnt even train an LLM, so comparing the two doesnt make sense. I will add comments to the example that explain how the data is defined

PR #9673, opened 2024-09-17
add `root_weight` to basicGNN for "skip connections"

puririshi98:


rusty1s:
Thanks, looks good. There is one last problem with `to_hetero` (which seems to be a bug in `ToHeteroTransformer`).

puririshi98:
> Thanks, looks good. There is one last problem with `to_hetero` (which seems to be a bug in `ToHeteroTransformer`).

i can take a look

PR #9731, opened 2024-10-24
Integrated SSMA aggregation

AlmogDavid:
Integrated the "Sequential Signal Mixing Aggregation" (SSMA) aggregation method from: https://arxiv.org/abs/2409.19414


Kh4L:
Looks great! Did you try it out with some OGB benchmarks?

I think it'd be nice to add an option to try it out in [ogbn_train.py ](https://github.com/pyg-team/pytorch_geometric/blob/f5c829344517c823c24abb08ce2fc7cf00ff29f7/examples/ogbn_train.py#L22) - in a followup PR, and see if we get the  accuracy boost claimed in Table 2 ( like the 15-20% boost in ogbn-products :rocket: )!

AlmogDavid:
> Looks great! Did you try it out with some OGB benchmarks from PyG examples?
> 
> I think it'd be nice to add an option to try it out in [ogbn_train.py ](https://github.com/pyg-team/pytorch_geometric/blob/f5c829344517c823c24abb08ce2fc7cf00ff29f7/examples/ogbn_train.py#L22) - in a followup PR, and see if we get the accuracy boost claimed in Table 2 ( like the 15-20% boost in ogbn-products 🚀 )!

Thanks!
You can find the experimental setting and to reproduce the results using our repository: https://almogdavid.github.io/SSMA/
There you can find full set of datasets and configurations to try out yourself, not sure if its appropriate to link to the repository from the documentation itself, but if you think it will be helpful I will do it.

AlmogDavid:
@wsad1 , @EdisonLeeeee . Is there some ETA for a review?

Thanks.

AlmogDavid:
@wsad1 , @EdisonLeeeee - Any updates?

AlmogDavid:
@rusty1s  - How can we progress this MR?

torch_geometric/nn/aggr/ssma.py
```diff
@@ -0,0 +1,195 @@
...
+                          self._max_neighbors)  # [N, #groups, #neighbors]
+
+        # Compute softmax over the neighbors based on the attention weights
+        # and the graph topology
+        edge_attention_ste = softmax(
+            F.leaky_relu(x_l[edge_index[0]] + x_r[edge_index[1]]) /
+            self._neighbor_att_temp, edge_index[1])  # [E, #groups, #neighbors]
+        return edge_attention_ste
+
+    def pre_aggregation_hook(self, module, inputs):
```
wsad1, line 114: Could you explain why the prehook is needed?
Maybe add to the documentation why this pre hook is needed, and why these opreations can't be implemented as part of the forward pass.

torch_geometric/nn/aggr/ssma.py
```diff
@@ -0,0 +1,195 @@
...
+from torch import Tensor
+
+from torch_geometric.nn.aggr import Aggregation
+from torch_geometric.utils import scatter, softmax
+
+
+class SSMAAggregation(Aggregation):
+    r"""Performs the Sequential Signal Mixing Aggregation (SSMA) method from
+    `"Sequential Signal Mixing Aggregation for Message Passing Graph
+    Neural Networks" <https://arxiv.org/abs/2409.19414>`_ paper.
```
wsad1, line 16: Add the forumulae for the aggregation , like we have for other aggregators https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.aggr.PowerMeanAggregation.html.

AlmogDavid: Done

torch_geometric/nn/aggr/ssma.py
AlmogDavid: Thanks for your comment.

The pre-hook is needed only when using the attention mechanism in SSMA. This function creates the edge weights for attention, which requires edge information (source + target). Since this information is unavailable in the forward pass, the pre-hook is necessary.

Based on your feedback:
	•	Added documentation in the code.
	•	Ensured the pre-hook is called **only** when use_attention == True.
	•	Updated tests to verify this case.

Let me know if further clarification is needed.

AlmogDavid:
@wsad1 thanks for the review, I addressed all of you comments.
Please let me know if more things are needed. Thanks.

torch_geometric/nn/aggr/ssma.py
```diff
@@ -0,0 +1,207 @@
...
+                          self._max_neighbors)  # [N, #groups, #neighbors]
+
+        # Compute softmax over the neighbors based on the attention weights
+        # and the graph topology
+        edge_attention_ste = softmax(
+            F.leaky_relu(x_l[edge_index[0]] + x_r[edge_index[1]]) /
+            self._neighbor_att_temp, edge_index[1])  # [E, #groups, #neighbors]
+        return edge_attention_ste
+
+    def pre_aggregation_hook(self, module, inputs):
```
wsad1, line 120: The hook will only get the arguments `forward` (gets)[https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html]. So the hook won't have access to `edge_index`.

wsad1: So Instead of making this a prehook, make this a static function that the user could call before calling forward.

wsad1 (commented):
Thanks for the contribution. 
I think its best to move this to [torch_geometric/contrib](https://github.com/pyg-team/pytorch_geometric/tree/master/torch_geometric/contrib) folder.

torch_geometric/nn/aggr/ssma.py
AlmogDavid: Thanks for the review. This is not a forward hook. it's a pre aggregation hook (propagate_forward_pre_hook: https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.MessagePassing.html?highlight=propagate_forward_pre_hook#torch_geometric.nn.conv.MessagePassing.register_propagate_forward_pre_hook) It does have access to edge index.

It also modifies private fields in the aggregation so I don't see how it can become a static method.

AlmogDavid:
Moved the aggregation to the contrib folder as suggested.

AlmogDavid:
@rusty1s What can we do to move this MR forward? I’ve addressed all @wsad1’s comments and implemented the requested changes. I also reached out to @EdisonLeeeee, but he doesn’t have time for a code review right now.

AlmogDavid:
Someone, any updates? @wsad1 @rusty1s

test/contrib/nn/aggr/test_ssma.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_representation():
+    ssma_agg = SSMAAggregation(in_dim=10, num_neighbors=3, mlp_compression=0.1,
+                               use_attention=False, n_heads=59, temp=0.2,
+                               att_feature="dummy", learn_affine=True)
+
+    rep = str(ssma_agg)
+    gen_rep = eval(rep)
+
+    assert str(gen_rep) == rep
```
wsad1, line 97:
```suggestion
```

wsad1 (approved):

AlmogDavid:
@rusty1s - Can you please review?

AlmogDavid:
@wsad1 @rusty1s  ?

AlmogDavid:
@akihironitta Can we merge?

AlmogDavid:
@akihironitta  @wsad1 - The MR is open for 9 months...
What can we do in order to push it?

AlmogDavid:
@akihironitta ....

PR #9737, opened 2024-10-26
[ENH] Allow users to pass `edge_weight` to `GraphUNet` model

### LINKED ISSUE #9736: Pass edge weights/attributes to GraphUNet

I noticed that the GraphUNet model doesn't make use of the edge attributes that I have calculated. I find it very odd that the model uses ones as the starting weights, as opposed to having the option to pass predefined weights to better inform the model on an initial pass.

### Alternatives

_No response_

### Additional context

_No response_

eurunuela:
Closes #9736.

This PR:

- Adds the `edge_weight` parameter to the `nn.models.GraphUNet` model's forward function, so that users can provide an initial weights vector.
- Makes `edge_weight` default to `None`.
- Initializes `edge_weight` to ones if it is `None`. This ensures functionality stays the same by default.
- If `edge_weight` is given by the user, its size is checked before running any of the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Allowed users to pass `edge_weight` to `GraphUNet` models ([#9737](https://github.com/pyg-team/pytorch_geometric/pull/9737))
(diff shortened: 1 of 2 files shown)


rusty1s (approved):

eurunuela:
Hi @rusty1s,

Thanks for looking into this. However, I don't think the PR was ready to be merged (my bad for making it ready on GitHub). There is an issue where the multiplication of the adjacency matrix with itself doesn't work because it is a sparse matrix. I will open a new issue with this.

PR #9748, opened 2024-10-29
Add support for `torch_delaunay` package in `Delaunay` transformation

ybubnov:
This patch adds support of torch_delaunay package, which works with Torch tensors. The new implementation uses `torch_delaunay` package if it is installed (by default), and then falls back to the `scipy` implementation otherwise.


rusty1s (commented):
Very cool :)

rusty1s (approved):

PR #9756, opened 2024-11-01
Fixing `edge_mask` handling for directed graphs in `k_hop_subgraph`

ryoji-kubo:
Example use case:

```python
from torch_geometric.utils import k_hop_subgraph
import torch

edge_index = torch.tensor([[1, 2, 3],
                           [0, 1, 1]])
# get the 2-hop neighbors of node 0 in the directed graph.
_, edge_index, _, edge_mask = k_hop_subgraph(0, 2, edge_index, relabel_nodes=False, directed=True)
```

This gives the following result:

Expected Outcome:
```python
>>> edge_index
tensor([[1, 2, 3],
        [0, 1, 1]])

>>> edge_mask
tensor([True,  True,  True])
```

Actual Outcome:
```python
>>> edge_index
tensor([[2, 3],
        [1, 1]])
>>> edge_mask
tensor([False,  True,  True])
```

This stems from the fact that the line `torch.index_select(node_mask, 0, row, out=edge_mask)` overwrites `edge_mask`, effectively only marking the edges used in the final hop as `True`.

To fix this, I have added `preserved_edge_mask ` that will mark the edges used in each hop as `True`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Fixed `k_hop_subgraph()` for directed graphs
 - Added the `WebQSPDataset` dataset ([#9481](https://github.com/pyg-team/pytorch_geometric/pull/9481))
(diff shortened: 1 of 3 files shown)


rusty1s (approved):
Thanks :) Also added a test.

PR #9773, opened 2024-11-11
Fix arrow direction in `_visualize_graph_via_networkx`

darabos:
Matplotlib arrows don't have a source and a destination. They have a text and a dot. As the [example](https://matplotlib.org/stable/gallery/text_labels_and_annotations/fancyarrow_demo.html) shows, a `<-` arrow points from the dot to the text and `->` points from the text to the dot.
Screenshot 2024-11-11 at 15 47 24
`_visualize_graph_via_networkx()` sets `xy=pos[src]` and `xytext=pos[dst]`, so we want arrows from the dot (`xy`) to the text (`xytext`). That's the `<-` arrow.

This can also be confirmed by comparing the visualization to the GraphViz version. The arrows go the other way. Or just looking at the data and the picture, haha! It took me a while to figure out that it's not my graph that's messed up! 😅

### DIFF
--- a/torch_geometric/visualization/graph.py
+++ b/torch_geometric/visualization/graph.py
@@ -132,7 +132,7 @@ def _visualize_graph_via_networkx(
             xy=pos[src],
             xytext=pos[dst],
             arrowprops=dict(
-                arrowstyle="->",
+                arrowstyle="<-",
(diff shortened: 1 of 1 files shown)


rusty1s (approved):
Nice catch :)

PR #9776, opened 2024-11-12
Added functionality of `FaceToEdge` transform to work for 3D tetrahedral mesh elements

Aiik:
Now transforms 2D triangular elements/faces with shape [3,n] to  as well as 3D tetrahedral elements with shape [4,n] to edges of shape [2,n]. 

Including pytest case for face input with shape [4,n]


rusty1s (approved):
Cool :)

PR #9798, opened 2024-11-19
Add `group_batch` functionality

tzuhanchang:
Tensor `src` is batched into groups according to `index` tensor in the given dimension `dim`.
Padding is applied to enforce consistent size of batched groups.

Examples:
```
>>> src = tensor([[0.1349, 0.8266],
...               [0.3651, 0.1737],
...               [0.0211, 0.7000],
...               [0.1971, 0.2903],
...               [0.4086, 0.7221]])
>>> index = torch.LongTensor([0,0,1,2,2])
>>> group_batch(src, index, dim=0, pad_size=3, return_mask=True)
(tensor([[[0.1349, 0.8266],
          [0.3651, 0.1737],
          [  -inf,   -inf]],
         [[0.0211, 0.7000],
          [  -inf,   -inf],
          [  -inf,   -inf]],
         [[0.1971, 0.2903],
          [0.4086, 0.7221],
          [  -inf,   -inf]]])
tensor([[[ True,  True],
         [ True,  True],
         [False, False]],
        [[ True,  True],
         [False, False],
         [False, False]],
        [[ True,  True],
         [ True,  True],
         [False, False]]]))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added `group_batch` functionality ([#9798](https://github.com/pyg-team/pytorch_geometric/pull/9798))
 - Added `nn.models.GLEM` ([#9662](https://github.com/pyg-team/pytorch_geometric/pull/9662))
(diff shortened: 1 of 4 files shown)


tzuhanchang:
@rusty1s

PR #9806, opened 2024-11-25
Large Graph Indexer WebQSP Refactor

zaristei:
Successor to [9666](https://github.com/pyg-team/pytorch_geometric/pull/9666), this:
- ~~updates the documentation to show how to utilize GNN RAG and~~(now handled by separate branch)
- updates WebQSP to help serve as a toy example for LargeGraphIndexer.
- fixes issues with LargeGraphIndexer running out of memory by introducing a default batch size and multithreading ability

~~currently blocked by a bug that causes the g_retriever.py example to get 1% less accuracy.~~ Bug is due to a fp32 precision issue related to batch kernels in Huggingface's transformers. Performance difference is too inconsequential to require a fix.

may also be the cause of low  retrieval precision in https://github.com/pyg-team/pytorch_geometric/pull/9846


puririshi98:
will merge this once zack joins full time and has bandwidth and compute to debug and ensure that merging his PR keeps test hit metric the same or higher (not lower as is now).

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -140,121 +164,223 @@ class KGQABaseDataset(InMemoryDataset):
...
+        import datasets
 
-        for dataset, path in zip(
-            [datasets['train'], datasets['validation'], datasets['test']],
+        # HF Load Dataset by dataset name if no path is specified
+        self.load_dataset_kwargs['path'] = self.load_dataset_kwargs.get(
+            'path', self.dataset_name)
+        raw_dataset = datasets.load_dataset(**self.load_dataset_kwargs)
+
+        # TODO: @riship should this be done in the base class?
```
puririshi98, line 249: yes

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -140,121 +164,223 @@ class KGQABaseDataset(InMemoryDataset):
...
+        self.required_splits = ['train', 'validation', 'test']
+
+        self.verbose = verbose
+        self.force_reload = force_reload
         super().__init__(root, force_reload=force_reload)
 
-        if split not in {'train', 'val', 'test'}:
+        # NOTE: Current behavior is to process the entire dataset, and only return the split specified by the user
```
puririshi98, line 207: if you call this dataset 3 times
```
WebQSP(split="train")
..
Webqsp(..
```
it just processes it once and then returns train on the first call, but the val and test ones just pull from the already processed data

puririshi98 (changes requested):

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -118,15 +129,28 @@ def retrieval_via_pcst(
...
 
+    # HACK Added so that the subset of nodes and edges selected can be tracked
+    data['node_idx'] = node_idx
+    data['edge_idx'] = edge_idx
+
     return data, desc
 
 
+def preprocess_triplet(triplet: TripletLike) -> TripletLike:
+    h, r, t = triplet
```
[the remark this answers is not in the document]
zaristei: Referring to this one creates a circular import, so I'm going to keep my copy in `web_qsp_dataset.py` for now.

PR #9810, opened 2024-11-27
ProteinMPNN_dataset

jdhenaos:
A new Dataset for ProteinMPNN model was added.

The Dataset was evaluated using the next piece of code:

```
from torch_geometric.loader import DataLoader

path = "./sample_data"
params = {
    "LIST"    : f"{path}/pdb_2021aug02_sample/list.csv",
    "VAL"     : f"{path}/pdb_2021aug02_sample/valid_clusters.txt",
    "DIR"     : f"{path}/pdb_2021aug02_sample",
    "DATCUT"  : "2030-Jan-01",
    "RESCUT"  : 3.5, #resolution cutoff for PDBs
    "HOMO"    : 0.70 #min seq.id. to detect homo chains
}

train_dataset = PMPNNDataset(root=path,params=params,set_type='train')
validation_dataset = PMPNNDataset(root=path,params=params,set_type='val')

train_loader = DataLoader(train_dataset, batch_size=32, shuffle=True,
                          num_workers=2)
validation_loader = DataLoader(validation_dataset, batch_size=32, shuffle=True,
                          num_workers=2)
```

### DIFF
--- a/torch_geometric/datasets/proteinmpnn_dataset.py
+++ b/torch_geometric/datasets/proteinmpnn_dataset.py
@@ -1,0 +1,223 @@
+from torch_geometric.data import (
+    Data,
+    InMemoryDataset,
+    download_url,
+    extract_tar,
+)
+from torch_geometric.io import fs
+import torch
+import os.path as osp
+import os
+import random
+from torch_geometric.utils import one_hot
+import pandas as pd
+import numpy as np
+import csv
+from dateutil import parser
+
+class PMPNNDataset(InMemoryDataset):
+
+  r"""Dataset used in the training example from the `"Robust deep learning based 
+  protein sequence design using ProteinMPNN"
+  <https://www.biorxiv.org/content/10.1101/2022.06.03.494563v1.full.pdf>'_paper.
+
+  Args:
+      root (str): Root directory where the dataset should be saved.
+      params (dict): Dictionary of parameters for dataset creation:
+                     LIST: Path to the table with metadata.
+                     VAL: Path to list of cluster IDs for model validation.
+                     DIR: Path to dataset.
+                     DATCUT: Date (YYY-MM-DD) threshold of sequence deposition.
+                     RESCUT: Resolution cutoff for PDBs.
+                     HOMO: Minimal sequence identity to detect homodimeric chains.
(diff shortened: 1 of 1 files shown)


torch_geometric/datasets/proteinmpnn_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                for f in files] or osp.join(self.raw_dir, files)
+
+    def download(self):
+        path = download_url(self.url, self.root)
+        extract_tar(path, self.root)
+        os.unlink(path)
+        osp.join(self.root)
+        fs.rm(self.raw_dir)
+
+    def build_training_clusters(self, params, debug):
```
xnuohz, line 83: debug mode should be removed.

torch_geometric/datasets/proteinmpnn_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                if k[0] in homo:
+                    masked.append(counter)
+
+            item_graph = Data(
+                seq=seq,
+                xyz=torch.cat(xyz, dim=0),
+                idx=torch.cat(idx, dim=0),
+                masked=torch.Tensor(masked).int(),
+                #label = self.item[0]
+            )
```
xnuohz, line 228: I didn't see `x` and `edge_index` in the dataset, ProteinMPNN's input only includes sequence data, right?

xnuohz (commented):
Left some comments. Plz add a dataset test and make CI pass:)

puririshi98:
I agree with @xnuohz reviews, i will do my own review once they have been addressed

torch_geometric/datasets/proteinmpnn_dataset.py
xnuohz: @jdhenaos seems like you didn't update according to my comments, also doesn't include model part in pyg version. so im not clear if the data class can fit the model. as such, i think currently this can't be mergable.
cc @puririshi98 wdyt?

jdhenaos: This needs a bit more work, I will update the code as soon as I can, and I will elaborate my point regarding the edge_index

PR #9817, opened 2024-11-30
On single-CPU systems, skip tests which are known not to work

sanvila:
Hello. While building this package on a single-CPU system I noticed that two tests fail.

`test_cpu_affinity_neighbor_loader()` fails in this way:

`E    ValueError: More workers (got 2) than available cores (got 1)`

and `test_multithreading_neighbor_loader()` fails in this way:

`E    ValueError: 'worker_threads' should be smaller than the total available number of threads 1 (got 2)`

I guess those tests are simply not suitable to be executed on single-CPU systems, so the proposed MR uses `os.cpu_count()` to test before executing them.

The failures may be reproduced easily by booting with `GRUB_CMDLINE_LINUX="nr_cpus=1"`.

Thanks.

### DIFF
--- a/test/loader/test_mixin.py
+++ b/test/loader/test_mixin.py
@@ -1,6 +1,7 @@
 import subprocess
 from time import sleep
 
+import psutil
 import pytest
 import torch
 
@@ -12,6 +13,8 @@ from torch_geometric.testing import onlyLinux, onlyNeighborSampler
 @pytest.mark.xfail(reason="TODO: Fix test")
 @onlyLinux
 @onlyNeighborSampler
+@pytest.mark.skipif(
+    psutil.cpu_count(logical=False) == 1, reason="Requires multiple CPU cores")
 @pytest.mark.parametrize('loader_cores', [None, [1, 2]])
 def test_cpu_affinity_neighbor_loader(loader_cores, spawn_context):
     data = Data(x=torch.randn(1, 1))
@@ -41,6 +44,8 @@ def init_fn(worker_id):
 
 @onlyLinux
 @onlyNeighborSampler
+@pytest.mark.skipif(
+    psutil.cpu_count(logical=False) == 1, reason="Requires multiple CPU cores")
 def test_multithreading_neighbor_loader(spawn_context):
     loader = NeighborLoader(
         data=Data(x=torch.randn(1, 1)),


sanvila:
Hello. On AWS instances of type `c7i.large` or `m7i.large`, having one core and 2 threads (2 vCPUs in total), this patch I proposed some time ago is not enough (still fails with the same error).

I've modified the patch to use `psutil.cpu_count` instead.

test/loader/test_mixin.py
```diff
@@ -12,6 +13,9 @@
 @pytest.mark.xfail(reason="TODO: Fix test")
 @onlyLinux
 @onlyNeighborSampler
+@pytest.mark.skipif(
+    psutil.cpu_count(logical=False) == 1,
+    reason="Not suitable for single-CPU systems")
```
akihironitta:
```suggestion
@pytest.mark.skipif(
    psutil.cpu_count(logical=False) == 1,
    reason="Requires multiple CPU cores")
```

test/loader/test_mixin.py
```diff
@@ -41,6 +45,9 @@ def init_fn(worker_id):
 
 @onlyLinux
 @onlyNeighborSampler
+@pytest.mark.skipif(
+    psutil.cpu_count(logical=False) == 1,
+    reason="Not suitable for single-CPU systems")
```
akihironitta:
```suggestion
    reason="Requires multiple CPU cores")
```

CHANGELOG.md
```diff
@@ -167,6 +167,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed import errors on `MessagePassing` modules with nested inheritance ([#8973](https://github.com/pyg-team/pytorch_geometric/pull/8973))
 - Fixed bug in multi XPU training ([#9456](https://github.com/pyg-team/pytorch_geometric/pull/9456))
 - Fixed TorchScript compilation error for `MessagePassing._check_input` on older torch versions ([#9564](https://github.com/pyg-team/pytorch_geometric/pull/9564))
+- Fixed tests on single-CPU systems ([#9871](https://github.com/pyg-team/pytorch_geometric/pull/9817))
```
akihironitta: No need to add an entry for test changes
```suggestion
```

akihironitta (approved):

sanvila:
Thank you. I'm also changing slightly the wording of the commit message to be in line with the wording of the code.

akihironitta:
> Thank you. I'm also changing slightly the wording of the commit message to be in line with the wording of the code.

All commits are squashed into one when merging the PR, so it doesn't matter much, but thank you!

PR #9820, opened 2024-12-03
Implement HIN2Vec Model

jesst0x:
As part of Stanford CS224W course, I implemented HIN2Vec model as per paper [**HIN2Vec: Explore Meta-paths in
    Heterogeneous Information Networks for Representation Learning**](https://dl.acm.org/doi/pdf/10.1145/3132847.3132953) to learn node embedding for heterogeneous graph. 

The model is in contrib package, and I also included unit test and example to use the model.

### DIFF
--- a/torch_geometric/contrib/nn/models/hin2vec.py
+++ b/torch_geometric/contrib/nn/models/hin2vec.py
@@ -24,6 +24,13 @@ class HIN2Vec(torch.nn.Module):
     embeddings in a heterogeneous graph are learned via a target sets of
     relationships in the form of meta-paths.
 
+    .. note::
+
+        For examples of using HIN2Vec, see
+        `examples/contrib/hin2vec.py
+        <https://github.com/pyg-team/pytorch_geometric/blob/master/examples/
+        contrib/hin2vec.py>`_.
+
     Args:
         edge_index_dict (Dict[Tuple[str, str, str], torch.Tensor]): Dictionary
             holding edge indices for each
@@ -246,7 +253,8 @@ class HIN2Vec(torch.nn.Module):
             edge_index_with_offset[0] += self.node_start[keys[0]]
             edge_index_with_offset[1] += self.node_start[keys[-1]]
             edge_id = torch.full((1, edge_index.size(1)),
-                                 self.edge_type_id_dict[keys])
+                                 self.edge_type_id_dict[keys],
+                                 device=edge_index.device)
             edge_index_with_offset = torch.cat(
                 (edge_index_with_offset, edge_id), dim=0)
             all_edge_index.append(edge_index_with_offset)
--- a/torch_geometric/contrib/nn/models/hin2vec.py
+++ b/torch_geometric/contrib/nn/models/hin2vec.py
@@ -352,13 +352,8 @@ class HIN2Vec(torch.nn.Module):
 class _BinaryStepFunction(Function):
     r"""Implement differentiable binary step function."""
     @staticmethod
-    def forward(input: Tensor) -> Tensor:
+    def forward(ctx: Any, input: Tensor) -> Tensor:
         return (input > 0).float()
-
-    @staticmethod
-    def setup_context(ctx: Any, inputs: Any, output: Any):
-        input, = inputs
-        ctx.save_for_backward(input)
 
     @staticmethod
     def backward(ctx: Any, grad_output: Tensor) -> Tensor:
--- a/torch_geometric/contrib/nn/models/hin2vec.py
+++ b/torch_geometric/contrib/nn/models/hin2vec.py
@@ -7,12 +7,7 @@ from torch.nn import Embedding, Sigmoid
 from torch.utils.data import DataLoader
 
 from torch_geometric.index import index2ptr
-from torch_geometric.typing import (
-    WITH_PYG_LIB,
-    WITH_TORCH_CLUSTER,
-    EdgeType,
-    NodeType,
-)
+from torch_geometric.typing import WITH_TORCH_CLUSTER, EdgeType, NodeType
 from torch_geometric.utils import sort_edge_index
 from torch_geometric.utils.num_nodes import maybe_num_nodes_dict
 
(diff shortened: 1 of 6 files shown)


jesst0x:
Hi, all my tests passed except one of previous PyTorch version (prev_pytest 1.13.0). The forward method is static, so it works as intended since all other version passed.

Can you please let me know if need to fix this test for prev version? Thanks

jesst0x:
Hi @rusty1s @wsad1, this is my first time submitting PR for **pytorch_geometric**. May I know what could be the expected turnaround time of this PR review? I am to trying see if anything that I can fix earlier to be in time for the course project.

Ref my previous comment, all checks passed except one of prev pytorch version test. I believe my current implementation is correct. That might be different behavior in prev version. 

Thanks

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+        # -1 indicates no edge in random walk
+        mask = edge_seq < 0
+        edge_seq = edge_seq.clamp(min=0, max=len(self.edge_id) - 1)
+
+        edge_id_seq = self.edge_id[edge_seq]
+        edge_id_seq[mask] = -1
+        return node_seq, edge_id_seq
+
+    def _extract_distinct_metapath(self, rw: Tensor) -> Dict[int, Tensor]:
+        r"""Extract distinct metapath within self.metapath_length hops from
```
zechengz, line 303:
```suggestion
        r"""Extract distinct metapath within `self.metapath_length` hops from
```

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+            neg_sample[:, 1] = neg_node_idx.long()
+        else:
+            neg_sample[:, 1] = torch.randint(0, self.node_count - 1,
+                                             (neg_sample.size(0), ))
+        if not self.allow_circle:
+            neg_sample = self._remove_circle(neg_sample)
+
+        return pos_sample, neg_sample
+
+    def _remove_circle(self, sample: Tensor) -> Tensor:
```
zechengz, line 227: cycle instead of circle?
```suggestion
    def _remove_cycle(self, sample: Tensor) -> Tensor:
```

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+    @staticmethod
+    def backward(ctx: Any, grad_output: Tensor) -> Tensor:
+        input, = ctx.saved_tensors
+        dx = input.clone()
+        mask = (-6 <= dx) & (dx <= 6)
+        dx[mask] = torch.sigmoid(dx[mask]) * (1 - torch.sigmoid(dx[mask]))
+        return grad_output * dx
+
+
+class _BinaryStepModule(torch.nn.Module):
```
zechengz, line 370: The unit test error seems related to this. Maybe we can use something like
```python
from packaging import version

TORCH_VERSION = version.parse(torch.__version__)
if TORCH_VERSION < version.parse("...")...
   class ...
else:
   class ...
```
to make sure it's also working for the older version?

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+        reg: str = 'step',
+        num_nodes_dict: Optional[Dict[NodeType, int]] = None,
+        allow_circle: bool = False,
+        same_neg_node_type: bool = False,
+        sparse: bool = False,
+    ):
+        super().__init__()
+
+        if metapath_length < 1:
+            raise ValueError("The metapath_length must be positive integer.")
```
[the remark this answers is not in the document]
jesst0x: Done.

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+
+        if walk_length <= metapath_length:
+            raise ValueError(
+                "The 'walk_length' is not longer than the 'metapath_length.'")
```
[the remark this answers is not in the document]
jesst0x: Done

torch_geometric/contrib/nn/models/hin2vec.py
```diff
@@ -0,0 +1,375 @@
...
+    def forward(input: Tensor) -> Tensor:
+        return (input > 0).float()
+
+    @staticmethod
+    def setup_context(ctx: Any, inputs: Any, output: Any):
+        input, = inputs
+        ctx.save_for_backward(input)
+
+    @staticmethod
+    def backward(ctx: Any, grad_output: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
jesst0x: Added dostring to explain implementation of differentiable binary step function.

torch_geometric/contrib/nn/models/hin2vec.py
jesst0x: Done

torch_geometric/contrib/nn/models/hin2vec.py
jesst0x: Updated to cycle

torch_geometric/contrib/nn/models/hin2vec.py
jesst0x: I figured out that the issue is  actually with `Torch.Autograd.Function` (_BinaryStepFunction) implementation. There are 2 ways to define `forward()` and `setup_context()`, ie. combined or separate. Pytorch recommend [second way ](https://pytorch.org/docs/stable/notes/extending.html#how-to-use), which I implemented. 

Apparently, this is not compatible to pytorch version earlier than 2.0 after some debugging and [checking](https://pytorch.org/docs/stable/notes/extending.html#how-to-use). I updated it to the combined method instead, to be compatible to both earlier and current version. 

Now, all the tests have passed.

jesst0x:
@zechengz Thank you for the review, really appreciate it. I have resolved all the comments, and all the tests have passed now. 

Please let me know if there is other things to change, or it is ready to merge. Thanks.

PR #9821, opened 2024-12-05
Rough draft meant to be made in our fork, feel free to delete

mattjhayes3:
rough draft pr, still under development

### DIFF
--- a/torch_geometric/nn/models/label_usage.py
+++ b/torch_geometric/nn/models/label_usage.py
@@ -15,11 +15,10 @@ class LabelUsage(torch.nn.Module):
     Args:
         split_ratio (float): Proportion of true labels to use as features 
             during training.
-        num_recycling_iterations (int):  Number of iterations the previously
+        num_recycling_iterations (int): Number of iterations the previously
             predicted softlabels are used as features.
-        return_tuple (bool):  Whether we would like to return the predicted 
-            labels from the two sets separately, with index 
-            corresponding to the split.
+        return_tuple (bool): Whether to return the label indices from the split
+            sets separately along with the output
         base_model: An instance of the model that will do the 
             inner forward pass.
         num_classes (int): Number of classes in dataset
@@ -59,7 +58,7 @@ class LabelUsage(torch.nn.Module):
         return torch.cat([features, onehot], dim=-1)
 
     def forward(self, x, edge_index, y, train_idx):
-        """
+        r"""
         Forward pass using label usage algorithm.
 
         Args:
@@ -87,5 +86,5 @@ class LabelUsage(torch.nn.Module):
 
         # return tuples if specified
         if self.return_tuple:
-            return output[train_labels_idx], output[train_pred_idx], train_pred_idx
+            return output, train_labels_idx, train_pred_idx
         return output
--- a/torch_geometric/nn/models/label_usage.py
+++ b/torch_geometric/nn/models/label_usage.py
@@ -15,10 +15,10 @@ class LabelUsage(torch.nn.Module):
         accordingly to include both features and classes.
 
     Args:
-        base_model: An instance of the model that will do the 
+        base_model: An instance of the model that will do the
             inner forward pass.
         num_classes (int): Number of classes in dataset
-        split_ratio (float): Proportion of true labels to use as features 
+        split_ratio (float): Proportion of true labels to use as features
             during training (default: :obj:'0.5')
         num_recycling_iterations (int): Number of iterations for the
             label reuse procedure to cycle predicted soft labels 
@@ -27,7 +27,6 @@ class LabelUsage(torch.nn.Module):
             train_pred_idx) otherwise returns prediction output 
             (default :obj:'False')
     """
-
     def __init__(
         self,
         base_model: torch.nn.Module,
@@ -37,7 +36,7 @@ class LabelUsage(torch.nn.Module):
         return_tuple: bool = False,
     ):
 
-        super(LabelUsage, self).__init__()
+        super().__init__()
         self.base_model = base_model
         self.num_classes = num_classes
         self.split_ratio = split_ratio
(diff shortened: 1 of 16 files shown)


torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+            edge_index (torch.Tensor or SparseTensor): The edge indices
+            y (torch.Tensor): Node label tensor 
+            train_idx (torch.Tensor): Global indices of all 
+                nodes in training set
+            batch (torch.Tensor, optional): Optional mini-batch specification
+                indicating global node indices in batch (default: :obj:'None')
+        """
+        # re-assign train_idx to be nodes in mini-batch
+        if batch is not None:
+            train_idx = batch
```
mattjhayes3, line 67: batch is never used again, so I'm not sure it makes sense to have both batch and train_idx parameters.  
We can just update the documentation to something like "Indices of the training set (or minibatch)"

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+            train_pred_idx) otherwise returns prediction output 
+            (default :obj:'False')
+    """
+
+    def __init__(
+        self,
+        base_model: torch.nn.Module,
+        num_classes: int,
+        split_ratio: float = 0.5,
+        num_recycling_iterations: int = 100,
```
mattjhayes3, line 35: 100 seems high, that would mean ~100x forward passes during training.  

I don't think paper says how many they used, and the only value besides zero I'm seeing in their repo is 1.  

1 does seem kind of low, I'd be interested to see how much extra forward passes slow things down on these datasets.  I'd like try values like [0,1,2,3,5,10,100], but we probably don't have time for all of those.  

Let's maybe change the default to 1 for now and see how much it slows things down on `benchmarks/citation` before trying a couple other values if we have time?

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+        train_idx: Tensor,
+        batch: OptTensor = None,
+        ):
+        r"""
+        Forward pass using label usage algorithm.
+
+        Args:
+            x (torch.Tensor): Node feature tensor 
+            edge_index (torch.Tensor or SparseTensor): The edge indices
+            y (torch.Tensor): Node label tensor
```
mattjhayes3: We should be explicit about the expected dimensions.  It looks like it's currently `(N,)` where N is the length of the first dimension of `x`

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+            train_idx = batch
+
+        # random masking to split train_idx based on split ratio
+        mask = torch.rand(train_idx.shape) < self.split_ratio
+        train_labels_idx = train_idx[mask]  # D_L: nodes with features and labels
+        train_pred_idx = train_idx[~mask]  # D_U: nodes to predict labels 
+
+        # add labels to features for train_labels_idx nodes
+        # zero value nodes in train_pred_idx
+        onehot = torch.zeros([x.shape[0], self.num_classes]).to(x.device)
```
mattjhayes3, line 77: We might consider making this initialization user-configurable.  I'd be interested to benchmark if using the mean label works better when number of reuse iterations is zero or low. 

something like `init: Union[Tensor, float, 'mean'] = 0 # How to initialize unlabeled examples. 'mean' computes the mean label of the train_idx. If a tensor is passed, it must be one-dimensional and the same shape as y.size(1)`

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+        onehot[train_labels_idx, y[train_labels_idx]] = 1  # create a one-hot encoding
+        feat = torch.cat([x, onehot], dim=-1)
+
+        # label reuse procedure
+        for _ in range(max(1, self.num_recycling_iterations)):
```
mattjhayes3, line 83: Apologize this gets a bit nitpicky:
1. As the loop currently is, I think it should be `max(1, self.num_recycling_iterations + 1)`.  one iteration of this loop never actually "recycles" the predictions, it does "Label Usage" but not "Label reuse"
2. Combining the first iteration and recycle iterations is elegant, but it does introduce slight inefficiencies in the extra softmax and cat on the last iteration.  It's possible PyG team will care, doing the "usage" outside of the "reusage" loop is only one extra line that might be worth it, or we could call out the decision in the PR and/or writeup.  
3. The paper's impl `feat[unlabel_idx, -n_classes:] = F.softmax(pred[unlabel_idx], dim=-1)` may be a bit more efficient as well because it avoids copying the x features.  It's also shorter, though a little less readable for non-experts

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+            output = self.base_model(feat, edge_index)
+            pred_labels = F.softmax(output, dim=1)
+            feat[train_pred_idx] = torch.cat(
+                [x[train_pred_idx], pred_labels[train_pred_idx]], dim=1
```
mattjhayes3: We need to set the predictions for the test and validation indices as well.  I think it's safe to just compute the complement of the train indices before the loop, something like
```
mask = torch.ones(x.size(0), dtype=torch.bool)
mask[train_label_idx] = False
unlabeled_idx = torch.arange(x.size(0))[mask]
```

torch_geometric/nn/models/label_usage.py
```diff
@@ -0,0 +1,92 @@
...
+    Args:
+        base_model: An instance of the model that will do the 
+            inner forward pass.
+        num_classes (int): Number of classes in dataset
+        split_ratio (float): Proportion of true labels to use as features 
+            during training (default: :obj:'0.5')
+        num_recycling_iterations (int): Number of iterations for the
+            label reuse procedure to cycle predicted soft labels 
+            (default: :obj:'100')
+        return_tuple (bool): If true, returns (output, train_label_idx,
```
mattjhayes3: The paper impl only takes gradient steps on the `train_pred_idx` and thus we may want to make this true by default or even remove the option.  I haven't fully thought through if gradient steps on the `train_label_idx` might be harmful for different losses

PR #9864, opened 2024-12-15
CS224W - Adds TransD KGE, and Bernoulli corruption strategy for all KGE

mattjhayes3:
Implements TransD and Bernoulli corruption strategy (used in TransD and TransH papers).

## Details
- Did not see any PRs for these ones yet :)
- Used the KGEModel base and tried to keep consistent with other KGE
- Tried to implement everything as efficiently as possible while following the paper
- Incorporated into `examples/kge_fb15k_237.py`, also adding duration calculation there
- The Bernoulli parameters are currently computed for each batch rather than the whole training set
    - This is a little slower but seemed to work much better in terms of metrics
    - Also seems a little more intuitive for the user to simply flip a bool as opposed to requiring they pass their whole training set for initialization, or requiring they call loader().
    - We might also try cumulating the statistics during training, but these initial experiments suggest this would be the worst of both worlds (slower as we won't know when first epoch has ended, ultimately similar to precomputing in terms of metrics)
- Happy to split Bernoulli into a separate PR, or make any other edits you might prefer.

## Benchmarks 
I was only able to compare it to TransE on `examples/kge_fb15k_237.py` and do 3 runs each, but even without any hyperparameter tuning, findings seem fairly consistent with paper, i.e. significant improvements in all metrics [[colab](https://colab.research.google.com/drive/1FnMzHqnn8db1r4TEnHkBmYr1GHvSWTKx?usp=sharing)]:
| Method/Evalset        | Mean Rank     | MRR           | Hits@10       |
|:----------------------|:--------------|:--------------|:--------------|
| TransD Bernoulli Val  | 176.19 ± 0.74 | 0.223 ± 0.000 | 0.408 ± 0.001 |
| TransD Bernoulli Test | 180.65 ± 0.71 | 0.220 ± 0.001 | 0.401 ± 0.002 |
| TransD Val            | 180.08 ± 0.37 | 0.211 ± 0.003 | 0.400 ± 0.002 |
| TransD Test           | 184.06 ± 1.78 | 0.210 ± 0.003 | 0.396 ± 0.002 |
| TransE Val            | 258.74 ± 5.76 | 0.221 ± 0.000 | 0.366 ± 0.004 |
| TransE Test           | 268.30 ± 6.39 | 0.217 ± 0.001 | 0.362 ± 0.004 |


mattjhayes3:
Sorry for all the extra reviewers!  Was trying to fix the timeline with suggestion in https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

Abandoning this in favor of https://github.com/pyg-team/pytorch_geometric/pull/9866

PR #9877, opened 2024-12-17
Graph Sage OGBN Example with Scheduler

RorryB:
This PR is to update the ogbn example to have a scheduler, improving performance from a test accuracy of 75.52% to 77.19%


puririshi98 (approved):
thanks for addressing my offline comments, this looks good now

PR #9887, opened 2024-12-23
Migrate old `pytorch-lightning` imports to modern `lightning` imports

### LINKED ISSUE #9811: `ValueError("Expected a parent")` when using new Lightning imports and PyG Lightning DataModule wrappers
### 🐛 Describe the bug

# Bug
When using modern Lightning imports (i.e. `from lightning import ...`) rather than old imports (i.e. `from pytorch_lightning import ...`) in a project mixing Lightning and PyG, the code crashes on a non-intuitive `ValueError("Expected a parent")`. I suspect the exact line that raises this message depends on the user application.

This seems to be caused by Lightning sometimes (but not always!) failing when import styles are mixed together (see [this issue](https://github.com/Lightning-AI/pytorch-lightning/issues/17485) for context). Given that PyG's Lightning wrappers use the old style, but most other codebases use the modern style, a conflict happens.

Below is the smallest working example I could come up with that reproduces the issue:
```python
import torch
from lightning import LightningModule, Trainer
from torch.nn import functional as F
from torch_geometric.data.lightning import LightningNodeData
from torch_geometric.nn import GCN


class GCNLitModule(LightningModule):
    def __init__(self, in_features: int, num_classes: int):
        super().__init__()
        self.model = GCN(in_channels=in_features, hidden_channels=16, num_layers=2, out_channels=num_classes)

    def forward(self, batch):
        return self.model(batch.x, batch.edge_index)

    def training_step(self, batch):
        out = self.forward(batch)
        loss = F.nll_loss(out, batch.y)
        return loss

    def configure_optimizers(self):
        return torch.optim.Adam(self.model.parameters(), lr=0.01)


if __name__ == "__main__":
    pyg_dataset = Planetoid(root="../data", name="Cora")
    lit_model = GCNLitModule(in_features=pyg_dataset.num_features, num_classes=pyg_dataset.num_classes)
    lit_datamodule = LightningNodeData(pyg_dataset[0], batch_size=32, loader="full")
    trainer = Trainer(max_epochs=10)
    trainer.fit(model=lit_model, datamodule=lit_datamodule)

```

# Alternative considered
In this specific example, the error can be fixed by reverting back to the old import style:
```diff
- from lightning import LightningModule, Trainer
+ from pytorch_lightning import LightningModule, Trainer
```

However, in more complex projects, e.g. with other dependencies using Lightning, it's not realistic (and often outside of the user's power) to revert all imports to the old style. For example, in the real project where I first encountered this error, troubleshooting lead me to edit imports in dependencies...

# Suggested fix
Given that the modern Lightning import style is not so new anymore (and rather widely adopted from I could see), I would suggest PyG update how they import Lightning. I was able to fix the example above (and the issue in my real project) by simply changing [PyG's Lightning imports](https://github.com/pyg-team/pytorch_geometric/blob/10047ba010cf17e72ec54ef8c1fb5067c8c54334/torch_geometric/data/lightning/datamodule.py#L13-L18) like below:
```diff
try:
-    from pytorch_lightning import LightningDataModule as PLLightningDataModule
+    from lightning import LightningDataModule as PLLightningDataModule
    no_pytorch_lightning = False
except ImportError:
    PLLightningDataModule = object  # type: ignore
    no_pytorch_lightning = True
```

### Versions

I was not able to obtain the full info about my environment because I'm using `uv` and the script crashes on a `uv` environment. But I am experiencing this with Python 3.12, and here is the relevant info about the torch-related dependencies:
```text
lightning==2.4.0
lightning-utilities==0.11.9
torch==2.5.1
torch-geometric==2.6.1
pytorch-lightning==2.4.0
```

nathanpainchaud:
Fixes #9811.

As the title says. PyTorch Lightning as been recommending a new import style (`import Lightning as L` rather than `import pytorch_lightning as pl`) since early 2023 now.

The old style was/is still supported for backwards compatibility, but both styles cannot be used at the same time (see [here](https://github.com/Lightning-AI/pytorch-lightning/discussions/16688#discussioncomment-8193205)). Since PyG uses the old style, this means that projects using PyG are stuck with the old style. Most importantly, it can cause clashes when projects rely on other dependencies that use the modern import style (see #9811).

Therefore, I would suggest that PyG migrates to the new style, like other projects have done (e.g. Optuna [here](https://github.com/optuna/optuna/issues/4935)).

Since the task didn't seem to massive, I've migrated over all of the Lightning that I found across the project, and standardized the import styles between examples and tests (to use the style recommended in Lightning's own docs).

### DIFF
--- a/examples/pytorch_lightning/gin.py
+++ b/examples/pytorch_lightning/gin.py
@@ -1,8 +1,10 @@
 import os.path as osp
 
-import pytorch_lightning as pl
 import torch
 import torch.nn.functional as F
+from lightning import LightningModule, Trainer
+from lightning.pytorch.callbacks import ModelCheckpoint
+from lightning.pytorch.strategies import DDPStrategy
 from torchmetrics import Accuracy
 
 import torch_geometric.transforms as T
@@ -11,7 +13,7 @@ from torch_geometric.datasets import TUDataset
 from torch_geometric.nn import GIN, MLP, global_add_pool
 
 
-class Model(pl.LightningModule):
+class Model(LightningModule):
     def __init__(self, in_channels: int, out_channels: int,
                  hidden_channels: int = 64, num_layers: int = 3,
                  dropout: float = 0.5):
@@ -71,10 +73,10 @@ if __name__ == '__main__':
     model = Model(dataset.num_node_features, dataset.num_classes)
 
     devices = torch.cuda.device_count()
-    strategy = pl.strategies.DDPStrategy(accelerator='gpu')
-    checkpoint = pl.callbacks.ModelCheckpoint(monitor='val_acc', save_top_k=1,
+    strategy = DDPStrategy(accelerator='gpu')
+    checkpoint = ModelCheckpoint(monitor='val_acc', save_top_k=1,
                                               mode='max')
-    trainer = pl.Trainer(strategy=strategy, devices=devices, max_epochs=50,
+    trainer = Trainer(strategy=strategy, devices=devices, max_epochs=50,
                          log_every_n_steps=5, callbacks=[checkpoint])
 
     trainer.fit(model, datamodule)
--- a/examples/pytorch_lightning/graph_sage.py
+++ b/examples/pytorch_lightning/graph_sage.py
@@ -1,8 +1,10 @@
 import os.path as osp
 
-import pytorch_lightning as pl
 import torch
 import torch.nn.functional as F
+from lightning import LightningModule, Trainer
+from lightning.pytorch.callbacks import ModelCheckpoint
+from lightning.pytorch.strategies import SingleDeviceStrategy
 from torch.nn import BatchNorm1d
 from torchmetrics import Accuracy
 
@@ -11,7 +13,7 @@ from torch_geometric.datasets import Reddit
 from torch_geometric.nn import GraphSAGE
 
 
-class Model(pl.LightningModule):
+class Model(LightningModule):
     def __init__(self, in_channels: int, out_channels: int,
                  hidden_channels: int = 256, num_layers: int = 2,
                  dropout: float = 0.5):
@@ -71,10 +73,10 @@ if __name__ == '__main__':
 
     model = Model(dataset.num_node_features, dataset.num_classes)
 
-    strategy = pl.strategies.SingleDeviceStrategy('cuda:0')
-    checkpoint = pl.callbacks.ModelCheckpoint(monitor='val_acc', save_top_k=1,
+    strategy = SingleDeviceStrategy('cuda:0')
+    checkpoint = ModelCheckpoint(monitor='val_acc', save_top_k=1,
                                               mode='max')
-    trainer = pl.Trainer(strategy=strategy, devices=1, max_epochs=20,
+    trainer = Trainer(strategy=strategy, devices=1, max_epochs=20,
                          callbacks=[checkpoint])
 
     trainer.fit(model, datamodule)
--- a/examples/pytorch_lightning/relational_gnn.py
+++ b/examples/pytorch_lightning/relational_gnn.py
@@ -1,11 +1,11 @@
 import os.path as osp
 from typing import Dict, List, Tuple
 
-import pytorch_lightning as pl
 import torch
 import torch.nn.functional as F
-from pytorch_lightning import LightningModule, Trainer
-from pytorch_lightning.callbacks import ModelCheckpoint
+from lightning import LightningModule, Trainer
+from lightning.pytorch.callbacks import ModelCheckpoint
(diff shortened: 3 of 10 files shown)


akihironitta (commented):
Hi @nathanpainchaud, thanks for working on this with the PR description :) Given comments in the linked issues, we should definitely do something about this.

For the reason stated in https://github.com/pyg-team/pytorch_geometric/discussions/9235#discussioncomment-11187670, could we:
* keep the import statements in examples as they are, and
* support both `pytorch_lightning` and `lightning` names in `torch_geometric/` to avoid a breaking change?

nathanpainchaud:
> Hi @nathanpainchaud, thanks for working on this with the PR description :) Given comments in the linked issues, we should definitely do something about this.

Thanks for realizing that something has to be done about this. I'm coming back to this PR to try and get the ball rolling again :wink: 

> For the reason stated in [#9235 (reply in thread)](https://github.com/pyg-team/pytorch_geometric/discussions/9235#discussioncomment-11187670), could we:
> 
> * keep the import statements in examples as they are, and
> * support both `pytorch_lightning` and `lightning` names in `torch_geometric/` to avoid a breaking change?

Considering your second point, if that's the route you want to take, a reply in the thread you pointed me to already offers a solution (used by Ray) to support both packages:
```python
try:
    from lightning import *
except ModuleNotFoundError:
    from pytorch_lightning import *
```

Given that the import statement themselves are problematic, I don't see a way around this issue without modifying them (your first point).

As a more general comment/opinion, I understand and sympathize with the burden on maintainers that Lightning's upstream changes caused. Still, I don't think I see a long-term viable solution other than to adopt the new recommendation, given that it's been widely adopted by the rest of the ecosystem now. And IMO, I don't see the change in dependencies/import style as a breaking change, considering that the current way of doing things is already broken for most people :sweat_smile:

akihironitta:
Hey @nathanpainchaud, sorry for the delay, and thank you so much for sharing your thoughts!

> Still, I don't think I see a long-term viable solution other than to adopt the new recommendation, given that it's been widely adopted by the rest of the ecosystem now.

Being officially recommended shouldn't be the only reason to adopt it, and it seems to me that it hasn't been widely adopted in the ecosystem after looking at download stats:
* https://pepy.tech/projects/pytorch-lightning?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=2.5.2%2C2.4.0%2C2.3.3%2C2.2.5
* https://pepy.tech/projects/lightning?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=2.5.2%2C2.4.0%2C2.3.3%2C2.2.5

>  I don't see the change in dependencies/import style as a breaking change, considering that the current way of doing things is already broken for most people

I totally agree that this is something we should fix. However, this is not just style change, and as far as I understand, it still is a breaking change for users who install PyTorch Lightning with `pip install pytorch-lightning` because `import lightning` will fail in this case. https://stackoverflow.com/questions/21703216/what-is-a-breaking-change-in-software

nathanpainchaud:
> Being officially recommended shouldn't be the only reason to adopt it, and it seems to me that it hasn't been widely adopted in the ecosystem after looking at download stats:
> 
> * https://pepy.tech/projects/pytorch-lightning?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=2.5.2%2C2.4.0%2C2.3.3%2C2.2.5
> * https://pepy.tech/projects/lightning?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=2.5.2%2C2.4.0%2C2.3.3%2C2.2.5

I would say that these stats are misleading at first glance. Yes pytorch-lightning (pl) as much more downloads, but that is because it's been around for longer than Lightning (L). If you look at the downloads for the last few releases (in the lower left), the downloads are more even (2.5.2 is 5.5M for pl vs 3.2M for L).

Aside from individual downloads and adoption by end-users in applications, I'd say that another big factor is adoption in the ecosystem of other tools working with (PyTorch) Lightning. On that front, all of the other libraries that I use and I've seen, like Optuna linked above, have migrated to the new import style. So PyG is more of a hold-out on that front.

> > I don't see the change in dependencies/import style as a breaking change, considering that the current way of doing things is already broken for most people
> 
> I totally agree that this is something we should fix. However, this is not just style change, and as far as I understand, it still is a breaking change for users who install PyTorch Lightning with `pip install pytorch-lightning` because `import lightning` will fail in this case. https://stackoverflow.com/questions/21703216/what-is-a-breaking-change-in-software

I understand your point. So I'll try to work on fixing the fix as I've described in [my comment on the original issue](https://github.com/pyg-team/pytorch_geometric/issues/9811#issuecomment-3200191285), and hopefully this will satisfy everyone :slightly_smiling_face:

PR #9896, opened 2024-12-27
Fix broken metrics in a `DistributedDataParallel` example

akihironitta:
Fixes a few issues:
* Fixes the same three calls to `dist.all_reduce(train_acc, op=dist.ReduceOp.SUM)` (introduced in #8880) that led to the wrong metrics.
* Avoids the `int(cuda_tensor)` call in every eval iteration to get rid of the D2H synchronization.
* Calls `optimizer.step()` at the end of each training step to release GPU memory for gradients before evaluation loops.
* Minor decorative changes:
  * Adds type annotations.
  * Adds a progress bar for the training loop.

### DIFF
--- a/examples/multi_gpu/distributed_sampling.py
+++ b/examples/multi_gpu/distributed_sampling.py
@@ -1,7 +1,6 @@
 import os
 import os.path as osp
 from math import ceil
-from tqdm import tqdm
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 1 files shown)


examples/multi_gpu/distributed_sampling.py
```diff
@@ -34,20 +40,25 @@ def forward(self, x: Tensor, edge_index: Tensor) -> Tensor:
...
-        total_correct += int((pred == y).sum())
+        total_correct += (pred == y).sum()
         total_examples += batch.batch_size
-    return torch.tensor(total_correct / total_examples, device=rank)
+
+    return total_correct / total_examples
 
 
-def run(rank, world_size, dataset):
+def run(rank: int, world_size: int, dataset: Reddit) -> None:
```
Kh4L, line 61:
```suggestion
def run(rank: int, world_size: int, dataset: Dataset) -> None:
```

Kh4L (changes requested):
Thanks for the PR!

PR #9905, opened 2024-12-30
Graphgym documentation completion

kolmiw:
I changed a single line of a comment. The compute loss function's description was cut in half mid-sentence. This change shouldn't change the code by any means though


akihironitta:
For note, the mypy failure is unrelated to the change of this PR.

PR #9911, opened 2025-01-02
Add new batch filtering method for efficiency purpose

leonardcaquot94:
This PR improves `DataBatch` and `HeteroDataBatch` filtering:

- **Keeps Original Batch Type:** Previously, `batch[mask]` used to return a data list, requiring reconversion with `Batch.from_data_list(batch[mask])`. Now, it directly returns a filtered Batch, preserving the original type.

- **Faster Filtering:** The old approach used to split the batch in order to apply the mask, which was inefficient, especially with batch reconstruction afterward. The new implementation applies the mask directly to all batch components, achieving up to 10x speedup.

**Note**: If the user still needs a data list, they can convert the result with batch[mask].to_data_list(), or use the previous method: batch.index_select(mask).


leonardcaquot94:
Here is a discussion related to this PR : https://github.com/pyg-team/pytorch_geometric/discussions/9681

Issue #9895, opened 2024-12-26
GlobalStorage#is_node_attr() breaks when cat_dim is a tuple


Wollents:
### 🐛 Describe the bug

Hi,

when adj in the GlobalStorage is a ** sparse tensor**, `_cat_dim_` would be a tuple. But I want to use the to_dgl(), it would throws an error. Because, when calling the method GlobalStorage#is_node_attr(), there are lines that make the application throw errows. 

```python
 cat_dim = self._parent().__cat_dim__(key, value, self)
 num_nodes, num_edges = self.num_nodes, self.num_edges
 if value.shape[cat_dim] != num_nodes:
     if value.shape[cat_dim] == num_edges:
         self._cached_attr[AttrType.EDGE].add(key)
     else:
         self._cached_attr[AttrType.OTHER].add(key)
     return False
```
The following snippet recreates the error:
``` python
import torch
from torch_geometric.data import Data
from torch_geometric.utils import to_dense_adj, to_dgl

edge_index = torch.tensor([
    [0, 1, 1, 2],
    [1, 0, 2, 1]
], dtype=torch.long)

adj = to_dense_adj(edge_index=edge_index).squeeze()
adj = adj.to_sparse_coo()

graph = Data(adj=adj, edge_index=edge_index)

dgl_graph = to_dgl(graph)
```

this example will throw the following error message:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/root/.pyenv/versions/3.9.21/envs/pygodEnv/lib/python3.9/site-packages/torch_geometric/utils/convert.py", line 539, in to_dgl
    for attr in data.node_attrs():
  File "/root/.pyenv/versions/3.9.21/envs/pygodEnv/lib/python3.9/site-packages/torch_geometric/data/data.py", line 215, in node_attrs
    return list(set(chain(*[s.node_attrs() for s in self.node_stores])))
  File "/root/.pyenv/versions/3.9.21/envs/pygodEnv/lib/python3.9/site-packages/torch_geometric/data/data.py", line 215, in <listcomp>
    return list(set(chain(*[s.node_attrs() for s in self.node_stores])))
  File "/root/.pyenv/versions/3.9.21/envs/pygodEnv/lib/python3.9/site-packages/torch_geometric/data/storage.py", line 530, in node_attrs
    return [key for key in self.keys() if self.is_node_attr(key)]
[...]

Issue #8709, opened 2024-01-02
NeighborLoader breaks when cat_dim is a tuple


viktor-ktorvi:
### 🐛 Describe the bug

Hi,

I batch my sparse matrices block diagonally so `cat_dim` is a tuple.  I want to use NeighborLoader but it expects `cat_dim` to not be a tuple internally and throws an error:

```
Traceback (most recent call last):
  File "C:\Users\todos\PycharmProjects\mlpf\examples\graph_sampling_bug.py", line 59, in <module>
    main()
  File "C:\Users\todos\PycharmProjects\mlpf\examples\graph_sampling_bug.py", line 55, in main
    subgraph = next(iter(subgraph_loader))
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch\utils\data\dataloader.py", line 630, in __next__
    data = self._next_data()
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch\utils\data\dataloader.py", line 674, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch\utils\data\_utils\fetch.py", line 54, in fetch
    return self.collate_fn(data)
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch_geometric\loader\node_loader.py", line 148, in collate_fn
    out = self.filter_fn(out)
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch_geometric\loader\node_loader.py", line 165, in filter_fn
    data = filter_data(  #
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch_geometric\loader\utils.py", line 156, in filter_data
    filter_node_store_(data._store, out._store, node)
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch_geometric\loader\utils.py", line 92, in filter_node_store_
    elif store.is_node_attr(key):
  File "C:\Users\todos\anaconda3\envs\mlpfenv_cu121_pyg_newest\lib\site-packages\torch_geometric\data\storage.py", line 808, in is_node_attr
[...]

PR #9927, opened 2025-01-09
Fix `GlobalStorage.is_node_attr` and `GlobalStorage.is_edge_attr` when passing tuples in `cat_dim`

erytheis:
The PR fixes the issue when calling is_node_attr and is_edge_attr for Data objects with sparse attributes that assume block-diagonal concatenation ( [#9895](https://github.com/pyg-team/pytorch_geometric/issues/9895) and in  [#8709](https://github.com/pyg-team/pytorch_geometric/issues/8709)). These cases raise an error due to cat_dim being a tuple, the change catches this case and looks at the 0-th dimension of the sparse tensor.


rusty1s:
I would assume that if `cat_dim` is given as a tuple, it can never be a node or edge attribute. Shall we just return `False` in these cases?

rusty1s (approved):

PR #9931, opened 2025-01-11
[BugFix] Fixing two `lightning` tests.

drivanov:
The tests `test_lightning_dataset` and `test_lightning_node_data` began failing after the recent update of `pytorch_lightning` to version 2.5. The cause of these failures is the new way for calculation of `str(datamodule)`

Since 2024-12-10 they have started to use a new method:
```
~/Projects/pytorch-lightning$ git blame src/lightning/pytorch/core/datamodule.py -L 249,253
9709c645c8 (Nikita Tatsch 2024-12-10 19:39:25 -0500 249)     def __str__(self) -> str:
9709c645c8 (Nikita Tatsch 2024-12-10 19:39:25 -0500 250)         """Return a string representation of the datasets that are set up.
9709c645c8 (Nikita Tatsch 2024-12-10 19:39:25 -0500 251) 
9709c645c8 (Nikita Tatsch 2024-12-10 19:39:25 -0500 252)         Returns:
9709c645c8 (Nikita Tatsch 2024-12-10 19:39:25 -0500 253)             A string representation of the datasets that are setup.
```

In 2.4 the operation `str(datamodule)` was performed by
```
304         def __repr__(self) -> str:
305  ->         kwargs = kwargs_repr(
306                 train_dataset=self.train_dataset,
307                 val_dataset=self.val_dataset,
308                 test_dataset=self.test_dataset,
309                 pred_dataset=self.pred_dataset,
310                 **self.kwargs,
311             )
312             return f'{self.__class__.__name__}({kwargs})'
```

This PR addresses the issue while maintaining backward compatibility, ensuring that these tests run correctly with `pytorch_lightning` versions earlier than 2.5.

### DIFF
--- a/test/data/lightning/test_datamodule.py
+++ b/test/data/lightning/test_datamodule.py
@@ -18,12 +18,12 @@ from torch_geometric.testing import (
     MyFeatureStore,
     MyGraphStore,
     get_random_edge_index,
+    has_package,
     onlyCUDA,
     onlyFullTest,
     onlyNeighborSampler,
     onlyOnline,
     withPackage,
-    has_package,
 )
 
 try:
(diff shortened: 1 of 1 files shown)


puririshi98 (approved):
thanks for the fix

PR #9953, opened 2025-01-15
Cugraph Examples Updates

guanxingithub:
We consolidate cugraph training with
Models: GAT, GCN, GraphSAGE
Datasets: ogbn-arxiv, ogbn-products, ogbn-papers100M
examples/ogbn_train_cugraph.py is the script for single gpu training
examples/multi_gpu/ogbn_train_cugraph_multigpu.py is the script for multiple gpu training.


puririshi98:
@guanxingithub, as mentioned before please delete the old cugraph examples that you based these new files on, otherwise there will be duplicates.

puririshi98:
also please attach logs of these scripts running succesfully on your setup for review

guanxingithub:
[h100_test_1gpu_1024_gpu.log](https://github.com/user-attachments/files/18445214/h100_test_1gpu_1024_gpu.log)
[h100_test_8gpu_1024_gpu.log](https://github.com/user-attachments/files/18445216/h100_test_8gpu_1024_gpu.log)

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+    for i, batch in enumerate(train_loader):
+        batch = batch.cuda()
+        optimizer.zero_grad()
+        out = model(batch.x, batch.edge_index)[:batch.batch_size]
+        y = batch.y[:batch.batch_size].view(-1).to(torch.long)
+        loss = F.cross_entropy(out, y)
+        loss.backward()
+        optimizer.step()
+
+        total_loss += float(loss) * y.size(0)
```
akihironitta, line 128: Let's not call `float()` to avoid unnecessary D2H sync every step.

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+        total_correct += int(out.argmax(dim=-1).eq(y).sum())
```
akihironitta, line 129: ditto

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+def test(model, loader):
+    model.eval()
+
+    total_correct = total_examples = 0
+    for batch in loader:
+        batch = batch.cuda()
+        out = model(batch.x, batch.edge_index)[:batch.batch_size]
+        y = batch.y[:batch.batch_size].view(-1).to(torch.long)
+
+        total_correct += int(out.argmax(dim=-1).eq(y).sum())
```
akihironitta, line 145: ditto

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+        epochs = args.epochs
+        for epoch in range(1, epochs + 1):
+            train_start = time.time()
+            loss, train_acc = train(model, train_loader)
+            train_end = time.time()
+            train_times.append(train_end - train_start)
+            inference_start = time.time()
+            train_acc = test(model, train_loader)
+            val_acc = test(model, val_loader)
+            test_acc = test(model, test_loader)
```
akihironitta, line 257: Let's not evaluate test accuracy every epoch. It should only be computed once at the end of this script.

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+            print(f'Epoch {epoch:02d}, Loss: {loss:.4f}, Approx. Train:'
+                  f' {train_acc:.4f} Time: {train_end - train_start:.4f}s')
+            print(f'Train: {train_acc:.4f}, Val: {val_acc:.4f}, '
+                  f'Test: {test_acc:.4f}')
+
+            times.append(time.time() - train_start)
+            if val_acc > best_val:
+                best_val = val_acc
+            if test_acc > best_test:
+                best_test = test_acc
```
akihironitta, line 271: ditto

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+    elif args.model == "SAGE":
+        print(f"Training {args.dataset} with GraphSAGE model.")
+        model = torch_geometric.nn.models.GraphSAGE(
+            dataset.num_features,
+            args.hidden_channels,
+            args.num_layers,
+            dataset.num_classes,
+        ).cuda()
+    else:
+        pass
```
akihironitta, line 205:
```suggestion
    else:
        raise ValueError(...)
```

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+
+    total_loss = total_correct = total_examples = 0
+    for i, batch in enumerate(train_loader):
+        batch = batch.cuda()
+        optimizer.zero_grad()
+        out = model(batch.x, batch.edge_index)[:batch.batch_size]
+        y = batch.y[:batch.batch_size].view(-1).to(torch.long)
+        loss = F.cross_entropy(out, y)
+        loss.backward()
+        optimizer.step()
```
akihironitta, line 126: Can we call `optimizer.zero_grad` at the end? Otherwise, the gradients are still left on the device during subsequent evaluation loops, I believe.

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,293 @@
...
+import torch
+from rmm.allocators.cupy import rmm_cupy_allocator
+from rmm.allocators.torch import rmm_torch_allocator
+
+# Must change allocators immediately upon import
+# or else other imports will cause memory to be
+# allocated and prevent changing the allocator
+rmm.reinitialize(devices=[0], pool_allocator=True, managed_memory=True)
+cupy.cuda.set_allocator(rmm_cupy_allocator)
+torch.cuda.memory.change_current_allocator(rmm_torch_allocator)
```
akihironitta, line 19: I think this is new even for most people who are somewhat familiar with PyTorch. Are there references (issues, docs, etc.) that users can read to find out more? If so, it'd be great if you could include them in the comment here?

akihironitta (commented):
Great work @guanxingithub! This is not blocking you from merging this PR, but what would be so nice before merging this PR would be to profile some parts of the scripts and make sure there's no basic inefficient code :)

examples/ogbn_train_cugraph.py
guanxingithub: total_loss += loss.item() * y.size(0)

examples/ogbn_train_cugraph.py
guanxingithub: total_correct += out.argmax(dim=-1).eq(y).sum().item()

examples/ogbn_train_cugraph.py
guanxingithub: total_correct += out.argmax(dim=-1).eq(y).sum().item()

examples/ogbn_train_cugraph.py
guanxingithub: Sure! remove it.

examples/ogbn_train_cugraph.py
guanxingithub: remove it

examples/ogbn_train_cugraph.py
guanxingithub: Sure!

examples/ogbn_train_cugraph.py
akihironitta: Let's not call `.item()` here. You can call it once at the end of a training epoch.
```suggestion
        total_loss += loss * y.size(0)
```

examples/ogbn_train_cugraph.py
akihironitta: Let's not call `.item()`. Instead, let's remove `int(...)` from this line and call `.item()` only once at the end of a training epoch.

examples/ogbn_train_cugraph.py
akihironitta: ditto

akihironitta (approved):
Nice work :)

examples/ogbn_train_cugraph.py
guanxingithub: Good suggestion!

examples/ogbn_train_cugraph.py
guanxingithub: Good suggestion. But at the beginning of the loop, especially first iteration, we had better to clean gradients left on the device. We can add optimizer.zero_grad after the whole loop so that the optimizer is cleaned up and ready for other training.

akihironitta: > But at the beginning of the loop, especially first iteration, we had better to clean gradients left on the device.

In what case gradients are left on device in the current script?

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,289 @@
...
+def test(model, loader):
+    model.eval()
+
+    total_correct = total_examples = 0
+    for batch in loader:
+        batch = batch.cuda()
+        out = model(batch.x, batch.edge_index)[:batch.batch_size]
+        y = batch.y[:batch.batch_size].view(-1).to(torch.long)
+
+        total_correct += out.argmax(dim=-1).eq(y).sum().item()
```
akihironitta, line 150: Any chance you could make sure this is addressed in all the scripts you touch in this PR?
```suggestion
        total_correct += out.argmax(dim=-1).eq(y).sum()
```

examples/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,289 @@
...
+        prep_time = round(time.perf_counter() - wall_clock_start, 2)
+        print("Total time before training begins (prep_time) =", prep_time,
+              "seconds")
+        print("Beginning training...")
+        val_accs = []
+        times = []
+        train_times = []
+        inference_times = []
+        best_val = 0.
+        start = time.time()
```
akihironitta, line 251: Can you use `time.perf_counter()` for consistency across scripts in this PR?
```suggestion
        start = time.perf_counter()
```

akihironitta (commented):
Nice work! I'd like these cugraph examples to be as performant as possible :) Let's merge this once comments are addressed and this PR is cleaned up :)

examples/ogbn_train_cugraph.py
guanxingithub: when optimizer is created, the default gradients maybe not zero. so when we use it first time, had better to reset it.

guanxingithub:
[ogbn-arxiv_cugraph_1gpu_profiling.txt](https://github.com/user-attachments/files/18510930/ogbn-arxiv_cugraph_1gpu_profiling.txt)

puririshi98:
@akihironitta reviewing the profile by @guanxingithub, it does not seem like there are any bottlenecks, i will leave to you to decide if we can merge if this investigation satisfies your concerns

akihironitta (commented):
Can you also make sure `README.md` also gets updated?

akihironitta (commented):
@puririshi98 I'm not sure you have already, but would you mind reviewing this PR as well?

puririshi98:
i have but ill do another omnday

puririshi98:
> Can you also make sure `README.md` also gets updated?

good catch

puririshi98 (approved):
LGTM, was also reviewed by Alexandria, our cugraph expert. @akihironitta lmk if anything else needed to merge

akihironitta (commented):
@guanxingithub @puririshi98 @alexbarghi-nv Thanks for bearing with me ❤️ I'll test this out and get it merged tonight!

examples/multi_gpu/ogbn_train_cugraph.py
```diff
@@ -0,0 +1,378 @@
...
+    )
+
+    cupy.cuda.Device(rank).use()
+    from rmm.allocators.cupy import rmm_cupy_allocator
+
+    cupy.cuda.set_allocator(rmm_cupy_allocator)
+
+    from cugraph.testing.mg_utils import enable_spilling
+
+    enable_spilling()
```
akihironitta, line 136: small: Can we avoid using `enable_spilling` in both scripts? otherwise, this script requires `pytest`.

akihironitta (approved):
Haven't managed to run, but LGTM!

examples/multi_gpu/ogbn_train_cugraph.py
akihironitta: sorry i meant we should replace it with its implementation inline:
https://github.com/rapidsai/cugraph/blob/b64b04fde6bab4ac1c810269c2101eff0e41cabd/python/cugraph/cugraph/testing/mg_utils.py#L205-L207

guanxingithub: Please let me know whether the implementation is what you expected? Thanks.

puririshi98: @guanxingithub the original enable spilling set `cudf.set_option("spill", True)`, curious why you choose to set it to False now

puririshi98: LGTM now thanks

PR #9955, opened 2025-01-15
Fixed a warning generated in `exemples/llm/glen.py` test.

drivanov:
The following warning has been fixed:
```
/usr/local/lib/python3.12/dist-packages/torch_geometric/data/in_memory_dataset.py:300: UserWarning: It is not recommended 
to directly access the internal storage format `data` of an 'InMemoryDataset'. If you are absolutely certain what you are 
doing, access the internal storage via `InMemoryDataset._data` instead to suppress this warning. Alternatively, you can 
access stacked individual attributes of every graph via `dataset.{attr_name}`.
  warnings.warn(msg)
```

### DIFF
--- a/examples/llm/glem.py
+++ b/examples/llm/glem.py
@@ -79,7 +79,7 @@ def main(args):
 
     dataset = PygNodePropPredDataset(f'ogbn-{dataset_name}', root=root)
     split_idx = dataset.get_idx_split()
-    data = dataset._data
+    data = dataset[0]
 
     tag_dataset = TAGDataset(root, dataset, hf_model,
                              token_on_disk=token_on_disk)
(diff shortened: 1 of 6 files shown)


examples/llm/glem.py
```diff
@@ -79,7 +79,7 @@ def main(args):
 
     dataset = PygNodePropPredDataset(f'ogbn-{dataset_name}', root=root)
     split_idx = dataset.get_idx_split()
-    data = dataset.data
+    data = dataset._data
```
akihironitta, line 82: It's usually a good idea to avoid exposing private attributes to users.
```suggestion
    data = dataset[0]
```

akihironitta (commented):
Nice catch!

examples/llm/glem.py
drivanov: Done. Thanks for the advice!

akihironitta (approved):

PR #9956, opened 2025-01-16
Add `LPFormer` model and example

HarryShomer:
This pull request adds an implementation of the SOTA Link Prediction model `LPFormer` from the KDD'24 paper - [LPFormer: An Adaptive Graph Transformer for Link Prediction](https://arxiv.org/pdf/2310.11009).

### Example

`python examples/lpformer.py`. 

I am able to replicate the reported performance on `ogbl-ppa`. 

The results are shown below. I use the same hyperparameters used in the paper. Results are over 5 seeds.

| Reported | PyG |
| ------------  | ------ |
|  63.32 ± 0.6 | 63.15 ± 0.9  |

### Testing

The tests for LPFormer can be found in `test/nn/models/test_lpformer.py`. They each pass.

### Docs

I've also included the documentation for the code. It builds and renders correctly with Sphinx with no errors.


HarryShomer:
@wsad1 @EdisonLeeeee Can I get an ETA on a review? Thanks!

puririshi98:
> @puririshi98 Thanks for the heads up!
> 
> I'm curious though, what's the alternative to torch_sparse? As far as I can tell, pyg still doesn't support torch.sparse for memory-efficient aggregations, which is why I use SparseTensor.
> 
> Could you let me know what I should use instead of torch_sparse?

there should be a pytorch native version, look into it, if you struggle to find an option let me know and i can try to check

HarryShomer:
@puririshi98 I've updated the code to remove the torch_sparse references.

For some reason I was under the impression that the MessagePassing layers could not accept torch.sparse for the edge_index argument. I think this may have been true for some older versions.

Please let me know if there's anything else needed on my side.

puririshi98:
will merge once we get the onnx issue fixed upstream and merge that in . aiming EoD tmrw

Issue #9709, opened 2024-10-14
add_random_edge triggers type error


Zero-Yi:
### 🐛 Describe the bug

Hi,

```python
import torch
from torch_geometric.utils import add_random_edge

edge_index = torch.tensor([[0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 4],
                            [1, 2, 3, 4, 0, 2, 3, 4, 0, 1, 4, 0, 1, 4, 0, 1, 2, 3]])

edge_index, added_edges = add_random_edge(edge_index, p=0.05, force_undirected=True, num_nodes=5)
```

And the error goes:
```bash
Traceback (most recent call last):
  File "/workspaces/mygcl/test.py", line 7, in <module>
    edge_index, added_edges = add_random_edge(edge_index, p=0.05, force_undirected=True, num_nodes=5)
  File "/usr/local/lib/python3.9/dist-packages/torch_geometric/utils/augmentation.py", line 230, in add_random_edge
    edge_index_to_add = negative_sampling(
  File "/usr/local/lib/python3.9/dist-packages/torch_geometric/utils/_negative_sampling.py", line 114, in negative_sampling
    return vector_to_edge_index(neg_idx, size, bipartite, force_undirected)
  File "/usr/local/lib/python3.9/dist-packages/torch_geometric/utils/_negative_sampling.py", line 377, in vector_to_edge_index
    col = offset[row].add_(idx) % num_nodes
RuntimeError: result type Float can't be cast to the desired output type Long
```

It seems not adding any edges, which is okay. But there might be some modifications needed in `negative_sampling` to prevent the runtime type error.

### Versions

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 23357  100 23357    0     0  82052      0 --:--:-- --:--:-- --:--:-- 82242
Collecting environment information...
PyTorch version: 2.1.0+cu118
Is debug build: False
CUDA used to build PyTorch: 11.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.3 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.35

[...]

PR #9957, opened 2025-01-17
Fix negative_sampling leading to runtime error because of zero sample size

erytheis:
Fixes `negative_sampling` leading to runtime error when the `sample_size` is 0 (for example when only 1 undirected edge is sampled [https://github.com/pyg-team/pytorch_geometric/issues/9709](https://github.com/pyg-team/pytorch_geometric/issues/9709) )


akihironitta (commented):
Mind adding a test case to cover what's being fixed?

PR #9992, opened 2025-01-29
Latest txt2kg

puririshi98:


puririshi98:
@rusty1s @akihironitta this is mostly ready for review, except i need to cleanup for the CI linters, but any early reviews would help getting this merged once those cleanups are done.

torch_geometric/nn/nlp/txt2kg.py
```diff
@@ -0,0 +1,321 @@
...
+                # Process chunks in parallel using multiple processes
+                num_procs = min(len(chunks), _get_num_procs())
+                meta_chunk_size = int(len(chunks) / num_procs)
+                in_chunks_per_proc = {
+                    j:
+                    chunks[j *
+                           meta_chunk_size:min((j + 1) *
+                                               meta_chunk_size, len(chunks))]
+                    for j in range(num_procs)
+                }
```
Kh4L, line 154: We should import `openai` here so it fails in the main process rather than in subprocesses:
```python
try:
    import openai
except ImportError:
    print("Failed to import `openai` package, please install it and rerun the script")
    sys.exit(1)
```

torch_geometric/nn/nlp/txt2kg.py
```diff
@@ -0,0 +1,321 @@
...
+def _chunk_to_triples_str_cloud(
+        txt: str, GLOBAL_NIM_KEY='',
+        NIM_MODEL="nvidia/llama-3.1-nemotron-70b-instruct",
+        post_text=SYSTEM_PROMPT) -> str:
+    global CLIENT_INITD
+    if not CLIENT_INITD:
+        # We use NIMs since most PyG users may not be able to run a 70B+ model
+        from openai import OpenAI
+        global CLIENT
+        CLIENT = OpenAI(base_url="https://integrate.api.nvidia.com/v1",
```
Kh4L, line 194: Would be nice to expose base_url through argparse

Kh4L (changes requested):

torch_geometric/nn/nlp/txt2kg.py
puririshi98: i will actually add this functionality myself but still would be good for you to benchmark openAI models vs our NIM models

puririshi98:
will make CI green and then merge after https://github.com/pyg-team/pytorch_geometric/pull/10126/ goes into main and I pull main into this branch

puririshi98:
@wsad1 @rusty1s @akihironitta plz lmk what needed to merge this before cutting 2.7 PyG next tues (9/2)

docs/source/_figures/flowchart.svg
wsad1, line 1: why add these?
Do you plan to use it later on in docs. 
Can we add these when its used.

torch_geometric/utils/rag/graph_store.py
```diff
@@ -0,0 +1,192 @@
...
+)
+from torch_geometric.typing import EdgeTensorType
+from torch_geometric.utils import index_sort
+
+
+class NeighborSamplingRAGGraphStore(LocalGraphStore):
+    """Neighbor sampling based graph-store to store & retrieve graph data."""
+    def __init__(  # type: ignore[no-untyped-def]
+        self,
+        feature_store: Optional[FeatureStore] = None,
```
wsad1, line 21: Why is the feature_store an input to the graph store?
Can't we just initialize the sampler outside the graph store?

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -130,124 +36,225 @@ class KGQABaseDataset(InMemoryDataset):
...
+        load_dataset_kwargs: Optional[Dict[str, Any]] = None,
+        retrieval_kwargs: Optional[Dict[str, Any]] = None,
     ) -> None:
+        self.split = split
         self.dataset_name = dataset_name
         self.use_pcst = use_pcst
+        self.load_dataset_kwargs = load_dataset_kwargs or {}
+        """
+        NOTE: If running into memory issues,
+        try reducing this batch size
```
wsad1: which batch size?

torch_geometric/datasets/web_qsp_dataset.py
```diff
@@ -130,124 +36,225 @@ class KGQABaseDataset(InMemoryDataset):
...
     """
     def __init__(
         self,
         dataset_name: str,
         root: str,
         split: str = "train",
         force_reload: bool = False,
+        verbose: bool = False,
         use_pcst: bool = True,
         use_cwq: bool = True,
```
wsad1: Is this argument used somewhere?

torch_geometric/sampler/neighbor_sampler.py
```diff
@@ -402,7 +401,7 @@ def sample_from_nodes(
     ) -> Union[SamplerOutput, HeteroSamplerOutput]:
         out = node_sample(inputs, self._sample)
         if self.subgraph_type == SubgraphType.bidirectional:
-            out = out.to_bidirectional(keep_orig_edges=self.keep_orig_edges)
+            out = out.to_bidirectional()
         return out
```
akihironitta: Can you remind us of why this is being removed?

akihironitta (commented):
Commented below, but it looks like this PR has a breaking change (which breaks our internal tests).

torch_geometric/sampler/neighbor_sampler.py
puririshi98: good catch, deep apologies it seems this was the outcome of a typo during merge conflict alignment.

torch_geometric/datasets/web_qsp_dataset.py
puririshi98: https://github.com/pyg-team/pytorch_geometric/blob/0ca289e125cf669dd7c00df789f1f43e050f3fd7/torch_geometric/datasets/web_qsp_dataset.py#L185-L195

torch_geometric/utils/rag/graph_store.py
puririshi98: @zaristei is the one who built this, quoting him
"For the first one, it needs to be in order to initialize the sampler properly, its an artifact of the fact that we don't fully separate fs and gs yet
for the second one, yes but also no. IIRC the sampler is usually coupled with the graph store thats utilizing it"

I think if there are concerns about the design of this @wsad1 could you file an issue and tag me and @zaristei and we can make sure to make a follow up PR for this. For now its functional and driving value growth across nvidia and our fortune 500 customers.

torch_geometric/datasets/web_qsp_dataset.py
wsad1: My bad should have been more specific. Iam talking about `use_cwq`.

torch_geometric/utils/rag/backend_utils.py
```diff
@@ -0,0 +1,444 @@
...
+    GraphStore,
+    LargeGraphIndexer,
+    TripletLike,
+)
+from torch_geometric.data.large_graph_indexer import EDGE_RELATION
+from torch_geometric.distributed import (
+    LocalFeatureStore,
+    LocalGraphStore,
+    Partitioner,
+)
```
akihironitta, line 37: I am confused about the fact that #10411 has suggested that we deprecate `torch_geometric.distributed`, but then this PR starts using the deprecated subpackage.

akihironitta (commented):
I've been trying to review this PR, but since this PR contains many lines of code change, it's forcing me to spend much more effort than necessary, and sorry I don't think I will have enough time til Tuesday which is when you'd like to merge this PR by.

torch_geometric/utils/rag/backend_utils.py
puririshi98: For any normal GNN workflow (not RAG) torch.dist has seen issues for all of our customers and never worked when i tested it originally myself. They also don't have support for this anymore. So we recommend deprecation of it in light of cuGraph which has support.

However, since these 3 modules are being **temporarily** used for this RAG workflow.
As such,  I left the deprecation warning off of these 3 modules in https://github.com/pyg-team/pytorch_geometric/pull/10411.

Note that i say temporarily because NVIDIA is working on an accelerate cuda version of this which will seemlesly be integrated into `RAGQueryLoader` as a backend that we will have on as default otherwise fallback to the torch.dist option but raise a warning that they should use CuGraph for speed and scale. That is the longterm vision and this is a placeholder until then.

puririshi98:
closing in light of https://github.com/pyg-team/pytorch_geometric/pull/10436

PR #9998, opened 2025-01-31
Fixing the URL for `InstructMol` dataset.

drivanov:
The previously used https://huggingface.co/datasets/OpenMol/PubChemSFT/blob/main/all_clean.json is no longer working but a functional alternative can be found on the [webpage](https://huggingface.co/datasets/OpenMol/PubChemSFT/blob/main/all_clean.json) under `Copy download link`: https://huggingface.co/datasets/OpenMol/PubChemSFT/resolve/main/all_clean.json


puririshi98 (approved):
thanks this is a clean safe fix.

PR #10002, opened 2025-02-05
Make computation of mesh laplacian more efficient

minnerbe:
This PR slightly changes the computation of the mesh laplacian to be more efficient. In essence, the introduced changes are:
- Remove the redundant computation of the triangle area. The cross product $|e_{ij} \times e_{jk}|$ computes (twice) the area of a triangle for all pairs of edges, but was computed 6 times for all triangles (3x for the stiffness, 3x for the mass).
- Aggregate triangle areas over faces instead of edges.
- Slightly optimize the number of operations in some expressions.

All tests pass. In addition, I tested the validity of the changes with random triangulations with 1e2-1e5 nodes while benchmarking. The corresponding benchmarking script is not included in this PR, but can be found [here](https://github.com/minnerbe/pytorch_geometric/blob/test_changes/test_changes.py). Since this doesn't change the observable behavior of the code, I did not include a line in the changelog.

The performance improvements vary across architectures and with the number of nodes of the triangulation. Below, you can find the speedup for a given architecture (rows) for different normalization methods (columns) with different numbers of nodes in the mesh (tables). This speedup is of course anecdotal, but the fact that there is speedup seems pretty consistent.

```
N=1e2             None    sym    rw
x86_64            1.29x   2.00x  2.02x
aarch64           1.37x   2.16x  2.21x
cuda              1.32x   2.07x  2.07x

N=1e4             None    sym    rw
x86_64            1.19x   2.24x  2.13x
aarch64           1.08x   1.99x  1.96x
cuda              1.28x   2.06x  2.06x

N=1e5             None    sym    rw
x86_64            1.08x   1.75x  1.83x
aarch64           1.03x   1.91x  1.95x
cuda              2.24x   2.95x  2.98x
```

### DIFF
--- a/torch_geometric/utils/mesh_laplacian.py
+++ b/torch_geometric/utils/mesh_laplacian.py
@@ -59,7 +59,7 @@ def get_mesh_laplacian(
     # 1. Compute edge vectors of edges opposite to each vertex
     e_21 = pos[face[2]] - pos[face[1]]
     e_02 = pos[face[0]] - pos[face[2]]
-    e_10 = - (e_21 + e_02) # this is equal to pos[face[1]] - pos[face[0]]
+    e_10 = -(e_21 + e_02)  # this is equal to pos[face[1]] - pos[face[0]]
 
     # 2. Compute areas of all triangles
     area_times_four =  2 * torch.norm(torch.cross(e_02, e_10, dim=1), dim=1)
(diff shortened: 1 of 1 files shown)


minnerbe:
Hi @wsad1! I know that this is probably not high up your priority list, but have you had a chance to look at these changes yet? Thanks for taking the time!

PR #10023, opened 2025-02-12
Removed `SAMPLED_OP` flag from `torch_geometric/typing.py`

pbielak:
The `SAMPLED_OP` flag was used some time ago in the softmax implementation (`torch_geometrics/utils/softmax.py`) and with commit: https://github.com/pyg-team/pytorch_geometric/commit/d7e03c71cda21d62bb465984fe9bf4488c6c6a5a its usage was removed. However, the flag is still present in the `torch_geometric/typing.py` file.


rusty1s:
I think we can keep it IMO. No strong pref.

pbielak:
Also IMO if we don't use a particular code fragment anymore, it should be removed ;)

PR #10039, opened 2025-02-17
Corrected license metadata

jamesmyatt:
See https://packaging.python.org/en/latest/specifications/pyproject-toml/#license


rusty1s (approved):

rusty1s:
Ty! Can we fix CI? :)

jamesmyatt:
Looks like flit has only just added support for the new metadata spec: https://github.com/pypa/flit/pull/715

PR #10049, opened 2025-02-19
add torch_geometric.nn.nlp to docs

puririshi98:
w/o this they dont appear in docs

image


akihironitta (approved):
You also need to edit docs/source I think

PR #10078, opened 2025-02-27
Add remote `openai` backend to `LLM`

Kh4L:


puririshi98 (changes requested):
this looks good at a high level but i would like to see it integrated into the G-retriever example with argparser flags, and see a succesfull run of both huggingface and openai backends to confirm correctness and reasonable results (with minimal difference between backends

puririshi98 (changes requested):
please also make it clear that when using the openai backend, it only support frozen LLM, while huggingface support LORA and full finetuning as wel.

It would also be cool if you could add support for LORA with the openai backend

puririshi98 (changes requested):
okay sounds good. LGTM id just ask that you run a simple test showing that the outputs are identical for the same frozen model wether its through NIMs or huggingface.

PR #10080, opened 2025-02-28
Make `datapipe.py` compatible with `torchdata==0.11` or later

drivanov:
With `torch 2.7`/`torchdata 0.11` releases, the `datapype.py` example fails 
```
python3 /workspace/examples/datapipe.py 
  File "/workspace/examples/datapipe.py", line 17, in <module>
    from torchdata.datapipes.iter import FileLister, FileOpener, IterDataPipe
ModuleNotFoundError: No module named 'torchdata.datapipes'
```
fails because some classes have been moved from `torchdata` to `torch` nd a few methods, including in_memory_cache, have been removed.

The current PR resolves this issue in that example.

**NOTES:**
Running this example with the option `--task mesh` was failing even with previous versions of  `torch`/`torchdata` packages 

- because 3991 out of 7981 are binary, not text files in the required "OFF" format;
- if the  `meshio` and `torch_claster` packages are not installed. 

Proposed changes address both these issues.


examples/datapipe.py
```diff
@@ -45,8 +70,12 @@ def __iter__(self):
 
         for path in self.dp:
             category = osp.basename(path).split('_')[0]
+            try:
+                mesh = meshio.read(path)
+            except Exception:
+                #print(f'Problem reading file "{path}", it will be skept')
```
Kh4L, line 76:
```suggestion
```

examples/datapipe.py
```diff
@@ -45,8 +70,11 @@ def __iter__(self):
 
         for path in self.dp:
             category = osp.basename(path).split('_')[0]
+            try:
+                mesh = meshio.read(path)
+            except Exception:
+                continue
```
akihironitta, line 76: Can we comment what exceptions happen in which case?

examples/datapipe.py
```diff
@@ -25,18 +32,36 @@ def molecule_datapipe() -> IterDataPipe:
     root_dir = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data')
     path = download_url(f'{url}/HIV.csv', root_dir)
 
-    datapipe = FileOpener([path])
-    datapipe = datapipe.parse_csv_as_dict()
+    datapipe = FileOpener([path], mode="rt")
+    # Convert CSV rows into dictionaries, skipping the header row
+    datapipe = datapipe.map(lambda file: (
+        dict(zip(["smiles", "activity", "HIV_active"], row))
+        for i, row in enumerate(csv.reader(file[1])) if i > 0 and row))
```
akihironitta, line 39: I'm not familiar with the current state of datapipes, but is this the recommended way?

examples/datapipe.py
```diff
@@ -25,18 +32,36 @@ def molecule_datapipe() -> IterDataPipe:
...
+        installed = True
+        for package in ['meshio', 'torch_cluster']:
+            if importlib.util.find_spec(package) is None:
+                installed = False
+                print(
+                    f"This example requires the package {package} to be installed."
+                )
+                print(f"Please run: 'pip install {package}'")
+        if not installed:
+            exit()
```
akihironitta, line 63: Let's raise an exception so that it exits with non-zero status

akihironitta: Can we put this at the top of this script? As a user, I'd like the script fail early if necessary packages are unavailable.

examples/datapipe.py
```diff
@@ -25,18 +32,36 @@ def molecule_datapipe() -> IterDataPipe:
...
+    # Convert CSV rows into dictionaries, skipping the header row
+    datapipe = datapipe.map(lambda file: (
+        dict(zip(["smiles", "activity", "HIV_active"], row))
+        for i, row in enumerate(csv.reader(file[1])) if i > 0 and row))
+
+    datapipe = IterableWrapper(chain.from_iterable(datapipe))
     datapipe = datapipe.parse_smiles(target_key='HIV_active')
-    datapipe = datapipe.in_memory_cache()  # Cache graph instances in-memory.
 
+    _, cached_datapipe = tee(datapipe)
```
akihironitta, line 44: How is it different from this?
```suggestion
    cached_datapipe = tee(datapipe, 1)
```

Also, can you remind me of why we need `tee`?

examples/datapipe.py
drivanov: I got this recommendation from ChatGPT. I tried different options, this one is the best.

examples/datapipe.py
drivanov: 1. With the change you proposed, this example fails:
```
<itertools._tee object at 0x7f3512666b80>
Iterating over all data...
  File "/workspace/examples/datapipe.py", line 131, in <module>
    for batch in datapipe:
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/datapipes/_hook_iterator.py", line 204, in wrap_generator
    response = gen.send(None)
               ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/datapipes/iter/grouping.py", line 97, in __iter__
    yield self.wrapper_class(batch)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_geometric/data/batch.py", line 97, in from_data_list
    batch, slice_dict, inc_dict = collate(
                                  ^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_geometric/data/collate.py", line 56, in collate
    out = cls(_base_cls=data_list[0].__class__)  # type: ignore
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_geometric/data/batch.py", line 36, in __call__
    globals()[name] = MetaResolver(name, (cls, base_cls), {})
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: type 'itertools._tee' is not an acceptable base type
This exception is thrown by __iter__ of Batcher(batch_size=32, drop_last=False)
```

2. I tested both approaches:

(a) With `tee`:
```
    _, cached_datapipe = tee(datapipe)
    datapipe = IterableWrapper(cached_datapipe)
```

(b) Without `tee`:
```
    datapipe = IterableWrapper(datapipe)
```

for both `molecule` and `mesh` values of the `--task` option. 

- For `molecule`, there is no significant difference.
- For `mesh`, using version (a) results in:

```
Iterating over all data...
Done! [68.57s]
Iterating over all data a second time...
Done! [0.00s]
```
Whereas version (b) results in:
```
Iterating over all data...
Done! [68.50s]
Iterating over all data a second time...
Done! [67.46s]
```
Since version (a) significantly improves performance on the second iteration for `mesh`, using it for both cases would be more appropriate for consistency.

akihironitta: > With the change you proposed, this example fails:

Sorry for the typo! What I wanted to ask is, what the difference is between:

```python
    _, cached_datapipe = tee(datapipe)
```

and

```python
    cached_datapipe, = tee(datapipe, 1)
```

To me, these are exactly the same, but your way just creates an extra unnecesary object.

akihironitta: https://docs.python.org/3/library/itertools.html#itertools.tee

drivanov: Yes, it works. Thank you for the great suggestion!

akihironitta (approved):

PR #10090, opened 2025-03-03
Fix Mean Rank to Start from 1 in `KGEModel`

chest760:
Mean Rank (MR) metrics should always be ≥ 1, but the current implementation allows rank values to start from 0, leading to incorrect mean rank scores. While adding +1 to the calculated mean_rank produces the correct MR, this approach might be misleading for users.
If my concern is incorrect or if this fix is minor, please feel free to close this PR.
Thanks!


Kh4L (approved):
Good catch, LGTM!

PR #10093, opened 2025-03-04
Adding orthogonal convolution

bkiani:
Including code which implements the orthogonal convolution from https://arxiv.org/abs/2410.05499 natively in pytorch geometric.


bkiani:
I am unsure how to add to the changelog.md file since I do not know my pull request number. Some help in this would be appreciated.

PR #10108, opened 2025-03-11
Add tests for RAG Tests including RagQADatasets

zaristei:
This PR adds a new category of tests for RAG related datasets and networks, with a runner that evaluates them with a separate set of dependencies. Right now, these tests run for:
`test_web_qsp_dataset.py` (also included in this PR)
`test_sentence_transformer.py` (currently disabled unless the full suite is enabled).


zaristei:
Will handle skipped dependencies in separate PR.

zaristei:
I also ran the full suite on Ubuntu 22.04 CUDA 12.4 Python 10. Logs here: [pytest_output.log](https://github.com/user-attachments/files/19199046/pytest_output.log)

puririshi98 (changes requested):
needs llm and gretriever unit tests as well

puririshi98 (changes requested):
this LGTM now but ill ask if @akihironitta and @rusty1s have any concerns since they host the CI.

pyproject.toml
```diff
@@ -74,13 +82,15 @@ full = [
     "scikit-learn",
     "ase",
     "captum<0.7.0",
+    "datasets",
```
akihironitta, line 85: Let's not include the new dependencies in `full` just like `graphgym` dependencies

.github/workflows/rag_testing.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          files: |
+            benchmark/**
+            conda/**
+            docker/**
+            docs/**
+            examples/**
+            graphgym/**
+            README.md
+            CHANGELOG.md
```
akihironitta, line 38: We could be a bit more specific here (e.g. only run this workflow when related files like `nn.nlp` are modified)

test/datasets/test_web_qsp_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+    assert len(dataset_val) == 5
+    assert str(dataset_val) == "KGQABaseDataset(5)"
+
+    dataset_test = KGQABaseDataset(root=tmp_path, dataset_name="TestDataset",
+                                   split="test", use_pcst=False)
+    assert len(dataset_test) == 5
+    assert str(dataset_test) == "KGQABaseDataset(5)"
+
+    # TODO: More rigorous tests to validate that values are correct
+    # TODO: Proper tests for PCST and CWQ
```
akihironitta, line 201:
```suggestion
    # TODO(zaristei): More rigorous tests to validate that values are correct
    # TODO(zaristei): Proper tests for PCST and CWQ
```
https://google.github.io/styleguide/pyguide.html#312-todo-comments

pyproject.toml
zaristei: Ok, then should RAG tests not run when full tests do?

.github/workflows/rag_testing.yml
zaristei: Updated the file check to include all the existing rag tests. When future tests are added, we need to make sure the corresponding files are included on this filter to catch the tests.

pyproject.toml
zaristei: Updated so that rag tests are not set to run when full tests run.

.github/workflows/rag_testing.yml
```diff
@@ -0,0 +1,51 @@
...
+  rag_pytest:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 40
+
+      # Skip workflow if only certain files have been changed.
```
akihironitta, line 25:
```suggestion
      # Run workflow if only certain files have been changed.
```

pyproject.toml
```diff
@@ -191,6 +199,9 @@ filterwarnings = [
     # Filter `pytorch_lightning` warnings:
     "ignore:GPU available but not used:UserWarning",
 ]
+markers = [
+    "rag: mark test as RAG test",
+]
```
akihironitta, line 204: neat!

akihironitta (approved):
Neat! The change in CI and the new test cases look good to me :) (I should spend more time catching up on what you and Rishi have been working on to understand the full picture of the pipeline though)

pyproject.toml
akihironitta: Sorry for the delay. Looks good :) We can promote these dependencies to `full` whenever it makes sense in the future :)

puririshi98 (approved):
LGTM

PR #9018, opened 2024-03-05
GPSE Implementation

semihcanturk:
[Graph Positional and Structural Encoder](https://arxiv.org/abs/2307.07107) implementation as per  #8310. Adapted from the original repository: https://github.com/G-Taxonomy-Workgroup/GPSE. This version is a standalone implementation that is decoupled from GraphGym, and thus aims for better accessibility and a smoother integration into PyG. While the priority of this PR is to enable loading and using pre-trained models in plug-and-play fashion, it also includes the custom loss function used to train the model. Nevertheless, it might be easier to use the original repository for pre-training and fine-tuning new GPSE models for the time being.  

This PR includes the following:

- `GPSE`: The main GPSE module, that generates learned encodings for input graphs.
- Several helper classes (`FeatureEncoder`, `GNNStackStage`, `IdentityHead`, `GNNInductiveHybridMultiHead`, `ResGatedGCNConvGraphGymLayer`, `Linear`, `MLP`, `GeneralMultiLayer`, `GeneralLayer`, `BatchNorm1dNode`, `BatchNorm1dEdge`, `VirtualNodePatchSingleton`) and wrapper functions (`GNNPreMP`, `GNNLayer`), all adapted from their GraphGym versions for compatibility and enabling the loading of weights pre-trained using the GraphGym/original version.
- The class method `GPSE.from_pretrained()` that returns a model with pre-trained weights from the original repository/Zenodo files.
- `GPSENodeEncoder`, a helper linear/MLP encoder that takes the GPSE encodings precomputed as`batch.pestat_GPSE` in the input graphs, maps them to a desired dimension and appends them to node features.
- `precompute_GPSE` , a function that takes in a GPSE model and a dataset, and precomputes GPSE encodings in-place for a given dataset using the helper function `gpse_process_batch`.
- The transform `AddGPSE`, which in similar fashion to `AddLaplacianEigenvectorPE` and `AddRandomWalkPE` adds the GPSE encodings to a given graph using the helper function `gpse_process`
- The testing modules `test/test_gpse.py` and `test/test_add_gpse.py`.
- The loss function `gpse_loss` and helper functions `cosim_col_sep` and `process_batch_idx` used in GPSE training.
- A comprehensive example in `examples/gpse.py` is provided as a separate PR in #10118.

This PR has been tested with the latest (25.01-py3) NVIDIA stack, and has works without any issues.

PR #10118, opened 2025-03-14
GPSE example

semihcanturk:
Following #9018, Provides a comprehensive example in `examples/gpse.py` to compute GPSE encodings and use them for a graph regression task on the ZINC dataset. Two methods to compute GPSE encodings is demonstrated:

- Through `precompute_GPSE`: Given a PyG dataset, computes GPSE encodings in-place  _once_ before training, without saving them to storage. Ideal if you want to compute the encodings only once per run (unlike a dataset transform) but do not want to save the pre-transformed dataset to storage (unlike a PyG pre-transform).

To run with default pretrained weights (molpcba):
```
python examples/gpse.py --gpse
```
To run with pretrained weights from any other dataset, please provide the pretraining dataset name from the available options as a kwarg:
```
python examples/gpse.py --gpse geom
```

- Through the `AddGPSE` transform: A PyG transform analogous to [AddLaplacianEigenvectorPE](https://pytorch-geometric.readthedocs.io/en/2.6.0/generated/torch_geometric.transforms.AddLaplacianEigenvectorPE.html#torch_geometric.transforms.AddLaplacianEigenvectorPE) and [AddRandomWalkPE](https://pytorch-geometric.readthedocs.io/en/2.6.0/generated/torch_geometric.transforms.AddRandomWalkPE.html#torch_geometric.transforms.AddRandomWalkPE), can be used as a pre-transform or transform to a PyG dataset.
```
python examples/gpse.py --gpse --as_transform
```

Using as a transform is not recommended as recomputing them for every batch in every epoch is quite inefficient; using it as a pre-transform or through `precompute_GPSE` is suggested instead. In either case, the `torch_geometric.nn.GPSENodeEncoder` is then used to compute a mapping of the GPSE encodings to the desired dimension, and append them to `batch.x` to prepare them as inputs to a GNN.

This PR has been tested with the latest (25.01-py3) NVIDIA stack, and has works without any issues.


puririshi98 (approved):
LGTM, thanks!

PR #10126, opened 2025-03-19
Bi-directional Sampling for NeighborSampler

zaristei:
This PR makes the following changes:

- Modifies NeighborSampler so that sampling can occur from `src` to `dst` node, and not just from `dst` to `src` node, which is the current default behavior.
- Adds a new sampler called `BidirectionalNeighborSampler`, which samples nodes/edges in both upstream and downstream directions for a directed graph, as if it were an undirected graph.
- Add `merge_with` and `collate` methods to the SamplerOutput dataclass, allowing for different SamplerOutputs to be merged together. This also becomes useful in scenarios where the user may want to sample using multiple different unique sampling methods.

This PR will remain in draft until methods are also implemented on the corresponding Hetero graph classes, and unittests have been written to cover new behaviors.

### DIFF
--- a/test/sampler/test_sampler_neighbor_sampler.py
+++ b/test/sampler/test_sampler_neighbor_sampler.py
@@ -1,46 +1,46 @@
 import pytest
 import torch
-
+from torch_geometric.testing import onlyNeighborSampler, MyGraphStore, MyFeatureStore
 from torch_geometric.data import Data, HeteroData
 from torch_geometric.sampler.base import NodeSamplerInput, SamplerOutput
 from torch_geometric.sampler.neighbor_sampler import NeighborSampler
-from torch_geometric.testing import (
-    MyFeatureStore,
-    MyGraphStore,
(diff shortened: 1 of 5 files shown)


puririshi98:
closing since matthias said theres an existing simple way to do this. please jsut put the fix into my branch:
https://github.com/pyg-team/pytorch_geometric/pull/9992

puririshi98 (changes requested):
LGTM but i will request that @akihironitta or @rusty1s review this

zaristei:
https://github.com/pyg-team/pytorch_geometric/pull/10200 should be merged first

puririshi98:
@zaristei to help me fix CI

puririshi98:
turns out this PR already included fixes from the other PR and so when they merged it duplicated them. reverting the merge just for that file seemed to fix things since the changes from his older pr still exist in main, should be good to merge once the unrelated onnx issue is fixed.

PR #10129, opened 2025-03-20
Fix for configmixin with explicit classes

fedelopez77:
Fix for configmixin with explicit classes

### DIFF
--- a/torch_geometric/config_mixin.py
+++ b/torch_geometric/config_mixin.py
@@ -73,9 +73,9 @@ def _recursive_config(value: Any) -> Any:
         return value.config()
     if is_torch_instance(value, ConfigMixin):
         return value.config()
-    if isinstance(value, (tuple, list)) or isinstance(value, ModuleList):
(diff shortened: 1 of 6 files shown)


torch_geometric/config_mixin.py
```diff
@@ -71,9 +73,9 @@ def _recursive_config(value: Any) -> Any:
         return value.config()
     if is_torch_instance(value, ConfigMixin):
         return value.config()
-    if isinstance(value, (tuple, list, Iterable)):
+    if isinstance(value, (tuple, list)) or isinstance(value, ModuleList):
```
rusty1s, line 76:
```suggestion
    if isinstance(value, (tuple, list, ModuleList)):
```

rusty1s (approved):

torch_geometric/config_mixin.py
fedelopez77: Done

PR #10133, opened 2025-03-20
Updates get_data() in tech_qa.py example to handle JSON corpus data if present

rlratzel:
Updates get_data() in the `tech_qa.py` example to handle JSON corpus data if present.

Tested on corpus files from `earnings_consulting_doc` using a breakpoint to confirm `get_data` properly extracted the text from the `corpus/*.json` files, then on `techqa` and confirmed the `corpus/*.txt` files were read as they were before.


examples/llm/tech_qa.py
```diff
@@ -139,9 +139,22 @@ def get_data():
...
-            text_contexts.append(f.read())
+
+    # Read corpus data. Prefer *.json files, fall back to txt files.
+    file_paths = glob(f"corpus/*.json")
+    if len(file_paths) > 0:
+        for file_path in file_paths:
+            with open(file_path, "r+") as f:
+                data = json.load(f)
+            doc_type = data[0]["document_type"]
+            if doc_type != "text":
```
puririshi98, line 150: please add comments about what the other types are (pdf i assume? maybe others, would be good to add)
otherwise LGTM

puririshi98: and i assume you tested this on the json file examples we have?

PR #10135, opened 2025-03-21
Adds checkpointing for KG creation to `tech_qa.py`

rlratzel:
Based on code from [here](https://github.com/pyg-team/pytorch_geometric/blob/bb3992d7d488af576c948c267ebfe79b97227496/examples/llm/hotpot_qa.py#L32), this adds a `--checkpointing` option to the `tech_qa.py` script which saves the progress of KG creation at a set interval, allowing for the restoration of the KG in order to resume from a point close to where the script quit unexpectedly.

This has only been "lightly" tested using `pdb` to step through and inspect state. This still needs a proper end-to-end run with an interruption to verify the KG is properly restored and creation resumed from a checkpoint.

### DIFF
--- a/examples/llm/tech_qa.py
+++ b/examples/llm/tech_qa.py
@@ -87,10 +87,11 @@ def parse_args():
     parser.add_argument(
         '--llm_generator_mode', type=str, default=LLM_GEN_MODE_DEFAULT,
         choices=["frozen", "lora",
-                 "full"], help="Wether to freeze the Generator LLM,\
+                 "full"], help="Whether to freeze the Generator LLM,\
                         use LORA, or fully finetune")
     parser.add_argument('--dont_save_model', action="store_true",
-                        help="Wether to skip model saving.")
+                        help="Whether to skip model saving.")
+    parser.add_argument('--checkpointing', action="store_true")
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


puririshi98 (changes requested):
this is great, 2 asks:
1) can you mimick this in hotpot_qa.py replacing the existing checkpointing setup with something closer to this
2) can you test this out in hotpot_qa and tech_qa

rlratzel:
@puririshi98 is there any reason we wouldn't checkpoint KG creation by default?  It seems relatively cheap.  Perhaps then we'd add an option to skip checkpointing (but I'd be tempted to wait and see if that's even necessary).

rlratzel:
Chatted with @puririshi98 offline:
To test, we'll run this on hotpot_qa and tech_qa using a subset of the original corpus to allow for much faster completion. These runs will be done without a checkpoint/restore to establish an accuracy baseline, then again with a checkpoint/restore to verify the accuracy compares to the baseline as expected.

puririshi98:
> @puririshi98 is there any reason we wouldn't checkpoint KG creation by default? It seems relatively cheap. Perhaps then we'd add an option to skip checkpointing instead (but I'd be tempted to wait and see if that's even necessary).

agreed

PR #10139, opened 2025-03-25
Create some Sanity tests for default NeighborSampler behavior

zaristei:
While working on [this PR](https://github.com/pyg-team/pytorch_geometric/pull/10126), I noticed that there are no sanity tests for the default NeighborSampler. Although NeigborLoader effectively tests a lot of this functionality, having lower level tests helps with ensuring consistency for each `sample` call of the sampler, without the batching logic of `NeighborLoader` getting involved.

### DIFF
--- a/torch_geometric/data/graph_store.py
+++ b/torch_geometric/data/graph_store.py
@@ -261,7 +261,8 @@ class GraphStore(ABC):
                 col = ptr2index(col)
 
             if attr.layout != EdgeLayout.CSR:  # COO->CSR
-                num_rows = attr.size[0] if attr.size is not None else int(row.max()) + 1
+                num_rows = attr.size[0] if attr.size is not None else int(
+                    row.max()) + 1
                 row, perm = index_sort(row, max_value=num_rows)
                 col = col[perm]
                 row = index2ptr(row, num_rows)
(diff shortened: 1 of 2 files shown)


torch_geometric/data/graph_store.py
```diff
@@ -261,7 +261,8 @@ def _edge_to_layout(
                 col = ptr2index(col)
 
             if attr.layout != EdgeLayout.CSR:  # COO->CSR
-                num_rows = attr.size[0] if attr.size else int(row.max()) + 1
+                num_rows = attr.size[0] if attr.size is not None else int(
```
puririshi98, line 264: why this change?

puririshi98 (changes requested):

torch_geometric/data/graph_store.py
zaristei: Without the change, this line throws `RuntimeError: bool value of Tensor with more than one value is ambiguous`. I suspect no one has been running or testing the `COO->CSR` transformation on `GraphStore` before I stumbled across this.

puririshi98: good find

puririshi98 (changes requested):
LGTM

puririshi98 (approved):
LGTM low risk changes adding extra reliability for future changes

Issue #283, opened 2019-05-07
Adding MeshCNN to PyTorch Geometric


ranahanocka:
It would be so awesome if you could add MeshCNN to PyTorch Geometric. We have just released the [code](https://github.com/ranahanocka/MeshCNN) (PyTorch)

rusty1s:
Thanks for letting me know. I see what I can do, although I would appreciate a pull request if you find the time.

jlevy44:
I second this! Nice repo @ranahanocka !

If I had more time, I'd help out!

omria503:
Hi,
I'm trying to implement MeshCNN with PyTorch Geometric. 
Since I'm new with PyTorch Geometric and the basic stuff of graphs I would appreciate your help and comments.
I would like to implement the first layer which is the mesh convolution:
According to the [MeshCNN paper](https://arxiv.org/pdf/1809.05910.pdf):
![image](https://user-images.githubusercontent.com/60508587/101246053-7d402d00-3719-11eb-829f-6df91d94ea48.png)

One important note is that each edge holds a feature vector, not each node, therefore we can think of it as the edges are the nodes in a dual graph:
![image](https://user-images.githubusercontent.com/60508587/101246143-1a02ca80-371a-11eb-8201-517249ed4df4.png)

I've planned to implement it as the following:
edge_index - a 2d Tensor with shape of 2xE, it holds all the edge connections to each other.
In the example above: [[0, 0, 0, 0, 0, 1, 2, 3, 4, ...], [0, 1, 2, 3, 4, 0, 0, 0, 0, ...]]
x is a tensor with all the features for all the connections.
In the message function I've created the actual features (e, |a-c|, a+c, |b-d|, b+d) and used nn.Linear layer instead of convolution layer. Next, an aggregation function applied - "add". Last, the update function will return the aggregation output.

Do you think this is the correct way (using linear layer + aggregation "add" instead of convolution)?

Thanks

rusty1s:
Yes, `Linear` layer and "sum" aggregation is the way to go to implement MeshCNN IMO. However, as far as I can see, edge indices needs to be sorted to guarantee correct computation of `e_1`, `e_2`, ...

Omri-L:
Hi,
I am planning to send a pull request in one of the next weeks for [MeshCNN ](https://arxiv.org/pdf/1809.05910.pdf)layers.
In order to make it easier for me, I will appreciate if you could advise about the following modules:

[...]

PR #10153, opened 2025-03-31
feat(transforms): add `mesh_to_dual_graph` transformation

michaelfortunato:
This commit introduces a new transformation that converts a mesh to its "dual".

The dual of a mesh was used implicitly in
[MeshCNN](https://arxiv.org/abs/1809.05910), and referenced in [PD-MeshNet](https://arxiv.org/abs/2010.12455).

An implementation for this transform can be found in [pymesh](https://pymesh.readthedocs.io/en/latest/api_misc.html?highlight=dual#pymesh.mesh_to_dual_graph).

The data for this simple test, including the output dual tensor we validate against, was constructed using Blender, which also provides a mesh_to_dual transform (can be found as a geometry node).

So, we may want to keep this transform outside of pytorch_geometric and depend on `pymesh`, or include it pytorch_geometric! I see arguments for both sides.

Let me know what you all think!
Best,
Michael


michaelfortunato:
In fact the dual use in MeshCNN has already been noted here #283.

Issue #9112, opened 2024-03-26
[Roadmap] Heterogeneous Graphs Explainability Support


Mellandd:
### 🚀 The feature, motivation and pitch

Explainability is a key feature of GNNs, which is already implemented in PyG. However, of all the features introduced, only a few have been adapted to heterogeneous graphs.

**Algorithms**: Of all the algorithms implemented for explainability, only the Captum algorithm is compatible with heterogeneous graphs. It would be interesting to adapt other specific algorithms for graphs such as ``GNNExplainer`` or ``PGExplainer``, and other algorithms such as ``AttentionExplainer``.

Moreover, the algorithms adapted by PyG could be extended with new algorithms that have been published over the years (for example, see this [survey](https://arxiv.org/abs/2306.01958)), but this isn't just for heterogeneous graphs. Maybe in the future, we could work on creating new algorithms simultaneously for heterogeneous graphs, without this gap.

- [x] Adaptation of ``GNNExplainer`` for heterogeneous graphs.
- [x] Adaptation of ``PGExplainer`` for heterogeneous graphs. 
- [ ] Adaptation of ``AttentionExplainer`` for heterogeneous graphs. 
- [ ] Implementation of new explainability algorithms. System to allow adaptation to heterogeneous graphs without additional work.

**Features**: Some features available in the explanations of homogeneous GNNs are missing in heterogeneous GNNs. For example, the ``visualize_graph`` method of ``Explanation`` is not available for ``HeteroExplanation``. This right now can be done with the ``get_explanation_subgraph`` method, and generating the plot by hand with NetworkX, but it would be nice to implement it to do it automatically.

- [ ] Implementation of missing features for HeteroExplanation, such as ``visualize_graph``.

**Metrics**: Currently, the available metrics such as Fidelity or Faithfulness are only available for homogeneous graphs, but those metrics could be adapted for heterogeneous graphs. To continue the work of #5628, we could think implementing new metrics for all kind of graphs, such as sparsity or stability (e.g. see https://arxiv.org/pdf/2012.15445.pdf)

- [ ] Adaptation of fidelity, faithfulness and characterization score for heterogeneous graphs.
- [ ] Sparsity metric for all graphs.
- [ ] Stability metric for all graphs.

rachitk:
[...]

PR #10158, opened 2025-04-02
Support hetero data in `GNNExplainer`

XinweiHe:


test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -291,3 +291,61 @@ def test_gnn_explainer_attentive_fp(check_explanation):
...
+        model_config=model_config,
+    )
+
+    explanation = explainer(
+        hetero_data.x_dict,
+        hetero_data.edge_index_dict,
+        target=target,
+        index=index,
+    )
+
```
wsad1, line 350:
```suggestion
    assert isinstance(explanation, HeteroExplanation)
```

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -291,3 +291,61 @@ def test_gnn_explainer_attentive_fp(check_explanation):
...
+    explanation.validate(raise_on_error=True)
```
wsad1, line 351: Additionally add a check to see `node_mask` for different node and `edge_mask` for differents edges is not `None` . 
Further call other methods in explanation like

test/explain/algorithm/test_gnn_explainer.py
```diff
@@ -291,3 +291,61 @@ def test_gnn_explainer_attentive_fp(check_explanation):
...
+        explanation_type=explanation_type,
+        node_mask_type=node_mask_type,
+        edge_mask_type=edge_mask_type,
+        model_config=model_config,
+    )
+
+    explanation = explainer(
+        hetero_data.x_dict,
+        hetero_data.edge_index_dict,
+        target=target,
```
wsad1, line 347: Can we test without passing `target` or for hetero do we always need a target?

wsad1 (approved):
Nice work.
Ok to merge once we add more tests.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -184,40 +335,135 @@ def _initialize_masks(self, x: Tensor, edge_index: Tensor):
         else:
             assert False
 
+    def _collect_gradients(self):
+        if self.is_hetero:
+            # Collect gradients for heterogeneous graph
+            self._collect_hetero_gradients()
+        else:
+            # Collect gradients for homogeneous graph
+            self._collect_homo_gradients()
```
akihironitta, line 344: I'm in favor of not adding redundant comments that don't add any info:
```suggestion
        if self.is_hetero:
            self._collect_hetero_gradients()
        else:
            self._collect_homo_gradients()
```

akihironitta (commented):
I think, ideally, we can create three new internal classes, (one for base class, and one for each of homogenous and heterogeneous cases), and write these actual implementations in these two subclasses. This isn't blocking given the timeline though :)

test/explain/algorithm/test_gnn_explainer.py
XinweiHe: `target is None` lots of the times. So this is good.

torch_geometric/explain/algorithm/gnn_explainer.py
```diff
@@ -69,101 +75,245 @@ def __init__(self, epochs: int = 100, lr: float = 0.01, **kwargs):
...
         self.edge_mask = self.hard_edge_mask = None
+        self.is_hetero = False
 
     def forward(
         self,
         model: torch.nn.Module,
-        x: Tensor,
-        edge_index: Tensor,
+        x: Union[Tensor, Dict[NodeType, Tensor]],
+        edge_index: Union[Tensor, Dict[EdgeType, Tensor]],
```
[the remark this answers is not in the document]
XinweiHe: ok

XinweiHe:
> I think, ideally, we can create three new internal classes, (one for base class, and one for each of homogenous and heterogeneous cases), and write these actual implementations in these two subclasses.

will do in a follow-up.

test/explain/conftest.py
```diff
@@ -52,14 +52,36 @@ def forward(self, x, edge_index):
...
+
+        self.lin = torch.nn.Linear(32, out_channels)
 
     def forward(self, x_dict, edge_index_dict,
                 additonal_arg=None) -> torch.Tensor:
-        return self.lin(self.graph_sage(x_dict, edge_index_dict)['paper'])
+        x = self.lin(self.graph_sage(x_dict, edge_index_dict)['paper'])
+
+        # Apply transformations based on model_config if available
+        if hasattr(self, 'model_config') and self.model_config:
```
akihironitta, line 78: When is `hasattr(self, 'model_config')` `False`?

test/explain/conftest.py
```diff
@@ -92,3 +114,44 @@ def _check_explanation(
...
+
+
+@pytest.fixture()
+def check_explanation_hetero():
+    def _check_explanation_hetero(
+        explanation: HeteroExplanation,
+        node_mask_type: Optional[MaskType],
+        edge_mask_type: Optional[MaskType],
+        hetero_data: HeteroData,
+    ):
```
akihironitta, line 131: Why are we defining it as a pytest fixture? It can just be a normal function in the test file

torch_geometric/explain/algorithm/gnn_explainer.py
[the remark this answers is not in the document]
akihironitta: neat!

akihironitta (approved):

test/explain/conftest.py
XinweiHe: removed

test/explain/conftest.py
XinweiHe: This can be used for other heterogeneous graph support for explainers e.g. `PGExplainer` and `AttentionExplainer`.

rutujagurav:
In what version of pyg is/would be the GNNExplainer support for hetero graphs available?

benstear:
Im also curious what version of pyg the GNNExplainer for heteroData will be available.

akihironitta:
@rutujagurav @benstear The change in this PR is included in 2.7.0.

PR #10164, opened 2025-04-03
Bidirectional Sampler to fix sampling edge cases

zaristei:
Implementation of homogenous bidirectional neighborsampler to allow for sampling from 'leaf' nodes in directed graphs.
Heterogeneous still to be implemented before merging into main.


puririshi98:
seems theres an easier fix as advised by matthias, closing for now, go with that

PR #10166, opened 2025-04-03
Add max_seq_length in SentenceTransformer

Kh4L:


torch_geometric/nn/nlp/sentence_transformer.py
```diff
@@ -32,6 +32,8 @@ def __init__(
         if self.tokenizer.pad_token is None:
             self.tokenizer.pad_token = self.tokenizer.eos_token
 
+        self.max_seq_length = self.model.config.max_position_embeddings
```
puririshi98, line 35: can you check how many tokens this comes out to for ModernBert or Roberta and add that as a comment?

puririshi98 (changes requested):

PR #10168, opened 2025-04-05
Added support for heterogenous graphs in `PGExplainer`

XinweiHe:
Note: This version of heterogenous data support use global aggregated node embedding to generate the explanation for each heterogeneous edge type. We could potentially support edge specific embedding as an option in the future as necessary.


wsad1:
Nice work. 
Can we test for [HGTConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HGTConv.html#torch_geometric.nn.conv.HGTConv) and [HanConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HANConv.html#torch_geometric.nn.conv.HANConv), to ensure that it works for Hetero GNNs not created by the to_hetero operation.

wsad1:
> I think those are not supported at the moment. Will support them soon.

Will it be part of this PR?

XinweiHe:
> Will it be part of this PR?

Yes.

wsad1 (approved):

PR #10169, opened 2025-04-06
Added support for heterogenous graphs in `AttentionExplainer`

XinweiHe:


wsad1 (commented):
Nice work. 
Can we test for [HGTConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HGTConv.html#torch_geometric.nn.conv.HGTConv) and [HanConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HANConv.html#torch_geometric.nn.conv.HANConv), to ensure that it works for Hetero GNNs not created by the to_hetero operation.

XinweiHe:
> Can we test for [HGTConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HGTConv.html#torch_geometric.nn.conv.HGTConv) and [HanConv](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.HANConv.html#torch_geometric.nn.conv.HANConv), to ensure that it works for Hetero GNNs not created by the to_hetero operation.

`HGTConv` and `HanConv` are not relevant here, because `AttentionExplainer` only works with GNN model with attentions, in particular, `GATConv`, `GATv2Conv` or `TransformerConv`.

We will support `HeteroConv` with these attention GNN models in this PR though.

wsad1:
> HGTConv and HanConv are not relevant here, because AttentionExplainer only works with GNN model with attentions, in particular, GATConv, GATv2Conv or TransformerConv.

`HanConv` and `HGTConv` use attention while message passing.

XinweiHe:
> HanConv and HGTConv use attention while message passing.

You are right but at the moment, attention explainer is only designed to handle the case where `alpha` exists as an input parameter in `def message(...)` function. Since this is not the setup for `HanConv` and `HGTConv` at the moment we would need to spend extra effort supporting them which would probably be the best to not mix with this initial heterogeneous graph support for attention explainer.

XinweiHe:
could we get this effort merge and create a follow-up ticket to handle the `HanConv` and `HGTConv` cases?

wsad1 (approved):
Looks good. 
Lets have a quick follow up for `HANConv` and other hetro GNNs.

PR #10200, opened 2025-04-16
Enable getting global indices from SamplerOutput

zaristei:
The current implementation of SamplerOutput returns local indices of node ids in the `node` list for the `row`,`col`, `batch`, `orig_row`, and `orig_col` fields. This leads to some confusion for using the edges/seed nodes retrieved from the sampler. This PR adds some helper methods that will derive the global indices for these fields given the information already stored within the SamplerOutput.
Unittests have been also written for both homogeneous and heterogeneous outputs.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -6,13 +6,13 @@ inputs:
     default: '3.9'
   torch-version:
     required: false
-    default: '2.7.0'
+    default: '2.8.0'
   cuda-version:
     required: false
     default: cpu
   torchvision-version:
     required: false
-    default: '0.22.0'
+    default: '0.23.0'
   full_install:
     required: false
     default: true
@@ -22,7 +22,7 @@ runs:
 
   steps:
     - name: Install uv
-      uses: astral-sh/setup-uv@v5
+      uses: astral-sh/setup-uv@v6
       with:
         python-version: ${{ inputs.python-version }}
         activate-environment: true
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,9 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions/setup"
     schedule:
       interval: "daily"
       time: "00:00"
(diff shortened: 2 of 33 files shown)


Kh4L (approved):
lgtm!

torch_geometric/sampler/utils.py
```diff
@@ -160,3 +160,48 @@ def remap_keys(
         k if k in exclude else mapping.get(k, k): v
         for k, v in inputs.items()
     }
+
+
+def local_to_global_node_idx(node_values: Tensor,
```
rusty1s, line 165: I don't think we need a helper function to do `node_values[local_indices]`

torch_geometric/sampler/utils.py
```diff
@@ -160,3 +160,48 @@ def remap_keys(
...
+        local_indices (Tensor): The local indices. (num_indices)
+
+    Returns:
+        Tensor: The values of the node_values tensor at the local indices.
+        (num_indices, feature_dim)
+    """
+    return torch.index_select(node_values, dim=0, index=local_indices)
+
+
+def global_to_local_node_idx(node_values: Tensor,
```
rusty1s, line 181: When would that be useful?

rusty1s (commented):
What's the purpose?

zaristei:
The purpose of these changes is to make it easier for users of `SamplerOutput` to access the global row/col ids, as well as global seed node ids from the sampler. Although the docstring indicates that the `row` and `col` fields are indices of nodes, it's not immediately obvious from that documentation that one can get the global indices by indexing into the nodes list, or that the same can be done for to get seed nodes from `batch`, for instance.

torch_geometric/sampler/utils.py
zaristei: In the downstream PR, I create a sampler that needs to merge together more than one SamplerOutput. This helper is useful for doing operations of that sort, while also being a exact reverse mapping of `local_to_global_node_idx`.

torch_geometric/sampler/utils.py
zaristei: I think its helpful to refactor as a separate function, because otherwise it's not immediately obvious to readers of the code that global indexing is functionally equivalent to locally indexing into the node field.

torch_geometric/sampler/utils.py
zaristei: [Link for reference](https://github.com/pyg-team/pytorch_geometric/pull/10126/files#diff-025dec2aa82ccd8573d9e0dc3ddda336a23edbb5454f5e6861236ecababf4f2cR267)

puririshi98 (approved):
merging as per discussions with @rusty1s , these are heavily tested through NVIDIA already

PR #10208, opened 2025-04-18
[Docs] Grammatical fixes in compile.rst

JeS24:
- Grammatical fixes at some places.
- Changed the first occurrence of "graph break" to "computational graph break" for clarity.


akihironitta:
Thank you for sending the fix! :pray:

JeS24:
@akihironitta Thanks for the merge.

> People usually call them graph breaks

True. I debated that edit for a while. My intention was to clearly differentiate between 'computational graphs' and 'combinatorial graphs', or graph datasets, since both appear in the context of PyG, even though they are fundamentally different objects. Just to avoid any possible confusion.

PR #10213, opened 2025-04-21
LLM: format prompt with hf prompt format if available

Kh4L:


torch_geometric/nn/nlp/llm.py
```diff
@@ -108,6 +108,28 @@ def __init__(
...
+                add_generation_prompt=False,
+            )
+            dummy_text_with_gen_prompt = self.tokenizer.apply_chat_template(
+                dummy_message,
+                tokenize=False,
+                add_generation_prompt=True,
+            )
+            self.gen_prompt = dummy_text_with_gen_prompt[len(dummy_text):]
+        else:
+            self.gen_prompt = None
```
puririshi98, line 131: LGTM in general, one question
in the case where gen_prompt is None since we do not have a template from huggingface, 
would we want to use something simple like we had before?
`prompt_template = """Answer this question based on retrieved contexts. Just give the answer without explanation.
`

Kh4L: woops, gen_prompt is a misnomer here, its just wether the tokenizer should append  the "assistant header" tokens to the or not, which I was planning to use

will replace that code

PR #10229, opened 2025-04-25
Add VectorRAG

Kh4L:
This PR decouples VectorRAG in a different class, offering with interface for non-graph based RAG + add DocumentRetriever VectorRAG.

### DIFF
--- a/examples/llm/txt2kg_rag.py
+++ b/examples/llm/txt2kg_rag.py
@@ -379,7 +379,7 @@ def train(args, data_lists):
                     # insert VectorRAG context
                     new_qs.append(
                         prompt_template.format(question=q,
-                                               context=batch.text_context[i]))
+                                               context= "\n".join(batch.text_context[i])))
                 batch.question = new_qs
 
                 optimizer.zero_grad()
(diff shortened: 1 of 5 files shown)


examples/llm/tech_qa.py
```diff
@@ -285,8 +290,7 @@ def make_dataset(args):
             data=(fs, gs), seed_nodes_kwargs={"k_nodes": knn_neighsample_bs},
             sampler_kwargs={"num_neighbors": [fanout] * num_hops},
             local_filter=make_pcst_filter(triples, model),
-            local_filter_kwargs=local_filter_kwargs, raw_docs=context_docs,
-            embedded_docs=embedded_docs, k_for_docs=args.k_for_docs)
+            local_filter_kwargs=local_filter_kwargs, vector_rag=vector_rag)
```
puririshi98, line 293: can we call this document_retriever instead of vector_rag since technically vector_rag would be vector retrieval augmented generation but this is just the retriever part, the generator is seperated in our pipeline.

puririshi98 (changes requested):
LGTM other than this minor thing

examples/llm/tech_qa.py
Kh4L: good point, i'd go for `vector_retriever` and i'd replace `local_filter` with `graph_retriever`

wdyt?

Kh4L: sure, i like `subgraph_filter`,  I don't have strong opinions when it comes to naming those 

yeah `graph_data` sg as well.. tbh PyG calling graphs `Data`/`data` is so confusing

PR #10230, opened 2025-04-26
Add Relational Deep Learning example

BlazStojanovic:
Uses `relbench` and `torchframe` to demonstrate relational deep learning end to end.


vid-koci (approved):
Approving. Any comments that I've had I've just directly changed in the code.

examples/rdl.py
```diff
@@ -0,0 +1,711 @@
...
+elif known_args.dataset == "rel-event":
+    args.add_argument(
+        "--task", type=str, default="user-repeat",
+        choices=["user-repeat", "user-ignore", "user-attendance"])
+elif known_args.dataset == "rel-avito":
+    args.add_argument("--task", type=str, default="user-visits",
+                      choices=["user-visits", "user-clicks", "ad-ctr"])
+
+# Data loader arguments
+args.add_argument("--time_attr", type=str, default="time")
```
wsad1, line 90: Is this always "time"? If it is can we drop it as an argument.

wsad1 (approved):
Nice work.

examples/rdl.py
BlazStojanovic: I think you're correct, removed the argument.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
+"""This is a demo of Relational Deep Learning code for RelBench dataset.
+For more details on RelBench and Relational Deep learning, please refer to
+https://arxiv.org/abs/2407.20060 and
+https://raw.githubusercontent.com/mlresearch/v235/main/assets/fey24a/fey24a.pdf
+This is NOT the official code for the experiments in these papers.
+To exactly reproduce the experimental results, from the linked papers, please
```
rusty1s, line 6:
```suggestion
To exactly reproduce the experimental results from the linked papers, please
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+see https://github.com/snap-stanford/relbench or
+https://github.com/kumo-ai/ContextGNN for link prediction.
```
rusty1s, line 8:
```suggestion
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+    args.add_argument(
+        "--task", type=str, default="study-outcome",
+        choices=["study-outcome", "study-adverse", "site-success"])
+elif known_args.dataset == "rel-event":
+    args.add_argument(
+        "--task", type=str, default="user-repeat",
+        choices=["user-repeat", "user-ignore", "user-attendance"])
+elif known_args.dataset == "rel-avito":
+    args.add_argument("--task", type=str, default="user-visits",
+                      choices=["user-visits", "user-clicks", "ad-ctr"])
```
rusty1s, line 87: This is both very clean and very rigor at the same time. IMO, we can just do `args.add_argument('--dataset')` and `args.add_argument('--task')`. We will fail later anyway on a mismatch between the two.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+
+# Data loader arguments
+args.add_argument("--batch_size", type=int, default=512)
+args.add_argument("--temporal_strategy", type=str, default="uniform",
+                  choices=["uniform", "last"])
+args.add_argument("--num_workers", type=int, default=0)
+args.add_argument("--persistent_workers", type=bool, default=False)
```
rusty1s, line 94: I think we can skip those and just set a reasonable default. Would prefer `num_workers > 0` and `persistent_workers=True`.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+
+# Model parameters
+args.add_argument("--num_neighbors", type=list, default=[128, 128])
+args.add_argument("--channels", type=int, default=128)
+args.add_argument("--out_channels", type=int, default=1)
```
rusty1s, line 99: No reason to set. It's task dependent.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+args.add_argument("--aggr", type=str, default="sum")
+args.add_argument("--norm", type=str, default="batch_norm")
+
+# Training parameters
+args.add_argument("--epochs", type=int, default=10)
+args.add_argument("--learning_rate", type=float, default=0.005)
```
rusty1s, line 105:
```suggestion
args.add_argument("--lr", type=float, default=0.005)
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+        default_stype_encoder_cls_kwargs (Dict[torch_frame.stype, Any]):
+            A dictionary mapping from :obj:`torch_frame.stype` object into a
+            tuple specifying :class:`torch_frame.nn.StypeEncoder` class and its
+            keyword arguments :obj:`kwargs`.
+    """
+    def __init__(
+        self,
+        channels: int,
+        node_to_col_names_dict: Dict[NodeType, Dict[torch_frame.stype,
+                                                    List[str]]],
```
rusty1s, line 156:
```suggestion
                                                    List[str]], ],
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+            be changed for different :class:`torch_frame_model_cls`.
+        default_stype_encoder_cls_kwargs (Dict[torch_frame.stype, Any]):
+            A dictionary mapping from :obj:`torch_frame.stype` object into a
+            tuple specifying :class:`torch_frame.nn.StypeEncoder` class and its
+            keyword arguments :obj:`kwargs`.
+    """
+    def __init__(
+        self,
+        channels: int,
+        node_to_col_names_dict: Dict[NodeType, Dict[torch_frame.stype,
```
rusty1s, line 155:
```suggestion
        col_names_dict: Dict[NodeType, Dict[torch_frame.stype,
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+                                                    List[str]]],
+        node_to_col_stats: Dict[NodeType, Dict[str, Dict[StatType, Any]]],
```
rusty1s, line 157:
```suggestion
        stats_dict: Dict[NodeType, Dict[str, Dict[StatType, Any]]],
```

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+        torch_frame_model_cls=ResNet,
+        torch_frame_model_kwargs: Dict[str, Any] = {
+            "channels": 128,
+            "num_layers": 4,
+        },
```
rusty1s, line 162: Usually, we don't want to instantiate mutable objects as part of init. It's probably okay for an example though.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+        default_stype_encoder_cls_kwargs: Dict[torch_frame.stype, Any] = {
+            torch_frame.categorical: (torch_frame.nn.EmbeddingEncoder, {}),
+            torch_frame.numerical: (torch_frame.nn.LinearEncoder, {}),
+            torch_frame.multicategorical: (
+                torch_frame.nn.MultiCategoricalEmbeddingEncoder,
+                {},
+            ),
+            torch_frame.embedding: (torch_frame.nn.LinearEmbeddingEncoder, {}),
+            torch_frame.timestamp: (torch_frame.nn.TimestampEncoder, {}),
+        },
```
rusty1s, line 172: Do we need to expose? Let's just set it.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+            Statistics of node features
+        num_layers (int): Number of GNN layers
+        channels (int): Hidden dimension size
+        out_channels (int): Output dimension size
+        aggr (str): Aggregation method for GNN
+        norm (str): Normalization method for MLP
+    """
+    def __init__(
+        self,
+        data: HeteroData,
```
rusty1s, line 404: Do we need to pass the data itself to the model? That's an anti-pattern.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+        higher_is_better = True
+    else:
+        raise ValueError(f"Unsupported task type: {task.task_type}")
+
+    return out_channels, loss_fn, tune_metric, higher_is_better
+
+
+def train(
+    model: Model,
+    loader_dict: Dict[str, NeighborLoader],
```
rusty1s, line 511: train_loader

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+
+    for split, table in [
+        ("train", train_table),
+        ("val", val_table),
+        ("test", test_table),
+    ]:
+        table_input = get_node_train_table_input(
+            table=table,
+            task=task,
+        )
```
rusty1s, line 616: Can we expose this logic here in the example? That's pretty crucial for the understanding.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+        val_metrics = task.evaluate(val_pred, val_table)
+        print(f"Epoch: {epoch:02d}, "
+              f"Train loss: {train_loss}, "
+              f"Val metrics: {val_metrics}")
+
+        if (higher_is_better and val_metrics[tune_metric] > best_val_metric
+            ) or (not higher_is_better
+                  and val_metrics[tune_metric] < best_val_metric):
```
rusty1s, line 686:
```suggestion
        if ((higher_is_better and val_metrics[tune_metric] > best_val_metric
            ) or (not higher_is_better
                  and val_metrics[tune_metric] < best_val_metric)):
```
for nicer formatting.

examples/rdl.py
```diff
@@ -0,0 +1,710 @@
...
+    print("Loading the best model state dictionary...")
+    model.load_state_dict(state_dict)
+    val_pred = test(
+        loader=loader_dict["val"],
+        model=model,
+        task=task,
+    )
+
+    val_metrics = task.evaluate(val_pred, val_table)
+    print(f"Validation metrics: {val_metrics}")
```
rusty1s, line 701: No need to run again?

BlazStojanovic:
@rusty1s Incorporated your comments

The example now includes the logic for going from task -> input to the model, and the only heavy lifting logic still borrowed from RelBench is:
```
    data, col_stats_dict = make_pkey_fkey_graph(
        ...
    )
```

Should we try to move this logic into the example as well? Might be tricker to do and personally I'm okay with this utility being imported from relbench.

examples/rdl.py
```diff
@@ -0,0 +1,773 @@
...
+        )
+        val_metrics = task.evaluate(val_pred, val_table)
+        print(f"Epoch: {epoch:02d}, "
+              f"Train loss: {train_loss}, "
+              f"Val metrics: {val_metrics}")
+
+        is_better_op = operator.gt if higher_is_better else operator.lt
+        if is_better_op(val_metrics[tune_metric], best_val_metric):
+            best_val_metric = val_metrics[tune_metric]
+            state_dict = copy.deepcopy(model.state_dict())
```
akihironitta, line 760: It doesn't matter much for this model size, but it'd be good to offload the state dict to CPU :)

akihironitta (approved):

examples/rdl.py
BlazStojanovic: ah, good point - feel free to add :)

puririshi98:
added a minor change to have a readme sentence about this example since i think its worth drawing attention to. feel free to modify as you all see fit

PR #10244, opened 2025-05-02
TXT2KG refactor v2 more

Kh4L:


examples/llm/txt2kg_rag.py
```diff
@@ -538,5 +583,8 @@ def eval(question: str, pred: str, correct_answer: str):
         sys.exit(1)
     print("Starting TechQA training with args: ", args)
     data_lists = make_dataset(args)
+    if os.path.exists("document_retriever.pt"):
```
puririshi98, line 586: not sure what the purpose of this is. if we regenerate dataset we'll be using the saved retriever inside the above function, 
```
        if os.path.exists("document_retriever.pt"):
            print("Loading document retriever from checkpoint...")
            vector_retriever = DocumentRetriever.load(
                "document_retriever.pt", model=model.encode,
                model_kwargs=model_kwargs)
            if args.k_for_docs != vector_retriever.k_for_docs:
                vector_retriever.k_for_docs = args.k_for_docs
        else:
            print("Creating document retriever...")
            vector_retriever = DocumentRetriever(context_docs,
                                                 k_for_docs=args.k_for_docs,
                                                 model=model.encode,
                                                 model_kwargs=model_kwargs)
            torch.save(vector_retriever, "document_retriever.pt")
            vector_retriever.save("document_retriever.pt")
```
and if we dont regen the dataset we have the saved tech_qa.pt, not sure why we want to to update the data lists

puririshi98 (changes requested):

examples/llm/txt2kg_rag.py
Kh4L: We want to update the dataset only if `k_for_docs` passed by the user is different from the one used to gen the dataset

just updated the PR maybe its a bit clearer now

Kh4L: > and if we dont regen the dataset we have the saved tech_qa.pt, not sure why we want to to update the data lists

when we want to experiement different k_for_docs (different vectorrag) without having to regen the whole graphrag

zaristei (approved):
LGTM so far

examples/llm/txt2kg_rag.py
puririshi98: okay got it

examples/llm/txt2kg_rag.py
```diff
@@ -536,7 +578,15 @@ def eval(question: str, pred: str, correct_answer: str):
         print("Error: wandb package not found but --wandb flag was used.")
         print("Please install wandb and rerun the script.")
         sys.exit(1)
+
     print("Starting TechQA training with args: ", args)
-    data_lists = make_dataset(args)
+    if os.path.exists("tech_qa.pt") and not args.regenerate_dataset:
+        print("Re-using Saved TechQA KG-RAG Dataset...")
+        data_lists = torch.load("tech_qa.pt", weights_only=False)
```
puririshi98, line 585: can we add a check that the vector retriever k doesnt match the current asked k fi we are going to update. if k is the same we dont need to. otherwise lgtm ill merge

puririshi98 (changes requested):

PR #10246, opened 2025-05-02
Remove `auto-skip-changelog` label

akihironitta:
It's confusing and hasn't been working properly. Now that we have `.github/workflows/auto-merge.yml`, we don't need this label at all.


rusty1s:
👍

PR #10281, opened 2025-05-17
Show available tasks if not provided in relbench example

puririshi98:
while adding this to our CI i noticed if you dont pass task it fails and then i had to go to https://github.com/snap-stanford/relbench/blob/main/relbench/tasks/__init__.py to find the options. some help strings would be good.
I am going to submit a followup PR soon

### DIFF
--- a/examples/rdl.py
+++ b/examples/rdl.py
@@ -576,7 +576,9 @@ def main():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument("--dataset", type=str, default="rel-f1",
                         choices=get_dataset_names())
-    parser.add_argument("--task", type=str, required=True)
+    parser.add_argument("--task", type=str, required=True,
+                        help="As an example, possible tasks for 'rel-f1'\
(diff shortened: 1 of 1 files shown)


examples/rdl.py
```diff
@@ -576,7 +576,9 @@ def main():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument("--dataset", type=str, default="rel-f1",
                         choices=get_dataset_names())
-    parser.add_argument("--task", type=str, required=True)
+    parser.add_argument("--task", type=str, required=True,
+                        help="As an example, possible tasks for 'rel-f1'\
+                        are ['driver-position', 'driver-dnf', 'driver-top3'])
```
BlazStojanovic, line 581: Maybe just link where people can find all the options

BlazStojanovic (approved):

akihironitta (approved):
Meant to set it to `None` so that users will always see the available tasks in the check below.

akihironitta:
@puririshi98 Thanks for setting up CI to test this on your side! Feel free to merge this :)

PR #10290, opened 2025-05-27
auto download techQA if no data avail

puririshi98:
rick/ralph to test e2e run using 
parser.add_argument(
        '--use_x_percent_corpus', default=100, type=int,
        help="Debug flag that allows user to only use a random percentage "
        "of available knowledge base corpus for RAG")
set to 5


examples/llm/txt2kg_rag.py
```diff
@@ -212,6 +212,36 @@ def _process_and_chunk_text(text, chunk_size, doc_parsing_mode):
 
 def get_data(args):
     # need a JSON dict of Questions and answers, see below for how its used
+    if not (os.path.exists(os.path.join(args.dataset, "train.json")) or os.path.join(args.dataset, "corpus")):
```
Kh4L, line 215: can you make `args.dataset` default to techqa and run this DL pass only if `args.dataset==techqa`?

Kh4L (changes requested):
left a small comment, other than that LGTM 🚀

nv-rliu (changes requested):
Add matching `"` to the end of line 156

examples/llm/txt2kg_rag.py
Kh4L: also need to move the files to the dataset dir

PR #10309, opened 2025-06-07
Test Coverage for RAG FeatureStore, GraphStore, and Vector RAG Loaders

zaristei:


torch_geometric/utils/rag/vectorrag.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, raw_docs: List[str],
 
         if self.embedded_docs is None:
             assert self.model is not None, "Model must be provided if embedded_docs is not provided"
+            self.model_kwargs = model_kwargs or {}
```
puririshi98, line 48: whats the point here if it cant be none due to the assert

torch_geometric/utils/rag/feature_store.py
```diff
@@ -137,7 +137,8 @@ def load_subgraph(
         # if it supported edge features
         edge_id = sample.edge
         edge_attr = self.edge_attr[edge_id]
-        return Data(edge_attr=edge_attr, edge_idx=edge_id)
+        edge_idx = torch.stack([sample.global_row, sample.global_col], dim=0)
```
puririshi98: can you explain why we need this change

puririshi98 (changes requested):

puririshi98 (changes requested):
please run these locally on our internal cluster and make sure these tests pass in our latest container

torch_geometric/utils/rag/feature_store.py
zaristei: oops, i might have conflated edge_index with edge_idx here. Let me confirm that edge_idx doesnt get used anywhere else in the code and can be safely removed. The edge_index is necessary to track the adjacency matrix of the new graph that gets returned, because otherwise it's not being tracked.

puririshi98 (changes requested):
also seems some of them fail in github CI, please put the appropriate decorators to make sure this doesnt happen

torch_geometric/utils/rag/feature_store.py
puririshi98: if you need to replace edge_idx= torch.stack([sample.global_row, sample.global_col], dim=0)
then do we no longer need line 138?
edge_id = sample.edge
if its unused should drop it, but we definitely want to make sure accuracy remains safe after these changes

puririshi98 (changes requested):

torch_geometric/utils/rag/vectorrag.py
puririshi98: @zaristei ?

PR #10314, opened 2025-06-09
[Bug] Preventing side effects in distributed tests by destroying process group …

drivanov:
in `test_molecule_gpt_datase`.

This PR addresses a test isolation issue by explicitly destroying the distributed process group at the end of the `test_molecule_gpt_dataset`. This is necessary because:

The test initializes a distributed environment (e.g., via `torch.distributed` backend implicitly or via PyG internals).

If the process group is not properly destroyed, subsequent tests may fail with errors like:

```
        group = group or _get_default_group()
        opts = AllgatherOptions()
        opts.asyncOp = async_op
>       work = group.allgather([tensor_list], [tensor], opts)
E       RuntimeError: No backend type associated with device type cpu
```
This issue was observed specifically on GB200 machines when running multiple CI tests in sequence.

For instance, this bug manifests when running:
```
pytest /opt/pyg/pytorch_geometric/test/datasets/test_molecule_gpt_dataset.py::test_molecule_gpt_dataset \
        /opt/pyg/pytorch_geometric/test/metrics/test_link_pred_metric.py
```
but **does not** appear when running the second test (test_link_pred_metric.py) independently or before the first.

This change ensures proper resource cleanup in CI and avoids test interference across modules that may share distributed state.


test/datasets/test_molecule_gpt_dataset.py
```diff
@@ -8,3 +10,7 @@ def test_molecule_gpt_dataset():
     assert str(dataset) == f'MoleculeGPTDataset({len(dataset)})'
     assert dataset.num_edge_features == 4
     assert dataset.num_node_features == 6
+
+    # Optional cleanup if distributed is initialized
+    if dist.is_initialized():
+        dist.destroy_process_group()
```
xnuohz: how about moving to `test/conftest.py`

drivanov: Done. 
@xnuohz: Thanks, it was a good suggestion!

xnuohz (approved):

Kh4L (approved):

PR #10366, opened 2025-07-15
make clamp range of `PowerMeanAggregation` adjustable

KaneGreen:
1. Add two args when class init to allow users to set their own `clamp` function range.
2. Change the default lower bound of `clamp` function from `0` to `0.0001`. This is because the value of `0` can easily cause `p` to become `NaN` after several forwards when `learn=True`.

### DIFF
--- a/torch_geometric/nn/aggr/basic.py
+++ b/torch_geometric/nn/aggr/basic.py
@@ -245,9 +245,14 @@ class PowerMeanAggregation(Aggregation):
         clamp_max (float, optional): Upper-bound of the range to be clamped
             to. There is no upper bound if set to :obj:`None`.
     """
-    def __init__(self, p: float = 1.0, learn: bool = False, channels: int = 1,
-                 clamp_min: Optional[float] = 1e-4,
-                 clamp_max: Optional[float] = 100., ) -> None:
+    def __init__(
+        self,
+        p: float = 1.0,
+        learn: bool = False,
+        channels: int = 1,
+        clamp_min: Optional[float] = 1e-4,
+        clamp_max: Optional[float] = 100.,
+    ) -> None:
         super().__init__()
 
         if not learn and channels != 1:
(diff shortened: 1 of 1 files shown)


torch_geometric/nn/aggr/basic.py
```diff
@@ -240,8 +240,14 @@ class PowerMeanAggregation(Aggregation):
             If set to a value greater than :obj:`1`, :math:`p` will be learned
             per input feature channel. This requires compatible shapes for the
             input to the forward calculation. (default: :obj:`1`)
+        clamp_min (float, optional): Lower-bound of the range to be clamped
+            to. There is no lower bound if set to None.
+        clamp_max (float, optional): Upper-bound of the range to be clamped
+            to. There is no upper bound if set to None.
```
rusty1s:
```suggestion
        clamp_min (float, optional): Lower-bound of the range to be clamped
            to. There is no lower bound if set to :obj:`None`.
        clamp_max (float, optional): Upper-bound of the range to be clamped
            to. There is no upper bound if set to :obj:`None`.
```

torch_geometric/nn/aggr/basic.py
```diff
@@ -240,8 +240,14 @@ class PowerMeanAggregation(Aggregation):
...
     """
-    def __init__(self, p: float = 1.0, learn: bool = False, channels: int = 1):
+    def __init__(self, p: float = 1.0, learn: bool = False, channels: int = 1,
+                 clamp_min: Optional[float] = 1e-4,
+                 clamp_max: Optional[float] = 100.):
```
rusty1s:
```suggestion
                 clamp_max: Optional[float] = 100., ) -> None:
```

rusty1s (approved):

KaneGreen:
Hello, I noticed pyg 2.7.0 released on Oct 15. And I hope to know is there anything blocking this PR to be merged and anything I need to do?

PR #10371, opened 2025-07-19
Add GNAN model to PyG

codearky:
This PR adds the GNAN (https://arxiv.org/pdf/2406.01317) model and a dedicated dataloader to PyG


puririshi98 (changes requested):
LGTM at a high level, but please lets move the example to examples/explain.
Please also add a sentence to examples/README.md to mention, "For examples on explainability with GNNs, see examples/explain."
Then please also update examples/explain/README.md to have a short explanation of what the new example does. 
Once thats done I will do a deeper review and can merge

puririshi98 (changes requested):
LGTM, but can you share a log of the the e2e run of `examples/explain/mgnan_graph_mutagenicity.py` inside the nvidia PyG container to make sure it works as expected:  https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pyg/tags?version=26.03-py3

puririshi98:
@codearky

codearky:
[mgnan-ngc-477482.full.log](https://github.com/user-attachments/files/28636754/mgnan-ngc-477482.full.log)
@puririshi98 Sorry for the delay, let me know if this satisfy the requirement

puririshi98 (approved):
LGTM

puririshi98:
@akihironitta to merge

torch_geometric/loader/mgnan_dataloader.py
akihironitta, line 1: We cannot place this new addition here since it's too specific to the use case. Since this dataloader is just a PyTorch dataloader with a simple custom `collate_fn`, we'd need to have this in the example script.

torch_geometric/llm/models/g_retriever.py
```diff
@@ -96,7 +96,7 @@ def encode(
         edge_attr: Optional[Tensor],
     ) -> Tensor:
         x = x.to(self.llm.device)
-        edge_index = edge_index.to(self.llm.device)
+        edge_index = edge_index.to(torch.int64).to(self.llm.device)
```
akihironitta: If you have to cast `edge_index` here, we must be doing something wrong on the caller side. I'd suggest fixing it there instead of here.

codearky: I removed this change as it has nothing to do with the pure GNAN case (it came from another layer of integration of GNAN with the G-Retriever pipeline - which is unnecessary in this PR and will be handled separately).

torch_geometric/loader/mgnan_dataloader.py
codearky: Refactored to use a plain PyTorch DataLoader with a custom collater (MGNANCollater) that now lives in the example script.

codearky:
Following recent changes I rerun the e2e example
[mgnan-ngc-499052.full.log](https://github.com/user-attachments/files/28709656/mgnan-ngc-499052.full.log)

puririshi98:
@akihironitta  to review

examples/explain/mgnan_graph_mutagenicity.py
```diff
@@ -0,0 +1,270 @@
...
+    in_channels = dataset.num_features
+
+    print(f"Dataset info: {num_classes} classes, {in_channels} features")
+    print(f"Dataset size: {len(dataset)} graphs")
+
+    # Simple 80/10/10 split
+    indices = list(range(len(dataset)))
+    random.shuffle(indices)
+    n_train = int(0.8 * len(indices))
+    n_val = int(0.1 * len(indices))
```
akihironitta:
```suggestion
    n_train = int(0.8 * len(dataset))
    n_val = int(0.1 * len(dataset))
```

examples/explain/mgnan_graph_mutagenicity.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # print("=" * 60)
```
akihironitta:
```suggestion
```

torch_geometric/contrib/nn/models/mgnan.py
```diff
@@ -0,0 +1,322 @@
...
+
+from torch_geometric.data import Batch, Data
+from torch_geometric.utils import scatter
+
+__all__ = [
+    'MGNAN',
+]
+
+
+def _init_weights(module: nn.Module, std: float = 1.0):
```
akihironitta:
```suggestion
def _init_weights(module: nn.Module, std: float = 1.0) -> None:
```

akihironitta (approved):
I skimmed through the changes, LGTM!

PR #10388, opened 2025-07-30
Adding separate() Method to Data and HeteroData for extracting connected components

jesseangelis:
**Summary**

This PR adds a new method, `separate()`, to both `HeteroData` and `Data` classes in PyTorch Geometric. It identifies and extracts disjoint connected components from a (heterogeneous or homogeneous) graph using a union-find algorithm.

**Motivation**

While PyG provides convenient utilities such as `subgraph`, `edge_subgraph`, and `node_type_subgraph`, it does not currently offer a built-in method to extract connected components from graphs. This functionality could be useful for:

- Preprocessing or cleaning sparse or noisy graphs
- Isolating disconnected structures
- Enabling mini-batching strategies for large, sparse datasets

This PR provides a general-purpose implementation that complements existing subgraph utilities.

**API**

```
components = data.separate()
components = hetero_data.separate()
```

- Returns: A `List[Data]` or `List[HeteroData]`, with each item corresponding to a connected component.
- Optional filtering: For heterogeneous graphs, users are encouraged to pre-filter using existing methods like `node_type_subgraph` or `edge_type_subgraph` if they wish to limit the types involved.

**Highlights**

- Implementation for both Data and `HeteroData`
- Union-Find algorithm ensures O($\alpha$(n)), where $\alpha$: Ackermann function
- Compatible with PyG idioms and retains all feature and edge attributes if present in the connected component
    
**Discussion Points**

Based on feedback from the original issue, this version avoids introducing new arguments like `allowed_edge_types` or `allowed_node_types` and instead suggests combining the new method with existing filtering utilities.

Happy to further extend or refine the method based on team preferences!

### DIFF
--- a/test/data/test_hetero_data.py
+++ b/test/data/test_hetero_data.py
@@ -594,6 +594,177 @@ def test_hetero_data_update():
                        other['paper', 'paper'].edge_index)
 
 
+def test_hetero_data_separate():
+    data = HeteroData()
+    data["red"].x = torch.tensor([[1.0], [2.0], [3.0], [4.0], [5.0]])
+    data["red"].y = torch.tensor([1, 2, 3, 4, 5])
+    data["red"].z = torch.tensor([[1.1, 1.2], [2.1, 2.2], [3.1, 3.2],
+                                  [4.1, 4.2], [5.1, 5.2]])
+    data["blue"].x = torch.tensor([[6.0], [7.0]])
+    data["red", "to",
+         "red"].edge_index = torch.tensor([[0, 1, 2, 3], [1, 0, 3, 2]],
+                                          dtype=torch.long)
+    data["red", "with", "red"].edge_index = torch.tensor([[1], [1]],
+                                                         dtype=torch.long)
+    split_data = data.separate()
+
+    assert isinstance(split_data, list)
+    assert len(split_data) == 5
+    assert isinstance(split_data[0], HeteroData)
+    assert isinstance(split_data[1], HeteroData)
+    assert isinstance(split_data[2], HeteroData)
+    assert isinstance(split_data[3], HeteroData)
+    assert split_data[0].node_types == ['red']
+    assert split_data[1].node_types == ['red']
+    assert split_data[2].node_types == ['red']
+    assert split_data[3].node_types == ['blue']
+    assert split_data[4].node_types == ['blue']
+    assert split_data[0].edge_types == [('red', 'to', 'red'),
+                                        ('red', 'with', 'red')]
+    assert split_data[1].edge_types == [('red', 'to', 'red')]
+    assert split_data[2].edge_types == []
+    assert split_data[3].edge_types == []
+    assert split_data[4].edge_types == []
+    assert torch.equal(split_data[0]["red"].x, torch.tensor([[1.0], [2.0]]))
+    assert torch.equal(split_data[0]["red"].y, torch.tensor([1, 2]))
+    assert torch.equal(split_data[0]["red"].z,
+                       torch.tensor([[1.1, 1.2], [2.1, 2.2]]))
+    assert torch.equal(split_data[1]["red"].x, torch.tensor([[3.0], [4.0]]))
+    assert torch.equal(split_data[1]["red"].y, torch.tensor([3, 4]))
+    assert torch.equal(split_data[1]["red"].z,
+                       torch.tensor([[3.1, 3.2], [4.1, 4.2]]))
+    assert torch.equal(split_data[2]["red"].x, torch.tensor([[5.0]]))
+    assert torch.equal(split_data[2]["red"].y, torch.tensor([5]))
+    assert torch.equal(split_data[2]["red"].z, torch.tensor([[5.1, 5.2]]))
+    assert torch.equal(split_data[3]["blue"].x, torch.tensor([[6.0]]))
+    assert torch.equal(split_data[4]["blue"].x, torch.tensor([[7.0]]))
+    assert torch.equal(split_data[0]["red", "to", "red"].edge_index,
+                       torch.tensor([[0, 1], [1, 0]]))
+    assert torch.equal(split_data[0]["red", "with", "red"].edge_index,
+                       torch.tensor([[1], [1]]))
+    assert torch.equal(split_data[1]["red", "to", "red"].edge_index,
+                       torch.tensor([[0, 1], [1, 0]]))
+
+
+@pytest.fixture
+def unmutable_obj():
+    class NotMutable:
+        def __getitem__(self, idx):
+            return 0
+
+    yield NotMutable()
+
+
+def test_hetero_data_heck_slicable_and_mutable(unmutable_obj):
+
+    # Case 1: All attributes are sliceable and mutable (should not raise)
+    data = HeteroData()
+    data['paper'].x = torch.zeros(3, 4)
+    data['author'].y = torch.ones(2, 5)
+    data['paper', 'to', 'author'].edge_index = torch.zeros(2, 2)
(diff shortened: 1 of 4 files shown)


torch_geometric/data/hetero_data.py
```diff
@@ -487,6 +487,149 @@ def validate(self, raise_on_error: bool = True) -> bool:
 
         return status
 
+    def separate(self) -> List[Self]:
```
wsad1:
```suggestion
    def connected_components(self) -> List[Self]:
```
Nit: rename to connected_components.

test/data/test_hetero_data.py
```diff
@@ -594,6 +594,177 @@ def test_hetero_data_update():
...
+                                                         dtype=torch.long)
+    split_data = data.separate()
+
+    assert isinstance(split_data, list)
+    assert len(split_data) == 5
+    assert isinstance(split_data[0], HeteroData)
+    assert isinstance(split_data[1], HeteroData)
+    assert isinstance(split_data[2], HeteroData)
+    assert isinstance(split_data[3], HeteroData)
+    assert split_data[0].node_types == ['red']
```
wsad1: Can we test the scenario, where a split has multiple node/edge types.
Plus a case where split return just one element i.e graph is has single connected component.

wsad1 (commented):
Left some comments.
Will take another look this week.

torch_geometric/data/hetero_data.py
```diff
@@ -487,6 +487,149 @@ def validate(self, raise_on_error: bool = True) -> bool:
...
+
+        # Group nodes by their representative parent
+        components_map = defaultdict(list)
+        for node, parent in self._parents.items():
+            components_map[parent].append(node)
+        del self._parents
+        del self._ranks
+
+        components: List[Self] = []
+        for nodes in components_map.values():
```
wsad1, line 542: Do you think we can reuse the `subgraph` function. 
Essentially convert `component` which is tuple of `NodeType` , `node id` a dict of `Dict[NodeType, Tensor of node ids]` and then pass this to `subgraph function`.

jesseangelis: Yes, that's a great suggestion. Reusing subgraph makes the code much cleaner and ensures all node and edge types are present in each component (including empty ones), which might be beneficial for downstream applications.

torch_geometric/data/hetero_data.py
```diff
@@ -487,6 +487,103 @@ def validate(self, raise_on_error: bool = True) -> bool:
 
         return status
 
+    def connected_components(self) -> List[Self]:
+        r"""Extracts connected components of the heterogeneous graph using
+        a union-find algorithm. The components are returned as a list of
+        :class:`~torch_geometric.data.HeteroData` objects. Only works if
+        attributes are sliceable (i.e. implement `__getitem__`) and mutable
+        (i.e. implement `__setitem__`).
```
wsad1:
```suggestion
        :class:`~torch_geometric.data.HeteroData` objects.
```

torch_geometric/data/data.py
```diff
@@ -904,6 +905,80 @@ def to_heterogeneous(
...
+            >>> tensor([[0, 1], [1, 0]])
+
+            print(components[1].x)
+            >>> tensor([[3.0], [4.0]])
+
+            print(components[1].y)
+            >>> tensor([[3.1], [4.1]])
+
+            print(components[1].edge_index)
+            >>> tensor([[0, 1], [1, 0]])
```
wsad1:
```suggestion
            components = data.connected_components()
            print(len(components))
            >>> 2

            print(components[0].x)
            >>> Data(x=[2, 1], y=[2, 1], edge_index=[2, 2])
```

torch_geometric/data/data.py
```diff
@@ -904,6 +905,80 @@ def to_heterogeneous(
 
         return data
 
+    def connected_components(self) -> List[Self]:
+        r"""Extracts connected components of the graph using a union-find
+        algorithm. The components are returned as a list of
+        :class:`~torch_geometric.data.Data` objects, where each object
+        represents a connected component of the graph. Only works if
+        attributes are slicable (i.e. has the __getitem__ method implemented)
+        and are mutable (i.e. has the __setitem__ method implemented).
```
wsad1:
```suggestion
        represents a connected component of the graph.
```

torch_geometric/data/hetero_data.py
```diff
@@ -487,6 +487,101 @@ def validate(self, raise_on_error: bool = True) -> bool:
...
+                [[0, 1, 2, 3], [1, 0, 3, 2]], dtype=torch.long
+            )
+
+            components = data.connected_components()
+            print(len(components))
+            >>> 4
+
+            print(type(components))
+            >>> <class 'list'>
+
```
wsad1:
```suggestion
```

torch_geometric/data/hetero_data.py
```diff
@@ -487,6 +487,101 @@ def validate(self, raise_on_error: bool = True) -> bool:
...
+                blue={x: tensor([[5.]])},
+                red, to, red={edge_index: tensor([[],[]])}
+            )
+
+            print(components[3])
+            >>> HeteroData(
+                red={x: tensor([[]])},
+                blue={x: tensor([[6.]])},
+                red, to, red={edge_index: tensor([[],[]])}
+            )
```
wsad1:
```suggestion
```

test/data/test_hetero_data.py
```diff
@@ -594,6 +594,275 @@ def test_hetero_data_update():
...
+    data["red", "with", "red"].edge_index = torch.tensor([[1], [1]],
+                                                         dtype=torch.long)
+    split_data = data.connected_components()
+
+    assert isinstance(split_data, list)
+    assert len(split_data) == 5
+    assert isinstance(split_data[0], HeteroData)
+    assert isinstance(split_data[1], HeteroData)
+    assert isinstance(split_data[2], HeteroData)
+    assert isinstance(split_data[3], HeteroData)
```
wsad1:
```suggestion
    for data in split_data:
        assert isinstance(data, HeteroData)
    
```

torch_geometric/data/data.py
```diff
@@ -1150,6 +1205,65 @@ def get_all_edge_attrs(self) -> List[EdgeAttr]:
...
+        This is necessary for the `separate` method to work correctly.
+        """
+        for key in self._store.keys():
+            if key in {'edge_index', 'adj', 'adj_t'}:
+                continue
+            if not hasattr(self[key], "__getitem__"):
+                raise TypeError(f"Attribute '{key}' does not support slicing.")
+            if not hasattr(self[key], "__setitem__"):
+                raise TypeError(f"Attribute '{key}' is not mutable.")
+
```
wsad1:
```suggestion
```

torch_geometric/data/data.py
```diff
@@ -1150,6 +1205,51 @@ def get_all_edge_attrs(self) -> List[EdgeAttr]:
...
+        """
+        root1 = self._find_parent(node1)
+        root2 = self._find_parent(node2)
+        if root1 != root2:
+            if self._ranks[root1] < self._ranks[root2]:
+                self._parents[root1] = root2
+                self._ranks[root2] += 1
+            elif self._ranks[root1] > self._ranks[root2]:
+                self._parents[root2] = root1
+                self._ranks[root1] += 1
```
wsad1:
```suggestion
            elif self._ranks[root1] > self._ranks[root2]:
                self._parents[root2] = root1
```

wsad1 (approved):
Thanks for the work @jesseangelis .

PR #10394, opened 2025-08-04
Add GoldenRAG Script

nv-rliu:
Based heavily off of `txt2kg_rag.py`

This PR adds a script to answer questions with golden contexts provided to an LLM Generator.

### DIFF
--- a/examples/llm/golden_rag.py
+++ b/examples/llm/golden_rag.py
@@ -162,7 +162,9 @@ if __name__ == '__main__':
     # Need to sanitize sensitive keys
     saved_NIM_KEY = args.NV_NIM_KEY
     args.NV_NIM_KEY = "********"
-    print(f"Starting {args.dataset} training with args: ", args)
+    print(
+        f"Starting {args.dataset} training with args:\n{json.dumps(args, indent=4, sort_keys=True)}"
+    )
     args.NV_NIM_KEY = saved_NIM_KEY
 
     eval_batch_size = args.eval_batch_size


puririshi98:
btw wont merge this, just want it in side branch for testing

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+
+    assert args.NV_NIM_KEY, "NVIDIA API key is required for TXT2KG and eval"
+
+    return args
+
+
+def get_data(args):
+    # need a JSON dict of Questions and answers, see below for how its used
+    json_path = Path(args.dataset) / "train.json"
+    Path(args.dataset) / "corpus"
```
rlratzel: Seems like this line isn't needed?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+
+    dataset_name = os.path.basename(args.dataset)
+    dataset_path = os.path.join(args.dataset, f"{dataset_name}.pt")
+    torch.save((total_data_list, max_chars_in_train_answer), dataset_path)
+
+    # NOTE: test and validation split have been removed. below code for all pairs
+    # torch.save((test_data, max_chars_in_train_answer), dataset_path)
+    # return total_data_list
+
+    return total_data_list[int(.8 * len(total_data_list)):]
```
rlratzel: Curious why not return the entire list?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+    [END_RETRIEVED_CONTEXTS]
+    """
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--NV_NIM_MODEL', type=str,
+                        default="nvidia/llama-3.1-nemotron-ultra-253b-v1",
+                        help="The NIM LLM to use for LLMJudge")
+    parser.add_argument('--NV_NIM_KEY', type=str, help="NVIDIA API key")
```
rlratzel, line 35: can the default be `os.environ["NV_NIM_KEY"]` or similar so we don't have to have our key visible in command history and on terminals when screen sharing, etc.?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+                question=test_batch.question, context=context)
+
+        # LLM generator inference step
+        # TODO: please check if this makes sense. context was "" in txt2kg_rag. Setting it equals to the context doc returned some weird output. A list[~260]
+        qs = [
+            q_with_context,
+        ]
+
+        pred = model.inference(question=qs,
+                               max_tokens=max_chars_in_train_answer / 3)
```
rlratzel, line 144: Can you comment why `/ 3`?  Is this an estimate used to convert number of chars to number of tokens (i.e. roughly 3 chars per token)?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+
+        eval_tuples.append((q, pred, test_batch.label))
+    for question, pred, label in tqdm(eval_tuples, desc="Eval"):
+        breakpoint()
```
rlratzel: Should this be removed?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+
+if __name__ == '__main__':
+    # for reproducibility
+    seed_everything(50)
+    args = parse_args()
+
+    # Need to sanitize sensitive keys
+    saved_NIM_KEY = args.NV_NIM_KEY
+    args.NV_NIM_KEY = "********"
+    print(f"Starting {args.dataset} training with args: ", args)
```
rlratzel: Can you consider this instead? (I'd like to do this in `txt2kg_rag.py` too)
```suggestion
    print(f"Starting {args.dataset} training with args:\n{json.dumps(args, indent=4, sort_keys=True)}")
```

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+                        help="Evaluation batch size")
+    parser.add_argument('--llm_generator_name', type=str,
+                        default="meta-llama/Meta-Llama-3.1-8B-Instruct",
+                        help="The LLM to use for Generation")
+    parser.add_argument(
+        '--llm_generator_mode', type=str, default="full",
+        choices=["frozen", "lora",
+                 "full"], help="Whether to freeze the Generator LLM,\
+                        use LORA, or fully finetune")
+    parser.add_argument('--dont_save_model', action="store_true",
```
rlratzel: This arg isn't used anywhere, is that an oversight or should it be removed?

examples/llm/golden_rag.py
```diff
@@ -0,0 +1,189 @@
...
+    qa_data = get_data(args)
+    print(" ==> Number of Docs:", len(qa_data))
+
+    # NOTE: test and validation split have been removed
+    # data_lists = {"test": []}
+    global max_chars_in_train_answer
+
+    total_data_list = []
+    for pair in tqdm(qa_data, desc="Building un-split dataset"):
+        max_chars_in_train_answer = max(len(pair['answer']),
```
rlratzel, line 81: I was reminded of this in @jnke2016's [PR](https://github.com/pyg-team/pytorch_geometric/pull/10413).
Should this be added to this loop too?
```
        if pair["is_impossible"]:
            continue
```

nv-rliu: To my knowledge, this field isn't generated with the latest `truequery.py`, therefore I haven't included it

examples/llm/golden_rag.py
nv-rliu: I like this.

examples/llm/golden_rag.py
nv-rliu: We can if we want. Right now we haven't pinned down an exact size of the data list we want to use for eval/testing. I vary this during my testing for getting a specific number of pairs. This is still testing and non production

examples/llm/golden_rag.py
nv-rliu: Sounds good

examples/llm/golden_rag.py
nv-rliu: @puririshi98 perhaps.

I just know that we have been testing max_tokens with custom values. Perhaps this is an error since we're counting the chars in our answers rather than tokens.

examples/llm/golden_rag.py
nv-rliu: Yes

rlratzel (approved):

PR #10410, opened 2025-08-11
Add Annotation to Stored Triples

nv-rliu:
## Goal
This PR adds support for storing the model name and a timestamp within the triples file for `txt2kg_rag` workflows. It also extends this functionality to KG checkpoints. It will detect if the retrieved triples are not consistent with the configuration of the current run.

### Checkpointing Looks like:

Directory before running:
```
user@machine:/SynthQA$ ls
corpus
config.yaml
train.json
```

After
```
user@machine:/SynthQA$ ls
corpus
config.yaml
train.json
llama-3.1-nemotron-70b-instruct--20250902_195331--checkpoint_kg.pt
```

### Stored Triples:
```
user@machine:/SynthQA$ ls
corpus
config.yaml
train.json
llama-3.1-nemotron-70b-instruct--20250902_195331--raw_triples.pt
```

When using stored, you will see
```
 -> Saved triples generated with: llama-3.1-nemotron-70b-instruct
Number of triples in our GraphDB = 164288
```


examples/llm/txt2kg_rag.py
```diff
@@ -322,8 +322,15 @@ def index_kg(args, context_docs):
     checkpoint_path = os.path.join(args.dataset, "checkpoint_kg.pt")
     if os.path.exists(checkpoint_path):
         print("Restoring KG from checkpoint...")
-        saved_relevant_triples = torch.load(checkpoint_path,
-                                            weights_only=False)
```
Kh4L, line 326: why did you remove `weights_only=False`?

i think you should leave that arg

Kh4L (changes requested):
LGTM overall, left a small comment

examples/llm/txt2kg_rag.py
nv-rliu: I thought `weights_only` was only useful if we want to extract model weights and biases, aka, tensors. Otherwise, isn't the argument not really useful in this situation? I can add it back if that's incorrect

Kh4L: AFAIK thats the case when you set `True`, here we want to load items that are not Tensors (W&B), so we need to set it to `False`, otherwise pytorch might complaint

examples/llm/README.md
```diff
@@ -1,11 +1,11 @@
...
+| [`g_retriever.py`](./g_retriever.py)   | Example for Retrieval-Augmented Generation (RAG) w/ GNN+LLM by co-training `LLAMA3` with `GAT` for answering questions based on knowledge graph information from the toy WebQSP dataset. We also have an [example repo](https://github.com/neo4j-product-examples/neo4j-gnn-llm-example) for integration with [Neo4j Graph DBs][neo4j.com] along with an associated [blog](https://developer.nvidia.com/blog/boosting-qa-accuracy-with-graphrag-using-pyg-and-graph-databases/) showing 2x accuracy gains over LLMs on real medical data.                                                                                                                                                                                                                                        |
```
puririshi98: we dont need these changes

examples/llm/txt2kg_rag.py
```diff
@@ -409,12 +418,17 @@ def make_dataset(args):
     triples = []
     raw_triples_path = os.path.join(args.dataset, "raw_triples.pt")
     if os.path.exists(raw_triples_path):
-        triples = torch.load(raw_triples_path, weights_only=False)
+        saved_data = torch.load(raw_triples_path)
```
puririshi98: plz save/load w format {llm-name}_{datetime}_[raw_triples.pt](http://raw_triples.pt/)

puririshi98 (changes requested):

puririshi98:
also plz make CI green, check the failures and fix

examples/llm/txt2kg_rag.py
nv-rliu: Ah okay. I thought it was okay to not use it since this Data object is never storing weights and biases. I'll add that back in.

examples/llm/README.md
nv-rliu: Don't know how this got affected. Removing

examples/llm/txt2kg_rag.py
nv-rliu: changed format

nv-rliu:
It looks like CI is failing here but I'm not sure why. The files I touched shouldn't affect the tests, right?

Kh4L:
> It looks like CI is failing here but I'm not sure why. The files I touched shouldn't affect the tests, right?

yeah i think we're safe skipping it

PR #10413, opened 2025-08-12
Add an agentic baseline

jnke2016:


puririshi98:
LGTM but wont merge it this should just exist as a side PR for benchmarking similar to golden RAG https://github.com/pyg-team/pytorch_geometric/pull/10394

examples/llm/aiq_baseline/rag_config.yml
rlratzel, line 1: What reads this file? I don't see it referenced anywhere in the script.

examples/llm/aiq_baseline/langchain_corpus_ingest.py
```diff
@@ -0,0 +1,260 @@
...
+    args = parser.parse_args()
+
+    (milvus_client, encoder_model) = main(
+        milvus_uri=args.milvus_uri, collection_name=args.collection_name,
+        dataset=args.dataset, llm_generator_name=args.llm_generator_name,
+        drop_collection=args.drop_collection,
+        embedding_model=args.embedding_model, chunk_size=args.chunk_size,
+        chunk_overlap=args.chunk_overlap, metric_type=args.metric_type)
+
+    if not with_react_agent:
```
rlratzel: Is this right?
```suggestion
    if not args.with_react_agent:
```

examples/llm/aiq_baseline/langchain_corpus_ingest.py
```diff
@@ -0,0 +1,260 @@
...
+    parser.add_argument('--with_react_agent', action="store_true",
+                        help="Use react_agent")
+    args = parser.parse_args()
+
+    (milvus_client, encoder_model) = main(
+        milvus_uri=args.milvus_uri, collection_name=args.collection_name,
+        dataset=args.dataset, llm_generator_name=args.llm_generator_name,
+        drop_collection=args.drop_collection,
+        embedding_model=args.embedding_model, chunk_size=args.chunk_size,
+        chunk_overlap=args.chunk_overlap, metric_type=args.metric_type)
```
rlratzel: It looks like you also have to pass `args.with_react_agent` based on the signature of `main`.

examples/llm/aiq_baseline/langchain_corpus_ingest.py
```diff
@@ -0,0 +1,260 @@
...
+
+def emb_text(text_lines, model):
+    embeddings = model.encode(text_lines)
+    return embeddings
+
+
+def main(*, milvus_uri: str, collection_name: str, dataset: str,
+         llm_generator_name: str, drop_collection: bool, embedding_model: str,
+         chunk_size: int, chunk_overlap: int, metric_type: str,
+         with_react_agent: bool):
```
rlratzel: Is this arg needed? I don't see it used in this function.

examples/llm/aiq_baseline/langchain_corpus_ingest.py
```diff
@@ -0,0 +1,260 @@
...
+    # as it automatically utilizes Milvus Lite to store all data in this file.
+
+    # If you have large scale of data, you can set up a more performant Milvus server on
+    # docker or kubernetes. In this setup, please use the server uri, e.g.http://localhost:19530, as your uri.
+    # If you want to use Zilliz Cloud, the fully managed cloud service for Milvus, adjust the uri and token,
+    # which correspond to the Public Endpoint and Api key in Zilliz Cloud.
+    milvus_client = MilvusClient(uri=milvus_uri)
+
+    # FIXME: For now, always drop collection prior to each run
+    if drop_collection:
```
rlratzel: Is this `FIXME` still accurate? It looks like `drop_collection` is passed in from the CLI and may not always be `True`.

examples/llm/aiq_baseline/langchain_corpus_ingest.py
```diff
@@ -0,0 +1,260 @@
...
+        question, answer = (data_point["question"], data_point["answer"])
+
+        print("The question = \n", question)
+        search_res = milvus_client.search(
+            collection_name=collection_name,
+            data=[
+                # Convert the question to an embedding vector
+                emb_text(question, encoder_model).tolist()[0]
+            ],  # Use the `emb_text` function to convert the question to an embedding vector
+            limit=10,  # Return top 3 results
```
rlratzel: Is this right?
```suggestion
            limit=3,  # Return top 3 results
```
or should we do this?
```suggestion
            limit=10,  # Return top 10 results
```

examples/llm/aiq_baseline/langchain_corpus_ingest.py
jnke2016: It was initially but I made and update to move that functionality in the test function instead. I just removed it as it is no longer necessary there.

examples/llm/aiq_baseline/langchain_corpus_ingest.py
jnke2016: right

examples/llm/aiq_baseline/rag_config.yml
jnke2016: This is how the file is read. `aiq run --config_file /pytorch_geometric/examples/llm/aiq_baseline/rag_config.yml --input "question"`

examples/llm/aiq_baseline/langchain_corpus_ingest.py
jnke2016: I picked the top result to match the `rag_config` file

PR #10417, opened 2025-08-19
Fix support for importing from Lightning package

### LINKED ISSUE #9811: `ValueError("Expected a parent")` when using new Lightning imports and PyG Lightning DataModule wrappers

nathanpainchaud:
Fixes #9811 

The fix proposed in #10404 was a step in the right direction by trying to import from the version installed by the user. However, as described in this comment, Lighting also installs a version of `pytorch_lightning`. So it should be checked second, as it will be available whether the user originally installed `lightning` or `pytorch_lightning`.

This is what the current PR does.

I tried implementing a test based on the example provided in #9811. However, I don't know how to dynamically (un)install packages (in this case `lightning`) in the current environment. If you want a test for the issue, and know of a way to install packages at runtime, please point it to me and I'll try to adapt my existing local tests.


akihironitta (approved):
@nathanpainchaud Thank you very much again for sending this fix. Could you document this in your PR description so that whenever someone sees this issue, we can refer to this PR description?

#### Before this PR

| how you import PL | `pip install lightning` | `pip install pytorch-lightning` |
| - | :-: | :-: |
| `import lightning` | ❌ | ✅ |
| `import pytorch_lightning` | ❌ | ✅ |

#### After this PR

| how you import PL | `pip install lightning` | `pip install pytorch-lightning` |
| - | :-: | :-: |
| `import lightning` | ✅ | ❌ |
| `import pytorch_lightning` | ❌ | ✅ |

nathanpainchaud:
> Thank you very much again for sending this fix.

Thank you for finally merging it :slightly_smiling_face: 

> Could you document this in your PR description so that whenever someone sees this issue, we can refer to this PR description?

@akihironitta Done. Let me know if you want to change the way I documented thigs.

PR #10424, opened 2025-08-28
Fix typo in `CODEOWNERS`

akihironitta:


akihironitta:
@puririshi98 Thanks for catching the typo in #9992!

PR #10368, opened 2025-07-15
Cleanup G-retriever Example and APIs

puririshi98:
Step 1 for merging: https://github.com/pyg-team/pytorch_geometric/pull/9992/

PR #10436, opened 2025-09-04
Reorganize TXT2KG and introduce `torch_geometric.llm`

puririshi98:
replaces https://github.com/pyg-team/pytorch_geometric/pull/9992
adds new torch_Geometric.llm for NVIDIA to mantain
example of how to use this in nvidia docker container:
git config --global credential.helper store; huggingface-cli login --token <insert_token>; cd /opt/pyg; pip uninstall -y torch-geometric; rm -rf pytorch_geometric; git clone -b latest-txt2kg https://github.com/pyg-team/pytorch_geometric.git; cd /opt/pyg/pytorch_geometric; pip install .; pip install openai

example to run:
python3 examples/llm/txt2kg_rag.py

thanks to @Kh4L @zaristei @rlratzel @rliu
for their contributions
image

image
This is thoroughly tested by many internal and external users and should be good to merge. we will continue to improve this pipeline in future PRs but this PR is ready.
Future PRs:

add the RAG CI job back after figuring out why tests pass but the CI gives a red X
add a non-toy dataset for the default data (needs to be released by NVIDIA on huggingface, waiting on that)
further improvements to user friendliness of example

I will work with @zaristei on refining RAGQueryLoader and Feature/Graph Store workflow as directed by @wsad1
Note: NVIDIA CI will run all unit tests (including these rag related ones) as well as the full txt2kg_rag.py example everytime we update our NVIDIA container. This will run on H100, B100, and A100 a few diff skus. Then NVIDIA QA also runs on almost every hardware SKU before each bi-monthly release.

(closed subPR: https://github.com/pyg-team/pytorch_geometric/pull/10368)
I had broken this PR down but then matthias said I could merge if he hasnt reviewed before i came back from vacay since we have tested and reviewed this so thoroughly at NVIDIA and externally on top of my involvement in most of the llm features of PyG.

NVIDIA's backing of this project spans many orgs all the way to the top and has fortune500 names adopting it left and right. Getting this merged in is the first step towards cementing PyG as THE framework for anything "Graph" at NVIDIA and I am leading this charge. We will continue to optimize as time goes and treat any issues that come up as p0s to fix asap. But merging this is an important first step.

### DIFF
--- a/torch_geometric/llm/utils/graph_store.py
+++ b/torch_geometric/llm/utils/graph_store.py
@@ -10,7 +10,6 @@ from torch_geometric.sampler import (
     NodeSamplerInput,
     SamplerOutput,
 )
-from torch_geometric.typing import EdgeTensorType
 from torch_geometric.utils import index_sort
 
 # A representation of an edge index, following the possible formats:
@@ -19,7 +18,7 @@ from torch_geometric.utils import index_sort
 #    * COO: (row, col)
 #    * CSC: (row, colptr)
 #    * CSR: (rowptr, col)
-EdgeTensorType = Union[Tensor, Tuple[Tensor, Tensor]]
+_EdgeTensorType = Union[Tensor, Tuple[Tensor, Tensor]]
 
 
 class NeighborSamplingRAGGraphStore(LocalGraphStore):
@@ -116,7 +115,7 @@ class NeighborSamplingRAGGraphStore(LocalGraphStore):
         return ret
 
     @property
-    def edge_index(self) -> EdgeTensorType:
+    def edge_index(self) -> _EdgeTensorType:
         """Gets the edge index of the graph.
 
         :return: The edge index as a tensor.
@@ -124,7 +123,7 @@ class NeighborSamplingRAGGraphStore(LocalGraphStore):
         return self.get_edge_index(*self.edge_idx_args, **self.edge_idx_kwargs)
 
     def put_edge_index(  # type: ignore[no-untyped-def]
-            self, edge_index: EdgeTensorType, *args, **kwargs) -> bool:
+            self, edge_index: _EdgeTensorType, *args, **kwargs) -> bool:
         """Stores an edge index in the graph store.
 
         :param edge_index: The edge index to store.
@@ -139,7 +138,7 @@ class NeighborSamplingRAGGraphStore(LocalGraphStore):
 
     # HACKY
     @edge_index.setter  # type: ignore
-    def edge_index(self, edge_index: EdgeTensorType) -> None:
+    def edge_index(self, edge_index: _EdgeTensorType) -> None:
         """Sets the edge index of the graph.
 
         :param edge_index: The edge index to set.


torch_geometric/typing.py
```diff
@@ -404,12 +404,15 @@ def __reduce__(self) -> tuple[Any, Any]:
...
-EdgeTensorType = Tuple[Tensor, Tensor]
+"""
+A representation of an edge index, following the possible formats:
+   * default: Tensor, size = [2, num_edges]
+   *     Tensor[0, :] == row, Tensor[1, :] == col
+   * COO: (row, col)
+   * CSC: (row, colptr)
+   * CSR: (rowptr, col)
+"""
+EdgeTensorType = Union[Tensor, Tuple[Tensor, Tensor]]
```
akihironitta: I haven't looked into the codebase to see where we use this type alias, but are you sure that this change is correct? Could we revert this?

.github/workflows/linting.yml
```diff
@@ -49,4 +50,4 @@ jobs:
       - name: Check type hints
         if: steps.changed-files-specific.outputs.only_changed != 'true'
         run: |
-          uv run --no-project mypy --cache-dir=/dev/null
+          uv run --no-project mypy --cache-dir=/dev/null --exclude llm
```
akihironitta: Let's set this in `pyproject.toml` so every dev is on the same page.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Adds TXT2KG class with examples ([#9992](https://github.com/pyg-team/pytorch_geometric/pull/9992))
```
```suggestion
- Added `torch_geometric.llm` and its examples ([#10436](https://github.com/pyg-team/pytorch_geometric/pull/10436))
```

akihironitta (commented):
LGTM as per offline discussion. I left some final comments before merging.

.github/workflows/linting.yml
puririshi98: sorry its not clear to me where to set this in pyproject.toml, could u suggest the required change?

torch_geometric/typing.py
puririshi98: the usage of this is for 
   * default: Tensor, size = [2, num_edges]
   *     Tensor[0, :] == row, Tensor[1, :] == col
we use this in our Graph Store implementation. I think this may be useful throughout pyg otherwise it forces users to only use Tuple[Tensor, Tensor] which is less commonly used nowadays than Tensor[2, num_edges]

If this has to be changed i can put this only where we use it but i think this has value to users in PyG, lmk what you want me to do.

akihironitta: Thanks, Rishi. I’ll check existing usage tomorrow.

.github/workflows/linting.yml
akihironitta: Can you add it to this list? https://github.com/pyg-team/pytorch_geometric/blob/fe2a16b18e0ecc4f4c8ad20001f31338dafa1cb3/pyproject.toml#L134

puririshi98: ty changed

torch_geometric/typing.py
akihironitta: I had a look at the codebase, and changing the type here means that we need to make sure the type is correct all over the codebase (, which is hard to detect in our CI because `mypy` is essentially disabled in `pyproject.toml`). An example of such a break is here:
https://github.com/pyg-team/pytorch_geometric/blob/fa6c220839801a0c426c2a6021aba0a990ac6921/torch_geometric/data/hetero_data.py#L1160-1179

I'll push commits directly now to revert this change.

akihironitta (approved):
As per our offline discussion amongst @rusty1s @wsad1 @puririshi98 and myself, we have decided to create a new subpackage `torch_geometric.llm` and assign @puririshi98  as its codeowner so that NVIDIA and the PyG community can iterate on the integration much more quickly even without needing to guarantee the same standard as the PyG core, e.g., thorough test coverage, documentation and code quality.

Thanks again @puririshi98 and the team for your patience and the exciting work! 🚀

PR #10439, opened 2025-09-05
[Refactor] Migrate ogbn_train_cugraph example to distributed cuGraph.

drivanov:
The single-GPU example `ogbn_train_cugraph.py` needs to be updated to use the distributed version of cuGraph.


puririshi98 (changes requested):
actualy lets have @alexbarghi-nv review

PR #10453, opened 2025-09-15
Improve Python 3.13 compatibility, MPS support, feature store equality, and test for k_hop_subsets_rough

DhyeyMavani2003:
### Summary
This PR fixes a Python 3.13 typing regression, improves Apple MPS support in encodings and scatter, and makes `TensorAttr` equality robust.

### Changes
- Python 3.13 typing compatibility:
  - Update `torch_geometric/inspector.py` to call `typing._eval_type(..., type_params=())` on 3.13+, with safe fallback for older versions.
- Robust equality for feature store metadata:
  - Implement `TensorAttr.__eq__` in `torch_geometric/data/feature_store.py` to safely compare `torch.Tensor`, `numpy.ndarray`, and `slice` fields, avoiding ambiguous boolean errors.
- MPS support for positional encoding:
  - In `torch_geometric/nn/encoding.py`, compute `logspace` on CPU, register as a non-persistent buffer, and move to the input device during `forward`. Resolves MPS missing op for `aten::logspace`.
- Explicit MPS behavior for scatter min/max:
  - In `torch_geometric/utils/_scatter.py`, raise `NotImplementedError` for `reduce` in [`min`, `max`, `amin`, `amax`] on MPS to match expected behavior on that backend.

### Rationale
- Fixes a `DeprecationWarning` treated as error on Python 3.13 when inspecting signatures via `typing._eval_type`.
- Prevents `RuntimeError: Boolean value of Tensor with more than one value is ambiguous` in equality checks for `TensorAttr`.
- Ensures positional encoding works on MPS by avoiding unsupported ops.
- Makes scatter’s min/max behavior deterministic on MPS and aligned with tests.

### Tests
- `test/utils/test_total_influence.py`: 2 passed (Python 3.13 typing fix validated)
- `test/data/test_feature_store.py::test_feature_store`: passed (robust `TensorAttr.__eq__`)
- `test/nn/test_encoding.py` on CPU and MPS: all passed (MPS-safe positional encoding)
- `test/utils/test_scatter.py::test_scatter_backward`: all passed, including `[min-mps]` and `[max-mps]` (explicit MPS handling)

### Backward compatibility
- `TensorAttr.__eq__` now defines explicit equality semantics (functional improvement; no breaking API change).
- `PositionalEncoding.frequency` is non-persistent and device-agnostic at runtime; behavior is unchanged aside from improved MPS support.

### DIFF
--- a/torch_geometric/data/feature_store.py
+++ b/torch_geometric/data/feature_store.py
@@ -65,6 +65,30 @@ class TensorAttr(CastMixin):
 
     # Convenience methods #####################################################
 
+    def __eq__(self, obj: object) -> bool:  # type: ignore[override]
+        r"""Safely compares two :class:`TensorAttr` objects.
+        Handles tensor/ndarray/slice comparisons without triggering ambiguous
+        boolean value errors.
+        """
+        if not isinstance(obj, TensorAttr):
+            return False
+
+        def _equal(a: Any, b: Any) -> bool:
+            # Handle torch.Tensor
+            if isinstance(a, torch.Tensor) and isinstance(b, torch.Tensor):
+                return torch.equal(a, b)
+            # Handle numpy.ndarray
+            if isinstance(a, np.ndarray) and isinstance(b, np.ndarray):
+                return bool(np.array_equal(a, b))
+            # Handle slice objects
+            if isinstance(a, slice) and isinstance(b, slice):
+                return (a.start, a.stop, a.step) == (b.start, b.stop, b.step)
+            return a == b
+
+        return (_equal(self.group_name, obj.group_name)
+                and _equal(self.attr_name, obj.attr_name)
+                and _equal(self.index, obj.index))
+
     def is_set(self, key: str) -> bool:
         r"""Whether an attribute is set in :obj:`TensorAttr`."""
         assert key in self.__dataclass_fields__
--- a/torch_geometric/data/feature_store.py
+++ b/torch_geometric/data/feature_store.py
@@ -65,10 +65,16 @@ class TensorAttr(CastMixin):
 
     # Convenience methods #####################################################
 
-    def __eq__(self, obj: object) -> bool:  # type: ignore[override]
+    def equal(self, obj: 'TensorAttr') -> bool:
         r"""Safely compares two :class:`TensorAttr` objects.
         Handles tensor/ndarray/slice comparisons without triggering ambiguous
         boolean value errors.
+
+        Args:
+            obj (TensorAttr): The other TensorAttr object to compare with.
+
+        Returns:
+            bool: True if both TensorAttr objects are equal, False otherwise.
         """
         if not isinstance(obj, TensorAttr):
             return False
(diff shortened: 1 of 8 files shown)


torch_geometric/inspector.py
```diff
@@ -430,7 +430,13 @@ def eval_type(value: Any, _globals: Dict[str, Any]) -> Type:
...
     if isinstance(value, str):
         value = typing.ForwardRef(value)
-    return typing._eval_type(value, _globals, None)  # type: ignore
+    # Python 3.13 deprecates calling `_eval_type` without `type_params`.
+    # Prefer the new signature and gracefully fall back for older versions.
+    try:  # Python >= 3.13
+        return typing._eval_type(value, _globals, None,
+                                 type_params=())  # type: ignore[attr-defined]
+    except TypeError:  # Python < 3.13
+        return typing._eval_type(value, _globals, None)  # type: ignore
```
akihironitta: Could we avoid try-except if we know how to branch the code path with running Python version? I think we can use `sys.version`.

akihironitta: Also, do you know how to reproduce the issue? It'd be very nice if we also add a test case in this PR.

torch_geometric/data/feature_store.py
```diff
@@ -65,6 +65,30 @@ class TensorAttr(CastMixin):
...
+        if not isinstance(obj, TensorAttr):
+            return False
+
+        def _equal(a: Any, b: Any) -> bool:
+            # Handle torch.Tensor
+            if isinstance(a, torch.Tensor) and isinstance(b, torch.Tensor):
+                return torch.equal(a, b)
+            # Handle numpy.ndarray
+            if isinstance(a, np.ndarray) and isinstance(b, np.ndarray):
+                return bool(np.array_equal(a, b))
```
akihironitta, line 88: Since it's somewhat expensive, I'd vote for avoiding implementing this as `__eq__`. Instead, I'd suggest adding it as a `TensorAttr.equal`.

> Implement `TensorAttr.__eq__` in `torch_geometric/data/feature_store.py` to safely compare `torch.Tensor`, `numpy.ndarray`, and `slice` fields, avoiding ambiguous boolean errors.

Could we write a quick test case?

torch_geometric/nn/encoding.py
```diff
@@ -61,7 +63,9 @@ def reset_parameters(self):
     def forward(self, x: Tensor) -> Tensor:
         """"""  # noqa: D419
         x = x / self.granularity if self.granularity != 1.0 else x
-        out = x.view(-1, 1) * self.frequency.view(1, -1)
+        freq = self.frequency if self.frequency.device == x.device else \
+            self.frequency.to(x.device)
+        out = x.view(-1, 1) * freq.view(1, -1)
```
akihironitta: > In `torch_geometric/nn/encoding.py`, compute logspace on CPU, register as a non-persistent buffer, and move to the input device during `forward`. Resolves MPS missing op for `aten::logspace`.

I don't think it makes sense to block the device just to avoid the MPS issue. Also, doesn't `PositionalEncoding(...).to('mps')` move the buffer to mps anyway, no?

akihironitta (commented):
@DhyeyMavani2003 Thank you very much for your contributions! It'd be great if we could have tests for these changes.

torch_geometric/inspector.py
DhyeyMavani2003: That's a great idea. Will be there in the next commit!

torch_geometric/data/feature_store.py
DhyeyMavani2003: Sounds great. I will add a quick test as well.

torch_geometric/nn/encoding.py
DhyeyMavani2003: @akihironitta You're completely right - I was overcomplicating it! I've removed the try-except and device checking in forward. Now it just checks `if device.type == 'mps'` upfront, uses persistent buffers (so `.to('mps')` works as you noted), and calls `self.to(device)` after init. Much simpler and follows standard PyTorch patterns. All the MPS tests are passing now.

DhyeyMavani2003:
Thank you for the thorough review; I just addressed all the comments! Can you please take a look again?

PR #10456, opened 2025-09-16
Test `torch_geometric.llm` in CI

puririshi98:
were issues in past that seemed unrelated, will fix if still there or hopefully just green CI


puririshi98:
need to make some change to LLM to test that it works

puririshi98:
it passes on my machine, not sure why fails here

puririshi98:
> Do we really need the `gc.collect` and `torch.cuda.empty_cache` calls?

was a feeble attempt at fixing the mysterious segfault (ive seen them fix similar issues before in my pytorch core work), will clean them up since their not needed. will do a clean PR when i have a solution to segfault

puririshi98:
okay now i know its the gretriever test causing is at least one of the tests causing the exit code 139

PR #10467, opened 2025-09-22
Dynamic batching: Check for both max num edges and max num nodes when adding graph.

speckhard:
The dynamic batching sampler currently is fed a maximum number of edges **or** nodes. This maximum number is used as a check when deciding whether to add a graph to a mini-batch. This PR changes the input to the class to be a maximum number of edges and a maximum number of nodes. The idea here is that both numbers are important when dictating the size of memory that a batch will occupy.

More discussion is found [here](https://github.com/pyg-team/pytorch_geometric/discussions/10466)


speckhard:
Fix `DynamicBatchSampler` tests and bugs for dynamic batching that includes both nodes/edges budget. The memory budget for a batch of nodes should be dependent not only on the number of graphs and number of nodes in the batch but also on the number of edges.

- Update tests from old `max_num`/`mode` API to `max_num_nodes`/`max_num_edges`
- Fix empty-batch guard (`or` -> `and`) that broke batching with edgeless data
- Fix `skip_too_big` not incrementing `num_processed`, causing duplicate samples
- Allow `max_num_edges=0` for fully unconnected graphs
- Add tests for dual constraints, validation, and edge skipping

Detailed Fixes

1. Empty-batch guard (or → and)

Issue: The previous logic if current_num_nodes == 0 or current_num_edges == 0 failed for edgeless data (where edges are always 0). The sampler thought the batch was permanently "empty" and refused to yield.

Fix: Changed to and. A batch is only empty if both nodes and edges are zero.

2. skip_too_big duplication bug

Issue: When skip_too_big=True triggered a continue, num_processed was not incremented. This caused the loop to restart at the same index, processing the same sample twice.

Note: This bug appears to have existed in the upstream code but was masked because previous tests always used num_steps, which stopped the iterator before the duplicates appeared. The new tests uncovered this latent issue.

speckhard:
Any chance someone might be able to add an opinion here or in the discussion? Tests are added and now passing. Total lines of code changed is very low.

PR #10478, opened 2025-10-03
Fix `test_train` to not rely on 'Sanity Checking' stdout in multi-GPU runs

drivanov:
The test `test_train` previously asserted on the presence of the "Sanity Checking" message in stdout. This was brittle because in multi-GPU/DistributedDataParallel runs, **only rank 0 prints this message**, so tests running on other ranks failed. 

This PR updates the test to:
- Remove the fragile stdout assertion.
- Assert trainer state (`!trainer.sanity_checking`, `current_epoch >= 0`).
- Use LoggerCallback to verify that both training and validation ran.

This makes the test deterministic and robust across single-GPU, multi-GPU, and CI environments.

### DIFF
--- a/test/graphgym/test_graphgym.py
+++ b/test/graphgym/test_graphgym.py
@@ -16,7 +16,7 @@ from torch_geometric.graphgym.config import (
     set_run_dir,
 )
 from torch_geometric.graphgym.loader import create_loader
-from torch_geometric.graphgym.logger import LoggerCallback, set_printing
+from torch_geometric.graphgym.logger import set_printing
 from torch_geometric.graphgym.model_builder import create_model
 from torch_geometric.graphgym.models.gnn import FeatureEncoder, GNNStackStage
 from torch_geometric.graphgym.models.head import GNNNodeHead
@@ -169,6 +169,7 @@ def destroy_process_group():
     if torch.distributed.is_initialized():
         torch.distributed.destroy_process_group()
 
+ggg = 0
 
 @onlyOnline
 @onlyLinux
@@ -194,12 +195,30 @@ def test_train(destroy_process_group, tmp_path, capfd):
     loaders = create_loader()
     model = create_model()
     cfg.params = params_count(model)
+    
+    # --- minimal logger callback that collects logs ---
+    class LoggerCallback(pl.Callback):
+        def __init__(self):
+            super().__init__()
+            self.logged = []
+
+        def on_train_batch_end(self, trainer, pl_module, outputs, batch, batch_idx):
+            self.logged.append({"type": "train", "step": trainer.global_step})
+
+        def on_validation_batch_end(
+            self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx=0
+        ):
+            self.logged.append({"type": "val", "step": trainer.global_step})
+
     logger = LoggerCallback()
-    trainer = pl.Trainer(max_epochs=1, max_steps=4, callbacks=logger,
-                         log_every_n_steps=1)
+    trainer = pl.Trainer(max_epochs=1, max_steps=4, callbacks=[logger],
+                         log_every_n_steps=1, enable_progress_bar=False)
     train_loader, val_loader = loaders[0], loaders[1]
     trainer.fit(model, train_loader, val_loader)
 
-    out, err = capfd.readouterr()
-    assert 'Sanity Checking' in out
-    assert 'Epoch 0:' in out
+    assert not trainer.sanity_checking
+    assert trainer.current_epoch >= 0
+    # ensure both train and val batches were seen
+    types = {entry["type"] for entry in logger.logged}
+    assert "val" in types, "Validation did not run"
+    assert "train" in types, "Training did not run"


test/graphgym/test_graphgym.py
```diff
@@ -170,6 +170,9 @@ def destroy_process_group():
         torch.distributed.destroy_process_group()
 
 
+ggg = 0
```
puririshi98: whats this for? can we remove it? otherwise lgtm

puririshi98 (changes requested):

test/graphgym/test_graphgym.py
drivanov: Done

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed Made `test_train` robust to multi-GPU runs by replacing stdout assertions with trainer state checks ([#10478](https://github.com/pyg-team/pytorch_geometric/pull/10478))
```
akihironitta: No need to add a changelog entry for non user facing chnages.
```suggestion
```

test/graphgym/test_graphgym.py
```diff
@@ -194,12 +194,30 @@ def test_train(destroy_process_group, tmp_path, capfd):
...
-                         log_every_n_steps=1)
+    trainer = pl.Trainer(max_epochs=1, max_steps=4, callbacks=[logger],
+                         log_every_n_steps=1, enable_progress_bar=False)
     train_loader, val_loader = loaders[0], loaders[1]
     trainer.fit(model, train_loader, val_loader)
 
-    out, err = capfd.readouterr()
-    assert 'Sanity Checking' in out
-    assert 'Epoch 0:' in out
+    assert not trainer.sanity_checking
```
akihironitta: If I understand this diff correctly, this `assert not trainer.sanity_checking` tests nothing. Previously, it tested that the code ran sanity checking, but with this PR, it doesn't. `sanity_checking` is only `True` when the trainer is running a few validation steps at the start of fit.

akihironitta: Can we just remove this assert? Testing whether it performed sanity checking doens't make much sense as a test anyway.

akihironitta (approved):

test/graphgym/test_graphgym.py
drivanov: This line was suggested by ChatGPT. The following explanation is adapted from its reasoning:

         assert not trainer.sanity_checking
could fail (i.e., `trainer.sanity_checking == True` after `trainer.fit()` finishes).

### **What would cause it**
**If training never moved past the sanity check**
- Example: an error in the training loop caused Lightning to stop during sanity checking.
- Then `trainer.fit()` would return early, and `trainer.sanity_checking` could remain **True**.

**If sanity checking was the only phase run**
- If `max_steps=0` or `max_epochs=0`, training effectively does nothing after the sanity check.
- Depending on version, `trainer.sanity_checking` might stay **True.**

**Misconfigured validation loop**
- If validation loader is empty or raises errors, Lightning can get stuck in a state where it ends sanity checking improperly.
- Edge cases in earlier Lightning versions sometimes left the flag not reset.

**Bug in Lightning**
- If the `sanity_checking` flag isn’t toggled back (it’s set **True** at the start of sanity check, **False** afterward).
- Rare, but regressions could happen across versions.

### **Normal expectation**

- After a successful `trainer.fit()`, **training always goes past sanity check**, so `trainer.sanity_checking` must be **False**.
- Therefore, this assertion is a robust guarantee that training ran at least one epoch/step beyond the pre-validation check.


### ⚡ **So in short:**
It would fail only if `trainer.fit(`) exited abnormally (error, misconfig, or 0 training steps) or Lightning had a bug in resetting the flag.

As we all know, ChatGPT isn't always accurate. Unfortunately, in this case, I am unable to assess the accuracy of its claims. Please advise.

puririshi98 (approved):
LGTM, cc @akihironitta

test/graphgym/test_graphgym.py
akihironitta: > After a successful `trainer.fit()`, training always goes past sanity check, so `trainer.sanity_checking` must be `False`.

@drivanov I'm not sure if you had a chance to read the LLM output, but asserting on `sanity_checking` is too loose to check a successful `trainer.fit()`.

test/graphgym/test_graphgym.py
```diff
@@ -194,12 +194,29 @@ def test_train(destroy_process_group, tmp_path, capfd):
...
+
+        def on_validation_batch_end(self, trainer, pl_module, outputs, batch,
+                                    batch_idx, dataloader_idx=0):
+            self.logged.append({"type": "val", "step": trainer.global_step})
+
     logger = LoggerCallback()
-    trainer = pl.Trainer(max_epochs=1, max_steps=4, callbacks=logger,
-                         log_every_n_steps=1)
+    trainer = pl.Trainer(max_epochs=1, max_steps=4, callbacks=[logger],
+                         log_every_n_steps=1, enable_progress_bar=False)
```
akihironitta:
```suggestion
    trainer = pl.Trainer(max_epochs=2, max_steps=4, callbacks=[logger],
                         log_every_n_steps=1, enable_progress_bar=False)
```

test/graphgym/test_graphgym.py
```diff
@@ -194,12 +194,29 @@ def test_train(destroy_process_group, tmp_path, capfd):
...
     train_loader, val_loader = loaders[0], loaders[1]
     trainer.fit(model, train_loader, val_loader)
 
-    out, err = capfd.readouterr()
-    assert 'Sanity Checking' in out
-    assert 'Epoch 0:' in out
+    assert trainer.current_epoch >= 0
```
akihironitta:
```suggestion
    assert trainer.current_epoch > 0
```

PR #10479, opened 2025-10-03
Add local LLM support and uncertainty estimation testing for PyG RAG

leochlon:
- Fixed import path
Corrected torch_geometric.nn.LLM → torch_geometric.llm.models.LLM to match the upstream module layout and avoid ImportError.

- Local TXT2KG support
Added --use_local_txt2kg flag so users can choose between OpenAI/NIM APIs and local Hugging Face models for knowledge graph extraction. Enables fully offline pipelines, avoids API costs, and improves reproducibility.

- Checkpoint validation
Properly validates cached triples when switching between local vs. API models. Prevents stale or mismatched KG states from silently propagating into training/inference.

- GPU utilization
Passes n_gpus parameter to TXT2KG’s internal LLM, ensuring multi-GPU hardware is actually used.

- Uncertainty-aware LLM
Extended LLM(..., uncertainty_estim=True) to compute EDFL/B2T/ISR metrics before generation:

1. Returns per-item uncertainty metrics: ISR, Δ̄, B2T, RoH bound, priors.
2. Optionally abstains ("[ABSTAIN]") if ISR < threshold.
3. During training, masks labels for low-ISR items, enabling GNN+LLM fine-tuning with calibrated refusal rather than random guessing

tests
```
# Test with actual context to see higher answer rates
python test/test_uncertainty_techqa.py \
    --use_uncertainty \
    --use_mock_context \
    --context_length 200 \
    --num_questions 20 \
    --abstain_on_low_isr \
    --isr_threshold 1.0 \
    --verbose

# Adjust threshold to be more/less conservative
python test/test_uncertainty_techqa.py \
    --use_uncertainty \
    --num_questions 20 \
    --isr_threshold 2.5 \
    --abstain_on_low_isr
```
Full pipeline with local models:
```
python examples/llm/txt2kg_rag.py \
    --use_local_txt2kg \
    --txt2kg_model "mistralai/Mistral-7B-Instruct-v0.3" \
    --llm_generator_name "mistralai/Mistral-7B-Instruct-v0.3" \
    --use_x_percent_corpus 0.5 \
    --num_gpus 1
```

### DIFF
--- a/torch_geometric/datasets/medshapenet.py
+++ b/torch_geometric/datasets/medshapenet.py
@@ -96,7 +96,8 @@ class MedShapeNet(InMemoryDataset):
 
         subset = []
         for dataset in list_of_datasets:
-            self.newpath = self.root + '/' + dataset.split("/")[1]
+            parts = dataset.split("/")
+            self.newpath = self.root + '/' + parts[1 if len(parts) > 1 else 0]
             if not os.path.exists(self.newpath):
                 os.makedirs(self.newpath)
             stl_files = msn_instance.dataset_files(dataset, '.stl')


puririshi98 (changes requested):
overall looks good, ill look deeper tmrw after your demo.
Simple first change is move the unit test to test/llm/models/test_llm.py as an additional unit test instead of this seperate file.

puririshi98:
plz make sure CI jobs are all green too

puririshi98:
should also updat eCHANGELOG.md

PR #10480, opened 2025-10-03
Enable local TXT2KG + uncertainty-aware LLM with EDFL/ISR

leochlon:
Fixed import path
Corrected torch_geometric.nn.LLM → torch_geometric.llm.models.LLM to match the upstream module layout and avoid ImportError.

Local TXT2KG support
Added --use_local_txt2kg flag so users can choose between OpenAI/NIM APIs and local Hugging Face models for knowledge graph extraction. Enables fully offline pipelines, avoids API costs, and improves reproducibility.

Checkpoint validation
Properly validates cached triples when switching between local vs. API models. Prevents stale or mismatched KG states from silently propagating into training/inference.

GPU utilization
Passes n_gpus parameter to TXT2KG’s internal LLM, ensuring multi-GPU hardware is actually used.

Uncertainty-aware LLM
Extended LLM(..., uncertainty_estim=True) to compute EDFL/B2T/ISR metrics before generation:

Returns per-item uncertainty metrics: ISR, Δ̄, B2T, RoH bound, priors.
Optionally abstains ("[ABSTAIN]") if ISR < threshold.
During training, masks labels for low-ISR items, enabling GNN+LLM fine-tuning with calibrated refusal rather than random guessing
tests

`
RAG_TEST=1 pytest test/llm/models/test_llm.py::test_llm_uncertainty -v
`


torch_geometric/llm/models/llm.py
```diff
@@ -135,22 +181,158 @@ def __init__(
...
+                    self._OllamaBackend = OllamaBackend
+                    self._AnthropicBackend = AnthropicBackend
+                except Exception:
+                    self._HFBackend = self._OllamaBackend = (
+                        self._AnthropicBackend) = None
+            except Exception:
+                self.uncertainty_estim = False
+                warnings.warn(
+                    "Uncertainty requested but `hallucination_toolkit` not "
+                    "importable. Proceeding without uncertainty.",
```
puririshi98, line 250: can you add a copypastable string for the pip install users should use to get it set up

puririshi98 (changes requested):

PR #10491, opened 2025-10-08
Add helpful error message when running rag_loader.py directly

Shubhf:
This PR improves developer experience by adding a clear error message at the top of rag_loader.py. Previously, running the file directly using:
`python torch_geometric/llm/rag_loader.py`
would lead to a confusing ModuleNotFoundError. This happens because direct execution breaks relative imports within the package.

The new message explicitly instructs users to run the module correctly via:
`python -m torch_geometric.llm.rag_loader`
This provides better guidance for new contributors working with the llm examples and aligns with Python’s recommended module execution practices.


puririshi98:
i dont think we need this (since we don't have it on any other pyg python files). generaly everything under torch_geometric are modules for import, not to be used as main. 
I appreciate the effort and do encourage you to continue submitting PRs that you think can help.

PR #10512, opened 2025-10-27
Fix `.llm` import trigger of `.distributed` depr warning

puririshi98:


puririshi98:
> I don't think that will prevent the warning since you still depend on `torch_geometric.distributed`.

im confused, where do you see the warning on your end. I tested importing pyg and dont see any warnings
```
Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch_geometric
>>> 
```

rusty1s:
you don't see them since DeprecationWarnings are not shown by default. It is visible in `pytest` though.

puririshi98:
> you don't see them since DeprecationWarnings are not shown by default. It is visible in `pytest` though.

is there a quick way to turn that on for interactive debugging?

rusty1s:
just call the warning as `warn("foo")`, without the stacklevel and category.

puririshi98:
> just call the warning as `warn("foo")`, without the stacklevel and category.

okay since we still use local graph/feature store i put the warnings where most would see them for initializing loaders because if your doing that you would probably benefit from using cugraph instead. hopefully this works for you. lmk if you recommend another way

rusty1s (approved):
Looks good. Do we need the warning in all three places?

torch_geometric/distributed/__init__.py
```diff
@@ -1,6 +1,7 @@
 from .dist_context import DistContext
 from .local_feature_store import LocalFeatureStore
 from .local_graph_store import LocalGraphStore
+
```
akihironitta:
```suggestion
```

PR #10522, opened 2025-11-06
make `coalesce` compile friendly

ValterH:
Update coalesce to be torch compiled without graph breaks.

### DIFF
--- a/torch_geometric/utils/_coalesce.py
+++ b/torch_geometric/utils/_coalesce.py
@@ -131,8 +131,8 @@ def coalesce(  # noqa: F811
     num_edges = edge_index[0].size(0)
     num_nodes = maybe_num_nodes(edge_index, num_nodes)
 
-    if num_nodes * num_nodes > torch_geometric.typing.MAX_INT64:
-        raise ValueError("'coalesce' will result in an overflow")
+    torch._check(num_nodes * num_nodes < torch_geometric.typing.MAX_INT64,
+                 "'coalesce' will result in an overflow")
 
     idx = edge_index[0].new_empty(num_edges + 1)
     idx[0] = -1
@@ -163,7 +163,7 @@ def coalesce(  # noqa: F811
     mask = idx[1:] > idx[:-1]
 
     # Only perform expensive merging in case there exists duplicates:
-    if mask.all():
+    if not torch.compiler.is_compiling() and mask.all():
         if edge_attr is None or isinstance(edge_attr, Tensor):
             return edge_index, edge_attr
         if isinstance(edge_attr, (list, tuple)):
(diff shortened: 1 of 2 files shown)


torch_geometric/utils/_coalesce.py
```diff
@@ -163,7 +163,7 @@ def coalesce(  # noqa: F811
     mask = idx[1:] > idx[:-1]
 
     # Only perform expensive merging in case there exists duplicates:
-    if mask.all():
+    if not torch.compiler.is_compiling() and mask.all():
```
ValterH: My thinking was: better to perform merging inside a kernel than to cause a graph break due to data-dependent branching. No change in eager execution.

torch_geometric/utils/_coalesce.py
```diff
@@ -131,8 +131,8 @@ def coalesce(  # noqa: F811
     num_edges = edge_index[0].size(0)
     num_nodes = maybe_num_nodes(edge_index, num_nodes)
 
-    if num_nodes * num_nodes > torch_geometric.typing.MAX_INT64:
-        raise ValueError("'coalesce' will result in an overflow")
+    torch._check(num_nodes * num_nodes < torch_geometric.typing.MAX_INT64,
+                 "'coalesce' will result in an overflow")
```
akihironitta: `<` should be `<=`
```suggestion
    torch._check(
        num_nodes * num_nodes <= torch_geometric.typing.MAX_INT64,
        "'coalesce' will result in an overflow",
    )
```

akihironitta (commented):
Any chance you could add a quick test case that reproduces the issue?

ValterH:
@akihironitta I added a test case, but had to remove the error message as `torch._check` requires the message to be callable and passing a lambda function causes a graph break during compilation.

torch_geometric/utils/_coalesce.py
rusty1s: Please use `torch_geometric.is_compiling()`.

torch_geometric/utils/_coalesce.py
```diff
@@ -131,7 +131,10 @@ def coalesce(  # noqa: F811
     num_edges = edge_index[0].size(0)
     num_nodes = maybe_num_nodes(edge_index, num_nodes)
 
-    if num_nodes * num_nodes > torch_geometric.typing.MAX_INT64:
+    if not torch.jit.is_scripting():
+        torch._check_value(num_nodes *
+                           num_nodes <= torch_geometric.typing.MAX_INT64)
+    elif num_nodes * num_nodes > torch_geometric.typing.MAX_INT64:
```
rusty1s: Can you explain the reasoning here?

rusty1s (approved):

torch_geometric/utils/_coalesce.py
ValterH: I did not want to break support for TorchScript and `torch._check_value` is not supported in TorchScript raising: `RuntimeError: Unknown builtin op: aten::_check_value.` Rewrote the logic to make this clearer.

PR #10595, opened 2026-02-05
 Fix runtime crash due to Half/Float mismatch in GNN+LLM training and inference.

drivanov:
**Summary**

This PR fixes a runtime crash observed in `examples/llm/txt2kg_rag.py` caused by FP16/FP32 dtype mismatches between graph node features and GNN model parameters during training, validation, and inference. 

The issue manifested as errors like:
```
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: Half != float
```
originating from GATConv / F.linear.

**Root cause**

- Graph node features (batch.x, and sometimes edge_attr) were silently produced in FP16 by upstream components (e.g. sentence embeddings, RAG utilities).
- GNN parameters (e.g. GATConv weights) are FP32 by default.
- PyTorch does not allow mixed FP16/FP32 inputs in F.linear, causing runtime failures.
- The problem affected multiple execution paths:
    - training
    - validation
    - inference (which bypasses get_loss)

AMP/autocast does not resolve this issue because the tensors were already FP16 before entering the model.

**Fix**

This PR normalizes graph tensor dtypes to FP32 immediately before GNN execution by explicitly casting:
- batch.x
- batch.edge_attr (when present)

A small helper is used to ensure this is applied consistently across:
- training
- validation
- inference

This guarantees dtype alignment at the model boundary and prevents future regressions when upstream components change precision.

The following `FutureWarning` was also corrected.
```
/workspace/examples/llm/txt2kg_rag.py:717: FutureWarning: torch.cuda.reset_max_memory_allocated now calls 
torch.cuda.reset_peak_memory_stats, which resets /all/ peak memory stat
```

### DIFF
--- a/examples/llm/g_retriever.py
+++ b/examples/llm/g_retriever.py
@@ -85,6 +85,12 @@ def load_params_dict(model, save_path):
     return model
 
 
+def normalize_batch_dtype(batch):
+    batch.x = batch.x.float()
+    if hasattr(batch, "edge_attr") and batch.edge_attr is not None:
+        batch.edge_attr = batch.edge_attr.float()
+
+
 def get_loss(model, batch, model_save_name="gnn+llm") -> Tensor:
     """Compute the loss for a given model and batch of data.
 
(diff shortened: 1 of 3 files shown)


puririshi98 (approved):
lgtm

PR #10600, opened 2026-02-11
Fix missing hyperedge weight application in HypergraphConv

Mr-Neutr0n:
## Summary

`HypergraphConv` documents the formula `X' = D^{-1} H W B^{-1} H^T X Theta`, where `W` is the diagonal hyperedge weight matrix. However, `hyperedge_weight` was only used to compute the node degree matrix `D` -- the weight matrix `W` itself was never applied to scale the intermediate edge representations between the two message passing steps.

This meant that providing non-uniform `hyperedge_weight` had an incomplete effect: it influenced the degree-based normalization (`D`) but did not scale the edge features as the formula requires. This fix applies `W` to the edge features after the node-to-edge aggregation and before the edge-to-node aggregation, correctly implementing the full formula.

**When `hyperedge_weight` is all ones (the default), behavior is unchanged.**

### Changes
- **`torch_geometric/nn/conv/hypergraph_conv.py`**: Apply `hyperedge_weight` to intermediate edge representation between the two `propagate` calls, with correct reshaping for both attention and non-attention modes.
- **`test/nn/conv/test_hypergraph_conv.py`**: Add test verifying that uniform weights match default behavior, and non-uniform weights produce different (correct) results.

## Test plan
- [x] Existing tests pass (output shapes unchanged)
- [x] New test verifies uniform weights match default behavior (`torch.allclose`)
- [x] New test verifies non-uniform weights produce meaningfully different output


Mr-Neutr0n:
Hi maintainers! Friendly ping on this PR — it fixes a real bug where `hyperedge_weight` was only partially applied in `HypergraphConv` (the `W` matrix was missing from the forward pass). The fix is complete with tests, CI is green, and the behavior is unchanged for the default (all-ones) case. Would appreciate a review when anyone has a moment. Thanks!

Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #10623, opened 2026-02-26
Improvement of the txt2kg example.

drivanov:
**Summary**

This PR refactors the multiprocessing logic in `torch_geometric.llm.models.txt2kg.py` to improve reproducibility and stability.

**Key Changes**

- Remove tmp-file IPC
  - Workers now return `List[(s, p, o)]` directly
  - Eliminates `/tmp/outs_for_proc_*` files
  - No filesystem side effects

- Fix spawn-related instability
  - Worker moved to module scope (picklable)
  - Explicit error propagation
  - Prevents silent `SIGTERM` cascades

- Ensure deterministic output
  - Global lexicographic sort of triples after aggregation
  - Output stable across runs and worker counts

**Result**
- Reproducible KG extraction
- Cleaner multiprocessing design
- No temp files
- Clearer failure modes
- No API changes.

Change in `llm.py` eliminates the following warning:
```
`torch_dtype` is deprecated! Use `dtype` instead!
```


puririshi98 (changes requested):
lgtm at a high level but please address these concerns:
1. The retry logic was silently deleted without replacement. The original code had up to 200 inner retries and 5 outer retries. This was there for a reason — NIM API calls over the network fail transiently. The new code has zero retry logic. A single transient HTTP error in any worker will now raise a RuntimeError and abort the entire job. For expensive, long-running KG extraction jobs, this is a significant regression in robustness.

2. _safe_worker swallows exceptions and returns a dict instead of raising. This is an anti-pattern with Pool.map. If a worker raises an exception, Pool.map will propagate it naturally to the main process — that's the correct behavior. The _safe_worker wrapper catches the exception, converts it to a dict, returns it as a "successful" result, and then the main process has to manually check every result for this sentinel value. This is more complex and fragile than just letting Pool.map propagate the exception. It also means if a worker silently returns a partial result before raising, it gets swallowed entirely.
 
3. The chunk distribution has an off-by-one / data loss bug that was present before and is not fixed. meta_chunk_size = int(len(chunks) / num_procs) with integer division means if len(chunks) is not divisible by num_procs, the last few chunks are silently dropped. For example, 10 chunks across 3 workers: sizes are 3, 3, 3 — the 10th chunk is never processed. This was a pre-existing bug but this PR was the right opportunity to fix it (e.g. using np.array_split or adjusting the slicing).

4. flat_triples.sort() sorts tuples lexicographically. This works, but the triples coming out of _parse_n_check_triples are lists, not tuples. Sorting lists also works, but the claim of "deterministic output" is only true within a single Python version and locale — list/tuple sort order is not guaranteed to be stable across Python versions for string content with mixed Unicode. Minor point, but worth noting for a library claiming reproducibility.

Aside from that please confirm you get an end 2 end run of txt2kg_rag on a small subsample using --use_x_percent_corpus

drivanov:
Done. 
Yes, the L1_TXT2KG_RAG_GPU test with the `--use_x_percent_corpus .003` parameter passes without problems.

puririshi98 (approved):
LGTM just make the CI green and we can merge

puririshi98 (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed multiprocessing logic in `txt2kg.py` and two `test_g_retriver*` tests ([#10623](https://github.com/pyg-team/pytorch_geometric/pull/10623))
```
akihironitta: @drivanov We don't need a changelog entry for test changes.
```suggestion
```

akihironitta (approved):

torch_geometric/llm/models/txt2kg.py
```diff
@@ -281,22 +287,44 @@ def _llm_then_python_parse(chunks, py_fn, llm_fn, **kwargs):
...
 
-def _multiproc_helper(rank, in_chunks_per_proc, py_fn, llm_fn, NIM_KEY,
-                      NIM_MODEL, ENDPOINT_URL):
-    out = _llm_then_python_parse(in_chunks_per_proc[rank], py_fn, llm_fn,
-                                 GLOBAL_NIM_KEY=NIM_KEY, NIM_MODEL=NIM_MODEL,
-                                 ENDPOINT_URL=ENDPOINT_URL)
-    torch.save(out, "/tmp/outs_for_proc_" + str(rank))
+def _multiproc_helper(rank, chunks_for_rank, py_fn, llm_fn, NIM_KEY, NIM_MODEL,
+                      ENDPOINT_URL, max_retries=MAX_NIM_RETRIES,
+                      base_delay=BASE_DELAY):
```
akihironitta:
```suggestion
def _multiproc_helper(rank, chunks_for_rank, py_fn, llm_fn, NIM_KEY, NIM_MODEL,
                      ENDPOINT_URL, max_retries=MAX_NIM_RETRIES,
                      base_delay=BASE_DELAY,):
```

torch_geometric/llm/models/txt2kg.py
```diff
@@ -137,61 +141,63 @@ def add_doc_2_KG(
...
-                    except:  # noqa
-                        pass
+        self.relevant_triples[key] = self._extract_relevant_triples(txt)
+
         # Increment the doc_id_counter for the next document
         self.doc_id_counter += 1
 
+    def _extract_relevant_triples(
+            self, txt: str, max_retries: int = MAX_OUTER_RETRIES,
+            retry_delay: float = RETRY_DELAY) -> List[Tuple[str, str, str]]:
```
akihironitta:
```suggestion
            retry_delay: float = RETRY_DELAY,) -> List[Tuple[str, str, str]]:
```

torch_geometric/llm/models/g_retriever.py
```diff
@@ -249,3 +250,11 @@ def __repr__(self) -> str:
                 f'  llm={self.llm},\n'
                 f'  gnn={self.gnn},\n'
                 f')')
+
+    def _align_dtype(self, x: torch.Tensor,
+                     llm_generator: torch.nn.Module) -> torch.Tensor:
```
akihironitta:
```suggestion
                     llm_generator: torch.nn.Module,) -> torch.Tensor:
```

04cb (approved):

drivanov:
@akihironitta :
I'm not entirely sure, but it seems there are some issues with the code coverage analysis script.

I added at least 10 new tests today, but I still see the same missing coverage for 17 lines 
```
Patch coverage is 73.01587% with 17 lines in your changes missing coverage
```
When I try to dig [deeper](https://app.codecov.io/gh/pyg-team/pytorch_geometric/pull/10623?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=pyg-team) and explore the links on that page, I see that 

- Commits have different number of coverage report uploads:
- BASE commit is 174 commits behind HEAD on master
- My latest commit is compared to code that is 11 months old.

 3. I also analyzed the [Missed Lines](https://app.codecov.io/gh/pyg-team/pytorch_geometric/pull/10623/indirect-changes) of

- `torch_geometric/llm/models/llm.py`
- `torch_geometric/llm/models/txt2kg.py`
- `torch_geometric/llm/models/g_retriever.py` 

      and with ChatGPT's help, I created a set of tests for them. When I add `pdb` breakpoints at the appropriate places, I see that the program actually hits them (for instance, [that one](https://app.codecov.io/gh/pyg-team/pytorch_geometric/pull/10623/blob/torch_geometric/llm/models/llm.py#L83)).
      However, the code coverage report still shows them as skipped, even though these new tests were [actually executed](https://github.com/pyg-team/pytorch_geometric/actions/runs/22977308320/job/66708960159?pr=10623):
```
      test/llm/models/test_llm.py::test_llm_prepare_inputs PASSED
      test/llm/models/test_llm.py::test_llm_single_prompt PASSED
      test/llm/models/test_llm.py::test_llm_variable_lengths PASSED 
 ```

akihironitta:
Thank you @drivanov for digging into what the issue is with codecov. I have seen this issue in the past, and it's likely due to their bug or our misconfiguration, so I'll merge this as is. Thank again for sending this patch!

PR #10628, opened 2026-03-04
Add from_relbench utility to convert RelBench databases to HeteroData

AJamal27891:
## Description
This PR is **Part 1 of 4** in splitting the monolithic Warehouse Intelligence system (#10353) into modular, composable pieces, as requested by the core maintainers.

This PR introduces the `from_relbench` utility to `torch_geometric.utils.relbench`. It allows users to convert complex, multi-table databases from the [RelBench](https://relbench.stanford.edu/) (Relational Deep Learning Benchmark) environment directly into PyG's native `HeteroData` format.

**Addressing Assessor Feedback:**
This specifically addresses @wsad1's feedback from #10353: *"why do we need a RelBenchDataset?"*
Based on that guidance, I have completely dropped the custom `RelBenchDataset` class wrapper. Instead, to align with PyG's stateless data processing philosophy and avoid reinventing the wheel, this PR only introduces a pure utility function. It decouples PyG's dataset classes from RelBench's internal state.

## Proposed Changes
- [x] Added `torch_geometric/utils/relbench.py` housing the `from_relbench` conversion utility.
- [x] Added exhaustive unit tests in `test/utils/test_relbench.py` utilizing dummy fallback flags to ensure rapid CI execution without massive database downloads.
- [x] Updated `CHANGELOG.md`.

*(Note: Parts 2, 3, and 4—covering the Warehouse Transforms, SAGEConv multi-task models, and LLM G-Retriever integrations—will follow in subsequent linked PRs once this foundational data layer is approved.)*

---
Ref: #10353 
Partially Closes #9839

PR #10633, opened 2026-03-09
Fix brittle test_summary_with_to_hetero_model assertion broken by tabulate 0.10.0

AJamal27891:
## Summary

`tabulate 0.10.0` (released 2026-03-04) improved its comma-integer detector causing values like `5,824` to be right-aligned instead of left-aligned. This broke `test_summary_with_to_hetero_model`, which asserted against a hardcoded left-aligned string.

This failure was introduced in `master` by `tabulate 0.10.0` and surfaces on any PR running CI against an environment with the latest PyPI release. It was noticed while working on #10628.

## Root Cause

`test_summary_with_to_hetero_model` is the **only** test affected because it is the only one where all rows in the `#Param` column are numeric. When tabulate sees a mixed numeric/text column (e.g. containing `--` dashes) it falls back to text alignment, so all other summary tests are unaffected.

PyG's `pyproject.toml` declares `tabulate` with no version pin, so CI always resolves the latest release from PyPI.

## Fix

Replace the exact string comparison with content-based assertions:
- Layer names checked via `assert layer in result_norm`  
- Parameter counts checked via `assert f'|{param}|' in result_norm` (pipe-bounded to prevent false positives from shape columns like `[100, 16]`)

A `_normalize_table()` helper strips per-cell padding before comparison, making the test permanently robust to future tabulate alignment changes.

```python
# Before — breaks whenever tabulate changes column alignment:
assert summary(model, ...) == expected[1:-1]

# After — tests content, not whitespace:
result_norm = _normalize_table(summary(model, ...))
assert '|5,824|' in result_norm
assert 'GraphModule' in result_norm
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed `test_summary_with_to_hetero_model` failing due to `tabulate 0.10.0` changing numeric column alignment; replaced brittle exact-string assertions with content-based checks using a `_normalize_table` helper
 - Fixed `return_attention_weights: bool` being not respected in `GATConv` and `GATv2Conv` ([#10596](https://github.com/pyg-team/pytorch_geometric/pull/10596))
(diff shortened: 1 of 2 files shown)


AJamal27891:
Closing — this was addressed in #10634. The content-based approach here is more robust long-term, but the immediate issue is resolved. Happy to revisit if the alignment changes again.

PR #10655, opened 2026-04-02
Fix kwargs validation in `test_llm.py::test_get_llm_kwargs`

drivanov:
The implementation of `test_llm.py::test_get_llm_kwargs` is not entirely correct. 
On a machine with less than 640 GB of GPU memory, this test will pass without issues—it will simply return a dictionary containing only (`revision`: `main`). However, on the machines with more 640 GB of GPU memory, the returned dictionary contains additional keys such as `max_memory`, `low_cpu_mem_usage`, `device_map` etc. 

That PR improves the validation of `kwargs`.


puririshi98 (approved):
lgtm

akihironitta (approved):

PR #10657, opened 2026-04-03
Fix `test_batch_norm` to check module attributes instead of relying on string representation

drivanov:
The `test_batch_norm` was recently changed in [PR#10647](https://github.com/pyg-team/pytorch_geometric/pull/10647) (**Fix CI failing with PyTorch 2.12+**), but it still fails with some PyTorh 2.12 versions, and `torch-2.12.0a0+0291f960b6.nvinternal.main.46693258` is one of them.

Current PR:
- Replaces brittle str(norm) checks with direct assertions on num_features, eps, momentum, affine, track_running_stats, weight, and bias.
- Removes dependence on `WITH_PT212` flag.
- Works correctly with both PyG wrapper `BatchNorm` and native `torch.nn.BatchNorm1d`.
- Version-agnostic and future-proof across PyTorch builds.

The fix was tested with :
- torch-2.12.0a0+0291f960b6.nvinternal.main.46693258
- torch 2.10.0a0+a36e1d39eb.nv26.1.42222806

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fix `test_batch_norm` to check module attributes instead of relying on string representation ([#10657](https://github.com/pyg-team/pytorch_geometric/pull/10657))
 - Removed an unnecessary device synchronization in `torch_geometric.utils.softmax` ([#10499](https://github.com/pyg-team/pytorch_geometric/pull/10499))
 - Fixed loading of legacy HuggingFace BERT checkpoints ([#10631](https://github.com/pyg-team/pytorch_geometric/pull/10631))
 - Fixed `return_attention_weights: bool` being not respected in `GATConv` and `GATv2Conv` ([#10596](https://github.com/pyg-team/pytorch_geometric/pull/10596))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
-- Fix `test_batch_norm` to check module attributes instead of relying on string representation ([#10657](https://github.com/pyg-team/pytorch_geometric/pull/10657))
 - Removed an unnecessary device synchronization in `torch_geometric.utils.softmax` ([#10499](https://github.com/pyg-team/pytorch_geometric/pull/10499))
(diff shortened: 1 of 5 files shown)


puririshi98 (approved):
LGTM

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies=[
     "fsspec",
     "jinja2",
     "numpy",
+    "packaging",
```
akihironitta: Isn't this only required for the PyTorch nightly release? I prefer adding it only to the CI step for now.
```suggestion
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fix `test_batch_norm` to check module attributes instead of relying on string representation ([#10657](https://github.com/pyg-team/pytorch_geometric/pull/10657))
```
akihironitta: No need to add a changelog entry for a test change.
```suggestion
```

test/nn/norm/test_batch_norm.py
```diff
@@ -3,25 +3,24 @@
...
     norm.reset_running_stats()
     norm.reset_parameters()
-    if WITH_PT212:
-        assert str(norm) == (f'BatchNorm(16, eps=1e-05, momentum=0.1, '
-                             f'affine={conf}, bias={conf}, '
-                             f'track_running_stats={conf})')
-    else:
-        assert str(norm) == (f'BatchNorm(16, eps=1e-05, momentum=0.1, '
-                             f'affine={conf}, '
-                             f'track_running_stats={conf})')
```
akihironitta, line 24: Is this change necessary?

akihironitta (approved):

test/nn/norm/test_batch_norm.py
drivanov: Yes, the crux of that problem was precisely that not all 2.12 versions of PyTorch added `bias=...` to `str(norm)`,  and `torch-2.12.0a0+0291f960b6.nvinternal.main.46693258` is just one of them.

akihironitta: @drivanov Is there any case you could update the PyTorch version in your internal CI to use a newer PyTorch nightly? https://github.com/pytorch/pytorch/commit/0291f960b6 seems one month old, and we're using the latest pytorch nightly in our CI. If it's super hard to do so on your side, I'm happy to adjust this test case.

drivanov: I think that it would be easier to adjust this test case.

PR #10674, opened 2026-04-24
Fix batched OnDiskDataset subset access

Karl-XZ:
Fix batched `OnDiskDataset` subset access via `multi_get()` and
`__getitems__()`.

`OnDiskDataset.get()` intentionally keeps raw database index semantics,
matching `Dataset.__getitem__()`, which already resolves subset-local indices
through `self.indices()` before calling `get()`.

The bug was in the batched path: `__getitems__()` delegated to `multi_get()`
without resolving subset-local batch indices to raw database indices. As a
result, `DataLoader(subset, ...)` could return the first N raw DB rows instead
of the subset or split rows.

This PR resolves indices in `multi_get()`, covering:
- subsets created by `index_select()`
- nested subsets
- split-specific `_indices`, such as `PCQM4Mv2`
- tensor and NumPy array indices
- empty batches

## Testing

```bash
python -m pytest -q test/data/test_on_disk_dataset.py -k "multi_get or direct_indices"
python -m pytest -q test/data/test_on_disk_dataset.py

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed inconsistent `OnDiskDataset` subset access after `index_select()` ([#10674](https://github.com/pyg-team/pytorch_geometric/pull/10674))
 - Fix MovieLens dataset incompatibility with `sentence-transformers>=5.0.0` ([#10668](https://github.com/pyg-team/pytorch_geometric/pull/10668)
 - Removed an unnecessary device synchronization in `torch_geometric.utils.softmax` ([#10499](https://github.com/pyg-team/pytorch_geometric/pull/10499))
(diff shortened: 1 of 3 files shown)


Karl-XZ:
I updated #10674 so that `get()` keeps raw DB index semantics, while the fix is applied to the batched path in `multi_get()` / `__getitems__()`.

The tests now cover `index_select()` subsets, nested subsets, direct `_indices` usage (e.g. split-specific datasets such as `PCQM4Mv2`), NumPy/tensor indices, empty batches, and `DataLoader(...)` batch fetching.

Locally verified with:
`python -m pytest -q test/data/test_on_disk_dataset.py`

lauritsf:
Hi, thanks for the contribution.
Since the indexing might be a bit confusing for people reading it, I think it might be worth spending a few lines to document the difference between getting raw db indices and getting them via e.g. `__getitems__`

Kind regards,

Karl-XZ:
Thanks, agreed. I added clarifying docstrings to make the distinction explicit: `get()` uses raw DB indices, while `multi_get()` / `__getitems__()` provide subset-aware batched access.

PR #10680, opened 2026-04-29
Fix LPFormer attention dimensions

BWAAEEEK:
Fixes LPFormer attention dimensions for multi-layer and multi-head configurations.

The transformer layer construction could reference an undefined `self.num_layers`, and stacked attention layers used incompatible pairwise feature dimensions after the first layer. Multi-head attention also normalized concatenated outputs with `out_channels` instead of `num_heads * out_channels`.

## Changes

- Use `num_transformer_layers` when identifying the final attention layer
- Set each `LPAttLayer` input/output dimension to match the expected feature flow across first, intermediate, and final layers
- Size `post_att_norm` correctly for concatenated multi-head outputs
- Add deterministic regression coverage for `num_transformer_layers in {1, 2, 3}` and `num_heads in {1, 2}`

## Tested with

```bash
/home/jovyan/venvs/pyg-pr-review/bin/python -m pytest -q test/nn/models/test_lpformer.py
```

### DIFF
--- a/test/nn/models/test_lpformer.py
+++ b/test/nn/models/test_lpformer.py
@@ -1,4 +1,5 @@
 import torch
+import pytest
 
 from torch_geometric.nn import LPFormer
 from torch_geometric.testing import withPackage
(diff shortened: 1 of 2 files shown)


BWAAEEEK:
Hi PyG team, just following up on this bug-fix PR.

This fixes two LPFormer dimension issues that are reproducible on `master`:
- `num_transformer_layers > 1` can reference an undefined `self.num_layers`
- `num_heads > 1` can fail in `LayerNorm` due to concatenated attention outputs

I also strengthened the regression test to cover `num_transformer_layers in {1, 2, 3}` and `num_heads in {1, 2}` with deterministic settings. All checks are green.

@rusty1s @akihironitta Would appreciate a review when you have bandwidth. Thanks!

PR #10692, opened 2026-05-12
Fix Python 3.14 compatibility in `config_store.py`

ddelgadovargas-cyber:
I got this error in Python3.14:
```console
test/test_config_store.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
config_store.py:262: in to_dataclass
    annotation = map_annotation(annotation, mapping=MAPPING)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

annotation = float | torch.Tensor | str
mapping = {<class 'torch.nn.modules.module.Module'>: typing.Any, <class 'torch.Tensor'>: typing.Any}

    def map_annotation(
        annotation: Any,
        mapping: Optional[Dict[Any, Any]] = None,
    ) -> Any:
        origin = getattr(annotation, '__origin__', None)
        args: Tuple[Any, ...] = getattr(annotation, '__args__', tuple())
        if origin in {Union, list, dict, tuple}:
            assert origin is not None
            args = tuple(map_annotation(a, mapping) for a in args)
            if type(annotation).__name__ == 'GenericAlias':
                # If annotated with `list[...]` or `dict[...]`:
                annotation = origin[args]
            else:
                # If annotated with `typing.List[...]` or `typing.Dict[...]`:
                annotation = copy.copy(annotation)
>               annotation.__args__ = args
E               AttributeError: readonly attribute

config_store.py:176: AttributeError
```

In Python 3.14, the `__args__` attribute of generic aliases (like `typing.Dict`) is read-only. The original code attempted to copy the annotation and mutate `__args__`. This change replaced this logic to use `origin[args]` directly, which is the standard way to construct these types in modern Python.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
-- Fix Python 3.14 compatibility in `config_store.py`
+- Fix Python 3.14 compatibility in `config_store.py`([#10692](https://github.com/pyg-team/pytorch_geometric/pull/10692))
 - Fix MovieLens dataset incompatibility with `sentence-transformers>=5.0.0` ([#10668](https://github.com/pyg-team/pytorch_geometric/pull/10668)
 - Removed an unnecessary device synchronization in `torch_geometric.utils.softmax` ([#10499](https://github.com/pyg-team/pytorch_geometric/pull/10499))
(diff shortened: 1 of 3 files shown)


ddelgadovargas-cyber:
This change was consolidated in #10726.

PR #10701, opened 2026-05-19
Drop torch.jit.script from SelectOutput and ConnectOutput

### LINKED ISSUE #10697: DeprecationWarning: `torch.jit.script` is deprecated. Please switch to `torch.compile` or `torch.export`.
### 🐛 Describe the bug

Tests fail:
```
torch_geometric/nn/pool/select/base.py:65: in <module>
    SelectOutput = torch.jit.script(SelectOutput)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.11/site-packages/torch/jit/_script.py:1488: in script
    warnings.warn(
E   DeprecationWarning: `torch.jit.script` is deprecated. Please switch to `torch.compile` or `torch.export`.
```

### Versions

Version: 2.7.0
pytorch-2.12.0
FreeBSD 15 STABLE

V-3604:
Fixes #10697.

`torch.jit.script` is deprecated in PyTorch 2.12 and emits a `DeprecationWarning` whether its called as a function or used as a decorator. The module level calls: 

    SelectOutput = torch.jit.script(SelectOutput)
    ConnectOutput = torch.jit.script(ConnectOutput)

in `select/base.py` and `connect/base.py` therefore raise the warning at import time, which `pyproject.toml`''s `error:.*torch_geometric.*` filter treats as a test error. There is a workaround in place (`ignore:.*torch.jit.*` further down in the filter list) but the deprecated call itself is still there.

This drops the scripting on both classes. They are now plain Python classes with type annotations and a custom `__init__`. `@dataclass(init=False)` is removed because it isn''t needed once we are not scripting the result (and the synthesized `__repr__` actually trips up TorchScript when downstream modules that return these types are scripted).

Verified on `torch==2.12.0`:

- import is clean (`-W error::DeprecationWarning` passes)
- `test/nn/pool/select/test_select_topk.py` and `test/nn/pool/connect/test_filter_edges.py` both pass (the `is_full_test()` JIT branches are not exercised by default, but I separately confirmed `torch.jit.script(FilterEdges())` still scripts correctly with the refactored `ConnectOutput`)
- broader `test/nn/pool/` suite: 28 passed, 16 skipped (unrelated optional deps)

The `ignore:.*torch.jit.*:DeprecationWarning` filter in `pyproject.toml` is left in place since the remaining `torch.jit.script(module)` calls in the `is_full_test()` paths still emit the same warning.

Screenshot:
Untitled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Removed `torch.jit.script` calls on `SelectOutput` and `ConnectOutput` to avoid `DeprecationWarning` on PyTorch 2.12 ([#10697](https://github.com/pyg-team/pytorch_geometric/issues/10697))
 - Fix MovieLens dataset incompatibility with `sentence-transformers>=5.0.0` ([#10668](https://github.com/pyg-team/pytorch_geometric/pull/10668)
 - Removed an unnecessary device synchronization in `torch_geometric.utils.softmax` ([#10499](https://github.com/pyg-team/pytorch_geometric/pull/10499))
 - Fixed loading of legacy HuggingFace BERT checkpoints ([#10631](https://github.com/pyg-team/pytorch_geometric/pull/10631))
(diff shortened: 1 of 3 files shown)


V-3604:
Closing in favor of #10711, which addresses the \	orch.jit.script\ deprecation (#10697) across the whole codebase, including \SelectOutput\ and \ConnectOutput\.

PR #10726, opened 2026-06-10
Fix config_store map_annotation on Python 3.14 (read-only __args__)

adi-kumo:
On Python 3.14, `config_store.map_annotation` raises `AttributeError: readonly attribute`: it assigns to `__args__`, which became read-only after the `typing.Union` unification ([CPython gh-105499](https://github.com/python/cpython/issues/105499)). This rebuilds the annotation instead of mutating it.

### Relationship to #10692
Same root cause as #10692 — thanks @ddelgadovargas-cyber. Opening this as an alternative because it:
1. **Preserves the `typing.*` form** — uses `copy_with(...)` so `typing.List[...]` stays `typing.List[...]` (rather than normalizing to `list[...]`), leaving the existing `test_map_annotation` assertions unchanged.
2. **Also fixes a latent bug on 3.10–3.13:** the original `copy.copy(alias); alias.__args__ = args` doesn't actually copy — `copy.copy()` on a `typing` alias returns the *same* interned object — so remapping one annotation mutates the shared global singleton (e.g. `typing.List[int]` becomes `typing.List[Any]` process-wide, and a later `Union` remap can crash). Rebuilding returns fresh objects and eliminates this.

Happy to consolidate into whichever approach maintainers prefer.

### Changes
- `Union`/`Optional` → `Union[args]`; other `typing.*` generics → `annotation.copy_with(tuple(args))` (legacy in-place path kept as a fallback); builtin `list[...]`/`dict[...]` path unchanged.
- Added `test_map_annotation` cases for `Union`/`Optional`/nested generics + Union-collapse-to-`Any`.

### Validation
`test/test_config_store.py` and `fill_config_store()` pass on Python **3.10 and 3.14**.

_(Draft — opening for discussion / alignment with #10692.)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
+
 ### Security
 
 ## [2.8.0] - 2026-06-05
--- a/test/test_config_store.py
+++ b/test/test_config_store.py
@@ -65,6 +65,10 @@ def test_map_annotation():
     assert map_annotation(List[Optional[int]], mapping) == List[Optional[Any]]
     assert map_annotation(Dict[str, Optional[int]],
                           mapping) == Dict[str, Optional[Any]]
+    # When every Union member maps to the same type, the rebuilt annotation
+    # collapses to canonical `Any` (the old in-place mutation left a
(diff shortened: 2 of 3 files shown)


adi-kumo:
Ready for review. Quick note on CI: the `pytest-full` reds are pre-existing on `master` (the scheduled full run fails the same 8 `jit.script` tests at this PR's base commit `286a2aa`) — unrelated to this change; the regular `pytest` lane is green. Validated on Python 3.10–3.12 and 3.14.

akihironitta (approved):

adi-kumo:
Thanks for the review @akihironitta! The only thing left is the required checks blocking merge — could you admin-merge / override when you get a chance, or let me know if there's a path you'd prefer?

PR #10755, opened 2026-07-23
Add `DenseGATv2Conv`

### LINKED ISSUE #10722: Feature Request: v2 version of DenseGATConv
I noticed that DenseGATConv exists, but I couldn't find a v2 version of it. Is there a dense equivalent of GATv2Conv available?

siddharthgaur1:
Closes #10722.

`DenseGATConv` exists but has no v2 counterpart, so there is currently no dense equivalent of `GATv2Conv`. This adds one, mirroring the relationship between `DenseGATConv` and `GATConv`.

### Implementation

GAT computes attention as `LeakyReLU(a_src · Wx_i + a_dst · Wx_j)` — the projection onto the attention vector happens before the non-linearity, which is the source of its static attention. GATv2 applies the non-linearity first:

```
alpha = a^T · LeakyReLU(W_l x_i + W_r x_j)
```

So the dense version uses two linear maps (`lin_l`, `lin_r`, optionally tied via `share_weights`) and a single `att` parameter, rather than one `lin` and two `att_*` vectors.

### Memory note

`DenseGATConv` can reduce to `[B, N, N, heads]` before the softmax. GATv2 cannot — the non-linearity applies to the full feature vector, so the intermediate is `[B, N, N, heads, out_channels]`, i.e. `out_channels`× more memory than `DenseGATConv`.

This is inherent to the GATv2 formulation rather than an artifact of this implementation, but since it matters on an already-`O(N^2)` dense layer, it's called out explicitly in the class docstring rather than left to be discovered.

### Scope

The parameter surface mirrors `DenseGATConv`, plus `share_weights` (GATv2-specific).

`edge_dim` and `residual` are deliberately not included: `DenseGATConv` carries a `# TODO Add support for edge features` and no dense conv implements `residual`, so supporting them here would make this layer inconsistent with the rest of the dense family. Happy to add either if you'd prefer.

### Tests

`test/nn/dense/test_dense_gatv2_conv.py` follows the existing `test_dense_gat_conv.py` pattern: weights are shared with the sparse `GATv2Conv` and the outputs are asserted equal, parametrized over `heads=[1, 4]`, `concat=[True, False]` and `share_weights=[False, True]` (8 combinations), plus masked-node zeroing, batch broadcasting, and the `is_full_test()` TorchScript check. All pass, including under `FULL_TEST=1`.

### Note on `nn/dense/__init__.py`

`conv_classes` / `pool_classes` are positional slices of `__all__`, and `docs/source/modules/nn.rst` iterates them to generate documentation. The slice bounds are updated so the new layer lands in `conv_classes` and the pooling group is unchanged — verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- Added `DenseGATv2Conv`, the dense counterpart of `GATv2Conv` ([#10755](https://github.com/pyg-team/pytorch_geometric/pull/10755))
 - Added `torch_geometric.contrib.nn.models.MGNAN` (M-GNAN, an extension of `GNAN` to multivariate shape functions) model plus an example ([#10371](https://github.com/pyg-team/pytorch_geometric/pull/10371))
 
 ### Changed
--- a/test/nn/dense/test_dense_gatv2_conv.py
+++ b/test/nn/dense/test_dense_gatv2_conv.py
@@ -1,0 +1,75 @@
+import pytest
+import torch
+
+from torch_geometric.nn import DenseGATv2Conv, GATv2Conv
+from torch_geometric.testing import is_full_test
+
+
+@pytest.mark.parametrize('heads', [1, 4])
+@pytest.mark.parametrize('concat', [True, False])
+@pytest.mark.parametrize('share_weights', [False, True])
+def test_dense_gatv2_conv(heads, concat, share_weights):
+    channels = 16
+    sparse_conv = GATv2Conv(channels, channels, heads=heads, concat=concat,
+                            share_weights=share_weights)
+    dense_conv = DenseGATv2Conv(channels, channels, heads=heads, concat=concat,
+                                share_weights=share_weights)
+    assert str(dense_conv) == f'DenseGATv2Conv(16, 16, heads={heads})'
+
+    # Ensure same weights and bias:
+    dense_conv.lin_l = sparse_conv.lin_l
+    dense_conv.lin_r = sparse_conv.lin_r
+    dense_conv.att = sparse_conv.att
+    dense_conv.bias = sparse_conv.bias
+
+    x = torch.randn((5, channels))
+    edge_index = torch.tensor([[0, 1, 1, 2, 3, 4], [1, 0, 2, 1, 4, 3]])
+
+    sparse_out = sparse_conv(x, edge_index)
+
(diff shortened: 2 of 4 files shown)


siddharthgaur1:
Rebased onto current master.

Flagging that PyG's Testing workflow hasn't run on any push here — the only green checks are Read the Docs and pre-commit.ci, so there's no actual test signal on this PR yet. I believe it needs the first-time-contributor approval to run workflows.

@akihironitta would you mind approving the run? Happy to adjust the API if `DenseGATv2Conv` should diverge from `DenseGATConv` anywhere.

siddharthgaur1:
@rusty1s — not a bump on this PR. Posting here rather than opening an issue because it seemed the wrong thing to file a meta-issue about, but it affects most of the open outside-contributor queue, so it seemed worth one message.

**Almost no PR from an outside contributor is getting a test run.** I sampled the 40 most recent open PRs and counted real check-runs via `GET /repos/pyg-team/pytorch_geometric/commits/{sha}/check-runs`:

| author_association | PRs | with a real test run |
|---|---|---|
| `NONE` (outside) | 31 | **1** |
| prior contributors | 9 | 8 |

30 open outside PRs have no test signal at all, the oldest from **2026-07-04**. The single exception is #10766, which carries the full 15-run `pytest / test` matrix — so the approval path works and does get used, just rarely.

**Why this is easy to miss:** those PRs don't look untested. Read the Docs and pre-commit.ci both report green on them, so the PR page shows check marks. But those are `StatusContext` entries, not `CheckRun` entries — `gh pr view --json statusCheckRollup` mixes the two, and the rollup reads as green. Counting `check-runs` separately is the only way to see it. I got this wrong myself twice before checking properly, which is what prompted the sample.

The consequence is a bit circular: a reviewer opening one of these sees no test results, so there's nothing to review against, so it waits — and the workflow that would produce those results is waiting on a reviewer.

This is GitHub's fork-PR approval setting (Settings → Actions → General → *Fork pull request workflows from outside collaborators*). If it's currently on **"Require approval for all outside collaborators"**, relaxing it to **"Require approval for first-time contributors"** would let repeat contributors run automatically while still gating genuinely new accounts. If it's deliberate as-is, that's completely fair — in that case it might be worth a line in the PR template saying a maintainer has to start CI, so contributors don't read the silence as rejection.

Disclosure: four of those 30 are mine (#10755, #10757, #10759, #10761), so I'm an interested party. I'd have raised it anyway — #10784 is another contributor's fix for a real `APPNP` dropout bug that I independently reproduced and verified, and it's been sitting since 08-28 with no run. Happy to be told this is working as intended and to stop.

PR #10757, opened 2026-07-26
Clarify radius/radius_graph docs: CPU uses k-d tree, GPU is brute-force

### LINKED ISSUE #10567: Clarify in radius and radius_graph that the implementation is O(N^2) on GPU
### 📚 Describe the documentation issue

Hi,

I saw the source in the doc using KDTree for CPU and assumed it'd be the same on the GPU. But the CUDA code is brute-force.

I hope this is clarified in the doc.

_No response_

siddharthgaur1:
## Summary
Closes #10567.

`radius`/`radius_graph` warn that the CPU path is biased towards certain quadrants when `max_num_neighbors` is set, and suggest moving to GPU as a workaround — but don't mention that the two backends use fundamentally different search strategies. This led to the linked issue: a user assumed the GPU path also uses a k-d tree, when in fact it's a brute-force `O(N * M)` kernel (verified in `pyg-lib`'s `radius_kernel.cu`, vs. the CPU path in `radius_kernel.cpp` which builds a `nanoflann` k-d tree).

Added a `.. note::` to both docstrings stating this explicitly, so the GPU suggestion in the existing warning isn't read as a strictly faster option for large point sets.

## Test plan
- [x] Docs-only change, no code path touched
- [x] Verified the complexity claim against `pyg-lib`'s CPU (`nanoflann` KD-tree) and CUDA (brute-force) kernels directly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- Clarified in the `radius` and `radius_graph` docstrings that the CPU implementation uses a `k`-d tree while the GPU implementation performs a brute-force `O(N * M)` search ([#PLACEHOLDER](https://github.com/pyg-team/pytorch_geometric/pull/PLACEHOLDER))
+
 ### Deprecated
 
 ### Removed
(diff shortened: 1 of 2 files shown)


siddharthgaur1:
Bumping this one too — it's a small docs-only change (`Shapes:` blocks for `TransformerConv`/`SplineConv`, part of the #3573 effort) and, like my other open PRs here, the Testing workflow has never run on it (only Read the Docs / pre-commit.ci show up). Looks like it's waiting on the first-time-contributor approval to trigger workflows. @akihironitta whenever you get a chance to approve the run on this batch, happy to address any feedback.

PR #10773, opened 2026-08-16
Fix timing methodology for GCN compile benchmark

drivanov:
**Motivation**

The GCN compile performance test was intermittently failing on some machines because of variability in the measured execution times. I investigated the failures and found that the benchmark timing was affected by compilation warm-up and timing of the evaluation phase.

This PR fixes the timing methodology to provide more stable and reliable training-time measurements.

**Changes**

- treat first 10 iterations as a warm-up before collecting timing measurements.
- Measure training time separately from evaluation time.

**Result**

- Timing variance for different epochs is now negligible.
- `torch.compile` shows approximately 2× speedup over the regular implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed timing methodology for GCN compile benchmark ([#10773](https://github.com/pyg-team/pytorch_geometric/pull/10773))
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
-- Fixed timing methodology for GCN compile benchmark ([#10773](https://github.com/pyg-team/pytorch_geometric/pull/10773))
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
 
 ### Security
--- a/examples/compile/gcn.py
+++ b/examples/compile/gcn.py
@@ -75,6 +75,7 @@ def test():
         accs.append(int((pred[mask] == data.y[mask]).sum()) / int(mask.sum()))
     return accs
 
+
 # Warm-up
 for i in range(0, 10):
     train()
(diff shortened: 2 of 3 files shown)


puririshi98 (approved):

examples/compile/gcn.py
```diff
@@ -76,12 +76,20 @@ def test():
...
+
 times = []
 for epoch in range(1, 201):
     start = time.time()
     loss = train()
-    train_acc, val_acc, test_acc = test()
+    torch.cuda.synchronize()
     times.append(time.time() - start)
+    train_acc, val_acc, test_acc = test()
+    torch.cuda.synchronize()
```
akihironitta: This doesn't make much sense because we already have D2H sync in each of `train` and `test`. We need to either keep it as is or try removing the syncs from `train` and `test`.

examples/compile/gcn.py
```diff
@@ -76,12 +76,20 @@ def test():
     return accs
 
 
+# Warm-up
+for _i in range(0, 10):
+    train()
+    test()
+    torch.cuda.synchronize()
+
```
akihironitta: I'm guessing we're using this script for internal testing, but IMO, this contaminates the example. Instead, we should simply ignore the first (few) epoch(s) in our internal testing setup if you think it's feasible.

examples/gcn.py
```diff
@@ -95,15 +95,23 @@ def test():
...
     loss = train()
+    torch.cuda.synchronize()
+    times.append(time.time() - start)
     train_acc, val_acc, tmp_test_acc = test()
+    torch.cuda.synchronize()
     if val_acc > best_val_acc:
         best_val_acc = val_acc
         test_acc = tmp_test_acc
     log(Epoch=epoch, Loss=loss, Train=train_acc, Val=val_acc, Test=test_acc)
-    times.append(time.time() - start)
```
akihironitta: ditto

examples/compile/gcn.py
drivanov: You were right: removing `torch.cuda.synchronize()` does not affect the variation in execution time between different epochs.

examples/compile/gcn.py
drivanov: Done

examples/gcn.py
drivanov: Done

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed timing methodology for GCN compile benchmark ([#10773](https://github.com/pyg-team/pytorch_geometric/pull/10773))
```
```suggestion
```

akihironitta (approved):

CHANGELOG.md
drivanov: I am not opposed to this change; however, we are getting:
```
Error: No update to CHANGELOG.md found!
```

PR #10775, opened 2026-08-19
Fix NormalizeScale for degenerate point clouds

tandede:
## Summary

- keep empty point clouds unchanged in `NormalizeScale`
- use a finite unit denominator when centering produces a zero-extent point cloud
- add regression tests for empty and fully coincident positions and document the fix in the changelog

## Problem

`NormalizeScale` centers positions and then multiplies them by the reciprocal of their maximum absolute coordinate. When all positions are identical, centering produces all zeros, so the reciprocal is infinite and the final `0 * inf` operation turns every coordinate into `NaN`. Empty position tensors fail earlier because `max()` has no reduction identity.

This can feed non-finite coordinates into downstream point-cloud models even though both inputs are valid degenerate graphs.

## Fix

Skip the reduction for empty position tensors. For non-empty tensors, replace a zero maximum absolute coordinate with a scalar one before computing the existing scale factor. The regular non-degenerate formula and output range remain unchanged, while coincident points stay centered at finite zeros. The tensor-only denominator selection also avoids a device synchronization.

## Testing

- `python -m pytest test/transforms/test_normalize_scale.py -q` (3 passed)
- `python -m pytest test/transforms -q` (151 passed, 14 skipped, 2 deselected)
- full test suite on PyTorch 2.12 with MPS fallback (5902 passed, 1166 skipped, 169 deselected)
- CPU and MPS forward/backward finite-value checks for coincident positions
- `pre-commit run --files torch_geometric/transforms/normalize_scale.py test/transforms/test_normalize_scale.py CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Fixed
 
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
+- Fixed `NormalizeScale` producing non-finite coordinates for degenerate point clouds
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


tandede:
Hi maintainers, just a gentle follow-up on this PR when you have time. It keeps `NormalizeScale` finite for empty and coincident point clouds and includes focused regression coverage. The reported documentation and pre-commit checks are passing, and I’m happy to address any feedback or adjust the implementation if needed. Thanks for maintaining PyTorch Geometric.

PR #10780, opened 2026-08-27
Fix normalize being ignored when average=True

AyushG-1210:
`torch_geometric.utils.total_influence` accepts a `normalize` argument that defaults to `True`, documented as:
> `normalize` (bool, optional): If `True`, normalize each hop-wise influence by
> the influence of hop 0.

However, it is only ever applied on the `average=True` path. Calling `total_influence(..., average=False)` silently returns **un-normalized** values, even though `normalize=True` by default and the docstring places no condition on it.

<details>
<summary> Reproduction: </summary>

```python
import torch
from torch_geometric.utils import total_influence

class GNN(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.conv1 = GCNConv(5, 6)
        self.conv2 = GCNConv(6, 7)
        return self.conv2(self.conv1(x, edge_index), edge_index)

torch.manual_seed(0)
data = Data(x=torch.randn(6, 5),
            edge_index=torch.tensor([[0, 1, 2, 3, 4], [1, 2, 3, 4, 5]]))
model = GNN()

def run(**kw):
    torch.manual_seed(42)
    return total_influence(model=model, data=data, max_hops=2,
                           num_samples=4, **kw)[0]

print(torch.equal(run(average=False, normalize=True),
                  run(average=False, normalize=False)))
```
On `master` this prints `True`, meaning the flag has no effect whatsoever.

</details>

## Fix:
Apply the normalization on the non-averaged path too, dividing each row by its own hop-0 influence:

```python
if average:
    avg_influence = avg_total_influence(allnodes, normalize=normalize)
elif normalize:  # normalize each row by its own hop-0 influence:
    avg_influence = allnodes / allnodes[:, 0:1]
else:
    avg_influence = allnodes
```

Before / after, `average=False`:

```
normalize=True (before, == raw)     normalize=True (after)
[[12.6886,  0.0000,  0.0000],       [[1.0000, 0.0000, 0.0000],
 [ 3.1721,  6.3443,  4.4861],        [1.0000, 2.0000, 1.4142],
 [ 3.1721,  6.3443,  4.4861],        [1.0000, 2.0000, 1.4142],
 [ 3.1721,  6.3443,  4.4861]]        [1.0000, 2.0000, 1.4142]]
```
Added two tests to `test/utils/test_total_influence.py`:

- `test_total_influence_normalize_average` pins the existing averaged behaviour (guards against regressing the path that already worked).
- `test_total_influence_normalize_no_average` asserts every row's hop-0 entry is `1.0` and that the result equals `raw / raw[:, 0:1]`.

The second test fails on `master` (`assert torch.allclose(I_norm[:, 0], torch.ones(num_samples))`)
and passes with this change.

## Additionally:
- `avg_total_influence`'s docstring was a verbatim copy of `influence_weighted_receptive_field`'s and described the wrong function.
- `:pyfunc:` → `:func:` (`:pyfunc:` is not a Sphinx role, so those cross-references were dead).
- Expanded the `normalize` docstring to state its behaviour in each mode.

### DIFF
--- a/test/utils/test_total_influence.py
+++ b/test/utils/test_total_influence.py
@@ -17,15 +17,16 @@ class GNN(torch.nn.Module):
         return x2
 
 
-def test_total_influence_smoke():
+def get_data():
     x = torch.randn(6, 5)
     edge_index = torch.tensor([[0, 1, 2, 3, 4], [1, 2, 3, 4, 5]])
+    return Data(x=x, edge_index=edge_index)
+
+
+def test_total_influence_smoke():
     max_hops = 2
     num_samples = 4
-    data = Data(
-        x=x,
-        edge_index=edge_index,
-    )
+    data = get_data()
     model = GNN()
     I, R = total_influence(
         model,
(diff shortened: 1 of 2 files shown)


AyushG-1210:
Any updates?

PR #10790, opened 2026-09-01
Fix the RelBench examples for `relbench>=3.0.1`

drivanov:
relbench 3.0 dropped `relbench.datasets`/`relbench.tasks`, so `examples/rdl.py`, `examples/relbench_example.py` and `examples/llm/relbench_gretriever.py` failed on import. This ports them to the 3.x API (`load_dataset` + `dataset.load_task`), lets relbench resolve free-form `--dataset`/`--task` (hosted names, Hub repos or local paths), tunes on the task's own default metric and rejects tasks the example can't handle before the slow materialization. `rdl.py` now keeps one materialized cache per dataset (`remove_columns=task.hidden_columns()`) instead of one per task, and puts every time column on `datetime64[ns]` (Arrow-backed ones via numpy) before relbench's `to_unix_time`, which otherwise mis-scales `datetime64[us]` data. Also fixes `--num_neighbors`, which `type=list` split into characters. Supersedes #10791, which should be closed. Tested on rel-f1 driver-position and driver-dnf with relbench 3.0.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed `example/rdl.py` for RelBench 3.0. ([#10790](https://github.com/pyg-team/pytorch_geometric/pull/10790))
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
 
 ### Security
--- a/examples/rdl.py
+++ b/examples/rdl.py
@@ -750,8 +750,7 @@ def main():
             device=device,
         )
 
-        val_metrics = task.evaluate(val_pred, val,
-                                    eval_metrics)  #task.get_table("val"))
+        val_metrics = task.evaluate(val_pred, val, eval_metrics)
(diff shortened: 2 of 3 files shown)


akihironitta (commented):
I think we don't need to support both RelBench 2 and 3, and we can simply require `relbench>=3.0.0` to run this example

drivanov:
@akihironitta : Here is a fix intended exclusively for RelBench >=3.0: [PR#10791](https://github.com/pyg-team/pytorch_geometric/pull/10791)

puririshi98:
Heads-up @drivanov: I pushed a squashed rework onto your `relbench_rdl` branch (3 commits on top of your head 808ebc1b). It ports all three RelBench examples to relbench>=3.0.1, fixes the binary-task `KeyError: 'roc_auc'` we hit on rel-f1, parses `--num_neighbors` as ints, shares one materialized cache per dataset via `remove_columns=task.hidden_columns()`, and normalizes timestamps to nanoseconds before `make_pkey_fkey_graph` (relbench's `to_unix_time` treats every datetime as ns). Please `git fetch && git reset --hard origin/relbench_rdl` rather than pull. Verified end to end on rel-f1 with relbench 3.0.1 on an A100; the container pin bump for our CI is in flight separately.

PR #10791, opened 2026-09-02
fix: adapt rdl example to RelBench 3.0 dataset API

drivanov:
**Summary**

Update `examples/rdl.py` to work with RelBench 3.0.

RelBench 3.0 changed the dataset loading API, so the example is updated to use the new API.

Without this fix, this test fails on NVidia version 26.09 when building with relbench 3.0:
```
  File "/workspace/examples/rdl.py", line 20, in <module>
    from relbench.datasets import get_dataset, get_dataset_names
ModuleNotFoundError: No module named 'relbench.datasets'
```

This is a replacement for [PR#10790](https://github.com/pyg-team/pytorch_geometric/pull/10790): **fix: make examples/rdl.py compatible with RelBench 2.x and 3.x**

### DIFF
--- a/examples/rdl.py
+++ b/examples/rdl.py
@@ -14,8 +14,10 @@ from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 
 import numpy as np
 import pandas as pd
+import sklearn.metrics as skm
 import torch
 import torch_frame
+from numpy.typing import NDArray
 from relbench import load_dataset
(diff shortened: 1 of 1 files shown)


puririshi98:
Closing in favor of #10790, which now carries the relbench>=3.0.1 port of all three RelBench examples (plus the binary-task metric fix and the shared cache). Thanks @drivanov.

PR #10797, opened 2026-09-04
fix: avoid tensor fill values in ONNX scatter path

### LINKED ISSUE #10327: "TypeError: full() received an invalid combination of arguments" in _scatter.py
### 🐛 Describe the bug

In line 106 of _scatter.py, we are building this tensor: 

Image

and get this error:
'''TypeError: full() received an invalid combination of arguments - got (size=tuple, fill_value=Tensor, device=torch.device, dtype=torch.dtype, ), but expected one of:
 * (tuple of ints size, Number fill_value, *, tuple of names names, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (tuple of ints size, Number fill_value, *, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)'''
 * 
However, the fill_value argument is expecting a Number type but src.min (or .max for that matter) returns a tensor. I tested it by calling .item on the src.min function and it passed in the number instead of the tensor and that seems to work. 

This call only happens when we're in the onnx export. For the other calls during training , we use the code in lines 101 - 104, and they work fine. I understand it's an edge case, or I might have just misunderstood what fill is doing. 

### Versions

[pip3] numpy==1.24.1
[pip3] torch==2.1.2
[pip3] torch-ema==0.3
[pip3] torch-geometric==2.6.1
[pip3] torchaudio==2.1.2
[pip3] torchdata==0.4.1+f9ecd8b
[pip3] torcheval==0.0.7
[pip3] torchmetrics==1.7.2
[pip3] torchvision==0.16.2
[conda] blas                                  1.0                 mkl
[conda] mkl                                   2020.2              256
[conda] numpy                                 1.24.1              py310hd02465a_0           conda-forge
[conda] pytorch                               2.1.2               py3.10_cuda12.1_cudnn8_0  pytorch
[conda] pytorch-cuda                          12.1                hde6ce7c_6                pytorch
[conda] pytorch-mutex                         1.0                 cuda                      pytorch
[conda] torch-ema                             0.3                 pypi_0                    pypi
[conda] torch-geometric                       2.6.1               pypi_0                    pypi
[conda] torchaudio                            2.1.2               pypi_0                    pypi
[conda] torchdata                             0.4.1               pyh8b8bddf_0              conda-forge
[conda] torcheval                             0.0.7               pypi_0                    pypi
[conda] torchmetrics                          1.7.2               pyhd8ed1ab_0              conda-forge
[conda] torchvision                           0.16.2              pypi_0                    pypi

LOGO127:
Fixes #10327

### Summary

The ONNX-specific min/max scatter path passed src.min() or src.max() (a zero-dimensional Tensor) as the fill_value to torch.full. Older PyTorch versions reject Tensor-valued fill arguments during ONNX export.

This change creates the same dynamically computed fill tensor with reshape(1).expand_as(src), preserving the source dtype and device without converting the value to a Python scalar.

### Tests

- pytest test/utils/test_scatter.py -k onnx_export -q
- pytest test/utils/test_scatter.py -q
- pytest test/metrics/test_link_pred_metric.py -q
- ruff check torch_geometric/utils/_scatter.py test/utils/test_scatter.py
- git diff --check

The local scatter test suite reports 23 passed and 10 skipped because the optional torch-scatter package is not installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed ONNX export for `scatter(reduce="min"/"max")` with older PyTorch versions
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
 
 ### Security
--- a/test/utils/test_scatter.py
+++ b/test/utils/test_scatter.py
@@ -90,7 +90,7 @@ def test_scatter_onnx_export(reduce, expected):
 
     with patch('torch_geometric.utils._scatter.is_in_onnx_export',
                return_value=True), patch('torch.full',
-                                          side_effect=full_scalar_only):
+                                         side_effect=full_scalar_only):
         out = scatter(src, index, dim=0, dim_size=3, reduce=reduce)
 
     assert torch.equal(out, torch.tensor(expected))
--- a/test/utils/test_scatter.py
+++ b/test/utils/test_scatter.py
@@ -1,4 +1,5 @@
 from itertools import product
+from unittest.mock import patch
 
 import pytest
 import torch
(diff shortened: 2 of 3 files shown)


LOGO127:
All checks are now green after adding the required `Unreleased` changelog entry. The fix remains limited to the ONNX `scatter(reduce="min"/"max")` path, with focused regression coverage for both reductions. Ready for review, and happy to adjust if needed.

LOGO127:
@rusty1s The required checks are green and the changelog follow-up is included. Since GitHub does not allow this account to submit a formal review request, could you take a look when convenient? Happy to adjust anything requested.

PR #10799, opened 2026-09-06
Fix group_argsort ranks for constant scores

Anormalm:
## Summary

Fix invalid group-local ranks when `group_argsort` receives constant
scores with interleaved group IDs.

Subtracting the minimum produces all zeros, and the subsequent
normalization divides by zero. The resulting NaNs prevent the group
offsets from separating groups correctly.

Replace only a zero denominator with one, preserving the existing
rank convention.

## Tests

Add regression coverage for negative, zero, and positive constant
scores; float32 and int64 inputs; both sorting directions; and both
return_consecutive settings.

Source-isolated CPU validation with Python 3.13.5 and PyTorch 2.10.0:
- Before: 24 failed, 2 passed.
- After: 26 passed.

The tested source files were verified against the pushed Git blob
hashes.


Anormalm:
Hi maintainers, could someone take a look at this and the companion fix in #10800, and approve the pending CI runs? Both are focused correctness fixes with regression tests: constant-score handling in group_argsort and negative-dimension handling in cumsum. Happy to make any requested changes. Thanks!

PR #10784, opened 2026-08-28
Fix cumulative edge dropout in APPNP

aswanth-07:
## Summary

`APPNP` now samples edge dropout from the original normalized edge weights on every propagation step. This prevents dropout masks and inverted-dropout scaling from accumulating across the `K` iterations.

Fixes #10783

## Implementation

The original edge weights are retained once per forward pass for dense edge indices, PyTorch sparse tensors, and `torch_sparse.SparseTensor` inputs. Each iteration still receives a fresh dropout result, while the adjacency structure and existing propagation behavior remain unchanged.

## Validation

- Added a deterministic regression test that verifies all three propagation steps receive the original weights for dense and PyTorch sparse inputs, with an optional test path for `torch_sparse.SparseTensor`.
- `$env:FULL_TEST='1'; $env:PYTEST_DISABLE_PLUGIN_AUTOLOAD='1'; py -3.13 -m pytest test/nn/conv/test_appnp.py -q`: 4 passed, 1 skipped (`torch_sparse` unavailable). This also exercises TorchScript compilation.
- `py -3.13 -m pytest test/nn/conv -q`: 278 passed, 381 skipped.
- `py -3.13 -m pre_commit run --files CHANGELOG.md torch_geometric/nn/conv/appnp.py test/nn/conv/test_appnp.py`: passed.
- The regression assertion fails on pristine `master` because the second and third dropout calls receive already-dropped weights.
- The author of #10803 independently applied this change and confirmed that separate dense and PyTorch sparse resampling tests pass, then closed #10803 as a duplicate in favor of this PR.

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.

PR #10803, opened 2026-09-08
Fix `APPNP` applying edge dropout cumulatively across propagation steps

### LINKED ISSUE #10783: APPNP applies edge dropout cumulatively across propagation steps
### 🐛 Describe the bug

I believe there is a bug in the forward pass of APPNP ([torch_geometric/nn/conv/appnp.py](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/conv/appnp.py)). 

In the current behaviour: https://github.com/pyg-team/pytorch_geometric/blob/c7ccd0751564153dcffcbff03aea2829fa8db57e/torch_geometric/nn/conv/appnp.py#L121

we have `edge_weight = F.dropout(edge_weight, p=self.dropout)`. This means that edge dropout is applied cumulatively across the K propagation steps, rather than independently at each step. This causes edge weights to vanish much faster than intended, effectively removing most edges after only a few iterations. This also means that weights of surviving edges increase due to the scaling factor (1/(1-p)) applied during dropout, which can blow up and cause NaNs.

Expected behaviour for the block where `isinstance(edge_index, Tensor)` and not `is_torch_sparse_tensor(edge_index)`
```
        original_edge_weight = edge_weight
        for _ in range(self.K):
            if self.dropout > 0 and self.training:
                    if is_torch_sparse_tensor(edge_index):
                        _, edge_weight = to_edge_index(edge_index)
                        edge_weight = F.dropout(edge_weight, p=self.dropout)
                        edge_index = set_sparse_value(edge_index, edge_weight)
                    else:
                        assert edge_weight is not None
                        edge_weight = F.dropout(original_edge_weight, p=self.dropout)
                else:
                    value = edge_index.storage.value()
                    assert value is not None
                    value = F.dropout(value, p=self.dropout)
                    edge_index = edge_index.set_value(value, layout='coo')
            x = self.propagate(edge_index, x=x, edge_weight=edge_weight)
            x = x * (1 - self.alpha)
            x = x + self.alpha * h
```

In the parts where edge_index is used to generate the weights I'm not sure on the fix.

### Versions

Versions of relevant libraries:
[pip3] numpy==2.3.3
[pip3] nvidia-cublas-cu11==11.11.3.6
[pip3] nvidia-cuda-cupti-cu11==11.8.87
[pip3] nvidia-cuda-nvrtc-cu11==11.8.89
[pip3] nvidia-cuda-runtime-cu11==11.8.89
[pip3] nvidia-cudnn-cu11==9.1.0.70
[pip3] nvidia-cufft-cu11==10.9.0.58
[pip3] nvidia-curand-cu11==10.3.0.86
[pip3] nvidia-cusolver-cu11==11.4.1.48
[pip3] nvidia-cusparse-cu11==11.7.5.86
[pip3] nvidia-nccl-cu11==2.21.5
[pip3] nvidia-nccl-cu13==2.31.2
[pip3] nvidia-nvtx-cu11==11.8.86
[pip3] pytorch-lightning==2.6.0
[pip3] torch==2.6.0+cu118
[pip3] torch-geometric==2.7.0
[pip3] torchaudio==2.6.0+cu118
[pip3] torchmetrics==1.8.2
[pip3] torchvision==0.21.0+cu118
[pip3] triton==3.2.0
[conda] Could not collect

siddharthgaur1:
Closes #10783.

## Problem

`APPNP.forward` wrote each step's dropout result back into `edge_index` / `edge_weight`, so step `k+1` sampled from step `k`'s already-thinned weights rather than from the normalized ones. Two consequences, both as the reporter described:

- an edge survives all `K` steps with probability `(1 - p) ** K` instead of `1 - p`, so most edges are gone after a few iterations;
- the `1 / (1 - p)` rescaling stacks on the shrinking set of survivors, so their weights grow without bound — which is where the NaNs come from.

The issue reports the dense branch and says the fix for the others wasn't clear. **All three input paths had it**, for the same reason:

| path | how it compounded |
|---|---|
| dense `Tensor` | `edge_weight = F.dropout(edge_weight, ...)` rebinds the name each step |
| `is_torch_sparse_tensor` | reads values back out of the `edge_index` it overwrote via `set_sparse_value` on the previous step |
| `SparseTensor` | same, through `storage.value()` after `set_value` |

## Fix

Rather than carrying the reported patch's `original_edge_weight` (which fixes only the dense path), the loop simply stops rebinding its inputs. Each step derives a local `step_edge_index` / `step_edge_weight` from the pristine normalized values and propagates with those, so dropout is drawn afresh every step on every path. `edge_index` and `edge_weight` are now read-only inside the loop, which is what makes all three branches correct at once.

## Tests

`test_appnp_dropout_is_resampled_each_step`, parametrized over all three adjacency types, captures the weights actually propagated at each step through `register_propagate_forward_pre_hook` and asserts:

- **no strict nesting** — each step keeps at least one edge the previous step had dropped. This is the exact signature of the bug: with compounding, survivors are always a subset of the previous step's.
- **each step keeps roughly half the edges** at `p=0.5`, rather than `1 / 2 ** k` of them.

Seeded, so it's deterministic rather than statistical. On `master` both runnable parametrizations fail with

```
AssertionError: step 1 kept no edge that step 0 had dropped, so dropout was
applied on top of the previous step rather than resampled
```

and pass with the fix. `test_appnp_dropout_leaves_input_untouched` pins that a forward pass doesn't consume the caller's `edge_weight` — it passes before and after, and guards the restructuring.

## Verification

- `test/nn/conv/test_appnp.py`: 5 passed, 1 skipped (the `torch_sparse` parametrization — `torch-sparse` isn't installed here, so that one path is covered by CI rather than by me locally).
- `test/nn/conv/`: 268 passed, 379 skipped, no regressions.
- `FULL_TEST=1` passes, so `torch.jit.script` still compiles the rewritten loop — worth stating since the change introduces a new `Adj`-typed local inside it.
- `yapf`, `flake8`, `isort` and `ruff` clean at the pinned pre-commit versions.

`APPNP` has no other in-tree callers, so the blast radius is the layer itself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NxphSQWkGcC4sgncbRU64C

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- Fixed `APPNP` applying edge dropout cumulatively across propagation steps instead of resampling it at each step ([#10803](https://github.com/pyg-team/pytorch_geometric/pull/10803))
 - Fixed `config_store` on Python 3.14, where `typing.Union`/`typing.Optional` annotations could no longer be remapped in-place via the now read-only `__args__` attribute
 
 ### Security
--- a/test/nn/conv/test_appnp.py
+++ b/test/nn/conv/test_appnp.py
@@ -1,10 +1,12 @@
+import pytest
 import torch
+from torch import Tensor
 
 import torch_geometric.typing
 from torch_geometric.nn import APPNP
 from torch_geometric.testing import is_full_test
 from torch_geometric.typing import SparseTensor
-from torch_geometric.utils import to_torch_csc_tensor
+from torch_geometric.utils import to_edge_index, to_torch_csc_tensor
 
 
 def test_appnp():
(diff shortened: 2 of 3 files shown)


siddharthgaur1:
Closing this as a duplicate — @aswanth-07's #10784 has been open since 08-28 and fixes the same bug. I checked the issue for assignees and comments before starting but did not search for an existing PR against it, which is entirely on me. Sorry for the noise, @aswanth-07.

Since I had the branch set up, I applied #10784 locally and verified it rather than just deferring to it:

- All three input paths behave: the tests I had written independently (asserting the surviving-edge set is not nested across steps, and that each step keeps ~50% of edges at `p=0.5`, rather than `1/2**k`) **pass against #10784** on both the dense and torch-sparse paths. The `torch_sparse.SparseTensor` path skips here — no `torch-sparse` installed — so that one is still only covered by CI.
- `FULL_TEST=1` passes, so `torch.jit.script` still compiles the rewritten loop. Worth stating explicitly since the change adds locals inside a scripted loop, and #10784 doesn't mention it.
- Full file: 4 passed, 1 skipped. `test/nn/conv/`: no regressions.

So #10784 looks correct and complete to me, from an independent implementation of the same fix. It just needs a maintainer — note its Testing workflow has never run either (its only green marks are Read the Docs and pre-commit.ci, which are external statuses, not check-runs), so a reviewer opening it sees no test signal.

PR #10805, opened 2026-09-08
Make the number of loader workers in `glem.py` configurable

puririshi98:
The three `NeighborLoader`s in `examples/llm/glem.py` hard-code `num_workers=12` with `persistent_workers=True` (36 forked sampler processes). In our CI the example hangs intermittently at the very first pretraining batch with the workers idle, and we want to run it with `--num_workers 0`. Default unchanged; persistent workers only when there are any.


puririshi98:
pre-commit.ci run

PR #10810, opened 2026-09-11
Fix multiprocessing deadlock with CUDA

drivanov:
Use the spawn multiprocessing context when creating workers for data loading:

**Problem**

Using the default multiprocessing method (fork on Linux) can cause deadlocks or hangs when worker processes are created after CUDA has been initialized.

Forking a process with an already-initialized CUDA runtime is unsafe because the child inherits the parent's CUDA-related state. This can leave multiprocessing workers in an inconsistent state and cause the data loader to hang. 

It's exactly what we observe in 
```
$ pytest /opt/pyg/pytorch_geometric/test/data/lightning/test_datamodule.py::test_lightning_node_data[None-neighbor]
```

On most machines, the test produces only the following warning:
```
  /usr/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=751) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

However, on machines equipped with GB300 GPUs, the same test results in an actual deadlock.

**Fix**

Explicitly use the spawn multiprocessing context whenever `num_workers > 0`. Unlike `fork`, `spawn` starts each worker as a fresh Python process, avoiding inheritance of the parent's CUDA runtime state and preventing the observed deadlock.

**Impact**

- Fixes the deadlock observed with GB300 GPUs.
- Removes the unsafe `fork()` usage for multiprocessing data-loader workers.
- Preserves the existing behavior when `num_workers=0`.
- Makes multiprocessing behavior explicit and safe when CUDA is involved.


drivanov:
The testing errors are expected because [PR#10779:](https://github.com/pyg-team/pytorch_geometric/pull/10779)  (**Fix HashTensor string keys with xxhash 4.0**) has not been merged yet.

